PR #121, opened 2020-08-11
Update README.md

nwl:
Replaced some stray syft entries with grype


alfredodeza:
Hey @nwl I actually think this PR is good, any insight as to why it was closed?

PR #124, opened 2020-08-12
Explicitly use PythonFormat to address PEP440 rules

### LINKED ISSUE #91: Python version parser is needed
The semantic versioning parsing that Grype has can't handle the non-semantic versions that Python allows. More specifically, anything that is after a formal release (e.g. `1.0.0-post1`) and pre-release variations like `1.0.0-dev1`.

The spec is:

```
[N!]N(.N)*[{a|b|rc}N][.postN][.devN]
```

And it is fully documented here https://www.python.org/dev/peps/pep-0440/

Fixing this issue, should allow https://github.com/anchore/grype/issues/90 to be correctly fixed for Python.

alfredodeza:
Closes #91 

By using the `PythonFormat`, the code can explicitly deal with the rules and constraints of PEP440 - luckily getting addressed by the Fuzzy implementation.

Requires https://github.com/anchore/grype-db-builder/pull/53

### DIFF
--- a/grype/version/constraint.go
+++ b/grype/version/constraint.go
@@ -17,8 +17,15 @@ func GetConstraint(constStr string, format Format) (Constraint, error) {
 		return newDebConstraint(constStr)
 	case RpmFormat:
 		return newRpmConstraint(constStr)
+	case PythonFormat:
+		// This is specific to PythonFormat so that it adheres to PEP440 and its odd corner-cases
+		// It is significantly odd enough, that the fuzzyConstraint is the best bet to compare versions.
+		// Although this will work in most cases, some oddities aren't supported, like:
+		// 1.0b2.post345.dev456 which is allowed by the spec. In that case (a dev release of a post release)
+		// the comparator will fail. See https://www.python.org/dev/peps/pep-0440
+		return newFuzzyConstraint(constStr, "python")
 	case UnknownFormat:
-		return newFuzzyConstraint(constStr)
+		return newFuzzyConstraint(constStr, "unknown")
 	}
 	return nil, fmt.Errorf("could not find constraint for given format: %s", format)
 }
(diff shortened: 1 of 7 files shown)


grype/version/constraint.go
```diff
@@ -17,6 +17,8 @@ func GetConstraint(constStr string, format Format) (Constraint, error) {
 		return newDebConstraint(constStr)
 	case RpmFormat:
 		return newRpmConstraint(constStr)
+	case PythonFormat:
+		return newFuzzyConstraint(constStr)
```
wagoodman, line 21: nit: a comment highlighting why we are using the fuzzy constraint instead of a PEP440 parser would help here. The test cases added for the fuzzy constraint are great for pointing out what is supported, pairing that with a comment here that lists some specific example PEP440 features that aren't supported could help in debugging potential issues later (doesn't need to be exhaustive, as it is a very full-featured PEP)

test/integration/match_coverage_test.go
```diff
@@ -91,7 +91,7 @@ func addPythonMatches(t *testing.T, theScope scope.Scope, catalog *pkg.Catalog,
 		Confidence:      1.0,
 		Vulnerability:   *vulnObj,
 		Package:         thePkg,
-		SearchKey:       "language[python] constraint[< 2.6.2 (semver)]",
+		SearchKey:       "language[python] constraint[< 2.6.2 (unknown)]",
```
wagoodman, line 94: question: is this a change to support the short term fix? Or should this match the "python" format string?

alfredodeza: Good question. Initially I thought it was because https://github.com/anchore/grype-db-builder/pull/53 hasn't merged yet, but this is using a mock store... I'll debug what is going on here, I had it with `python` initially but the tests wanted `unknown`

grype/version/constraint.go
alfredodeza: I added a comment with some of the oddities that this might not pick up, as well as a link to PEP440

test/integration/match_coverage_test.go
alfredodeza: I've updated this. What happened was that the `fuzzyConstraint` didn't have access to whatever created the constraint. As it was implemented, the `String()` would return `(unknown)`. I've updated the constructor to accept a hint and use it.

alfredodeza:
@luhring @wagoodman I've updated this, since it has new changes would love a second (final?) look if possible

PR #136, opened 2020-08-17
Match on `setup.py` files

alfredodeza:
I thought this would take much more changes, but it looks like all that is needed is to add the new package type.

Blocked until https://github.com/anchore/syft/pull/150 is merged - this PR will need to update the dependencies to use the new changes.

```text
$ go run main.go -vv  dir://../devml
[...]
[0001] DEBUG searching for vulnerability matches for pkg=Pkg(type=python-setup, name=click, version=1.1.1)
[0001] DEBUG found 1 vulnerabilities for pkg=Pkg(type=python-setup, name=click, version=1.1.1)
[0001] DEBUG   └── vuln='CVE-2015-8768' type='Fuzzy Match' key='cpe[cpe:2.3:*:*:click:1.1.1:*:*:*:*:python:*:*] constraint[none (unknown)]' foundBy='python-matcher')
[0001] DEBUG searching for vulnerability matches for pkg=Pkg(type=egg, name=devml, version=0.5.1)
```

### DIFF
--- a/test/integration/match_coverage_test.go
+++ b/test/integration/match_coverage_test.go
@@ -18,7 +18,7 @@ import (
 
 func getPackagesByPath(t *testing.T, theScope scope.Scope, catalog *pkg.Catalog, thePath string) []*pkg.Package {
 	t.Helper()
-	refs, err := theScope.FilesByGlob(thePath)
+	refs, err := theScope.Resolver.FilesByGlob(thePath)
 	if err != nil {
 		t.Fatalf("could not get ref by path %q: %+v", thePath, err)
 	}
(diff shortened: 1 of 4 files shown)


test/integration/match_coverage_test.go
```diff
@@ -18,7 +18,7 @@ import (
 
 func getPackagesByPath(t *testing.T, theScope scope.Scope, catalog *pkg.Catalog, thePath string) []*pkg.Package {
 	t.Helper()
-	refs, err := theScope.FilesByGlob(thePath)
+	refs, err := theScope.Resolver.FilesByGlob(thePath)
```
alfredodeza, line 21: this was tricky to figure out... it was breaking because of changes in: https://github.com/anchore/syft/pull/148/

PR #143, opened 2020-08-28
cyclonedx presenter

### LINKED ISSUE #39: Presenter for CycloneDX Vulnerability Extensions
Support output of a resulting scan as CycloneDX w/Vulnerability Extension: https://cyclonedx.org/ext/vulnerability

This may include support for BOM descriptor and/or dependency graph, but isn't required.

This issue is about output only, not support for importing CycloneDX BOMs.

alfredodeza:
Implements the CycloneDX presenter. See https://cyclonedx.org/ext/vulnerability/

Closes #39


grype/presenter/cyclonedx/vuln-extension.go
```diff
@@ -0,0 +1,73 @@
...
+
+// Source is the origin of the vulnerability, like Github Advisories or NVD, along
+// with a URL constructed with the vulnerability ID
+type Source struct {
+	Name string `xml:"v:name,attr"`
+	URL  string `xml:"v:url"`
+}
+
+// Raiting has information about the intensity of a vulnerability
+type Raiting struct {
```
luhring, line 22: Should this be `Rating`? (spelling)

luhring: (ditto comment for all other references to this and related property/variable names)

grype/presenter/cyclonedx/presenter.go
```diff
@@ -0,0 +1,85 @@
...
+	_, err = output.Write([]byte(xml.Header))
+	if err != nil {
+		return err
+	}
+	_, err = output.Write(xmlOut)
+	if err != nil {
+		return err
+	}
+
+	_, err = output.Write([]byte("\n"))
```
luhring, line 83: Not blocking / for consideration — I'm wondering if it'd be more idiomatic to Go, and possibly less verbose, to use an **encoder** for this, given that our objective is to write output to an `io.Writer`. It seems to me that the `Marshal` functions are for when you want to store bytes (return a `[]byte`), and the `Encode` methods are for when you want to _write_ bytes (to a `Writer`). (Same pattern as the `json` std lib package.)

For reference: https://golang.org/pkg/encoding/xml/#NewEncoder

@wagoodman Want to make sure you see this, too.

wagoodman: good call, I can head back to syft an update there

grype/presenter/cyclonedx/vuln-extension.go
wagoodman: indeed, in order to catch all out-of-spec issues like this it would be ideal to add testing against the schema in the same/similar way done in syft https://github.com/anchore/syft/pull/163 . The extra addition would be to add the [vulnerability extension XSD](https://github.com/CycloneDX/specification/blob/master/schema/ext/vulnerability-1.0.xsd) to test against.

grype/presenter/cyclonedx/document.go
```diff
@@ -0,0 +1,149 @@
...
+
+// Source: https://github.com/CycloneDX/specification
+
+// Document represents a CycloneDX Vulnerability Document.
+type Document struct {
+	XMLName       xml.Name               `xml:"bom"`
+	XMLNs         string                 `xml:"xmlns,attr"`
+	Version       int                    `xml:"version,attr"`
+	SerialNumber  string                 `xml:"serialNumber,attr"`
+	Components    []Component            `xml:"components"`
```
wagoodman, line 24: I lied to you the last we spoke, I learned that `components>component` is necessary for marshaling as well. This hints to the XML lib to behave like so:
```xml
<components>
    <component>thing1</component>
    <component>thing2</component>
</components>
```
instead of like so:
```xml
<component>thing1</component>
<component>thing2</component>
```

grype/presenter/cyclonedx/presenter.go
```diff
@@ -0,0 +1,84 @@
...
+				Type:    "container",
+				Name:    imageID,
+				Version: versionStr,
+			},
+		}
+	default:
+		return fmt.Errorf("unsupported source: %T", src)
+	}
+
+	encoder := xml.NewEncoder(output)
```
alfredodeza, line 69: @luhring this has been addressed

alfredodeza:
Finally got everything passing. My goodness this was tricky to get right. Thank you @luhring for the final review!

PR #146, opened 2020-09-08
Change search key json output to a map

### LINKED ISSUE #145: Replace search key with a map
**What would you like to be added**: Change the search key to a `map[string]string`

**Why is this needed**: Currently the existing search key is a string that needs to be parsed, which would be unnecessary as a map.

Note: this is a follow up from https://github.com/anchore/grype/pull/137#discussion_r474756009

wagoodman:
This is a follow up on https://github.com/anchore/grype/pull/137#discussion_r474756009 ; Today the `SearchKey` for a grype Match is an arbitrary string. To allow for more programmatic usefulness of the field, changing the key to a map of items that are used in the search is more useful (as no string parsing is required to interpret the key).

Closes #145


luhring (commented):
I like the change. I could use more clarity on how this new data structure is intended to be used...

Could we do one or both of these things?

1. Document this field with a doc comment, spelling out how it should be used
2. If possible use more explicit types (instead of `map[string]interface{}`)

luhring (approved):
Makes sense. Looks good!

PR #149, opened 2020-09-09
Shell completion via Cobra utility

dakaneye:
Add completion script, ValidArgsFunction to root command to list docker images using docker go SDK, and update README

### DIFF
--- a/cmd/completion.go
+++ b/cmd/completion.go
@@ -1,0 +1,63 @@
+package cmd
+
+import (
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+// completionCmd represents the completion command
+var completionCmd = &cobra.Command{
+	Use:   "completion [bash|zsh|fish|powershell]",
+	Short: "Generate a shell completion for Grype (listing local docker images)",
+	Long: `To load completions (docker image list):
+
+Bash:
+
+$ source <(grype completion bash)
+
+# To load completions for each session, execute once:
+Linux:
+  $ grype completion bash > /etc/bash_completion.d/grype
+MacOS:
+  $ grype completion bash > /usr/local/etc/bash_completion.d/grype
+
+Zsh:
+
+# If shell completion is not already enabled in your environment you will need
+# to enable it.  You can execute the following once:
+
+$ echo "autoload -U compinit; compinit" >> ~/.zshrc
+
+# To load completions for each session, execute once:
+$ grype completion zsh > "${fpath[1]}/_grype"
+
+# You will need to start a new shell for this setup to take effect.
+
+Fish:
+
+$ grype completion fish | source
+
+# To load completions for each session, execute once:
+$ grype completion fish > ~/.config/fish/completions/grype.fish
+`,
+	DisableFlagsInUseLine: true,
+	ValidArgs:             []string{"bash", "fish"},
+	Args:                  cobra.ExactValidArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		var err error
+		switch args[0] {
+		case "bash":
+			err = cmd.Root().GenBashCompletion(os.Stdout)
+		case "fish":
+			err = cmd.Root().GenFishCompletion(os.Stdout, true)
+		}
+		if err != nil {
+			panic(err)
+		}
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(completionCmd)
+}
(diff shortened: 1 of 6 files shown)


cmd/completion.go
```diff
@@ -0,0 +1,63 @@
...
+	ValidArgs:             []string{"bash", "zsh", "fish", "powershell"},
+	Args:                  cobra.ExactValidArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		switch args[0] {
+		case "bash":
+			cmd.Root().GenBashCompletion(os.Stdout)
+		case "zsh":
+			cmd.Root().GenZshCompletion(os.Stdout)
+		case "fish":
+			cmd.Root().GenFishCompletion(os.Stdout, true)
```
alfredodeza, line 54: why does this need `true` here?

dakaneye: This was the example Cobra said to use for the generation command so I was following their lead here. 

Upon further inspection of the library code, it looks like this dictates what type of command cobra actually uses to do the shell completion. If it's false, it will use a different command that doesn't include completion result completions? Idk it's a bit confusing:
```go
const (
	// ShellCompRequestCmd is the name of the hidden command that is used to request
	// completion results from the program.  It is used by the shell completion scripts.
	ShellCompRequestCmd = "__complete"
	// ShellCompNoDescRequestCmd is the name of the hidden command that is used to request
	// completion results without their description.  It is used by the shell completion scripts.
	ShellCompNoDescRequestCmd = "__completeNoDesc"
)
```

wagoodman:
Bash completion works well, though I'm running into issues with `zsh` with this:
```
$ grype <tab><tab>                                  
_arguments:comparguments:325: invalid option definition: (-o --output)-o[report output formatter, options=[json table cyclonedx]]:
```

dakaneye:
Hmm, this actually doesn't seem to be working. I was printing the docker images well via the debug utility:
```
samdacanay@SDACANAY ~ % grype __complete anch
anchore-local-docker-registry:latest
anchore-engine:dev
anchore-engine:builder
anchore/anchore-engine-dev:latest
anchore/enterprise-dev:latest
anchore/anchore-on-prem-ui-dev:latest
anchore/enterprise:v2.4.0
anchore/enterprise-ui:v2.4.0
anchore/engine-db-preload:v0.7.2
:0
Completion ended with directive: ShellCompDirectiveDefault
```
But with Bash I'm not seeing images, just files:
```
bash-3.2$ grype <TAB><TAB>
.CFUserTextEncoding
.DS_Store
...
```
And with Zsh I'm seeing the subcommands:
```
samdacanay@SDACANAY ~ % grype <TAB><TAB>
completion  -- Generate a shell completion for Grype (listing local docker images)
db          -- vulnerability database operations
help        -- Help about any command
version     -- show the version
```

dakaneye:
OK @alfredodeza  @luhring @wagoodman kinda went down a rabbit hole with this stuff, but it was fun so who cares!?

In summary:
* Cobra has 2 bugs with ZSH completion
    * It doesn't escape square brackets in option descriptions (weirdly doesn't affect bash or fish completions)
    * It doesn't generate a script with support for the dynamic noun completion that I've added to list local docker repo tags (both bash and Fish generated completion scripts do)
* I've opened an [issue](https://github.com/spf13/cobra/issues/1226) with Cobra (let's see what they say!)
* I removed the ability to generate Zsh and Powershell from the completion utility for now. (Removed Powershell because I didn't have a way to test)
    * I decided to do it this way instead of hardcoding a generation script that can do the dynamic completion for the following reasons:
        * We didn't support Zsh before either
        * This is mostly a UX improvement for grype, and didn't want to invest too much time doing that generation, with the hopes that cobra can fix the issue in the near future

dakaneye:
OK, got a response from Cobra, they said it's fixed in their master branch, but has not yet been released. Since we're on 1.0.0, that's why it's not working. No idea when they'll release next but at least it validates that I'm not crazy because their master branch says it should work! You can follow the question in my linked issue above. For now should be good to go though. Once this gets approved I can do the same for syft b/c it should be useful there as well.

wagoodman (approved):
good catch with the fix on cobra's master branch

dakaneye:
@wagoodman awesome! Didn't know we could just refer to a commit on the repo when including other modules!

luhring:
That's awesome news re Cobra!

For now at least, PowerShell isn't a high priority for us, so no worries there. For future reference, PowerShell is cross-platform, so it should be something we'd all be able to test with.

I'm excited to have this completion functionality available!

PR #150, opened 2020-09-10
Add positional argument validation for DB commands

wagoodman:
Fixes a crash when no archive is given for db import and additionally is more explicit about positional arguments on remaining db commands.

### DIFF
--- a/cmd/db_check.go
+++ b/cmd/db_check.go
@@ -11,6 +11,7 @@ import (
 var dbCheckCmd = &cobra.Command{
 	Use:   "check",
 	Short: "check to see if there is a database update available",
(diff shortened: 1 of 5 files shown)


dakaneye (approved):
Good to be explicit for expected input! LGTM

luhring:
Just a post mortem curiosity — why do we pass `args` as an argument into functions we define that discard the `args` argument? (Is this something we could simplify by explicitly not expressing a need for `args` further upstream?)

E.g.:

```go
err := runDbStatusCmd(cmd, args)
```

```go
func runDbStatusCmd(_ *cobra.Command, _ []string) error {
```

PR #152, opened 2020-09-15
Change kebab case to camelCase

dakaneye:
Did a simple find/replace to fix the keys, but if any of these are stored as JSON persistently, it will require update there too

### DIFF
--- a/.bouncer.yaml
+++ b/.bouncer.yaml
@@ -4,3 +4,6 @@ permit:
   - Apache.*
   - MPL.*
   - ISC
+ignore-packages:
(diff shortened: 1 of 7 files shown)


wagoodman (approved):

luhring (approved):

wagoodman:
Since the version of syft has been updated, the packageurl repo is now a dependency of grype. However, an exception is made here for that dependency in syft: https://github.com/anchore/syft/blob/main/.bouncer.yaml#L7-L9

The same exception will need to be made in grype.

PR #153, opened 2020-09-16
Rename Result object to Matches

wagoodman:
Polishing before release --the "Results" object was never meant to be the final name. This PR finalizes a couple of the TODOs by renaming `Result` -> `Matches` and migrating to the `match` package.

### DIFF
--- a/grype/match/matches.go
+++ b/grype/match/matches.go
@@ -1,26 +1,21 @@
-package result
+package match
 
 import (
-	"github.com/anchore/grype/grype/match"
 	"github.com/anchore/syft/syft/pkg"
 )
 
-// TODO: consider moving this to the pkg/match under matches.go
-
-// TODO: consider renaming to Matches
-
-type Result struct {
-	byPackage map[pkg.ID][]match.Match
+type Matches struct {
+	byPackage map[pkg.ID][]Match
 }
 
-func NewResult() Result {
-	return Result{
-		byPackage: make(map[pkg.ID][]match.Match),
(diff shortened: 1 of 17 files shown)


grype/presenter/json/presenter_test.go
```diff
@@ -142,7 +141,7 @@ func TestJsonPresenter(t *testing.T) {
 		},
 	}
 
-	results := result.NewResult()
+	results := match.NewResult()
```
alfredodeza, line 144: shouldn't this now be `match.NewMatch()` ?

grype/match/matches.go
```diff
@@ -1,57 +1,52 @@
...
-type Result struct {
-	byPackage map[pkg.ID][]match.Match
+type Matches struct {
+	byPackage map[pkg.ID][]Match
 }
 
-func NewResult() Result {
-	return Result{
-		byPackage: make(map[pkg.ID][]match.Match),
+func NewResult() Matches {
```
alfredodeza, line 11: so this, I think, would need to change to `NewMatch()` or perhaps `NewMatches()` ?

grype/presenter/json/presenter_test.go
wagoodman: great catch, I'll update

grype/match/matches.go
wagoodman: agreed, `NewMatches`

alfredodeza (approved):

PR #154, opened 2020-09-16
Add CycloneDX schema validation

### LINKED ISSUE #148: Add CycloneDX Schema Validation
**What would you like to be added**: Similar to https://github.com/anchore/syft/pull/163, we should be testing the output of the grype CycloneDX report against a XML schema validator that includes the [vulnerability extension](https://github.com/CycloneDX/specification/blob/master/schema/ext/vulnerability-1.0.xsd).

**Why is this needed**: This is to ensure we do not release CycloneDX presenter changes that generate non-compliant XML without manual verification.

alfredodeza:
Found two bugs in the output thanks to `xmllint`. Fixed those along with getting the schemas in place for validation.

Closes #148 

Requires https://github.com/anchore/go-testutils/pull/3 to be merged, which fixes a problem with loading golden images.


grype/presenter/cyclonedx/test-fixtures/snapshot/TestCycloneDxPresenter_CycloneDX_Directory_Presenter.golden
```diff
@@ -55,7 +55,7 @@
     </component>
   </components>
   <bd:metadata>
-    <bd:timestamp>2020-08-31T15:39:55-04:00</bd:timestamp>
+    <bd:timestamp>2020-09-17T08:53:32-04:00</bd:timestamp>
     <bd:tool>
       <bd:vendor>anchore</bd:vendor>
       <bd:name>syft</bd:name>
```
wagoodman, line 61: this should be `grype`, so `NewBomDescriptor()` in  `vuln-extension.go` should be used in `document.go`

wagoodman (changes requested):
currently xmllint is not using the vulnerability extension XSD (`v` namespace), so it is possible for non-compliant output to not be validated (I believe only the cycloneDX XSD and bom extension are being referenced)

grype/presenter/cyclonedx/test-fixtures/snapshot/TestCycloneDxPresenter_CycloneDX_Directory_Presenter.golden
alfredodeza: Although the BOM is produced by Syft, you believe that this should be swapped for Grype?

alfredodeza:
> currently xmllint is not using the vulnerability extension XSD (`v` namespace), so it is possible for non-compliant output to not be validated (I believe only the cycloneDX XSD and bom extension are being referenced)

Very good catch. I thought it was, but was just confused because the namespace is now declared - passing validation. Adding the Vulnerability schema identified issues that I am now fixing

grype/presenter/cyclonedx/document.go
```diff
@@ -69,7 +73,21 @@ func NewVulnerability(m match.Match, p vulnerability.MetadataProvider) (Vulnerab
 	}
 
 	rating.Score = score
-	rating.Severity = metadata.Severity
+
+	r := Ratings{
+		Rating: []Rating{rating},
+	}
+
+	// The schema does not allow "Negligible", only allowing the following:
```
alfredodeza, line 81: @wagoodman there is some disparity here, not sure why `Negligible` is not allowed by the schema but captured in the database

grype/presenter/cyclonedx/test-fixtures/snapshot/TestCycloneDxPresenter_CycloneDX_Directory_Presenter.golden
alfredodeza: This got updated

grype/presenter/cyclonedx/vuln-extension.go
```diff
@@ -52,10 +57,15 @@ type Vulnerability struct {
...
 
+// Vulnerabilities contains all the vulnerabilities for a component
+type Vulnerabilities struct {
+	Vulnerability []Vulnerability `xml:"v:vulnerability,omitempty"`
+}
+
 // Component represents the a single software/package that has vulnerabilities.
 type Component struct {
 	syftCDX.Component
-	Vulnerabilities []Vulnerability `xml:"v:vulnerabilities"`
```
alfredodeza, line 58: This change allows passing the validation, basically it is nesting vulnerability into vulnerabilities

grype/presenter/cyclonedx/test-fixtures/snapshot/TestCycloneDxPresenter_CycloneDX_Directory_Presenter.golden
wagoodman: I thought about the "is it syft or grype" and I think though it's true that syft is producing the work, it would be confusing for someone to run grype and see that it was generated by syft. Open to opinions on this though, I feel there are good arguments either way.

grype/presenter/cyclonedx/document.go
wagoodman: The DB build process allows any values from upstream and does not do any validation around this --upstream is the source of truth on "correctness". Upstream "Negligible" is valid, it happens to be in CycloneDX specifically downstream where this is not allowed. It may be tempting to change upstream (in the DB build process) to eliminate this discrepancy, however, this would implicitly couple CycloneDX's opinions with what is "valid" in the DB, which feels wrong --presentation concerns/opinions/restrictions should say with presenters.

grype/presenter/cyclonedx/test-fixtures/snapshot/TestCycloneDxPresenter_CycloneDX_Directory_Presenter.golden
alfredodeza: I changed it to reflect `grype`. We can always update this later if we feel it isn't working well.

grype/presenter/cyclonedx/document.go
alfredodeza: That's right, this is CycloneDX being opinionated about this being invalid. I checked and other systems like Github's SARIF report allows it, so I vote to just mangle it here for Cyclone

grype/presenter/cyclonedx/vuln-extension.go
wagoodman: This approach is still leaving empty tags in the generated output:
```xml
    <component type="library">
      <name>base-passwd</name>
      <version>3.5.47</version>
      <v:vulnerabilities></v:vulnerabilities>
    </component>
```
where more correct behavior would be to not include empty tags:
```xml
    <component type="library">
      <name>base-passwd</name>
      <version>3.5.47</version>
    </component>
```
... both approaches are within-spec relative to CycloneDX, but the latter is preferred.

The `omitempty` struct tag option for `xml` has the following behavior (from https://golang.org/pkg/encoding/xml/#Marshal):
```
a field with a tag including the "omitempty" option is omitted
  if the field value is empty. The empty values are false, 0, any
  nil pointer or interface value, and any array, slice, map, or
  string of length zero.
``` 
which means that empty structs are treated as non-empty values (so `omitempty` will not omit empty `Vulnerability` instances). If instead the type was `[]Vulnerability` or `*Vulnerabilities` I believe the xml tags would be omitted.

This is the case for at least `rating` and `severity` tags (maybe more)

alfredodeza: Ahhh interesting, I thought I was catching them but I guess it wasn't sufficient. Will address this

wagoodman (approved):
looks great! We can come back later for package-url enhancements (which the spec does not require)

grype/presenter/cyclonedx/document.go
alfredodeza: Just a note for the sake of clarity: this ended up coming from Anchore itself, where it produces "Negligible". CycloneDX does not allow it because it isn't part of the allowed ratings as per https://nvd.nist.gov/vuln-metrics/cvss

PR #156, opened 2020-09-18
add --fail-on threshold support

### LINKED ISSUE #129: Create a switch for exiting non-zero if vulnerabilities are found
Users are going to want to use `grype` in their CI/CD pipelines. Typically, this will occur right after they've built a container image.

_Pipelines rely on the exit status of processes to know whether the pipeline has "failed" or should continue with subsequent steps. If a process (like a linter, a compiler, a test runner, etc.) exits 1 (or some non-zero number), that communicates that this particular CI job has failed._

With `grype`, users should be able to ask that the process exit non-zero when vulnerabilities are found, so that their pipelines fail — intentionally.

Example:

```
$ grype --enforce my-image
... a list of vulnerabilities ...
$ echo $?
1
```

The exact name of the switch itself is up for discussion, but should be both concise and easily understood by others.

It's important that this behavior be *opt-in*. By default, `grype` should exit non-zero only when a runtime error has occurred. There are plenty of scenarios in which a user would expect `grype` to exit `0` even when vulnerabilities are present (for example, when the output of grype is being piped to a downstream process).

wagoodman:
Adds the following functionalities:
- the ability to comb through the discovered vulnerabilities and set a non-0 exit code when a vulnerability at or above the given user threshold is found (provided by `--fail-on`).
- a `ExpectedErr` struct for creating errors that are known to be produced from grype. There is room to expand this in the future.

Note: the drop in coverage is somewhat fictitious since a test was added in a package that has no existing tests. 

Closes #129

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -190,8 +190,8 @@ func startWorker(userInput string, failOnSeverity *vulnerability.Severity) <-cha
 		// determine if there are any severities >= to the max allowable severity (which is optional).
 		// note: until the shared file lock in sqlittle is fixed the sqlite DB cannot be access concurrently,
 		// implying that the fail-on-severity check must be done before sending the presenter object.
-		if aboveAllowableSeverity(failOnSeverity, results, metadataProvider) {
-			errs <- grypeerr.ErrAboveAllowableSeverity
+		if hitSeverityThreshold(failOnSeverity, results, metadataProvider) {
+			errs <- grypeerr.ErrAboveSeverityThreshold
 		}
 
 		bus.Publish(partybus.Event{
@@ -209,9 +209,9 @@ func runDefaultCmd(_ *cobra.Command, args []string) error {
 	return ux(errs, eventSubscription)
 }
 
-// aboveAllowableSeverity indicates if there are any severities >= to the max allowable severity (which is optional)
-func aboveAllowableSeverity(failOnSeverity *vulnerability.Severity, results result.Result, metadataProvider vulnerability.MetadataProvider) bool {
-	if failOnSeverity != nil {
+// hitSeverityThreshold indicates if there are any severities >= to the max allowable severity (which is optional)
+func hitSeverityThreshold(thresholdSeverity *vulnerability.Severity, results result.Result, metadataProvider vulnerability.MetadataProvider) bool {
+	if thresholdSeverity != nil {
 		var maxDiscoveredSeverity vulnerability.Severity
 		for m := range results.Enumerate() {
 			metadata, err := metadataProvider.GetMetadata(m.Vulnerability.ID, m.Vulnerability.RecordSource)
@@ -224,7 +224,7 @@ func aboveAllowableSeverity(failOnSeverity *vulnerability.Severity, results resu
 			}
 		}
 
-		if maxDiscoveredSeverity >= *failOnSeverity {
+		if maxDiscoveredSeverity >= *thresholdSeverity {
 			return true
 		}
 	}
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"runtime/pprof"
@@ -10,7 +11,9 @@ import (
 
 	"github.com/anchore/grype/grype"
 	"github.com/anchore/grype/grype/event"
+	"github.com/anchore/grype/grype/grypeerr"
 	"github.com/anchore/grype/grype/presenter"
+	"github.com/anchore/grype/grype/result"
 	"github.com/anchore/grype/grype/vulnerability"
 	"github.com/anchore/grype/internal"
 	"github.com/anchore/grype/internal/bus"
@@ -68,13 +71,18 @@ var rootCmd = &cobra.Command{
 		}
 
 		if err != nil {
-			log.Errorf(err.Error())
+			var grypeErr grypeerr.ExpectedErr
+			if errors.As(err, &grypeErr) {
+				fmt.Fprintln(os.Stderr, format.Red.Format(grypeErr.Error()))
+			} else {
+				log.Errorf(err.Error())
+			}
 			os.Exit(1)
 		}
 	},
 	ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
 		// Since we use ValidArgsFunction, Cobra will call this AFTER having parsed all flags and arguments provided
-		dockerImageRepoTags, err := ListLocalDockerImages(toComplete)
+		dockerImageRepoTags, err := listLocalDockerImages(toComplete)
 		if err != nil {
 			// Indicates that an error occurred and completions should be ignored
 			return []string{"completion failed"}, cobra.ShellCompDirectiveError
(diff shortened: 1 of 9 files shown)


grype/grypeerr/errors.go
```diff
@@ -0,0 +1,6 @@
+package grypeerr
```
dakaneye, line 1: Nit: Would it be worth hyphenating this to make it more readable? I know the k8s client has a hyphen in the package name, so should be ok?

cmd/root.go
```diff
@@ -179,12 +204,34 @@ func startWorker(userInput string) <-chan error {
...
-	errs := startWorker(userInput)
+	errs := startWorker(userInput, appConfig.FailOnSeverity)
 	ux := ui.Select(appConfig.CliOptions.Verbosity > 0, appConfig.Quiet)
 	return ux(errs, eventSubscription)
 }
 
-func ListLocalDockerImages(prefix string) ([]string, error) {
+// hitSeverityThreshold indicates if there are any severities >= to the max allowable severity (which is optional)
+func hitSeverityThreshold(thresholdSeverity *vulnerability.Severity, results result.Result, metadataProvider vulnerability.MetadataProvider) bool {
+	if thresholdSeverity != nil {
```
dakaneye, line 214: could reduce the nesting in this method a bit:
```
if thresholdSeverity == nil {
    return true
}
```

dakaneye (approved):

grype/grypeerr/errors.go
```diff
@@ -0,0 +1,6 @@
...
+
+var (
+	// ErrAboveSeverityThreshold indicates when a vulnerability severity is discovered that is above the given --fail-on severity value
+	ErrAboveSeverityThreshold = NewExpectedErr("discovered vulnerabilities at or above the severity threshold")
```
luhring, line 5: I would probably move specific errors out of the general error package and closer to where the error is relevant. For example, if the severity checking logic is in the `cmd` package right now, that's where I'd put this error.

grype/grypeerr/errors.go
luhring: It's okay for hyphens to appear in module paths, but not package names. If we inserted a hyphen so that we have `package grype-err`, the code wouldn't compile.

[Effective Go](https://golang.org/doc/effective_go.html#package-names) (another great read for aspiring Go devs, btw!) puts forth the convention of packages that use "lower case, single-word names; there should be no need for underscores or mixedCaps". I don't think we need to apply this rule to 100% of cases, but it's worth keeping in mind.

In short, I'm okay with leaving the code as written. If we want to consider alternative approaches, one might be to name this package just "err" or something. So full path would be `github.com/anchore/grype/grype/err`. This would likely lead to aliasing the package when imported throughout our code, for example:

```go
import (
    grypeerr "github.com/anchore/grype/grype/err"
)
```

So I'm not sure it's a net win to change it.

cmd/root.go
luhring: Agreed. Also, I think this could potentially be further simplified by having the severity variable not be a pointer type.

internal/ui/etui/ephemeral_tui.go
```diff
@@ -73,16 +75,25 @@ func OutputToEphemeralTUI(workerErrs <-chan error, subscription *partybus.Subscr
 	ctx := context.Background()
 	grypeUIHandler := grypeUI.NewHandler()
 
-eventLoop:
+	var errResult error
 	for {
 		select {
-		case err := <-workerErrs:
+		case err, ok := <-workerErrs:
```
luhring, line 81: GoLand underlined `workerErrs` with this warning: "Receive may block because of 'nil' channel". May be worth checking for `nil` ahead of time.

internal/ui/etui/ephemeral_tui.go
```diff
@@ -73,16 +75,25 @@ func OutputToEphemeralTUI(workerErrs <-chan error, subscription *partybus.Subscr
...
+					errResult = err
+					continue
+				}
 				return err
 			}
+			if !ok {
+				// worker completed
+				workerErrs = nil
+			}
 		case e, ok := <-events:
```
luhring, line 93: GoLand underlined `events` with this warning: "Receive may block because of 'nil' channel". May be worth checking for `nil` ahead of time.

internal/ui/etui/ephemeral_tui.go
```diff
@@ -73,16 +75,25 @@ func OutputToEphemeralTUI(workerErrs <-chan error, subscription *partybus.Subscr
...
 
-eventLoop:
+	var errResult error
 	for {
 		select {
-		case err := <-workerErrs:
+		case err, ok := <-workerErrs:
 			if err != nil {
+				if errors.Is(err, grypeerr.ErrAboveSeverityThreshold) {
+					errResult = err
```
luhring, line 84: Definitely up for discussion --> if I'm reading this correctly, it looks like the logic around `errResult` is such that errors that aren't the latest error are lost forever. Is that true? If so, would it be worth collecting surfaced errors for reporting, if we're not displaying them immediately?

internal/ui/etui/ephemeral_tui.go
```diff
@@ -112,15 +123,15 @@ eventLoop:
...
 				// this is the last expected event
-				break eventLoop
+				events = nil
 			}
 		case <-ctx.Done():
-			if ctx.Err() != nil {
-				log.Errorf("cancelled (%+v)", err)
-			}
-			break eventLoop
+			return grypeerr.NewExpectedErr("cancelled: %w", ctx.Err())
```
luhring, line 129: Nit: sp: "canceled"

grype/grypeerr/errors.go
wagoodman: I agree with both of you --I don't like `grypeerr` but also agree with Dan's points and don't see a better alternative. The `err` name is problematic since, when imported, would be overridden by the most popular local variable name for errors within functions. I'll leave it as is for now.

cmd/root.go
wagoodman: 👍 guard clases are best!

grype/grypeerr/errors.go
wagoodman: Agreed, I'll move this over to `cmd`

internal/ui/etui/ephemeral_tui.go
wagoodman: per our conversation, this section probably needs more love than this PR is willing to give

internal/ui/etui/ephemeral_tui.go
wagoodman: this is here to prevent return of an error and short circuit the UI logic --per our conversation, this section needs a second pass on design. It is OK that we are capturing the most recent `errResult` since this is only there to indicate `ErrAboveSeverityThreshold` --will not scale though.

internal/ui/etui/ephemeral_tui.go
wagoodman: [ditto](https://github.com/anchore/grype/pull/156#discussion_r492236148)

grype/grypeerr/errors.go
wagoodman: per our conversation, let's iterate on this --we should try not to have a global sentinel errors package, but probably isn't a show stopper for now (especially since we don't have a better pattern off hand yet)

PR #160, opened 2020-09-24
Add OCI support + use URI schemes

### LINKED ISSUE #157: Update CLI syntax for input source specification
See [the Slack discussion](https://anchore.slack.com/archives/C0131B70Q7M/p1600796387138500) for more context and motivation for this change.

Schemes (e.g. "dir", "docker") should be delimited from the rest of the URI by a colon (`:`), not by a colon and two slashes (`://`). (It's okay if an individual scheme requires two slashes to begin the scheme-specific portion of the URI.)

For example, if users want to scan the current directory, then instead of typing:

```
$ grype dir://.
```

Users would type:

```
$ grype dir:.
```

In addition, if we can detect that an input string that lacks a scheme specification (e.g. no `dir:`) is most likely to be a filesystem path as opposed to another specification, we could allow an even easier syntax for an implicit specification of input source:

```
$ grype .
```

wagoodman:
Adds the ability to scan OCI images from tar archive or directory:
```
# explicitly
grype oci-archive:path/to.tar
grype oci-dir:path/to/dir

# or implicitly
syft path/to/tar/or/dir
```

Also hinted in the example above, this PR incorporates changing the URL approach for specifying the source transport to a URI oriented approach (see https://github.com/anchore/stereoscope/pull/41 PR description for more details). TL;DR is `://` is being replaced with `:` to fall in line with the URI RFC.

Lastly, migrates to using the stereoscope fixture utils instead of the functions from `go-testutils`.

Closes #157


go.mod
```diff
@@ -33,3 +33,9 @@ require (
 	golang.org/x/crypto v0.0.0-20200622213623-75b288015ac9
 	gopkg.in/yaml.v2 v2.3.0
 )
+
+replace github.com/anchore/stereoscope => ../stereoscope
+
+replace github.com/anchore/syft => ../syft
+
+replace github.com/anchore/go-testutils => ../go-testutils
```
luhring, line 41: ditto replace comment in other PR

wagoodman: [ditto](https://github.com/anchore/syft/pull/178#discussion_r494362831)

README.md
```diff
@@ -45,19 +46,23 @@ grype <image> --scope all-layers
...
-grype dir://path/to/dir
+grype path/to/dir
 ```
 
-By default Grype shows a summary table, however, a more detailed `json` format is also available.
-
+The output format for Grype is configurable as well:
 ```
-grype <image> -o json
+grype <image> -o <format>
```
alfredodeza, line 58: this is a great way to format the help

README.md
```diff
@@ -45,19 +46,23 @@ grype <image> --scope all-layers
...
 
 ```
-# scan a docker image tar (from the result of "docker image save ... -o image.tar" command)
-grype docker-archive://path/to/image.tar
+# scan a container image archive (from the result of `docker image save ...`, `podman save ...`, or `skopeo copy` commands)
+grype path/to/image.tar
 
 # scan a directory
-grype dir://path/to/dir
+grype path/to/dir
```
alfredodeza, line 53: I understand this has probably been discussed a lot, but I thought these changes were to put some order in the usage of schemes, not generalize them to be able to infer the destination. Specifically this line in the in this PR's subject: 

>  TL;DR is :// is being replaced with : to fall in line with the URI RFC.

Automatically inferring is going to be problematic for integrations because it was the way to detect what the source type is (dir, tar, docker, etc..). If `grype` allows these inputs, then the integrations need to enforce a constraint, by either not allowing these free-styling of input, or separating input in the form of configuration.

That would mean that instead of these:

```yaml
    - name: Run directory Scan Action
      uses: anchore/scan-action@v1
      with:
        source: "dir://tests/python"

    - name: Run Image Scan Action
      uses: anchore/scan-action@v1
      with:
        source: "alpine:3.12"
```

The action would require prefixing the scheme or separating the options like this:

```yaml
    - name: Run directory Scan Action
      uses: anchore/scan-action@v1
      with:
        dir-source: "tests/python"

    - name: Run Image Scan Action
      uses: anchore/scan-action@v1
      with:
        img-source: "alpine:3.12"
```

wagoodman: What prevents grype from inferring from generic input? That is, what prevents this:
```
    - name: Run directory Scan Action
      uses: anchore/scan-action@v1
      with:
        source: "tests/python"

    - name: Run Image Scan Action
      uses: anchore/scan-action@v1
      with:
        source: "alpine:3.12"
```
or more explicitly from the user:
```
    - name: Run directory Scan Action
      uses: anchore/scan-action@v1
      with:
        source: "dir:tests/python"

    - name: Run Image Scan Action
      uses: anchore/scan-action@v1
      with:
        source: "docker:alpine:3.12"
```
Which would be the same as grype options:
```
grype docker:alpine:3.12
grype alpine:3.12
grype dir:tests/python
grype tests/python
```
Is there something in particular with the integration that it requires knowledge of the source type?

alfredodeza: It isn't that there is something that prevents it. But the integration must know if the source is a path or an image for reporting.

wagoodman: I think that implicit input from a user is OK here, as the user does not need to understand explicit schemes. Also this does not block extensions from enforcing explicit usage of schemes (either by prepending strings, string validation, or otherwise). I think your comments are quite notable, but mostly impact the integration implementation, not the CLI surface area on grype meant for the user.

alfredodeza (approved):

PR #163, opened 2020-09-25
Change root of JSON presenter to a mapping (instead of a sequence)

wagoodman:
Changes the JSON presenter from an enumeration of vulnerability findings...
```
[
 {
   "vulnerability": {
    "id": "CVE-1999-0001",
    ...
    }
  },
  ...
]
```

...to a JSON mapping at the root of the document:
```
{
 "vulnerabilities": [
    {
     "vulnerability": {
      "id": "CVE-1999-0001",
      ...
    },
    ...
   ],
   ...
}
```

And additionally adds the image or directory source information (just as in syft).


grype/presenter/json/test-fixtures/snapshot/TestJsonDirsPresenter.golden
```diff
@@ -0,0 +1,98 @@
...
+    "foundBy": [
+     "the-cataloger-1"
+    ],
+    "locations": [
+     "/some/path/pkg1"
+    ]
+   }
+  }
+ ],
+ "directory": "/some/path"
```
alfredodeza, line 97: Having the golden files is a good way to visualize how this output is shaping up. I think this is going to be very helpful for consumers. One thing that jumps at me is that a consumer would need to check for the key existence instead of asking for a value that advertises what the source is. 

There is at least one potential caveat in the case of a directory being a one-to-one mapping like in here: If we want to add any other key information to the directory scan besides the path, this would need to have a breaking change to include that. 

The pattern(s) that the JSON presenter follows in other parts of the object is what I would like to see. For example, in the `"artifact"` object, there are a few keys that advertise what their values are:

```json
"name": "package-1",
"type": "deb",
```

Which is great to consume, as the consumer doesn't need to figure out what it is dealing with. This is an example on how that would look like if it _didn't_ advertise the values as it is today:

```json
"deb": "package-1",
```

And that is the similarity that I see with the proposed approach in this PR. Can we consider making this new section a generic name that encompasses `directory`,`image` and any other future source? I don't have any strong opinions on the key name, but following what `grype` has internally this could be:

```json
  "source": {
    "type": "directory",
    "target": {
      "path": "/some/path"
    }
  }
```

For an image it could probably be something like:

```json
  "source": {
    "type": "directory",
    "target": {
      "tag": "latest",
      "repo": "alpine:3.12",
    }
  }
```

wagoodman:
from an offline chat, changing `vulnerabilities` to `matches`

grype/presenter/json/test-fixtures/snapshot/TestJsonDirsPresenter.golden
wagoodman: Agreed, I'll incorporate in syft!

luhring (approved):

PR #168, opened 2020-09-25
Add first issue/PR welcome message action

robertp:


.github/workflows/first-pr-issue-message.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+on: [pull_request, issues]
+
+jobs:
+  greeting:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/first-interaction@v1
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
```
zhill, line 11: Is this a one-time use token for each build? If it were exposed would that be a problem?

zhill (commented):
Looks reasonable to me. My only question is the token and any possible exposure.

.github/workflows/first-pr-issue-message.yaml
robertp: @zhill This is what github has to say about it (from reading this previously I'm assuming it's not a risk, but I'm not 100% certain that it could not be used if it were exposed/leaked); the token does get read/write access to quite a bit on the repo:

"GitHub automatically creates a GITHUB_TOKEN secret to use in your workflow. You can use the GITHUB_TOKEN to authenticate in a workflow run.

When you enable GitHub Actions, GitHub installs a GitHub App on your repository. The GITHUB_TOKEN secret is a GitHub App installation access token. You can use the installation access token to authenticate on behalf of the GitHub App installed on your repository. The token's permissions are limited to the repository that contains your workflow. For more information, see "Permissions for the GITHUB_TOKEN."

Before each job begins, GitHub fetches an installation access token for the job. The token expires when the job is finished."

https://docs.github.com/en/free-pro-team@latest/actions/reference/authentication-in-a-workflow#about-the-github_token-secret

https://docs.github.com/en/free-pro-team@latest/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token

wagoodman (approved):

PR #177, opened 2020-10-08
Incorporate gemspec cataloger

wagoodman:
Pulls in the work from https://github.com/anchore/syft/issues/197 with the gemspec cataloger


Toure (approved):
All test passed.

PR #211, opened 2020-11-18
Remove powershell description from completion command

KeisukeYamashita:
## What
I've removed the `powershell` completion command because it is not supported.


wagoodman (approved):
thanks for the fix!

PR #222, opened 2020-12-08
Fix distroNamespace mapping to only use major version for select distros

### LINKED ISSUE #221: UBI-based images do not generate any vulnerabilities
**What happened**:
When scanning a UBI-based image such as `registry.access.redhat.com/ubi8/ubi:latest`, the utility identifies 191 packages installed but 0 vulnerabilities.

I expected that UBI would match the RedHat vulnerability information just as it does with CentOS images.

`grype registry.access.redhat.com/ubi8/ubi:latest`
`grype centos:latest`

RedHat UBI image uses the same packages as RedHat Enterprise Linux and thus CentOS. Likely the utility itself isn't aware of the similarity.

**Environment**:
- Output of `grype version`:
```
Application:          grype
Version:              0.6.0
BuildDate:            2020-12-03T22:35:30Z
GitCommit:            1a75295d702a5f80d02d45a6263f96925673c775
GitTreeState:         clean
GoVersion:            go1.14.12
Compiler:             gc
Supported DB Schema:  1
```
- OS (e.g: `cat /etc/os-release` or similar): MacOS 11 Big Sur

wagoodman:
The RedHat, CentOS, AmazonLinux, and OracleLinux distros should be using a namespace with explicitly only the major version. Today there are cases where we also use the minor version if the raw data surfaces it (e.g. `rhel:8.3` should really be `rhel:8`). This PR fixes this behavior.

Additionally this corrects the mapping for fedora, which should map to `rhel` as well. A future PR will additionally fix the version mapping for this case (captured in https://github.com/anchore/grype/issues/223).

Closes #221

### DIFF
--- a/grype/vulnerability/namespace.go
+++ b/grype/vulnerability/namespace.go
@@ -37,22 +37,27 @@ func githubJavaNamer(p *pkg.Package) []string {
 // TODO: expand with namer mapping to be more generic?
 func distroNamespace(d distro.Distro) string {
 	// TODO: can we drive this from information from grype-db? that would be ideal...
-	var distroStr string
-	switch d.Type {
-	case distro.CentOS, distro.RedHat:
-		distroStr = "rhel"
-	case distro.AmazonLinux:
-		distroStr = "amzn"
-	case distro.OracleLinux:
-		distroStr = "ol"
-	case distro.Alpine:
-		// XXX this assumes that a major and minor versions will always exist in Segments
-		segments := d.Version.Segments()
-		return fmt.Sprintf("%s:%d.%d", d.Type.String(), segments[0], segments[1])
-	default:
-		distroStr = d.Type.String()
+	var versionSegments []int
+	if d.Version != nil {
+		versionSegments = d.Version.Segments()
 	}
-	return fmt.Sprintf("%s:%s", strings.ToLower(distroStr), d.FullVersion())
+
+	if len(versionSegments) > 0 {
+		switch d.Type {
+		// derived from https://github.com/anchore/anchore-engine/blob/5bbbe6b9744f2fb806198ae5d6f0cfe3b367fd9d/anchore_engine/services/policy_engine/__init__.py#L149-L159
+		case distro.CentOS, distro.RedHat, distro.Fedora:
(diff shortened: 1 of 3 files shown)


grype/vulnerability/namespace.go
```diff
@@ -37,20 +37,27 @@ func githubJavaNamer(p *pkg.Package) []string {
...
-		distroStr = "amzn"
-	case distro.OracleLinux:
-		distroStr = "ol"
-	case distro.Alpine:
-		// XXX this assumes that a major and minor versions will always exist in Segments
-		segments := d.Version.Segments()
-		return fmt.Sprintf("%s:%d.%d", d.Type.String(), segments[0], segments[1])
-	default:
-		distroStr = d.Type.String()
+	var distroStr = d.Type.String()
```
luhring, line 40: Nit: You might consider moving this closer to where it's used to make the code a bit easier to follow. If I'm reading this correctly, this variable isn't used until the last line of the function.

grype/vulnerability/namespace.go
```diff
@@ -37,20 +37,27 @@ func githubJavaNamer(p *pkg.Package) []string {
...
+	if d.Version != nil {
+		versionSegments = d.Version.Segments()
+	}
+
+	if len(versionSegments) > 0 {
+		switch d.Type {
+		// derived from https://github.com/anchore/anchore-engine/blob/5bbbe6b9744f2fb806198ae5d6f0cfe3b367fd9d/anchore_engine/services/policy_engine/__init__.py#L149-L159
+		case distro.CentOS, distro.RedHat, distro.Fedora:
+			// TODO: there is no mapping of fedora version to RHEL latest version (only the name)
+			return fmt.Sprintf("rhel:%d", versionSegments[0])
```
luhring, line 52: I like the new shape of this flow, with returning directly once the needed values are known!

luhring (approved):
Nice!

grype/vulnerability/namespace.go
wagoodman: given the variations, this did seems a little easier to read 👍

PR #224, opened 2020-12-11
Decouple from syft objects where possible

### LINKED ISSUE #220: Decouple grype from syft-specific data structures
The matching process does not need a lot of information as input. Today we not only accept syft JSON as input, but also require all of the data it produces to populate the data structures.

It would be ideal to create adapters for the required syft concepts in grype (such as `pkg.Package`) to separate the concerns of grype and syft as well as enable future efforts to provide the same minimal information necessary to grype without requiring the syft JSON format (e.g. accept cycloneDX as input). 

Note: we should be excluding heavy information in the new adapter, specifically the `pkg.Package.Metadata` field.

This also implies eliminating the usage of syft data structures where they are not necessary (such as `pkg.Catalog`). 

Note: we should still allow for the referencing of syft constants or enum-like variable/constants (such as `pkg.Type` and their static definition in syft) --it is probably not worth decoupling from theses basic value definitions.

wagoodman:
This minimizes the required document input shape to the minimal amount of information and opens the door to accepting alternative SBOM documents as input.

- Updates to syft v0.10.0 (changes JSON shape and requires cycloneDX updates)
- Removes use of syft `pkg.Catalog` (except for tests)
- Add `pkg` to adapt the syft `pkg.Package` and minimal necessary metadata (select fields from dpkg, rpm, java)
- Removes `grype.Catalog` and replaces with a `pkg.Provider` abstraction for sourcing packages generically
- Removes requirement to hash the DB on grype startup, saving ~1 second on startup (for me).

Closes #220


grype/matcher/rpmdb/matcher.go
```diff
@@ -19,15 +21,15 @@ var rpmPackageNamePattern = regexp.MustCompile(`(?P<name>^[a-zA-Z0-9\-]+)-\d+\.`
 type Matcher struct {
 }
 
-func (m *Matcher) PackageTypes() []pkg.Type {
-	return []pkg.Type{pkg.RpmPkg}
+func (m *Matcher) PackageTypes() []syftPkg.Type {
+	return []syftPkg.Type{syftPkg.RpmPkg}
```
luhring, line 25: (for later) It might be nice to have a `ToSlice` kind of helper method on Syft's `Pkg.Type` type

grype/pkg/package.go
```diff
@@ -0,0 +1,106 @@
...
+	id        ID
+	Name      string            // the package name
+	Version   string            // the version of the package
+	Locations []source.Location // the locations that lead to the discovery of this package (note: this is not necessarily the locations that make up this package)
+	Language  pkg.Language      // the language ecosystem this package belongs to (e.g. JavaScript, Python, etc)
+	Licenses  []string
+	Type      pkg.Type    // the package type (e.g. Npm, Yarn, Python, Rpm, Deb, etc)
+	CPEs      []pkg.CPE   // all possible Common Platform Enumerators
+	PURL      string      // the Package URL (see https://github.com/package-url/purl-spec)
+	Metadata  interface{} // This is NOT the syft metadata! Only the select data needed for vulnerability matching
```
luhring, line 25: Very helpful comment!

grype/pkg/syft_json_provider.go
```diff
@@ -0,0 +1,213 @@
...
+		// the caller is explicitly hinting to use the given reader as input
+		reader = config.reader
+	} else {
+		// try and get a reader from the description
+		if strings.HasPrefix(config.userInput, "sbom:") {
+			// the user has explicitly hinted this is an sbom, if there is an issue return the error
+			filepath := strings.TrimPrefix(config.userInput, "sbom:")
+			sbomReader, err := os.Open(filepath)
+			if err != nil {
+				return nil, Context{}, fmt.Errorf("user hinted 'sbom:' but could read SBOM file: %w", err)
```
luhring, line 30: Should this be "but **_couldn't_** read ..."?

grype/pkg/syft_json_provider.go
```diff
@@ -0,0 +1,213 @@
...
+	Metadata     interface{}      `json:"metadata"`
+}
+
+// packageMetadataUnpacker is all values needed from Package to disambiguate ambiguous fields during json unmarshaling.
+type packageMetadataUnpacker struct {
+	MetadataType pkg.MetadataType `json:"metadataType"`
+	Metadata     json.RawMessage  `json:"metadata"`
+}
+
+// JavaMetadata encapsulates all Java ecosystem metadata for a package as well as an (optional) parent relationship.
```
luhring, line 85: (nit) this and the following doc comments won't be picked up by dev tooling since the names don't match exactly

luhring (approved):

grype/pkg/syft_json_provider.go
wagoodman: lol, woops 👍

grype/pkg/syft_json_provider.go
wagoodman: yup, copy-pasta problem, I'll adjust

PR #226, opened 2020-12-21
actions: alert scan-action when there is a new release of grype

alfredodeza:
I'm not sure how we've debugged these actions before, seems like there is a lot of potential for this to be slightly off and miss out. Throwing it out for ideas/feedback.


luhring (approved):

alfredodeza:
This doesn't look like the right approach anymore. The create-pr action might be better on the scan-action side of things, where a PR can be created (or updated) to include the latest release of grype

PR #228, opened 2021-01-12
Add automatic completion install to brew formula

sagikazarmark:


luhring:
@sagikazarmark This looks great. Thank you!

We've actually been reworking our release pipeline, including how the homebrew formula gets generated and tested. We'll figure out when the best time to merge this is — meaning before or after those pipeline updates land. But this is great, and it'd be great to have this for [Syft](https://github.com/anchore/syft) as well.

wagoodman (approved):
This looks great, thanks!!

PR #234, opened 2021-02-11
Refactor constraint expression parser to allow for quoted versions

wagoodman:
The existing constraint expression parsing assumed that splitting an expression by `||` is good enough to split OR'd expressions, however versions may have these delimiters within quoted text --the same goes for parenthetical statements. This PR rectifies fortifies the existing parser and allows for quoted versions to have arbitrary characters.

### DIFF
--- a/grype/version/constraint_expression.go
+++ b/grype/version/constraint_expression.go
@@ -1,8 +1,10 @@
 package version
 
 import (
+	"bytes"
 	"fmt"
 	"strings"
+	"text/scanner"
 )
 
 type constraintExpression struct {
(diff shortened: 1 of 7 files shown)


grype/version/constraint_unit.go
```diff
@@ -3,53 +3,44 @@ package version
...
-		if err != nil {
-			return nil, fmt.Errorf("unable to parse constraint operator: %+v", err)
-		}
-		result = append(result, constraintUnit{
-			rangeOperator: op,
-			version:       pair["version"],
-		})
+	op, err := parseOperator(match["operator"])
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse constraint operator: %+v", err)
```
alfredodeza, line 38: which operator? is it possible to add `match["operator"]` here?

grype/version/constraint_unit.go
```diff
@@ -3,53 +3,44 @@ package version
...
+	"strconv"
 	"strings"
+
+	"github.com/anchore/grype/internal"
 )
 
 // operator group only matches on range operators (GT, LT, GTE, LTE, E)
 // version group matches on everything except for whitespace and operators (range or boolean)
-var constraintPartPattern = regexp.MustCompile(`(?P<operator>[><=]*)\s*(?P<version>[^<>=\s,|]+)`)
+var constraintPartPattern = regexp.MustCompile(`\s*(?P<operator>[><=]*)\s*(?P<version>.+)`)
```
alfredodeza, line 14: not sure I follow why we allow any number of leading whitespaces

grype/version/constraint_expression.go
```diff
@@ -61,3 +71,49 @@ func (c *constraintExpression) satisfied(other *Version) (bool, error) {
...
+			captureAndGroup()
+		case currentToken == "(" || currentToken == ")":
+			return nil, fmt.Errorf("parenthetical expressions are not supported yet")
+		case currentToken != "|":
+			buf.Write([]byte(currentToken))
+		}
+		lastToken = currentToken
+		tokenRune = scnr.Scan()
+	}
+	captureVersionOperatorPair()
```
alfredodeza, line 115: I haven't seen this anonymous function usage before, and I'm having trouble following what is doing. Is it any different from adding the `if` conditions from the anonymous functions here?

grype/version/constraint_unit.go
wagoodman: I changed the expectations of what input can be given to constraintUnit, previously is was not possible to provide input with leading spaces, now it is (there is a matching test case to cover this as well).

grype/version/constraint_expression.go
wagoodman: indeed, no different that what you mentioned... it was to save on code repetition was all. For these two lines in particular it is possible for the loop to have been processing a token (since there will not necessarily be trailing `,` or `||` tokens) so this is to capture whatever is left in the buffer and the last `and` group into the final results.

alfredodeza: Ahhh of course! that makes sense. Thanks for clarifying this, I'd not seen this pattern before in Go.

alfredodeza (approved):

PR #241, opened 2021-03-02
Fix intake of SBOM from JSON

### LINKED ISSUE #235: Can't use syft JSON output as input
I'm seeing a set of related error cases when trying to have Grype read Syft's JSON output as its input (in lieu of Grype invoking Syft to catalog an image or directory).

As an initial setup, from `~/Desktop`, I ran `syft -o json alpine:latest > ./alpine.json`. This produced a 39KB JSON file that looked like what I'd expect.

Here are the cases I tried, all of which I'd expect to result in a successful vulnerability analysis from Grype:

**A) Piping to STDIN:**
```
$ cat ./alpine.json | grype
ERROR failed to catalog: unable to process input for scanning: ''
```

**B) Specifying an absolute path:**
```
$ grype sbom:/Users/dan/Desktop/alpine.json
ERROR failed to catalog: unable to process input for scanning: 'sbom:/Users/dan/Desktop/alpine.json'
```

**C) Specifying a path with tilde for the home directory:**
```
$ grype sbom:~/Desktop/alpine.json
ERROR failed to catalog: user hinted 'sbom:' but couldn't read SBOM file: open ~/Desktop/alpine.json: no such file or directory
```

**D) Specifying a relative path explicitly from the current directory:**
```
$ grype sbom:./alpine.json
ERROR failed to catalog: unable to process input for scanning: 'sbom:./alpine.json'
```

**E) Specifying a relative path implicitly from the current directory:**
```
$ grype sbom:alpine.json
ERROR failed to catalog: could not fetch image 'sbom:alpine.json': pull failed: Error response from daemon: pull access denied for sbom, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```

luhring:
This PR:

- fixes #235
- decomposes Syft provider logic into discrete tasks

### Notes:

This PR also adds the beginnings of "CLI-level" tests — to ensure that the built binary behaves correctly from a CLI context, such as the handling of data piped to the command's STDIN.

What's done so far:

- introduced a new "cli" make target to run these tests
- introduces tests to cover the cases listed in #235 (with the exception of `~` home directory processing)
- target snapshot builds for CLI tests
- included these tests in the CI pipeline

What's left for consideration:

- factoring out generalized CLI test helpers to a new directory or project

### DIFF
--- a/grype/pkg/provider.go
+++ b/grype/pkg/provider.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 
 	"github.com/anchore/grype/internal"
@@ -15,7 +16,7 @@ var errDoesNotProvide = fmt.Errorf("cannot provide packages from the given sourc
 
 type providerConfig struct {
 	userInput string
-	scopeOpt  *source.Scope
+	scopeOpt  source.Scope
 	reader    io.Reader
 }
 
@@ -28,29 +29,43 @@ func Provide(userInput string, scopeOpt source.Scope) ([]Package, Context, error
 		syftProvider, // important: we should try syft last
 	}
 
-	// aggregate stdin into a buffer that can be used across multiple providers dynamically
-	var previousStdin bytes.Buffer
+	// capture stdin bytes, so they can be used across multiple providers
+	capturedStdin := bytesFromStdin()
 
-	for _, p := range providers {
-		cfg := providerConfig{
-			userInput: userInput,
-			scopeOpt:  &scopeOpt,
-		}
+	for _, provide := range providers {
+		config := determineProviderConfig(userInput, scopeOpt, capturedStdin)
 
-		if internal.IsPipedInput() && userInput == "" {
-			// this is a hint to all providers that there is already a reader available, don't try and derive one from user input
-			// however, the user input may be useful in situations where the reader isn't provided
-
-			// this reader is a combination of previous bytes read from stdin by other providers as well as what still is
-			// available from stdin. The Tee reader is to ensure that any read bytes from stdin are preserved.
-			cfg.reader = io.MultiReader(&previousStdin, io.TeeReader(os.Stdin, &previousStdin))
-		}
-
-		packages, ctx, err := p(cfg)
+		packages, ctx, err := provide(config)
 		if !errors.Is(err, errDoesNotProvide) {
 			return packages, ctx, err
 		}
 	}
 
 	return nil, Context{}, errDoesNotProvide
+}
+
+func bytesFromStdin() []byte {
+	if internal.IsPipedInput() {
+		capturedStdin, err := ioutil.ReadAll(os.Stdin)
+		if err != nil {
+			return nil
+		}
+
+		return capturedStdin
+	}
+
+	return nil
+}
+
+func determineProviderConfig(userInput string, scopeOpt source.Scope, stdin []byte) providerConfig {
+	config := providerConfig{
+		userInput: userInput,
+		scopeOpt:  scopeOpt,
+	}
+
+	if len(stdin) > 0 {
+		config.reader = bytes.NewReader(stdin)
+	}
+
+	return config
 }
--- a/grype/pkg/syft_json_provider.go
+++ b/grype/pkg/syft_json_provider.go
@@ -2,11 +2,12 @@ package pkg
 
 import (
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io"
 	"os"
 	"strings"
+
+	"github.com/mitchellh/go-homedir"
 
 	"github.com/anchore/grype/grype/cpe"
 	"github.com/anchore/syft/syft/distro"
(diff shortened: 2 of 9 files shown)


test/cli/sbom_input_test.go
```diff
@@ -0,0 +1,130 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	cases := []struct {
+		name string
+		path string
+	}{
+		{
+			"absolute path",
```
alfredodeza, line 29: I wasn't aware that you could define structs without their field names. When is this style preferred?

grype/pkg/syft_json_provider.go
```diff
@@ -160,10 +185,6 @@ func (p *partialSyftPackage) UnmarshalJSON(b []byte) error {
 			PomGroupID:    group,
 			ManifestName:  name,
 		}
-	case "":
```
alfredodeza, line 163: was this part of the issue as well?

grype/pkg/syft_json_provider.go
```diff
@@ -20,33 +21,57 @@ import (
...
+	if err == nil {
+		return sbom, nil
+	}
+
+	// no usable SBOM is available
+	return nil, errDoesNotProvide
+}
+
+func openSbom(path string) (*os.File, error) {
+	expandedPath, err := homedir.Expand(path)
```
alfredodeza, line 60: more of an open-ended question, but how does this expand when using `sudo`? is it acceptable to expand to `/root` instead of `/home/myusername` ?

grype/pkg/syft_json_provider.go
```diff
@@ -20,33 +21,57 @@ import (
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to open SBOM: %w", err)
```
alfredodeza, line 62: I think this error should be specific to expanding the path, not to open the SBOM

test/cli/sbom_input_test.go
luhring: I don't think I have a good answer to that question offhand. In this case, I had two reasons for this syntax, 1) the field names appeared immediately before the data (in the struct type literal), and 2) in my editor (GoLand), the field names actually do appear as hints next to the field values.

Having said all that, I don't mind changing to include field names 😃

grype/pkg/syft_json_provider.go
luhring: Yes — I forget the exact execution path that was causing trouble, but I can dig that up if you'd like

grype/pkg/syft_json_provider.go
luhring: Interesting question... I would guess so. The homedir library tries its best to rely on the OS's own behavior for "home directory" lookup. (E.g. see here -> https://github.com/mitchellh/go-homedir/blob/af06845cf3004701891bf4fdb884bfe4920b3727/homedir.go#L89)

grype/pkg/syft_json_provider.go
luhring: I can make that change. 👍 

This is one area I think it'd be good to seek alignment on among the group. Errors are interesting... (especially in Go). If function A calls function B, and thus has to handle getting an error back from function B, which function reports that there was a problem with function B's responsibility? Lately, I've been preferring function B to do this, since that consolidates the message to function B, and any number of callers of function B can rely on this error message rather than have them all say something like "there was a problem with function B" in their own function bodies.

In this specific case, if you look into the code for the function `homedir.Expand(path)`, you'll notice it returns errors like `"cannot expand user-specific home dir"`. So in my function, `openSbom`, rather than `fmt.Errorf("error expanding the home dir in the path": %w", err)`, which would produce:

```
ERROR: error expanding the home dir in the path: cannot expand user-specific home dir
```

I decided to have _my_ error reflect the failure in _my function's_ job, which is to open an SBOM.

Again, happy to change it, but just wanted to share those thoughts!

grype/pkg/syft_json_provider.go
alfredodeza: This can be tricky in system calls using shell and interpreting back into code (I've run into issues like this in Python). Just tried it out with a quick test and it checks out. Thanks for the link the source code!

grype/pkg/syft_json_provider.go
alfredodeza: That explanation makes sense to me, and I agree with all your points. The caller of this function mentions something similar which seemed repetitive:

```go
    return nil, fmt.Errorf("unable to use specified SBOM: %w", err)
```

Re-using a bit your example, that would mean it would produce this error:

```
ERROR: unable to use specified SBOM: unable to open SBOM: cannot expand user-specific home dir
```

This reads somewhat repetitive. I don't have a good suggestion or alternative here, but that repetition is what made me think that perhaps the messaging should change. 

However! I'm fine with whatever you decide, this is not a hard comment :)

luhring: That's an interesting point. I'll think about that. I think my thinking was that there could be multiple reasons why an SBOM is unusable (e.g. invalid JSON syntax). `getSyftJSON` doesn't care why the SBOM is unusable, it just needs to know that it's not usable. The inner error `unable to open SBOM...` in `openSbom` is a specific reason why the SBOM was unusable — specifically that the SBOM couldn't be opened.

alfredodeza (approved):

wagoodman (approved):
nice addition 💯

grype/pkg/provider.go
```diff
@@ -28,29 +29,43 @@ func Provide(userInput string, scopeOpt source.Scope) ([]Package, Context, error
...
-			scopeOpt:  &scopeOpt,
-		}
-
-		if internal.IsPipedInput() && userInput == "" {
-			// this is a hint to all providers that there is already a reader available, don't try and derive one from user input
-			// however, the user input may be useful in situations where the reader isn't provided
-
-			// this reader is a combination of previous bytes read from stdin by other providers as well as what still is
-			// available from stdin. The Tee reader is to ensure that any read bytes from stdin are preserved.
-			cfg.reader = io.MultiReader(&previousStdin, io.TeeReader(os.Stdin, &previousStdin))
```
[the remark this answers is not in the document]
luhring: The main goal here was to simplify. To be honest, I was having trouble keeping track of the logic, which was notable to me because the lack of availability of SBOM JSON readers was central to these bugs. For now, we only have one provider that makes use of the stdin bytes reader, so the `multireader` + `teereader` weren't being leveraged beyond providing an `io.Reader` for stdin. I could see this changing if we see a need for more providers, and at that point, it'd be good to reassess the symmetry we're seeing among the providers.

If there's a more straightforward way to improve this code, I'd welcome that!

Issue #242, opened 2021-03-04
remove v1-based package


alfredodeza:
So that we normalize on a single import path everywhere. This change will allow to not depend on the schema version for the import path. 

This follows the following issue from grype-db: https://github.com/anchore/grype-db/issues/13

PR #243, opened 2021-03-04
refactor v1-based imports to db

alfredodeza:
Requires https://github.com/anchore/grype-db/pull/14 to be available first, so that the dependency can be bumped.

Fixes https://github.com/anchore/grype/issues/242


alfredodeza:
Tests are expected to break until the go module gets updated to point to a post-refactor grype-db

grype/presenter/json/presenter_test.go
```diff
@@ -188,7 +188,7 @@ func TestJsonImgsPresenter(t *testing.T) {
 	}
 
 	// TODO: add me back in when there is a JSON schema
-	// validateAgainstV1Schema(t, string(actual))
+	// validateAgainstdbSchema(t, string(actual))
```
luhring, line 191: nit: I'd suggest capitalizing at least the "D" in "db" to make it more identifiable as a distinct word from "Against"

luhring (approved):

grype/presenter/json/presenter_test.go
alfredodeza: I blame my search and replace regex :D will fix

alfredodeza: This has been fixed... with an improved and polished regular expression

Makefile
```diff
@@ -100,11 +100,6 @@ lint: ## Run gofmt + golangci lint checks
 	$(eval MALFORMED_FILENAMES := $(shell find . | grep -e ':'))
 	@bash -c "[[ '$(MALFORMED_FILENAMES)' == '' ]] || (printf '\nfound unsupported filename characters:\n$(MALFORMED_FILENAMES)\n\n' && false)"
 
-.PHONY: validate-schema
-validate-schema:
-	# ensure the codebase is only referencing a single grype-db schema version, multiple is not allowed
-	python test/validate_schema.py
```
alfredodeza, line 106: @wagoodman besides the simple search and replace, this is part that got removed. There is no need to check if we are importing multiple schemas. Wanted to make sure this was visible to you before merging

wagoodman: yup yup, that's correct 👍

wagoodman (approved):

PR #247, opened 2021-03-15
Suppress version constraint token scanner errors

wagoodman:
The `text/scanner` is used for constraint expression tokenization. If no error handler is provided to the scanner then errors are reported to stdout by default. This PR adds a handler which ignores all scanner tokenization errors, which is safe under the current context (see the added comments for details). 

Closes #241

### DIFF
--- a/grype/version/constraint_expression.go
+++ b/grype/version/constraint_expression.go
@@ -95,6 +95,20 @@ func scanExpression(phrase string) ([][]string, error) {
 	}
 
 	scnr.Init(strings.NewReader(phrase))
+
+	scnr.Error = func(*scanner.Scanner, string) {
+		// scanner has the ability to invoke a callback upon tokenization errors. By default, if no handler is provided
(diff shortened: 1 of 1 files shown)


luhring (approved):
Weird af!

alfredodeza:
I'm guessing there is no easy way to test this one out right? Otherwise, I would love to catch this with a test, brownie points if it can catch _other_ unexpected output as well as the one this fixes.

wagoodman:
@alfredodeza I thought about what kind of test I would write and how to write it. It can't easily be done in unit testing without affecting other tests, so that's not an option. Adding a new CLI-level test to account for it was something I considered, but felt it to be a little heavy handed for testing if the stdlib scanner configuration was correct (make certain this specific error text does not appear).

I think we can add a more generic test like you described last (check that output matches the given output, fail if anything else is given), but can be added in a follow up PR as a CLI test to verify grype scan output for multiple test-fixtures. That test would add a lot of value outside of this specific behavior being fixed on this PR.

PR #261, opened 2021-03-25
Move docker login after keychain creation

wagoodman:
Currently the release step logs into docker, however, this needs to occur after creating the keychain used for mac signing.


Btodhunter (approved):
lgtm

PR #284, opened 2021-04-09
Add support for templated output

### LINKED ISSUE #251: Allow user-defined output formats
It would be great to allow user-specified output formats (such as by templating) so that Grype could support arbitrary text formats.

This would mean that Grype could immediately support downstream consumers' needs, even if no presenter specific to that consumer's input format had been developed and incorporated into the Grype code.

Notably, this would especially help with Grype integrations into other systems, which would enable more native workflows that actively depend on Grype.

(This is also related to https://github.com/anchore/syft/issues/152)

luhring:
Closes #251 

For the user, this PR:
- Adds a new "template" output format, using a user-supplied Go template (via `-t <path/to/template-file>`)

For the developer, this PR:
- moves more "presenter" concerns into the "presenter domain" (its package and sub-packages), which:
  - reduces the number of interactions between the `cmd` package and the `presenter` package
  - allows more presenter symbols to be un-exported
- creates a native presenter "config" object, which is validated within the presenter domain
- introduces a `models` package within the `presenter` package, for consolidating data models shared by multiple presenters
- consolidates test data generation code that was repeated in tests for multiple presenter implementations into the `models` package
- changes the presenter flow such that presenters don't depend on a particular data structure for representing app config


go.mod
```diff
@@ -30,6 +31,7 @@ require (
 	github.com/spf13/cobra v1.0.1-0.20200909172742-8a63648dd905
 	github.com/spf13/viper v1.7.0
 	github.com/stretchr/testify v1.6.0
+	github.com/vbatts/tar-split v0.11.1 // indirect
```
wagoodman, line 34: I don't see this being used anywhere, was this intentional?

grype/presenter/config.go
```diff
@@ -0,0 +1,44 @@
...
+			AvailableFormats)
+	}
+
+	if format == templateFormat {
+		if outputTemplateFile == "" {
+			return Config{}, fmt.Errorf("must specify path to template file when using %q output format",
+				templateFormat)
+		}
+
+		// TODO: Should we also validate access to the template file and the template's syntax, too?
```
wagoodman, line 27: we can add this one day; I think it's acceptable to get this feature in without requiring up-front validation

grype/presenter/config.go
```diff
@@ -0,0 +1,44 @@
+package presenter
+
+import "fmt"
+
+// Config is the presenter domain's configuration data structure.
+type Config struct {
```
wagoodman, line 6: open question: do we want to also include the analysis result data on the config object (matches, image source metadata, etc)? (that would be my preference at a casual glance)

grype/presenter/config.go
```diff
@@ -0,0 +1,44 @@
...
+	format           format
+	templateFilePath string
+}
+
+// ValidatedConfig returns a new, validated presenter.Config. If a valid Config cannot be created using the given input,
+// an error is returned.
+func ValidatedConfig(output, outputTemplateFile string) (Config, error) {
```
wagoodman, line 13: nit/question: why not name this `NewConfig` instead?

grype/presenter/config.go
wagoodman: we chatted about this offline --we can explore this suggestion later 👍

wagoodman (approved):

go.mod
luhring: This was _not_ intentional! Good catch!

grype/presenter/config.go
luhring: Okay, sounds good

grype/presenter/config.go
luhring: Sweet 👍

grype/presenter/config.go
luhring: I wanted to signal that config validation was occurring here, as opposed to a more transparent constructor whose job is "create an object and use these values". I started with `NewConfig` but wanted to express to the consumer that this function tackles a noteworthy step in the flow of config into the domain. That being said, I'm not married to it. Are there reasons for `NewConfig` we should consider here?

grype/presenter/presenter.go
```diff
@@ -19,14 +19,17 @@ type Presenter interface {
...
-	case TablePresenter:
+func GetPresenter(presenterConfig Config, matches match.Matches, packages []pkg.Package, context pkg.Context,
+	metadataProvider vulnerability.MetadataProvider, appConfig interface{}) Presenter {
+	switch presenterConfig.format {
+	case jsonFormat:
+		return json.NewPresenter(matches, packages, context, metadataProvider, appConfig)
+	case tableFormat:
 		return table.NewPresenter(matches, packages, metadataProvider)
-	case CycloneDxPresenter:
+	case cycloneDXFormat:
```
alfredodeza, line 29: good change

alfredodeza (changes requested):

grype/presenter/config.go
```diff
@@ -0,0 +1,44 @@
...
+	format := parse(output)
+
+	if format == unknownFormat {
+		return Config{}, fmt.Errorf("unsupported output format %q, supported formats: %+v", output,
```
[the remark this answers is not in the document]
luhring: 👍

README.md
```diff
@@ -60,9 +60,26 @@ grype <image> -o <format>
...
+
+- Define your format as a [Go text template](https://golang.org/pkg/text/template/), and save this template as a file.
+
+- Set the output format to "template" (`-o template`). 
+
+- Specify the path to the template file (`-t ./path/to/custom.template`).
+
+- Grype's template processing uses the same data models as the `json` output format — so if you're wondering what data is available as you author a template, you can use the output from `grype <image> -o json` as a reference.
+
+**Example:** You could make Grype output data in CSV format by writing a Go template that renders CSV data and then running `grype <image> -o ~/path/to/csv.tmpl`.
```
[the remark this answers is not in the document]
luhring: Great suggestion, adding now

grype/presenter/config.go
```diff
@@ -0,0 +1,44 @@
...
+		// TODO: Should we also validate access to the template file and the template's syntax, too?
+
+		return Config{
+			format:           format,
+			templateFilePath: outputTemplateFile,
+		}, nil
+	}
+
+	if outputTemplateFile != "" {
+		return Config{}, fmt.Errorf("user specified template file %q, but "+
```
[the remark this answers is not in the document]
luhring: Great point, removed

grype/presenter/config.go
alfredodeza: I would think that a constructor can do validation, and not using the "NewXXX" convention would make it harder (although not horribly so) to discover. No strong opinions here, but is seems that @luhring needs some reassurance. I'll defer to @wagoodman

alfredodeza (approved):

PR #285, opened 2021-04-12
NEVER MERGE: concept illustration: adjust code for less concurrency in execution flow

luhring:
☠️  This is a concept illustration branch, not intended to be merged ever. **The only reason this is a PR at all is so I can attach a commentary to a branch!** 🔥 

This branch:

1. Introduces a dedicated `grype.Analysis` data structure to encapsulate data that tends to be passed around in groups of multiple parameters (i.e. matches, packages, context, metadata provider, and app config).

1. Pares back the ETUI's responsibility within the main execution flow, without introducing any user-observable changes in experience*. This results in:
    - a straightforward flow in the root command — preliminary checks, analysis (vulnerability matching), and handling of the analysis (checking if the severity threshold was crossed, and presenting the analysis)
    - the ability for presenters to have access to the original analysis at the time `Present(...)` is called, which impacts the `Presenter` interface and usage semantics
    - a more singular responsibility for ETUI code: handling _dynamic_ updates to the screen (as opposed to _any_ updates to the screen, which would include normal printing of data)

1. Allows there to be _just a single_ interaction between "main" (the root command) and the presenter package — the task of getting a presenter. This can now be done before the analysis data is available, which means that the presenter domain can validate the configuration received from the user and use polymorphism to return a single, abstracted presenter object that the consumer can use for the rest of the app's runtime.

1. Attempts to decompose functionality within the ETUI domain and adjust how concurrent operations are invoked and synchronized. This attempt was necessitated by the need for the ETUI to "yield back" to its consumer, as opposed to finishing out app execution on its own.

1. Looks like a tornado went through the code.

*There seems to be a race condition where output is sent after the frame is closed, which is exposed more readily in the current implementation than it was before.


luhring:
Had a great conversation with @wagoodman. I pushed a new commit that tries out some of the thoughts from that discussion.

One such idea was "separate out _formatting_ from _presenting_"...


**Formatting**: Taking data from the `grype.Analysis` data structure and formatting it using a particular method (e.g. formatting as JSON, or by using a provided template). Note that this kind of work _necessarily_ knows about Grype objects.

**Presenting**: Writing bytes to an arbitrary `io.Writer`, and returning an `error` if there's a problem. Note that this abstraction is _completely generalizable_, and it has nothing to do with Grype types.

Given this separation, a natural pipeline of work can occur:

1a. Determine the format to use (i.e. using configuration from the user).
1b. Meanwhile, generate a Grype analysis (by running the vulnerability matching logic).
2. Create a generalized "presenter"... by applying the _Format_ to the _Analysis_.
3. Pass this Presenter to whomever! They need not know about Grype concerns, and can invoke this functionality at any point, using any `io.Writer`.

Let's look at the code for this.

1a — getting a usable `Format`, via upfront config validation:

```go
format, err := presenter.DetermineFormat(appConfig.Output, appConfig.OutputTemplateFile)
```

1b — getting the actual Grype `Analysis`:

```go
analysis, err := analyzeWithUI(userInput)
```

2 — applying the `Format` to the `Analysis`:

```go
present := presenter.Apply(format, analysis)
```

3 — using the presenter (this can be done by any downstream consumer!):

```go
err = present(os.Stdout)
```

In looking at the code, you'll notice a major shift — this approach favors functions as the abstraction of work, instead of `interface`s as the abstraction of `struct`s. In drafting this, I tried several approaches before this one, and it just became clear that any use of `struct`s to represent steps in this pipeline were contrived and/or not adding value. It turns out that any required state can be encapsulated in a function, which can be passed freely downstream without consequences. In hindsight, it makes sense that as we turn this "presenting" process into more of a pipeline, the individual stages become pure functions.

Another observation about this approach is that we're able to lean more on familiar English meaning — such as by referring to concepts like "JSON" and "cycloneDX" as **_"formats"_**. A format, in essence, is a way for input data to be rendered as output. This turns out to be a nice use of a function! And now, "Presenters" are literally the final step in the pipeline, i.e. the entities that present the formatted analysis to a designated `io.Writer`.

PR #290, opened 2021-04-13
Add ability to pull images directly from a registry

### LINKED ISSUE #264: Ability to pull image directly from a registry (without the Docker daemon)
It would be advantageous if Grype could pull images directly from a container registry, in addition to its existing capabilities (relying on the Docker daemon or finding data locally).

Advantages of this method:

1. Grype can run within a container image and fetch container images without the user needing to expose the Docker daemon to the running container.
2. This can reduce wait times for users since the image would need to be pulled from the registry to the local system, rather than have an additional required step after the pull (i.e. retrieving the image from the Docker daemon). This improvement would be especially noticeable for larger images.

Open questions for consideration:

1. How will Grype handle authentication?
2. How will Grype handle private/custom CAs for registry access?
3. How will we decide when Grype will use this method of pulling an image vs. relying on the Docker daemon?
4. Should we consider any kind of local caching for images pulled directly from a registry?

wagoodman:
Adds the ability to pull images directly from a registry, no container runtime is required:
```
grype registry:reop/myimage:latast
```

Authentication and connection options are configurable under the new `registry` configuration section. A single set of credentials may be provided via environment variables, multiple credentials must be provided via a config file. See the README for the details.

Closes #264


luhring (approved):

test/cli/registry_auth_test.go
```diff
@@ -0,0 +1,78 @@
...
+			name: "fallback to keychain",
+			args: []string{"-vv", "registry:localhost:5000/something:latest"},
+			assertions: []traitAssertion{
+				assertInOutput("source=OciRegistry"),
+				assertInOutput("localhost:5000/something:latest"),
+				assertInOutput("no registry credentials configured, using the default keychain"),
+			},
+		},
+		{
+			name: "use creds",
```
alfredodeza, line 25: Two more cases here where password without username and username without password would be great

test/cli/registry_auth_test.go
```diff
@@ -0,0 +1,78 @@
...
+		t.Run(test.name, func(t *testing.T) {
+			cmd, stdout, stderr := runGrypeCommand(t, test.env, test.args...)
+			for _, traitAssertionFn := range test.assertions {
+				traitAssertionFn(t, stdout, stderr, cmd.ProcessState.ExitCode())
+			}
+			if t.Failed() {
+				t.Log("STDOUT:\n", stdout)
+				t.Log("STDERR:\n", stderr)
+				t.Log("COMMAND:", strings.Join(cmd.Args, " "))
+			}
```
alfredodeza, line 75: Would love to see this normalized somehow. We already have a Go testing repository, maybe we could use that?

test/cli/utils_test.go
```diff
@@ -0,0 +1,118 @@
...
+	cmd := getGrypeCommand(t, args...)
+	if env != nil {
+		cmd.Env = append(cmd.Env, envMapToSlice(env)...)
+	}
+	var stdout, stderr bytes.Buffer
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+
+	// ignore errors since this may be what the test expects
+	cmd.Run()
```
alfredodeza, line 61: Ah... I see, so in regards to my previous comment for normalizing, this could be handled by a wrapper to this function that when the command fails, the reporting is done automatically. Additionally, a separate wrapper function would be one that failure is expected and so it doesn't check `t.Failed()`. Naming is tricky here, but something like: `checkGrypeCommand` when failure shouldn't happen, and keep `runGrypeCommand` when it doesn't matter?

alfredodeza (changes requested):

test/cli/registry_auth_test.go
wagoodman: I can add that on 👍

test/cli/registry_auth_test.go
wagoodman: agreed, we only recently added these patterns, I think we can abstract out some of the common patterns to https://github.com/anchore/go-testutils soon 👍 .

test/cli/utils_test.go
wagoodman: a clarification: `t.Failed()` is with regard to the test status, not the status of the grype process or it's exit code.

PR #293, opened 2021-04-14
Modify registry credential verification

wagoodman:
Enforces that user and password OR a token must be present --no partial values allowed now. (Follow up to https://github.com/anchore/grype/pull/290#discussion_r613207047)

### DIFF
--- a/internal/config/registry.go
+++ b/internal/config/registry.go
@@ -29,7 +29,7 @@ func (cfg *registry) parseConfigValues() {
 		os.Getenv("GRYPE_REGISTRY_AUTH_PASSWORD"),
 		os.Getenv("GRYPE_REGISTRY_AUTH_TOKEN")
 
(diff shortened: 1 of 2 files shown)


internal/config/registry.go
```diff
@@ -42,8 +42,8 @@ func (cfg *registry) parseConfigValues() {
 	}
 }
 
-func hasNonEmptyCredentials(authority, password, token string) bool {
-	return authority != "" && password != "" || authority != "" && token != ""
+func hasNonEmptyCredentials(authority, username, password, token string) bool {
+	return authority != "" && password != "" && username != "" || authority != "" && token != ""
```
alfredodeza, line 46: this is great, I didn't realize that by spotting some extra test cases it would surface this as a needed change, thanks for addressing it

alfredodeza (approved):

PR #294, opened 2021-04-14
Safely join paths derived from tar headers

wagoodman:
We should not trust values derived from tar headers when crafting file paths. This adjusts the tar processing for DB import / update to consider this and error out when this case is found.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ RESET := $(shell tput -T linux sgr0)
 TITLE := $(BOLD)$(PURPLE)
 SUCCESS := $(BOLD)$(GREEN)
 # the quality gate lower threshold for unit test total % coverage (by function statements)
-COVERAGE_THRESHOLD := 50
+COVERAGE_THRESHOLD := 47
 BOOTSTRAP_CACHE="c7afb99ad"
 
 ## Build variables
--- a/internal/file/tar.go
+++ b/internal/file/tar.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strings"
 )
 
 const (
@@ -15,6 +16,29 @@ const (
 	MB
 	GB
 )
+
+type errZipSlipDetected struct {
+	Prefix   string
+	JoinArgs []string
+}
+
+func (e *errZipSlipDetected) Error() string {
+	return fmt.Sprintf("potential zip slip attack: prefix=%q dest=%q", e.Prefix, e.JoinArgs)
+}
+
+// safeJoin ensures that any destinations do not resolve to a path above the prefix path.
+func safeJoin(prefix string, dest ...string) (string, error) {
+	joinResult := filepath.Join(append([]string{prefix}, dest...)...)
+	cleanJoinResult := filepath.Clean(joinResult)
+	if !strings.HasPrefix(cleanJoinResult, filepath.Clean(prefix)) {
+		return "", &errZipSlipDetected{
+			Prefix:   prefix,
+			JoinArgs: dest,
+		}
+	}
+	// why not return the clean path? the called may not be expected it from what should only be a join operation.
+	return joinResult, nil
+}
 
 func UnTarGz(dst string, r io.Reader) error {
 	gzr, err := gzip.NewReader(r)
@@ -39,7 +63,10 @@ func UnTarGz(dst string, r io.Reader) error {
 			continue
 		}
 
-		target := filepath.Join(dst, header.Name)
+		target, err := safeJoin(dst, header.Name)
+		if err != nil {
+			return err
+		}
 
 		switch header.Typeflag {
 		case tar.TypeDir:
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -14,7 +14,7 @@ RESET := $(shell tput -T linux sgr0)
 TITLE := $(BOLD)$(PURPLE)
 SUCCESS := $(BOLD)$(GREEN)
 # the quality gate lower threshold for unit test total % coverage (by function statements)
-COVERAGE_THRESHOLD := 50
+COVERAGE_THRESHOLD := 47
```
alfredodeza, line 17: this is surprising, that even after adding tests it is dropping

internal/file/tar_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// looks like there isn't a helper for this yet? https://github.com/stretchr/testify/issues/497
```
alfredodeza, line 11: a missed opportunity from testify

internal/file/tar.go
```diff
@@ -16,6 +17,29 @@ const (
 	GB
 )
 
+type errZipSlipDetected struct {
+	Prefix   string
+	JoinArgs []string
+}
+
+func (e *errZipSlipDetected) Error() string {
+	return fmt.Sprintf("potential zip slip attack: prefix=%q dest=%q", e.Prefix, e.JoinArgs)
```
alfredodeza, line 26: Can this be changed to say what isn't allowed instead? I understand that this addresses the zip slip attack, but it assumes I grasp exactly what that is and how to fix it. For example:
```suggestion
	return fmt.Sprintf("paths are not allowed to resolve outside their prefix (%q). Destination: %q", e.Prefix, e.JoinArgs)
```

Probably butchering the message but anything that clarifies here would be great

internal/file/tar_test.go
```diff
@@ -0,0 +1,85 @@
...
+		{
+			prefix: "/a/place",
+			args: []string{
+				"somewhere/../else",
+			},
+			expected:     "/a/place/else",
+			errAssertion: assert.NoError,
+		},
+		{
+			prefix: "/a/../place",
```
alfredodeza, line 43: ah nice, it handles this case which the suggestion in the alert didn't

alfredodeza (changes requested):

internal/file/tar_test.go
luhring: I followed a chain of linked issues/PRs, and it looks like maybe they have actually added this now? https://pkg.go.dev/github.com/stretchr/testify@v1.7.0/assert#ErrorAs

Makefile
wagoodman: the problem is that the go tooling does not count packages that have 0 tests. I added tests to a package that previously had no tests, so now the rest of the package is being counted. So though coverage is technically going up, it's being counted down :( (this is odd, but has happened a couple times before as well).

internal/file/tar_test.go
wagoodman: nice! I just need to update testify.... good find 👍

wagoodman: interesting, using it seems to be perplexing though:
```
=== RUN   TestSafeJoin_PositiveCases//a/place:[../_../]
    tar_test.go:86: 
        	Error Trace:	tar_test.go:86
        	Error:      	Should be in error chain:
        	            	expected: "paths are not allowed to resolve outside their prefix (\"/a/place\"). Destination: [\"../\" \"../\"]"
        	            	in chain: "paths are not allowed to resolve outside their prefix (\"/a/place\"). Destination: [\"../\" \"../\"]"
        	Test:       	TestSafeJoin_PositiveCases//a/place:[../_../]
```

shout out if you see the difference? (I don't see it)

In the meantime I'm going to leave the existing helper since it's generating an assertion which helps with the test table setup anyway.

luhring: @wagoodman Yeah, I think you're right about the usefulness in the assertion helper function. In case it's useful, here's one implementation of that function that seems to work well:

```go
func assertErrorIsZipSlipDetected(t assert.TestingT, actualErr error, _ ...interface{}) bool {
	var expected *errZipSlipDetected
	return assert.ErrorAs(t, actualErr, &expected)
}
```

(and then down in the test case you'd see this...)

```go
{
// ...
	expected:     "",
	errAssertion: assertErrorIsZipSlipDetected,
},
```

wagoodman: awesome! I think I'll leave the existing variant in place, but your reference implementation here shows what I did wrong (I passed the address for a value, not a pointer type). 🦖 🌮

luhring: 🦖 🌮

alfredodeza (approved):
Looks good to me, but I defer to @luhring for a final pass

luhring:
@alfredodeza Thanks — this looks good to me. I'd probably remove this comment

> looks like there isn't a helper for this yet? https://github.com/stretchr/testify/issues/497

given what we know now, and consider using the testify assertion even w/ the wrapping function @wagoodman has. But this isn't urgent.

PR #299, opened 2021-04-15
Add grype db namespace indication in match details

### LINKED ISSUE #283: Expose the explicit record source for each match
Today there is a `RecordSource` field in grype-db which should be exposed in the match results to help identify exactly which FeedService FeedGroup the vulnerability data was derived from.

wagoodman:
Adds a new `grypeDbNamespace` field to the match details indicating the namespace within grype-db that the match was found in.

Closes #283

### DIFF
--- a/grype/matcher/common/cpe_matchers.go
+++ b/grype/matcher/common/cpe_matchers.go
@@ -58,8 +58,9 @@ func FindMatchesByPackageCPE(store vulnerability.ProviderByCPE, p pkg.Package, u
 						"cpe": cpe.BindToFmtString(),
 					},
 					SearchMatches: map[string]interface{}{
-						"cpes":       cpesToString(vuln.CPEs),
-						"constraint": vuln.Constraint.String(),
+						"grypeDbNamespace":  vuln.Namespace,
+						"cpes":              cpesToString(vuln.CPEs),
+						"versionConstraint": vuln.Constraint.String(),
 					},
 				})
 			}
--- a/grype/matcher/common/cpe_matchers.go
+++ b/grype/matcher/common/cpe_matchers.go
@@ -58,7 +58,7 @@ func FindMatchesByPackageCPE(store vulnerability.ProviderByCPE, p pkg.Package, u
 						"cpe": cpe.BindToFmtString(),
 					},
 					SearchMatches: map[string]interface{}{
-						"grypeDbNamespace":  vuln.Namespace,
+						"namespace":         vuln.Namespace,
 						"cpes":              cpesToString(vuln.CPEs),
 						"versionConstraint": vuln.Constraint.String(),
 					},
--- a/grype/matcher/common/utils_test.go
+++ b/grype/matcher/common/utils_test.go
@@ -1,0 +1,33 @@
+package common
+
+import (
+	"testing"
+
+	"github.com/anchore/grype/grype/match"
+	"github.com/anchore/grype/grype/vulnerability"
+	"github.com/go-test/deep"
+	"github.com/scylladb/go-set"
+	"github.com/stretchr/testify/assert"
+)
+
+func assertMatchesWithoutVulnData(t testing.TB, expected, actual []match.Match) {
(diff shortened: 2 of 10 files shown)


grype/matcher/common/cpe_matchers.go
```diff
@@ -58,8 +58,9 @@ func FindMatchesByPackageCPE(store vulnerability.ProviderByCPE, p pkg.Package, u
 						"cpe": cpe.BindToFmtString(),
 					},
 					SearchMatches: map[string]interface{}{
-						"cpes":       cpesToString(vuln.CPEs),
-						"constraint": vuln.Constraint.String(),
+						"grypeDbNamespace":  vuln.Namespace,
```
alfredodeza, line 61: technically this is not a namespace from grype DB though right? Can it be `Namespace` instead? or what is the benefit of mentioning grype-db?

wagoodman: I originally was going to start with `namespace` then realized that it's not clear what this is referring to... so I clarified it by adding `grypeDb` in front. I'm alright with `namespace`? I  didn't feel very strongly either way.

grype/vulnerability/vulnerability.go
```diff
@@ -13,6 +13,7 @@ type Vulnerability struct {
 	CPEs           []pkg.CPE
 	ID             string
 	RecordSource   string
+	Namespace      string
```
alfredodeza, line 16: Since I saw `Namespace` here, I assumed the rest would follow with the same convention. My preference would be to use `Namespace` throughout, but if we are not, then normalize for `grypeDbNamespace`. Keeping both reduces clarity

wagoodman: renamed back to `namespace` 👍

grype/matcher/common/utils_test.go
```diff
@@ -0,0 +1,33 @@
...
+	"testing"
+
+	"github.com/anchore/grype/grype/match"
+	"github.com/anchore/grype/grype/vulnerability"
+	"github.com/go-test/deep"
+	"github.com/scylladb/go-set"
+	"github.com/stretchr/testify/assert"
+)
+
+func assertMatchesWithoutVulnData(t testing.TB, expected, actual []match.Match) {
```
luhring, line 13: This helper function looks really useful. It took me a while to figure out what it does and how (still chewing on it actually). I have a couple of thoughts on how to improve that, but I'm open to discussion/suggestions!

First, it looks like this is comparing matches, including a single field of vulnerability data: ID. It would be ideal to find a name that reflects that. Perhaps something like `assertMatchesUsingIDsForVulnerabilities`. I don't love it, but the idea is to describe how vuln data is being compared within the matches (e.g., instead of signaling that vuln data is being disregarded altogether in the assertion).

Second, the implementation is taking me a while to wrap my head around. One question I still have is: what's being caught in the stringset comparison that's not caught in the earlier diffing? Another question I have is: does the order of the actual slice of matches matter? Looking at the `for` loop, my understanding is leaning toward yes. If that's true, would something like this work as the function body? (And if this is insufficient, it'd help to better understand why)

```go
for idx, a := range actual {
	// the vulnerability comparison should use only the vulnerability ID
	a.Vulnerability = vulnerability.Vulnerability{ID: a.Vulnerability.ID}
	assert.Equal(t, expected[idx], a)
}
```

luhring (changes requested):
Looks great in general! One function took me a bit - would be good to discuss strategies for shortening time to comprehension

alfredodeza (approved):
My one comment got addressed. Happy to defer to @luhring here

grype/matcher/common/utils_test.go
wagoodman: this makes sense --except for the deep.Equal replacement, I'll adjust the PR. (I found assert.Equal and deep.Equal were behaving a slightly differently in this context, didn't investigate any further for now)

luhring (approved):
Looks good!

PR #302, opened 2021-04-16
Add DB information to json descriptor block

### LINKED ISSUE #270: Add database information to the JSON output
The JSON `descriptor` section provides contextual information about the tool and its configuration when it was invoked. What it does not provide is details about the DB state (data age, DB checksum, schema version, etc). It would be ideal to capture this information. Note: the engine integration requires a DB checksum.

wagoodman:
Adds a new `db` section to the `descriptor` block in the json output:
```json
 "descriptor": {
  "name": "grype",
  "version": "[not provided]",
  "configuration": {
   ...
  },
  "db": {
   "dataAge": "2021-04-16T08:24:52Z",
   "schemaVersion": 2,
   "location": "/Users/wagoodman/Library/Caches/grype/db/2",
   "checksum": "sha256:5ef1699f8fded28fca78d2e8699754ce2fd3b0a0e2dbb10a80dcb39a461b9521",
   "error": null
  }
 }
}
```

Additionally updates the DB status command to match the same information:
```
$ grype db status
Location:  /Users/wagoodman/Library/Caches/grype/db/2
Built:     2021-04-16 08:24:52 +0000 UTC
Schema:    2
Checksum:  sha256:5ef1699f8fded28fca78d2e8699754ce2fd3b0a0e2dbb10a80dcb39a461b9521
Status:    valid
```

Lastly, the `--schema` option has been removed from the `db status` command since it is no longer used by the nightly DB publishing job.

Closes #270


alfredodeza (approved):

grype/db/status.go
```diff
@@ -3,9 +3,9 @@ package db
 import "time"
 
 type Status struct {
-	Age                   time.Time
-	CurrentSchemaVersion  int
-	RequiredSchemaVersion int
-	Location              string
-	Err                   error
+	Age           time.Time `json:"dataAge"`
```
luhring, line 6: Usually "age" is a duration — in seconds, days, years, etc. It looks like this is showing the date (a point in time) that the database was built, and I see it's referred to with the "Built:" field when running `grype db status`. Could we adjust this name?

PR #303, opened 2021-04-19
Adjust Alpine matching to lean on NVD matches when possible

### LINKED ISSUE #281: Alpine matching should use NVD as primary source
**Context**
Today the alpine matching is roughly this:

1. Find all Alpine SecDB matches for a given package name and version
2. Find all NVD matches for a given package (speculative) CPEs
3. Only keep matches where the CVE is common to both the NVD matches and the Alpine SecDB matches

The primary problem with this is that SecDB is only intended to be used as an indication of when a fix is available and should not be used to identify vulnerable versions (per the Alpine maintainers [reference needed?]).

An additional problem with this approach is that SecDB only tracks incorporated fixes, thus, there is a period of time where there may be a discoverable NVD match available before a fix is ported (and indicated in SecDB) --currently grype is not taking advantage of these potential NVD matches

**Proposal**
Update the APKDB matcher to surface NVD matches and remove matches that have a fixed-in indication for each matched package-version pairing for the discovered CVE.

Specifically update matching to: `set(NVDMatches) - set(SecDBFixedInVersions)`

Also consider a slightly more accurate model: `[set(NVD) | set(SecDBVulnerable)] - set(SecDBFixedInVersions)`

wagoodman:
As suggested in #281 , the matching logic has been adjusted to `[set(NVD) | set(SecDBVulnerable)] - set(SecDBFixedInVersions)`. Specifically:
- If an NVD match is found the record is surfaced if there is no secdb fix found for the same package-vulnerability pairing.
- If a secDB record is found, that record is preferred over the NVD record to preserve fixed-in information.

This PR additionally fixes the behavior of empty version constraints. An empty constraint indicates that any version will satisfy the constraint. During testing of this fix it was found that not all constraint objects behaved as such --this PR added tests to all constraint objects and adjusted where necessary to fix that behavior.

Closes #281


grype/matcher/apk/matcher.go
```diff
@@ -24,36 +24,42 @@ func (m *Matcher) Match(store vulnerability.Provider, d *distro.Distro, p pkg.Pa
...
-		// by this point all matches have been verified to be vulnerable within the given package version relative to the vulnerability source
-		_, ok := secDbCandidates[cve]
-		if ok {
-			// this is a match, use the NVD records as the primary record source (no need to merge)
-			matches = append(matches, cpeCandidatesForCve...)
+	// all secDB matches have been added, new we will add additional unique matches from CPE source (NVD et al.)
+	for id, cpeMatch := range cpeMatchesByID {
+		// by this point all matches have been verified to be vulnerable within the given package version relative to the vulnerability source.
+		// now we will add unique CPE candidates that were not found in secdb.
+		if _, exists := secDbMatchesByID[id]; !exists {
```
zhill, line 60: I may be misreading this, but I don't think this is the correct check for the set logic in the issue description. The condition to add a CPE match is when either a.) there is no CVE/ID in the secDB at all (Alpine makes no statement on the vuln), or b.) there is a vuln record in the secDB but the package is vulnerable.

As I read this section it is filtering out duplicates, but not acting like a filter to remove the false positive condition of a CPE match but a FixVersion record in the SecDB showing that the package was fixed. I think the secDbMatchesByID lookup should be replaced by a function that determines if the VulnID (CVE) is present at all in the SecDB (e.g. a new function `secDbHasVulnID(id)`)

zhill (changes requested):
The filtering logic looks like it removed duplicates rather than the stated set logic in the PR description.

grype/matcher/apk/matcher.go
zhill: Looking at the grype-db store reader interface, the `store.GetVulnerabilityMetadata()` may work for this use, if that doesn't return an entry for the Alpine namespace, then the NVD record should be used.

wagoodman: You're right, this partially covers the set logic, specifically this implements:`set(NVD) | [set(SecDBVulnerable) - set(SecDBFixedInVersions)]` . I see how to adjust to `[set(NVD) - set(SecDBFixedInVersions)] | [set(SecDBVulnerable) - set(SecDBFixedInVersions)]` (which is functionally equivalent) without using the `GetVulnerabilityMetadata`. The only downside with the adjustment I'm about to put in is that there is no guarantee based on the DB schema that there couldn't be multiple records for a single CVE and package (even though we know from the data that doesn't really happen), so the solution will be a little more verbose.

zhill: Ack. Just for double clarity the conditions we should have tests for:

Matches included in final output (assume CVE-123):
1. CVE-123 match found in Alpine SecDB based on package version < fixVersion
2. CVE-123 match found in NVD, Alpine SecDB has no entry for CVE-123

Matches **excluded** from final output:
1. CVE-123 match found in NVD and Alpine SecDB has entry for CVE-123, but SecDB indicates package is not vulnerable because package version >= fix version or package names do not match

I think that encodes the logic, but if there is a case I missed let's enumerate it explicitly with a test.

wagoodman: agreed --already have test cases for 1 & 2, just added the excluded case you mention above (3), and a few more

grype/matcher/apk/matcher.go
```diff
@@ -23,39 +26,109 @@ func (m *Matcher) Type() match.MatcherType {
...
-
 	// find Alpine SecDB matches for the given package name and version
 	secDbMatches, err := common.FindMatchesByPackageDistro(store, d, p, m.Type())
 	if err != nil {
 		return nil, err
 	}
 
-	for _, secDbMatch := range secDbMatches {
-		secDbCandidates[secDbMatch.Vulnerability.ID] = append(secDbCandidates[secDbMatch.Vulnerability.ID], secDbMatch)
+	cpeMatches, err := m.cpeMatchesWithoutSecDbFixes(store, d, p)
```
luhring, line 35: I think I'm getting confused on the use of "CPE" vs. "NVD". My understanding of the "set operation" required for correct alpine matching was that it should specifically use NVD (in addition to Alpine's SecDB), as opposed to any data source that uses CPEs. With that assumption, I was expecting to see references to "NVD" here, as opposed to "CPE".

Having said that, I do also see references to NVD in this implementation. It would help my understanding to see consistency among these usages, so the answer to this question becomes clear: "Does Alpine matching _need_ NVD in particular? Or will any CPE-based data source do fine, and NVD is just one example of a data source that meets this requirement?"

grype/matcher/apk/matcher.go
```diff
@@ -23,39 +26,109 @@ func (m *Matcher) Type() match.MatcherType {
...
+		return nil, err
+	}
+
+	// keep all secdb matches, as this is an authoritative source
+	matches = append(matches, secDbMatches...)
+
+	// add additional unique matches from CPE source that is unique from the SecDB matches
+	secDbMatchesByID := matchesByID(secDbMatches)
+	cpeMatchesByID := matchesByID(cpeMatches)
+	for id, cpeMatchesForID := range cpeMatchesByID {
```
luhring, line 46: This `for` block could be broken out as its own function as an opportunity to express a noteworthy piece of the set operation — i.e. "get the CPE matches not found in Alpine SecDB". That function could return a slice that gets appended to the caller's match slice.

luhring (changes requested):

grype/matcher/apk/matcher.go
wagoodman: > Does Alpine matching need NVD in particular? Or will any CPE-based data source do fine?...

Any CPE-based data source will do fine 👍 

The intention of the `vulnerability.StoreAdapter` is to abstract away data source distinctions when asking for vulnerabilities (e.g. "get me vulnerabilities from SecDB") and instead frame it in a more agnostic way (e.g. "get me vulnerabilities for my distro"... where any and all sources will do) . The other matchers using the common CPE matcher are oriented the same way. In this way, when more data sources are added and indexed by CPEs, we don't need to make code adjustments.

luhring: Okay, cool thanks! So that means the instances of "CPE" make sense. In that case, it's the remaining references like "find NVD matches" that were making me doubt the data source agnosticism.

zhill: I think we can consider NVD or similar as "upstream" sources as opposed to vendor sources. That is the logical distinction for the matching behavior.

grype/matcher/apk/matcher_test.go
```diff
@@ -70,42 +64,123 @@ func TestNoSecDBMatch(t *testing.T) {
...
+
+	for _, diff := range deep.Equal(expected, actual) {
+		t.Errorf("diff: %+v", diff)
 	}
 
 }
 
-func TestMatches(t *testing.T) {
-	// NVD and Alpine's secDB both have the same CVE ID for the package so it matches
+func TestBothSecdbAndNvdMatches(t *testing.T) {
```
zhill, line 99: Related to this, but not quite the same the same test would be if the CVE ID is present in both NVD and SecDB, but the CPE match for NVD and package name for SecDB aren't the same (e.g. SecDB entry is for different package name). I think in that case, the logic will correctly remove the NVD match, but there isn't a test for that condition. Do you thin it warrants a specific test or the logic covered by these tests already?

zhill (commented):
A question about a test, but it may already be covered. Otherwise looks good.

grype/matcher/apk/matcher_test.go
```diff
@@ -70,42 +64,123 @@ func TestNoSecDBMatch(t *testing.T) {
...
+	actual, err := m.Match(provider, &d, p)
+	assert.NoError(t, err)
+
+	for _, diff := range deep.Equal(expected, actual) {
+		t.Errorf("diff: %+v", diff)
+	}
+}
+
+func TestNvdOnlyMatches(t *testing.T) {
+	// NVD and Alpine's secDB both have the same CVE ID for the package
```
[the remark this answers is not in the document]
wagoodman: nope --copy-pasta. I'll adjust

grype/matcher/apk/matcher_test.go
wagoodman: The nuance seems important given the difference in product-name and package-names we see, and this can be easily tested --I'll add a case for it 👍

luhring (approved):

grype/matcher/apk/matcher.go
```diff
@@ -23,39 +26,115 @@ func (m *Matcher) Type() match.MatcherType {
...
 		return nil, err
 	}
 
-	for _, cpeMatch := range cpeMatches {
-		cpeCandidates[cpeMatch.Vulnerability.ID] = append(cpeCandidates[cpeMatch.Vulnerability.ID], cpeMatch)
+	cpeMatchesByID := matchesByID(cpeMatches)
+
+	// remove cpe matches where there is an entry in the secDB for the particular package-vulnerability pairing, and the
+	// installed package version is >= the fixed in version for the secDB record.
+	secDbVulnerabilities, err := store.GetByDistro(*d, p)
```
zhill, line 60: This returns all from the distro that match the package, but not doing an actual constraint check, correct?

grype/matcher/apk/matcher.go
```diff
@@ -23,39 +26,115 @@ func (m *Matcher) Type() match.MatcherType {
...
+		secDbVulnerabilitiesForID, exists := secDbVulnerabilitiesByID[id]
+		if !exists {
+			// does not exist in secdb, so the CPE record(s) should be added to the final results
+			finalCpeMatches = append(finalCpeMatches, cpeMatchesForID...)
+			continue
+		}
+
+		// there is a secdb entry...
+		for _, vuln := range secDbVulnerabilitiesForID {
+			// ...is there a fixed in entry? (should always be yes)
```
zhill, line 86: I think this implements the logic, but seems redundant to me. The SecDB match in Match() should catch that case and if there isn't a match from SecDB then the NVD match should be ignored. What case does this catch that isn't caught by the earlier secDB matching pass?

zhill (commented):
A question on possible redundant code, but otherwise looks good.

zhill (approved):
See question about redundant code, but not a blocker.

grype/matcher/apk/matcher.go
wagoodman: correct 👍

grype/matcher/apk/matcher.go
wagoodman: technically yes, this is redundant. However, since we're pulling out entries from the DB via `GetByDistro` (instead of using any earlier results) we are making no assumptions about the vulnerabilities that are returned. This means also checking explicitly that the package is vulnerable for the specific vulnerability details returned... which means we should check the version constraint.

The same is true for a list of vulnerabilities. We know that since SecDB is the only backing dataset, there should only be one vulnerability for each ID for each package, not multiple. However, the query makes no assumptions there, so we treat the return as if multiple could be returned, even though it shouldn't happen under the current circumstances.

PR #311, opened 2021-05-06
Include references back to Syft

luhring:
Signed-off-by: Dan Luhring <dan.luhring@anchore.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 [![License: Apache-2.0](https://img.shields.io/badge/License-Apache%202.0-blue.svg)](https://github.com/anchore/grype/blob/main/LICENSE)
 [![GitHub go.mod Go version](https://img.shields.io/github/go-mod/go-version/anchore/grype.svg)](https://github.com/anchore/grype)
 
-A vulnerability scanner for container images and filesystems. [Easily install the binary](#installation) to try it out. Works with [Syft](https://github.com/anchore/syft), the powerful SBOM tool for container images and filesystems.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,7 +7,7 @@
 [![License: Apache-2.0](https://img.shields.io/badge/License-Apache%202.0-blue.svg)](https://github.com/anchore/grype/blob/main/LICENSE)
 [![GitHub go.mod Go version](https://img.shields.io/github/go-mod/go-version/anchore/grype.svg)](https://github.com/anchore/grype)
 
-A vulnerability scanner for container images and filesystems. [Easily install the binary](#installation) to try it out.
+A vulnerability scanner for container images and filesystems. [Easily install the binary](#installation) to try it out. Works with [Syft](https://github.com/anchore/syft), the powerful SBOM tool for container images and filesystems.
```
alfredodeza, line 10: nit: it is common to spell out acronyms the first time they are introduced

README.md
```diff
@@ -55,6 +55,19 @@ grype path/to/image.tar
 grype dir:path/to/dir
 ```
 
+Use [Syft](https://github.com/anchore/syft) SBOMs for even faster vulnerability scanning in Grype:
+
+```
+# Just need to generate the SBOM once
```
alfredodeza, line 61: This looks great but it would be nice to break out the comment out of the code block

luhring: Discussed in Slack — this is consistent with how we approach our examples today, but we should heavily consider refreshing out READMEs and expanding into a docs site

README.md
luhring: I think this makes it a bit wordy, TBH. But I'll try it out for now!

alfredodeza (approved):

PR #313, opened 2021-05-11
Add package sorting for artifacts in json document

### LINKED ISSUE #245: json output should be sorted
Sorting in the output would make it easier to check differences between separate outputs. Since the order is not kept, the diff can be several thousand lines, making it hard to detect actual differences.

wagoodman:
Sorts the JSON presenter output by vulnerability ID, package name, package version, then package type. This will help with more stable output when comparing two documents.

Closes #245

### DIFF
--- a/grype/match/matches_test.go
+++ b/grype/match/matches_test.go
@@ -9,7 +9,7 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestMatchesSort(t *testing.T) {
+func TestMatchesSortMixedDimensions(t *testing.T) {
 	first := Match{
 		Vulnerability: vulnerability.Vulnerability{
 			ID: "CVE-2020-0010",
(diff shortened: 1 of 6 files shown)


grype/presenter/models/document.go
```diff
@@ -37,7 +37,7 @@ func NewDocument(packages []pkg.Package, context pkg.Context, matches match.Matc
 	metadataProvider vulnerability.MetadataProvider, appConfig interface{}, dbStatus interface{}) (Document, error) {
 	// we must preallocate the findings to ensure the JSON document does not show "null" when no matches are found
 	var findings = make([]Match, 0)
-	for m := range matches.Enumerate() {
+	for _, m := range matches.Sorted() {
```
alfredodeza, line 40: is there going to be a situation where we wouldn't want sorted packages by default? Why does it need to be explicitly called out?

grype/match/matches_test.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	"testing"
+
+	"github.com/anchore/grype/grype/pkg"
+	"github.com/anchore/grype/grype/vulnerability"
+	syftPkg "github.com/anchore/syft/syft/pkg"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMatchesSort(t *testing.T) {
```
alfredodeza, line 12: It isn't clear to me what cases this is covering. While trying to ensure there was coverage for the sorting, I had to go back and forth to catch which ones where for the same ID, then for the same package name and so on. This is not only time-consuming, but if I had to fix (or update) the sorting, I wouldn't be sure how to add a specific case that covers the new change. 

Going through the matches, it seems there is no case that catches same ID, same package name, and same Version.

I'd suggest breaking the test cases up so that the conditions they are testing are explicit.

alfredodeza (changes requested):

grype/match/matches_test.go
wagoodman: I kept the original test for a multi-dimensional sort test, but also broke out testing into pairs within the same dimension. 👍

alfredodeza (approved):

PR #320, opened 2021-05-24
Add java virtual path to package metadata

wagoodman:
The java virtual path is akin to the `Package.Locations` field except there is additional info that can track location information within nested jars with arbitrary depth. This brings further clarity as to which package exactly was matched on and exactly where it is. No other package type has this requirement, only java.

Testing was added to cover mapping of all package metadata from syft to the grype equivalent.

### DIFF
--- a/grype/pkg/package_test.go
+++ b/grype/pkg/package_test.go
@@ -1,0 +1,229 @@
+package pkg
+
+import (
+	"testing"
+
+	"github.com/scylladb/go-set/strset"
+
+	"github.com/anchore/syft/syft/file"
+	syftPkg "github.com/anchore/syft/syft/pkg"
+	"github.com/scylladb/go-set"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestPackageMetadataExtraction(t *testing.T) {
+	tests := []struct {
+		name     string
+		syftPkg  syftPkg.Package
+		metadata interface{}
+	}{
+		{
+			name: "dpkg-metadata",
+			syftPkg: syftPkg.Package{
+				MetadataType: syftPkg.DpkgMetadataType,
+				Metadata: syftPkg.DpkgMetadata{
+					Package:       "pkg-info",
+					Source:        "src-info",
+					Version:       "version-info",
+					SourceVersion: "src-version-info",
+					Architecture:  "arch-info",
+					Maintainer:    "maintainer-info",
+					InstalledSize: 10,
+					Files: []syftPkg.DpkgFileRecord{
+						{
+							Path: "path-info",
+							Digest: &file.Digest{
+								Algorithm: "algo-info",
+								Value:     "digest-info",
+							},
+							IsConfigFile: true,
+						},
+					},
+				},
+			},
+			metadata: DpkgMetadata{
+				Source: "src-info",
+			},
+		},
+		{
+			name: "rpmdb-metadata",
+			syftPkg: syftPkg.Package{
+				MetadataType: syftPkg.RpmdbMetadataType,
+				Metadata: syftPkg.RpmdbMetadata{
+					Name:      "name-info",
+					Version:   "version-info",
+					Epoch:     30,
+					Arch:      "arch-info",
+					Release:   "release-info",
+					SourceRpm: "src-rpm-info",
+					Size:      40,
+					License:   "license-info",
+					Vendor:    "vendor-info",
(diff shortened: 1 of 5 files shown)


grype/pkg/package_test.go
```diff
@@ -0,0 +1,229 @@
...
+
+	"github.com/scylladb/go-set/strset"
+
+	"github.com/anchore/syft/syft/file"
+	syftPkg "github.com/anchore/syft/syft/pkg"
+	"github.com/scylladb/go-set"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestPackageMetadataExtraction(t *testing.T) {
```
luhring, line 14: nit: I might consider a rename to the convention of naming the function under test explicitly, optionally suffixed with the kind of testing being done, e.g. something like `TestNew_MetadataExtraction`.

grype/pkg/package_test.go
```diff
@@ -0,0 +1,229 @@
...
+	observedMetadataTypes := set.NewStringSet()
+	expectedMetadataTypes := set.NewStringSet()
+	for _, ty := range syftPkg.AllMetadataTypes {
+		expectedMetadataTypes.Add(string(ty))
+	}
+
+	// run all of our cases
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			observedMetadataTypes.Add(string(test.syftPkg.MetadataType))
```
luhring, line 217: I think this might just be a case of the Mondays....... but I'm not clear on what's being tested with `observedMetadataTypes`

grype/pkg/package_test.go
```diff
@@ -0,0 +1,229 @@
...
+					Source:   "a",
+					Checksum: "a",
+				},
+			},
+			metadata: nil,
+		},
+	}
+
+	// capture each observed metadata type, we should see all of them relate to what syft provides by the end of testing
+	observedMetadataTypes := set.NewStringSet()
```
luhring, line 208: nit: I'd declare `observedMetadataTypes` down in the stanza that starts to populate the set (same scope, just repositioned down a few lines). I don't feel strongly about this, but it helped me mentally to picture setting up "expected things" first, and then starting to observe.

grype/pkg/package_test.go
```diff
@@ -0,0 +1,229 @@
...
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			observedMetadataTypes.Add(string(test.syftPkg.MetadataType))
+			assert.Equal(t, test.metadata, New(&test.syftPkg).Metadata)
+		})
+	}
+
+	// did we see all possible metadata types? if not, then there is an uncovered case and this test should error out
+	if !expectedMetadataTypes.IsEqual(observedMetadataTypes) {
+		t.Errorf("did not observe all possible package metadata types: missing: %+v extra %+v",
```
luhring, line 224: nit: Given the colon after "missing", I'm wondering if there should be a colon after "extra"

luhring (approved):
Looks good! Just some items for clarity

grype/pkg/package_test.go
wagoodman: the comments within the test comment on what is being tested with those values:
```
// capture each observed metadata type, we should see all of them relate to what syft provides by the end of testing
...
// did we see all possible metadata types? if not, then there is an uncovered case and this test should error out
```

luhring: @wagoodman Okay thanks, this is helping...

So we don't want Syft to have more metadata types than are tested here in Grype.

I'm trying to explore in my head why I took so long to understand that. (Of course, part of it is just me 😄 )

I think one factor is that we were testing two things at once. It could be a future improvement is to extract that assertion logic into its own function whose name states intent (e.g. `assertGrypeCaseForEverySyftMetadataType` or something). It also sounds like "expected/observed" maps to "Syft/Grype" — a change to `syftMetadataTypes` and `grypeMetadataTypes` (or similar) might help readers understand why we're concerned with tracking and diffing these sets.

Also, if I'm understanding correctly, there shouldn't be a case where Grype is testing more metadata types than are coming from Syft, since Grype wouldn't be able to reference symbols from Syft that Syft doesn't have in the currently used version. If we had messed up in Syft and left symbols around that Syft wasn't using, I guess this could open up the possibility for this case.

One fun fact, possibly useful / possibly not, there's an `assert.ElementsMatch` function that can compare slices in such a way that ignores order, which could help us avoid building sets in some cases.

PR #328, opened 2021-05-30
Report multiple match details for matches

wagoodman:
To accommodate CPE matching, where multiple CPEs may be used for a search input, the match object now displays multiple match details. 

Summary of changes:
- matchDetails now is an array, allowing for multiple details to be attached to one match. This is important for CPE matching only (for now), not distro matching.
- Version extracted from the search CPE is used in CPE searches, the package version is only used when there is no version in the given CPE
- multiple matchDetail records are merged together if they result in the same MatchedOn result and have the same SearchedBy.Namespace
- `MatchedOn.Namespace` has been moved to `SearchedBy.Namespace`, an oversight from a previous PR.

### DIFF
--- a/grype/match/match.go
+++ b/grype/match/match.go
@@ -2,6 +2,7 @@ package match
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/anchore/grype/grype/pkg"
 	"github.com/anchore/grype/grype/vulnerability"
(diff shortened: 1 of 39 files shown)


grype/match/match.go
```diff
@@ -10,20 +11,41 @@ import (
...
+
+type Details struct {
+	SearchedBy interface{} // The specific attributes that were used to search (other than package name and version) --this indicates "how" the match was made.
+	MatchedOn  interface{} // The specific attributes on the vulnerability object that were matched with --this indicates "what" was matched on / within.
+	Matcher    MatcherType // The matcher object that discovered the match.
+	Confidence float64     // The certainty of the match as a ratio (currently unused, reserved for future use).
+}
+
+type Fingerprint struct {
+	VulnerabilityID        string
```
wagoodman, line 27: from offline review: unexport these fields

grype/matcher/common/cpe_matchers.go
```diff
@@ -2,79 +2,206 @@ package common
...
+
+	return true
+}
+
+// FindMatchesByPackageCPE retrieves all vulnerabilities that match the generated CPE
+func FindMatchesByPackageCPE(store vulnerability.ProviderByCPE, p pkg.Package, upstreamMatcher match.MatcherType) ([]match.Match, error) {
+	matchesByFingerprint := make(map[match.Fingerprint]match.Match)
+	for _, cpe := range p.CPEs {
+		// prefer the CPE version, but if npt specified use the package version
+		pkgVersion := cpe.Version
```
wagoodman, line 62: rename pkgVersion

grype/matcher/common/cpe_matchers.go
```diff
@@ -2,79 +2,206 @@ package common
...
+		return existingDetails
 	}
 
-	return stringers
+	// could not merge with another entry, append to the end
+	existingDetails = append(existingDetails, newDetails)
+	return existingDetails
+}
+
+func keepMatchingCPEs(pkgVersion *version.Version, allCPEs []syftPkg.CPE) (matchedCPEs []syftPkg.CPE) {
```
wagoodman, line 167: rename with version in the name here (filterCPEsByVersion or keepCPEsByVersion)

grype/pkg/syft_json_provider.go
```diff
@@ -83,7 +83,7 @@ type partialSyftPackage struct {
 
 // packageBasicMetadata contains non-ambiguous values (type-wise) from pkg.Package.
 type packageBasicMetadata struct {
-	ID        string            `json:"id"`
+	ID        string            `json:"ID"`
```
wagoodman, line 86: this may have been from a find-replace, I should double check this

wagoodman: from an over zealous find-replace... removing

luhring (approved):
Looking good! Thanks for the Zoom sync, that helped tremendously 🥇

PR #330, opened 2021-06-02
Correct validate-grype-test-target check

wagoodman:
Currently the check will always fail, this PR splits the check and the grep command to get the data into separate lines and ensures the if condition is always using bash (some systems may default to sh).


Btodhunter (approved):
lgtm!

PR #331, opened 2021-06-02
Fix RPM epoch comparison logic

wagoodman:
This PR incorporates the changes from https://github.com/anchore/syft/pull/427 but additionally ensures that when comparing two RPM versions, [if an epoch is not provided then it is assumed to be 0](https://github.com/rpm-software-management/rpm/issues/450) (a change from previous behavior, where the epoch was ignored entirely if missing from one of the versions being compared).

Note: until the syft PR is merged and incorporated here PR checks will continue to fail

Related to https://github.com/anchore/syft/issues/408


grype/pkg/package_test.go
```diff
@@ -226,3 +226,10 @@ func TestNew_MetadataExtraction(t *testing.T) {
 		)
 	}
 }
+
+func intRef(i ...int) *int {
```
luhring, line 230: Same comment as Syft PR

luhring (changes requested):
One small request stemming from the corresponding Syft PR. Other than that looks great

luhring (approved):

grype/version/rpm_version.go
```diff
@@ -11,24 +11,26 @@ import (
...
 }
 
 func newRpmVersion(raw string) (rpmVersion, error) {
-	fields := strings.SplitN(raw, ":", 2)
-
-	var epoch *int
+	var fields = strings.SplitN(raw, ":", 2)
+	var err error
+	// when the epoch is not included, should be considered to be 0 during comparisons
+	// see https://github.com/rpm-software-management/rpm/issues/450
```
alfredodeza, line 23: wow  what a find

alfredodeza (approved):

PR #344, opened 2021-06-23
Modify Document struct Components tag to use lower case 'component'

rxl7906:
The Grype scanner produces a CycloneDX Vulnerability Document which is in XML output with lower case component tags.

However the "Document" struct from this library has a Components tag that uses an upper case "Component". When the CycloneDX XML output is passed into a Document struct, it results in an empty Components array.

Fixing this issue https://github.com/anchore/grype/issues/345.

Signed-off-by: Robin Li <lrobin@vmware.com>


luhring (approved):
Good catch!

PR #359, opened 2021-07-01
Catch unclean go.sum files in pipeline

### LINKED ISSUE #358: Pipelines allow unclean go.sum files that block our release pipeline
In neither of our primary pipelines (for PRs, and for releases) do we catch "unclean" `go.sum` files, where "unclean" is defined as having excess line items that would not be present after running `go mod tidy`.

We should catch this scenario in our Static Analysis pipeline (which runs for PRs), which is as far upstream we can detect this problem, short of changes to developer environments. We have most of the pieces in place to do this. But since we run our `check-go-mod-tidy` make target after having run `go mod tidy`, we aren't catching the problem. To correctly catch the problem in the Static Analysis pipeline, we need to:

- [x] remove the `go mod tidy` command from the bootstrap step (**Note**: this leaves `go mod download` as the last Go command in the bootstrapping process)
- [x] modify `go-mod-tidy-check.sh` to detect and error out upon finding a diff between _"A"_ and _"B"_, where: _"A"_ is the state of `go.sum` after running **`git checkout go.sum`**, and _"B"_ is state of `go.sum` after running **`go mod tidy`**. (**Note**: We need to ensure no destructive git changes, since developers can run this script on their local environments.)

Additionally, a change is needed in the Release worfklow. It's possible that the bootstrapping process can leave an "unclean" `go.sum` file on disk via a known [bug](https://github.com/golang/go/issues/45332) with `go mod download`. This file would get "tidied" when goreleaser begins to execute. However, this would put git into a dirty state, and a dirty git state would cause gorelease not to release (as we expect). (Note: It looks like this `go mod download` bug has been fixed and [backported to Go 1.16.5](https://github.com/golang/go/issues?q=milestone%3AGo1.16.5+label%3ACherryPickApproved+go.sum).)

- [x] Run the `go mod tidy` command just before invoking goreleaser

luhring:
Fixes #358 

This PR, once merged, is expected to unblock Grype's release process.

### DIFF
--- a/.github/scripts/go-mod-tidy-check.sh
+++ b/.github/scripts/go-mod-tidy-check.sh
@@ -1,19 +1,13 @@
 #!/usr/bin/env bash
 set -eu
 
-BACKUPS_DIR=$(mktemp -d "TEMP-backups-XXXXXXXXX")
-GIT_HEAD_STATE_DIR=$(mktemp -d "TEMP-git-head-state-XXXXXXXXX")
+ORIGINAL_STATE_DIR=$(mktemp -d "TEMP-original-state-XXXXXXXXX")
 TIDY_STATE_DIR=$(mktemp -d "TEMP-tidy-state-XXXXXXXXX")
 
-trap "cp -v ${BACKUPS_DIR}/* ./ && rm -fR ${BACKUPS_DIR} ${GIT_HEAD_STATE_DIR} ${TIDY_STATE_DIR}" EXIT
+trap "cp -v ${ORIGINAL_STATE_DIR}/* ./ && rm -fR ${ORIGINAL_STATE_DIR} ${TIDY_STATE_DIR}" EXIT
 
-echo "Backing up files from working tree..."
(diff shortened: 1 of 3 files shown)


.github/scripts/go-mod-tidy-check.sh
```diff
@@ -1,27 +1,37 @@
 #!/usr/bin/env bash
-set -u
+set -eu
```
wagoodman, line 2: may need to confirm this: if you `set -e` doesn't that prevent trap functionality for exit? (You may need to also add the `ERR` trap)

luhring: The trap command will still be executed, regardless of exiting with an error or exiting cleanly — is that what you mean?

wagoodman: cool, couldn't remember, wanted to double check 👍

wagoodman (approved):

PR #365, opened 2021-07-12
Always include severity in cyclonedx ratings

### LINKED ISSUE #364: cyclonedx doesn't report severity if feed has no CVSS
**What happened**:
When scanning `nginx:1.16`, based on debian:10 (or any image where the CVEs are in a non-RHEL namespace) the cyclonedx report has empty `<v:ratings></v:ratings>` because the feed has no CVSS info, even though CVE metadata does include a severity.

If CVE metadata includes severity, the cyclonedx should include 
```xml
<v:ratings>
  <v:rating>
    <v:severity>VALUE</v:severity>
  </v:rating>
</v:ratings>
```

`grype nginx:1.16 -o cyclonedx | grep rating`

Severity is checked inside a cvss loop (https://github.com/anchore/grype/blob/main/grype/presenter/cyclonedx/vulnerability.go#L103-L110). If metadata.Cvss is empty, a rating with only severity could be constructed.

`-o table` output does report the severity in the absence of cvss

**Environment**:
Application:          grype
Version:              0.13.0
BuildDate:            2021-06-02T01:57:12Z
GitCommit:            3d21b8397d65770d292184b09a4f676bce6f3ec8
GitTreeState:         clean
GoVersion:            go1.16.4
Compiler:             gc
Supported DB Schema:  3

xtreme-conor-nosal:
Some feeds don't include CVSS but still report severity of a CVE

Fix #364 and #366

### DIFF
--- a/grype/presenter/cyclonedx/vulnerability.go
+++ b/grype/presenter/cyclonedx/vulnerability.go
@@ -36,7 +36,7 @@ type Source struct {
 
 // Rating has information about the intensity of a vulnerability
 type Rating struct {
-	Score    Score  `xml:"v:score"`
+	Score    *Score `xml:"v:score,omitempty"`
 	Severity string `xml:"v:severity,omitempty"`
 	Method   string `xml:"v:method,omitempty"`
 	Vector   string `xml:"v:vector,omitempty"`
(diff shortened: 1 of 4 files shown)


grype/presenter/cyclonedx/vulnerability.go
```diff
@@ -91,24 +103,17 @@ func NewVulnerability(m match.Match, p vulnerability.MetadataProvider) (Vulnerab
...
+
 		method, err := cvssVersionToMethod(cvss.Version)
-		if err != nil {
+		if err == nil {
+			rating.Method = method
+			rating.Vector = cvss.Vector
+		} else {
 			log.Errorf("unable to parse CVSS version: %v", err)
 			// do not halt execution if one CVSS fails to provide an accurate Version
-			continue
```
wagoodman, line 98: was dropping this continue statement intentional?

wagoodman:
@xtreme-conor-nosal good update overall 🥇  , thanks for the contribution!

I added a test to this branch that covers at least the cases that a rating is minimally included even if there is no CVSS information provided --we can build on test cases here even if the cases I added ends up being deleted (based on the below potential changes).

I think a small change may be in order here so that the cycloneDX report is more accurate. In hindsight it doesn't make sense to report out the same severity for each CVSS discovered (something that was introduced before this PR to be clear), but instead, we should report out severity on a rating only when we positively have that value.

To be more specific:
1. We should probably change the `for _, cvss := range metadata.Cvss {...}` loop to **_not_** report out `severity`, since the `CVSS` struct does not have severity (and the metadata severity may conflict with the CVSS value). A future enhancement could use the CVSS information itself to infer severity, but there is no need to do that now (if we do this, it should be a follow up PR).
2. We should report out a separate `Rating` instance to report out only the `Severity` field populated (and no other fields).  This implies that the `Rating.Score` object should now have an `,omitempty` struct tag.

Let me know what you think, @xtreme-conor-nosal .

grype/presenter/cyclonedx/vulnerability.go
xtreme-conor-nosal: Yes, so the rating can still be appended with the severity and score (on the assumption that partial data is better than no data)

wagoodman (approved):
Thanks for the extra lift on the refactor opportunity --looks great 👍

PR #367, opened 2021-07-12
Omit empty scores from cyclonedx report

### LINKED ISSUE #366: cyclonedx reports a score of 0 if CVE has no CVSS
**What happened**:
While working on https://github.com/anchore/grype/issues/364 to report severity in the absence of CVSS, the serialized rating includes a Score object because the serialization hints do not `omitempty`.

Reporting a score of 0 seems misleading, so I would expect an empty score to be omitted.

with https://github.com/anchore/grype/pull/365 `grype nginx:1.16 -o cyclondex | grep -A5 -i score`

I'm not sure if this issue would come up without 365 (i.e. a CVSS that does not include a score), but the cyclonedx spec considers all fields optional.

xtreme-conor-nosal:
Fixes #366

Signed-off-by: Conor Nosal <cnosal@vmware.com>

### DIFF
--- a/grype/presenter/cyclonedx/vulnerability.go
+++ b/grype/presenter/cyclonedx/vulnerability.go
@@ -36,7 +36,7 @@ type Source struct {
 
 // Rating has information about the intensity of a vulnerability
 type Rating struct {
-	Score    Score  `xml:"v:score"`
+	Score    Score  `xml:"v:score,omitempty"`
 	Severity string `xml:"v:severity,omitempty"`
 	Method   string `xml:"v:method,omitempty"`
 	Vector   string `xml:"v:vector,omitempty"`
@@ -44,9 +44,9 @@ type Rating struct {
 
 // Score provides the different ways to measure how serious a vulnerability is
 type Score struct {
-	Base           float64 `xml:"v:base"`
-	Impact         float64 `xml:"v:impact"`
-	Exploitability float64 `xml:"v:exploitability"`
+	Base           float64 `xml:"v:base,omitempty"`
+	Impact         float64 `xml:"v:impact,omitempty"`
+	Exploitability float64 `xml:"v:exploitability,omitempty"`
 }
 
 // Advisories holds all the links for a vulnerability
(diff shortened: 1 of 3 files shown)


grype/presenter/cyclonedx/vulnerability.go
```diff
@@ -36,17 +36,17 @@ type Source struct {
 
 // Rating has information about the intensity of a vulnerability
 type Rating struct {
-	Score    Score  `xml:"v:score"`
+	Score    Score  `xml:"v:score,omitempty"`
```
wagoodman, line 39: Since the type of the `Score` field is not a pointer, this struct tag essentially has no effect. For it to be truly optional we would need to change from `Score` to `*Score`.

grype/presenter/cyclonedx/vulnerability.go
```diff
@@ -36,17 +36,17 @@ type Source struct {
...
 	Method   string `xml:"v:method,omitempty"`
 	Vector   string `xml:"v:vector,omitempty"`
 }
 
 // Score provides the different ways to measure how serious a vulnerability is
 type Score struct {
-	Base           float64 `xml:"v:base"`
-	Impact         float64 `xml:"v:impact"`
-	Exploitability float64 `xml:"v:exploitability"`
+	Base           float64 `xml:"v:base,omitempty"`
```
wagoodman, line 47: I'm a little torn on making `Base` optional, though the [cycloneDX vuln extension spec](https://cyclonedx.org/schema/ext/vulnerability/1.0) does list it as optional, if a score (the struct) is provided at all, the base score is semantically non-optional.

grype/presenter/cyclonedx/vulnerability.go
xtreme-conor-nosal: In my testing it did have an effect once the fields inside Score had omitempty, but I can double check that. Or would you prefer the pointer either way to be more explicit that the struct might not be present?

xtreme-conor-nosal:
Closing due to overlap with #365

PR #371, opened 2021-07-20
Bump grype-db version to remove branch strategy

### LINKED ISSUE #369: Integrate latest grype-db change (branch-removal) with grype

wagoodman:
This PR makes the appropriate changes to incorporate https://github.com/anchore/grype-db/pull/31 

This includes:
- now grype directly uses the grypeDB reader provided (the store abstraction has been removed)
- namespace determination is now provided by grype-db, so has been deleted from grype

Additionally adds a python script to assert that only one schema version is being used at a time. Note: this adds the requirement that python2 or python3 be on the system running static analysis checks.

Closes #369


test/validate-grype-db-schema.py
```diff
@@ -0,0 +1,50 @@
...
+            print("  %s" % location)
+    print()
+
+
+def validate(schema_to_locations):
+    schema_versions_found = schema_to_locations.keys()
+    try:
+        for x in schema_versions_found:
+            int(x)
+    except Exception:
```
luhring, line 23: (not a Python expert by any means 😄 ) Is catching a bare `Exception` something that we should try to avoid?

luhring (commented):
Looks good — had one question for you, though

test/validate-grype-db-schema.py
wagoodman: yup 👍 I'll scope this a bit narrower

luhring (approved):

PR #391, opened 2021-08-19
Support gomod configuration in goreleaser

xtreme-conor-nosal:
goreleaser added support for a `gomod.proxy` configuration stanza in [v0.162.0](https://github.com/goreleaser/goreleaser/commit/8306b946d3ee38a25d30c46bcaa7b12071a7256e)

The [docs](https://goreleaser.com/customization/gomod/) indicate --snapshot will ignore the setting so the flag is replaced with the implied `--skip-publish --skip-validate --skip-announce`.

I looked into adding an explicit proxy config into .goreleaser.yaml, but it appears the goreleaser cli doesn't allow for dynamic overrides. --config is repeatable, but the second file replaces the config instead of merging.

`make bootstrap build` should behave the same, but these changes allow for:

`
make bootstrap

echo "gomod:
  proxy: true
  env:
  - GOPROXY=myproxy.com" >> .goreleaser.yaml

make build
`
  

Signed-off-by: Conor Nosal <cnosal@vmware.com>


xtreme-conor-nosal:
The test is failing because the runner checks out the repo with `no-tags` , and `goreleaser release` expects a tag to be present for non-snapshot builds.

Without fully knowing how you use `make snapshot` internally, would it be reasonable for that task to invoke `goreleaser build` instead of `goreleaser release`? Or update `make build` to invoke `goreleaser build` instead of `snapshot` ?

wagoodman:
@xtreme-conor-nosal using `goreleaser build` is a great idea --I pushed your suggestion to the branch.

Additionally I rebased and bumped to the version of goreleaser to the latest version as of today.

wagoodman (approved):
Thanks for the update! Shout out if the adjustments I made negatively affect your use case and we can adjust.

PR #392, opened 2021-08-20
Bump untar file size threshold

wagoodman:
Today the tar helper is used to extract grype DB files from archives. However, it is possible for the grype DB to reach large sizes on development builds (up to 2GB currently). For this reason the maximum allowable file size in a tar needs to be raised (this threshold exists today to mitigate decompression bomb attacks).

### DIFF
--- a/internal/file/tar.go
+++ b/internal/file/tar.go
@@ -15,8 +15,8 @@ const (
 	KB = 1 << (10 * iota)
 	MB
 	GB
-	// Why 5GB? This is somewhat of an arbitrary threshold, however, we need to keep this at at minimum 2GB
-	// to accommodate possible grype DB sizes.
+	// limit the tar reader to 5GB per file to prevent decompression bomb attacks. Why 5GB? This is somewhat of an
+	// arbitrary threshold, however, we need to keep this at at minimum 2GB to accommodate possible grype DB sizes.
 	decompressionByteReadLimit = 5 * GB
 )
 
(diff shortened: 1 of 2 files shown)


internal/file/tar.go
```diff
@@ -93,3 +95,13 @@ func UnTarGz(dst string, r io.Reader) error {
 		}
 	}
 }
+
+func copyWithLimits(writer io.Writer, reader io.Reader, byteReadLimit int64, target string) error {
+	// limit the tar reader to 5GB per file to prevent decompression bomb attacks
```
luhring, line 100: I think the code would be more clear without this comment, actually. It looks like the "5GB" value has been parametrized, not set explicitly in this scope. And the comment above for `decompressionByteReadLimit` does a nice job of explaining this value's motivation 🙌

internal/file/tar.go
```diff
@@ -93,3 +95,13 @@ func UnTarGz(dst string, r io.Reader) error {
 		}
 	}
 }
+
+func copyWithLimits(writer io.Writer, reader io.Reader, byteReadLimit int64, target string) error {
```
luhring, line 99: I've had a bit of trouble figuring out what `target` is, after looking at this function and its tests. Is this a file path?

luhring (changes requested):

internal/file/tar.go
wagoodman: looks like I migrated the code without updating the comment --I'll update

internal/file/tar.go
wagoodman: it is a path within the archive, I'll adjust the name

luhring (approved):
Looks great! Thanks!

Issue #354, opened 2021-07-01
Grype generates different results when using syft directly or if getting syft input from pipe


wlfshmn:
**What happened**:

I normally invoke grype based on stored output from syft as a performance improvement as I use the syft output for multiple checks and this saves me additional invocations of syft.

I noticed however that the virtualPAth data from syft on java archives does not transfer over when using the piped invocation but does get populated correctly when syft is used internally in grype. Investigation shows that there are also differences in the cpu generation in the two scenarios.

**What you expected to happen**:

I expect the two invocation scenarios to result in the same output, assuming the same version of syft is used internally in grype and int he piped invocation.

**How to reproduce it (as minimally and precisely as possible)**:

```
$ syft springio/gs-spring-boot-docker:latest -o json | /usr/local/bin/grype -o json > grype-with-pipe-from-syft.txt
$ grype springio/gs-spring-boot-docker:latest -o json > grype-direct.txt
$ diff grype-direct.txt grype-with-pipe-from-syft.txt | wc -l
896
```
Files attached.
[grype-with-pipe-from-syft.txt.gz](https://github.com/anchore/grype/files/6749178/grype-with-pipe-from-syft.txt.gz)
[grype-direct.txt.gz](https://github.com/anchore/grype/files/6749177/grype-direct.txt.gz)
[diff.txt.gz](https://github.com/anchore/grype/files/6749176/diff.txt.gz)

**Anything else we need to know?**:

**Environment**:
```
$ syft version
Application:   syft
Version:       0.15.1
BuildDate:     2021-04-22T20:55:47Z
GitCommit:     032cecbe7f19ca6f6f07d80f8dc902530e4c7502
GitTreeState:  clean
Platform:      linux/amd64
GoVersion:     go1.16.3
Compiler:      gc

$ grype version
Application:          grype
Version:              0.13.0
BuildDate:            2021-06-02T01:57:12Z
GitCommit:            3d21b8397d65770d292184b09a4f676bce6f3ec8
GitTreeState:         clean
Platform:             linux/amd64
GoVersion:            go1.16.4
Compiler:             gc
Supported DB Schema:  3

Syft 0.15. was used to match the version included in this version of Grype. Updating to the most recent syft makes no differences to the cpe or virtualPath properties, and only adds additional information like licenses that were missing in both cases before.

dakaneye:
[...]

PR #393, opened 2021-08-23
Ensure that virtual path is reported for java archives.

dakaneye:
Improved logging and added a String() method too. 
Also, added a test case for parsing

Partially addressed #354 

### DIFF
--- a/grype/pkg/syft_json_provider_test.go
+++ b/grype/pkg/syft_json_provider_test.go
@@ -127,6 +127,7 @@ func TestParseSyftJSON(t *testing.T) {
 				}(),
 			},
 		},
+		springImageTestCase,
 	}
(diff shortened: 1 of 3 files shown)


grype/pkg/syft_json_provider_test.go
```diff
@@ -154,3 +155,81 @@ func TestParseSyftJSON(t *testing.T) {
 		})
 	}
 }
+
+// Note that the fixture has been modified from the real syft output to include less packages, CPEs, layers,
+// and package IDs are removed so that the test case variable isn't unwieldingly huge.
```
wagoodman, line 160: 🙏

wagoodman (approved):
💯

PR #396, opened 2021-08-23
Upgrade syft to 0.21.0. Fixes: #385

### LINKED ISSUE #385: Reporting vulnerabilities for NPM dependencies from lock files that should be excluded
**What happened**: Grype is outputting vulnerabilities that match packages that shouldn't have been included in the SBOM, due to https://github.com/anchore/syft/issues/431

**What you expected to happen**: Syft shouldn't be including dependencies found in nested `yarn.lock` and `package-lock.json` files, and thus those dependencies should not be matched against within Grype.

For more user context see: https://anchorecommunity.slack.com/archives/C027JE5M345/p1629292882017100

kzantow:
Upgrade to syft v0.21.0, which includes a change to skip certain files in `node_modules` directories, and fixes #385


luhring (approved):
🎉

PR #403, opened 2021-08-31
Add integration test for MSRC matching

wagoodman:
Adds a test which exercises passing a syft SBOM with windows KB info that enables matching against mock MSRC vulnerability data.

### DIFF
--- a/test/integration/match_by_sbom_document_test.go
+++ b/test/integration/match_by_sbom_document_test.go
@@ -1,0 +1,71 @@
+package integration
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/anchore/grype/grype"
+	"github.com/anchore/grype/grype/match"
+	"github.com/anchore/grype/grype/vulnerability"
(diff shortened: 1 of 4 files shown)


test/integration/match_by_sbom_document_test.go
```diff
@@ -0,0 +1,71 @@
...
+			},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			provider := vulnerability.NewProviderFromStore(newMockDbStore())
+			matches, _, _, err := grype.FindVulnerabilities(provider, fmt.Sprintf("sbom:%s", test.fixture), source.SquashedScope, nil)
+			assert.NoError(t, err)
+			details := make([]match.Details, 0)
```
luhring, line 55: Super **nit** because this is a test: but if you want, you can allocate the capacity based on the length of `matches.Sorted()`.

luhring (approved):

PR #406, opened 2021-09-02
Upgrade config, UI, and command package patterns

wagoodman:
Syft is several months ahead regarding patterns for configuration management and `cmd` package needs. This PR brings the differences closer to 0.

Specifically:
- Splits up the configuration into multiple files and adds the `parseConfigValues ` & `parseConfigValues` methods on each config struct that needs them.
- Adds the event loop that separates common event handling (signals, worker errors, and bus events) concerns from the UI implementations.
- Changes the main cobra execution path to prefer raising errors (use `RunE` over `Run`).
- Reorganize `ui` package to a more flat structure


internal/config/application.go
```diff
@@ -0,0 +1,231 @@
...
+			switch v := cfg.CliOptions.Verbosity; {
+			case v == 1:
+				cfg.Log.LevelOpt = logrus.InfoLevel
+			case v >= 2:
+				cfg.Log.LevelOpt = logrus.DebugLevel
+			default:
+				cfg.Log.LevelOpt = logrus.ErrorLevel
+			}
+		}
+	}
```
luhring, line 129: nit: this block could be factored out as a function that performs a single task: determining the log level

internal/config/registry.go
```diff
@@ -22,7 +24,14 @@ type registry struct {
...
 
-func (cfg *registry) parseConfigValues() {
+func (cfg registry) loadDefaultValues(v *viper.Viper) {
+	v.SetDefault("registry.insecure-skip-tls-verify", false)
+	v.SetDefault("registry.insecure-use-http", false)
+	v.SetDefault("registry.auth", []RegistryCredentials{})
+}
+
+// nolint: unparam
+func (cfg *registry) parseConfigValues() error {
```
luhring, line 34: question: I looked up `unparam` on the [golangci-lint docs](https://golangci-lint.run/usage/linters/) — could it be we no longer need this? The descrption of `unparam` is:

>Reports unused function parameters

test/cli/json_descriptor_test.go
```diff
@@ -19,9 +19,9 @@ func TestJsonDescriptor(t *testing.T) {
 				"GRYPE_CHECK_FOR_APP_UPDATE": "false",
 			},
 			assertions: []traitAssertion{
-				assertInOutput(`"CheckForAppUpdate": false`), // app config
-				assertInOutput(`"db": {`),                    // db status
-				assertInOutput(`"built":`),                   // db status
+				assertInOutput(`"check-for-app-update": false`), // app config
+				assertInOutput(`"db": {`),                       // db status
+				assertInOutput(`"built":`),                      // db status
```
luhring, line 24: question: What do the comments (both `db status`) on these two lines mean?

luhring (approved):
Looks good! Thanks for tackling this. I left a couple of comments and questions for discussion, but none of these block the merge itself.

test/cli/json_descriptor_test.go
wagoodman: I'll add more verbiage here. Essentially this is saying that these assertions are for the `db` block in the json output.

wagoodman:
In making the above updates I noticed that I missed adding struct tags in some config options --I've updated https://github.com/anchore/grype/pull/406/commits/999aaa78ca3a98116d082bc1fb27bfd5642ad43d

internal/config/application.go
wagoodman: I broke out the remaining option parsing blocks into functions

internal/config/registry.go
wagoodman: actually it is still required such that this fulfills the `parser` interface (unparam applies to return values as well, in this case is always nil)

internal/config/application.go
```diff
@@ -91,13 +91,43 @@ func (cfg Application) loadDefaultValues(v *viper.Viper) {
...
+	// parse application config options
+	for _, optionFn := range []func() error{
+		cfg.parseScopeOption,
+		cfg.parseLogLevelOption,
+		cfg.parseFailOnOption,
+	} {
+		if err := optionFn(); err != nil {
+			return err
+		}
+	}
```
luhring, line 103: Nice idea! 👍

internal/config/database.go
```diff
@@ -10,10 +10,10 @@ import (
...
 
 type database struct {
-	Dir                   string `mapstructure:"cache-dir"`
-	UpdateURL             string `mapstructure:"update-url"`
-	AutoUpdate            bool   `mapstructure:"auto-update"`
-	ValidateByHashOnStart bool   `mapstructure:"validate-by-hash-on-start"`
+	Dir                   string `yaml:"cache-dir" json:"cache-dir" mapstructure:"cache-dir"`
+	UpdateURL             string `yaml:"update-url" json:"update-url" mapstructure:"update-url"`
+	AutoUpdate            bool   `yaml:"auto-update" json:"auto-update" mapstructure:"auto-update"`
+	ValidateByHashOnStart bool   `yaml:"validate-by-hash-on-start" json:"validate-by-hash-on-start" mapstructure:"validate-by-hash-on-start"`
```
luhring, line 16: Blocking: for JSON, we use `camelCase`

test/cli/json_descriptor_test.go
```diff
@@ -19,9 +19,9 @@ func TestJsonDescriptor(t *testing.T) {
 				"GRYPE_CHECK_FOR_APP_UPDATE": "false",
 			},
 			assertions: []traitAssertion{
-				assertInOutput(`"check-for-app-update": false`), // app config
-				assertInOutput(`"db": {`),                       // db status
-				assertInOutput(`"built":`),                      // db status
+				assertInOutput(`"check-for-app-update": false`), // assert existence of the app config block
+				assertInOutput(`"db": {`),                       // assert existence of the db status block
+				assertInOutput(`"built":`),                      // assert existence of the db status block
```
luhring, line 24: Very helpful, thanks!

luhring (changes requested):

internal/config/database.go
luhring: Disregard, and I'll approve again: I see that we tend to use `snake-case` for our configuration JSON.

So given that, I'd suggest that as we think about important but breaking changes we want to make for 1.0, we consider using a consistent JSON convention between our inputs and outputs. 🙏

luhring (approved):

internal/config/database.go
wagoodman: that makes sense 💯

PR #407, opened 2021-09-03
Alpine match source pkg

### LINKED ISSUE #343: Alpine matching should include source indirection matching
Today DPKG does matching against source package information:
https://github.com/anchore/grype/blob/main/grype/matcher/dpkg/matcher.go#L30-L34

We should be doing the same for the alpine matcher:
https://github.com/anchore/grype/blob/main/grype/matcher/apk/matcher.go

zburstein:
This PR allows grype to match with source packages for alpine. It adds Metadata for APK packages that have OriginPackage reference. This is then used in the apk matcher to match against the origin package. 

After discussion with @wagoodman, it was decided that we would match the source just like we do with the actual package itself. In alpine, this is a multi step process. First it matches against secdb and then uses the cpes to also find matches within nvd. Matching the source package against secdb simply requires changing the name on the indirectPackage, but matching against nvd is a little more complicated. We decided to take all the cpes that are generated for the package and replace the package name with the source package. These cpes are then used to match against nvd

This PR also updates the version of go-version and adds a test for the bug that was fixed there: https://github.com/anchore/go-version/pull/2

Fixes #343

### DIFF
--- a/grype/matcher/apk/matcher.go
+++ b/grype/matcher/apk/matcher.go
@@ -1,7 +1,9 @@
 package apk
 
 import (
+	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/anchore/grype/grype/match"
 	"github.com/anchore/grype/grype/matcher/common"
@@ -10,6 +12,8 @@ import (
 	"github.com/anchore/grype/grype/vulnerability"
 	"github.com/anchore/syft/syft/distro"
 	syftPkg "github.com/anchore/syft/syft/pkg"
+	"github.com/jinzhu/copier"
+	"github.com/scylladb/go-set/strset"
 )
 
 type Matcher struct {
@@ -26,22 +30,19 @@ func (m *Matcher) Type() match.MatcherType {
 func (m *Matcher) Match(store vulnerability.Provider, d *distro.Distro, p pkg.Package) ([]match.Match, error) {
 	var matches = make([]match.Match, 0)
 
-	// find Alpine SecDB matches for the given package name and version
-	secDbMatches, err := common.FindMatchesByPackageDistro(store, d, p, m.Type())
+	// direct matches with package
+	directMatches, err := m.findApkPackage(store, d, p)
 	if err != nil {
 		return nil, err
 	}
+	matches = append(matches, directMatches...)
 
-	cpeMatches, err := m.cpeMatchesWithoutSecDbFixes(store, d, p)
+	// indirect matches with package source
+	indirectMatches, err := m.matchBySourceIndirection(store, d, p)
 	if err != nil {
 		return nil, err
 	}
-
-	// keep all secdb matches, as this is an authoritative source
-	matches = append(matches, secDbMatches...)
-
-	// keep only unique CPE matches
-	matches = append(matches, deduplicateMatches(secDbMatches, cpeMatches)...)
+	matches = append(matches, indirectMatches...)
 
 	return matches, nil
 }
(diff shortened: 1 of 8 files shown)


grype/matcher/apk/matcher.go
```diff
@@ -26,22 +29,19 @@ func (m *Matcher) Type() match.MatcherType {
 func (m *Matcher) Match(store vulnerability.Provider, d *distro.Distro, p pkg.Package) ([]match.Match, error) {
 	var matches = make([]match.Match, 0)
 
-	// find Alpine SecDB matches for the given package name and version
-	secDbMatches, err := common.FindMatchesByPackageDistro(store, d, p, m.Type())
+	//direct matches with package
```
luhring, line 32: (repeated for all similar cases) For readability, it's nice to have a space between the `//` and the comment text. I see this is already done in some places, and I'd recommend making all comments like this

grype/matcher/apk/matcher.go
```diff
@@ -135,3 +135,81 @@ func vulnerabilitiesByID(vulns []vulnerability.Vulnerability) map[string][]vulne
 
 	return results
 }
+
+func (m *Matcher) findApkPackage(store vulnerability.Provider, d *distro.Distro, p pkg.Package) ([]match.Match, error) {
```
luhring, line 139: Praise: It's awesome to break this out into a distinct function! 🎉

grype/matcher/apk/matcher.go
```diff
@@ -135,3 +135,81 @@ func vulnerabilitiesByID(vulns []vulnerability.Vulnerability) map[string][]vulne
...
+	var matches = make([]match.Match, 0)
```
luhring, line 140:
```suggestion
	var matches []match.Match
```

This syntax does _almost_ the same thing, except 1) it's a bit more idiomatic for empty slices, and 2) it helps to avoid unnecessary allocation

luhring: Also — nit: You might consider moving the declaration of `matches` closer to where it's first used

grype/matcher/apk/matcher.go
```diff
@@ -135,3 +135,81 @@ func vulnerabilitiesByID(vulns []vulnerability.Vulnerability) map[string][]vulne
...
+
+func (m *Matcher) matchBySourceIndirection(store vulnerability.Provider, d *distro.Distro, p pkg.Package) ([]match.Match, error) {
+	metadata, ok := p.Metadata.(pkg.ApkMetadata)
+	if !ok {
+		return nil, nil
+	}
+
+	// ignore packages without source indirection hints or where source name is identical to package name
+	if metadata.OriginPackage == "" || metadata.OriginPackage == p.Name {
+		return []match.Match{}, nil
```
luhring, line 170:
```suggestion
		return nil, nil
```

(empty slices in Go are expressed as `nil`, unless there's a special reason for an allocated instance)

grype/matcher/apk/matcher.go
```diff
@@ -135,3 +135,81 @@ func vulnerabilitiesByID(vulns []vulnerability.Vulnerability) map[string][]vulne
...
+	}
+
+	// use source package name for exact package name matching
+	var indirectPackage pkg.Package
```
luhring, line 174: Is this a comment for this line of code? I don't think I follow

grype/matcher/apk/matcher.go
```diff
@@ -135,3 +135,81 @@ func vulnerabilitiesByID(vulns []vulnerability.Vulnerability) map[string][]vulne
...
+	err := copier.Copy(&indirectPackage, p)
+	if err != nil {
+		return nil, fmt.Errorf("failed to copy package: %w", err)
+	}
+
+	// use the source package name
+	indirectPackage.Name = metadata.OriginPackage
+
+	//For each cpe, replace pkg name with origin and add to set
+	CPEStrings := strset.New()
```
luhring, line 185:
```suggestion
	cpeStrings := strset.New()
```

grype/matcher/apk/matcher.go
```diff
@@ -135,3 +135,81 @@ func vulnerabilitiesByID(vulns []vulnerability.Vulnerability) map[string][]vulne
...
+		CPEStrings.Add(updatedCPEString)
+	}
+
+	//With each entry in set, convert string to CPE and update indirectPackage CPEs
+	updatedCPEs := make([]syftPkg.CPE, 0)
+	for _, CPEString := range CPEStrings.List() {
+		updatedCPE, _ := syftPkg.NewCPE(CPEString)
+		updatedCPEs = append(updatedCPEs, updatedCPE)
+	}
+	indirectPackage.CPEs = updatedCPEs
```
luhring, line 197: I'm wondering if this function could be improved by factoring out this section of code. As I'm reading this function, it seems like the function's task can be decomposed into other identifiable tasks, e.g.:

1. compute an "indirect package"
2. try to use that package to find matches
3. map these matches back to the original package as "indirect matches"

Is this a correct summary? And if it is, could we factor out step 1 into its own function?

Let me know if this would be easier to think through in a sync up!

grype/matcher/apk/matcher.go
```diff
@@ -135,3 +135,81 @@ func vulnerabilitiesByID(vulns []vulnerability.Vulnerability) map[string][]vulne
...
+
+	//For each cpe, replace pkg name with origin and add to set
+	CPEStrings := strset.New()
+	for _, cpe := range indirectPackage.CPEs {
+		updatedCPEString := strings.Replace(cpe.BindToFmtString(), p.Name, indirectPackage.Name, -1)
+		CPEStrings.Add(updatedCPEString)
+	}
+
+	//With each entry in set, convert string to CPE and update indirectPackage CPEs
+	updatedCPEs := make([]syftPkg.CPE, 0)
```
luhring, line 192:
```suggestion
	var updatedCPEs []syftPkg.CPE
```

grype/matcher/apk/matcher.go
```diff
@@ -135,3 +135,81 @@ func vulnerabilitiesByID(vulns []vulnerability.Vulnerability) map[string][]vulne
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to find vulnerabilities by apk source indirection: %w", err)
+	}
+
+	// we want to make certain that we are tracking the match based on the package from the SBOM (not the indirect package)
+	// however, we also want to keep the indirect package around for future reference
+	for idx := range matches {
+		matches[idx].Package = p
+
+		if matches[idx].Type == match.ExactDirectMatch {
```
luhring, line 209: Is it important that we perform this check first before setting `matches[idx].Type` to `match.ExactIndirectMatch`? If so, why?

luhring (changes requested):
Hey Zane! Thanks for taking this on. I synced up with Alex, and we agree this is quite a challenging task, but I'm impressed you're tackling it!

Let me know if you want to talk through any of the comments on this — more than happy to chat. 😃

grype/matcher/apk/matcher.go
zburstein: I believe that this check needs to be done because the matches slice contains findings against both secdb and nvd for source package. The ones found against secdb for source package should be changed to ExactIndirectMatch, while the ones found against nvd for source pacakge should remain FuzzyMatches. There is no type that reflect a fuzzy match made against source package so @wagoodman suggested I make https://github.com/anchore/grype/issues/408 to reflect this.

grype/matcher/apk/matcher.go
zburstein: Thanks!

grype/matcher/apk/matcher.go
zburstein: Good to know! Updated both suggestions!

grype/matcher/apk/matcher.go
zburstein: All makes sense! Moved building the indirect package into its own function

grype/matcher/apk/matcher.go
zburstein: I think i misplaced this on a copy paste

grype/matcher/apk/matcher.go
```diff
@@ -135,3 +135,90 @@ func vulnerabilitiesByID(vulns []vulnerability.Vulnerability) map[string][]vulne
...
+			matches[idx].Type = match.ExactIndirectMatch
+		}
+	}
+
+	return matches, nil
+}
+
+func buildIndirectPackage(p pkg.Package, metadata pkg.ApkMetadata) (pkg.Package, error) {
```
luhring, line 197: Could we get the `metadata` value within this function's body (akin to what's being done in lines 163-171 in the current state), and then the function needs only the `p pkg.Package` parameter? My understanding is that `metadata` is only needed as a means of computing the indirect package...

grype/matcher/apk/matcher.go
luhring: Great improvement — one question here (I should've just put in this thread): https://github.com/anchore/grype/pull/407#discussion_r704735378

grype/matcher/apk/matcher.go
luhring: Ahhhh — that makes complete sense. I forgot about the "fuzzy" value that's possible here, and it sounds like this is a good solution given the non-ideal state of the pre-#408 world.

luhring (approved):
Nice work! 🎉

Issue #315, opened 2021-05-17
Add support for Suse Linux Enterprise Server (SLES)


nwl:
Add support for generating list of CVEs for SLES (and OpenSUSE assuming it uses the same feed?) based on published data at https://www.suse.com/support/security/

wagoodman:
required syft changes:
-  we already added [opensuse as a distro type](https://github.com/anchore/syft/blob/main/syft/distro/type.go#L19) for leap ... we could additionally add the rolling release variant, tumbleweed (a single line addition + tests)

required grype changes:
- we can use the same RPMDB matcher we have today without changes (the package type is used to figure the right macher to use, so we're set there). Other than that, We just need to add a namespace mapper (a few of lines).
 
Once the feed data is available in grype-db we'll be able to pick this up 👍

luhring:
Ticket in the feed service that this work depends on: https://anchore.atlassian.net/browse/ENTERPRISE-265

nwl:
This can be closed no?

luhring:
Validated using image `registry.suse.com/suse/sle15:15.1.6.2.508`, seeing vulnerabilities with data source values like `https://www.suse.com/security/cve/CVE-2021-22946`.

PR #413, opened 2021-09-08
Update grype-db dependency, add some SLES tests

dspalmer99:
This PR updates the version of the syft library used in grype to support SLES distro vuln data. I'm marking this as a draft PR until that data is available in the feed service and I can test with it.

This Pr makes the following changes:

- Bumps the grype-db version to get the SLES distro changes
- Adds a unit and integration tests to verify that matching against SLES distro vulns works as expected.

See anchore/grype#315 for more context on this change.

Signed-off-by: Dan Palmer <dan.palmer@anchore.com>


grype/matcher/common/distro_matchers_test.go
```diff
@@ -53,6 +53,33 @@ func (pr *mockDistroProvider) stub() {
 			},
 		},
 	}
+	pr.data["sles:12.5"] = map[string][]vulnerability.Vulnerability{
+		// direct...
+		"sles": {
```
wagoodman, line 58: nit: the key to this map should be a package name. What you've selected here will work given the package name used to validate is called `sles`, however, it's a little surprising the see the distro name in place of the package name.

grype/matcher/common/distro_matchers_test.go
```diff
@@ -53,6 +53,33 @@ func (pr *mockDistroProvider) stub() {
...
+			{
+				Constraint: version.MustGetConstraint("< 2014.1.5-6", version.RpmFormat),
+				ID:         "CVE-2014-fake-4",
+				Namespace:  "sles:12.5",
+			},
+		},
+		// indirect...
+		"sles-ltss": {
```
wagoodman, line 66: nit: this key isn't being used and could probably be removed. It looks like the above `neutron-devel` case isn't needed either (I think this was added back when there was an intention of doing source-package-indirect searching down in the common distro matcher, but they settled up in each distro matcher instead (deb & rpm)).

wagoodman (approved):

grype/matcher/common/distro_matchers_test.go
dspalmer99: I've made that change.

grype/matcher/common/distro_matchers_test.go
dspalmer99: I've made that change as well, and I removed the existing, unused `neutron-devel` test data.

PR #420, opened 2021-09-16
Enhance version cmd

spiffcs:
To Test run
```
make bootstrap
make build
```

Locally you should see syft version populated in the build:
Screen Shot 2021-09-16 at 10 13 07 AM

When you run:
`./snapshot/grype-macos_darwin_amd64/grype version`

You should see syft version populated in the output:
Screen Shot 2021-09-16 at 10 18 06 AM

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ BOOTSTRAP_CACHE="c7afb99ad"
 DISTDIR=./dist
 SNAPSHOTDIR=./snapshot
 GITTREESTATE=$(if $(shell git status --porcelain),dirty,clean)
-SYFTVERSION=$(shell go list -m all | grep github.com/anchore/syft | awk '{print $2}')
+SYFTVERSION=$(shell go list -m all | grep github.com/anchore/syft | awk '{print $$2}')
 OS := $(shell uname)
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -21,6 +21,7 @@ BOOTSTRAP_CACHE="c7afb99ad"
 DISTDIR=./dist
 SNAPSHOTDIR=./snapshot
 GITTREESTATE=$(if $(shell git status --porcelain),dirty,clean)
+SYFTVERSION=$(shell go list -m all | grep github.com/anchore/syft | awk '{print $$2}')
```
luhring, line 24: I don't know if this is any safer than what you've already done, but there's a "machine readable" flow using JSON, too, like this:

```
go list -m -json all | jq --raw-output 'select(.Path == "github.com/anchore/syft").Version'
```

luhring (approved):
Nice addition! 🎉

Makefile
spiffcs: Sorry I just saw this comment. Happy to make the change if the taskrunners have jq installed. I used awk because I was more certain that would be available in the workflow by default than something like jq.

PR #421, opened 2021-09-17
Update KB constraint to not satisfy if raw constraint is empty.

Vijay-P:
Per validation with legacy MSRC matching in Engine and current understanding of how MSRC vulnerabilities are reported, an image should not be vulnerable to a vulnerability if the match constraint is empty. This PR updates the kb constraint to match this understanding.

### DIFF
--- a/grype/version/kb_contraint.go
+++ b/grype/version/kb_contraint.go
@@ -37,14 +37,10 @@ func (c kbConstraint) supported(format Format) bool {
 }
 
 func (c kbConstraint) Satisfied(version *Version) (bool, error) {
-	if c.raw == "" && version != nil {
-		// an empty constraint is always satisfied
-		return true, nil
+	if c.raw == "" {
+		// an empty constraint is never satisfied
+		return false, nil
 	} else if version == nil {
-		if c.raw != "" {
-			// a non-empty constraint with no version given should always fail
-			return false, nil
-		}
 		return true, nil
 	}
 
--- a/grype/version/kb_contraint_test.go
+++ b/grype/version/kb_contraint_test.go
@@ -7,8 +7,9 @@ import (
 
 func TestVersionKbConstraint(t *testing.T) {
 	tests := []testCase{
-		{version: "", constraint: "", satisfied: true},
+		{version: "", constraint: "", satisfied: false},
 		{version: "", constraint: "foo", satisfied: false},
+		{version: "878787", constraint: "", satisfied: false},
 		{version: "1", constraint: "foo", satisfied: false},
 		{version: "1", constraint: "1", satisfied: true},
 		{version: "878787", constraint: "979797 || 101010 || 878787", satisfied: true},


spiffcs:
LGTM given the context in the PR - Is there an issue this is linked to in github?

Vijay-P:
> LGTM given the context in the PR - Is there an issue this is linked to in github?

There is not an issue. We discovered this yesterday evening when Dan Nurmi was testing msrc images with the new grype integration.

spiffcs:
No worries - just making sure I'm on the same page here is `MSRC` <- (Microsoft Security Response Center) image?
Is there a sample image you guys tested yesterday I can run to validate this change against?

grype/version/kb_contraint.go
```diff
@@ -37,14 +37,10 @@ func (c kbConstraint) supported(format Format) bool {
 }
 
 func (c kbConstraint) Satisfied(version *Version) (bool, error) {
-	if c.raw == "" && version != nil {
```
wagoodman, line 40: nit: since this is surprising relative to other constraints defined, can we add more details in comments as to why this is different?

grype/version/kb_contraint.go
```diff
@@ -37,14 +37,10 @@ func (c kbConstraint) supported(format Format) bool {
...
-		// an empty constraint is always satisfied
-		return true, nil
+	if c.raw == "" {
+		// an empty constraint is never satisfied
```
luhring, line 41: @wagoodman What are your thoughts here? My understanding is that this is inconsistent with how version constraints work everywhere else in Grype, and if so, I'm concerned this might be confusing to Grype devs.

luhring: Whoa — https://github.com/anchore/grype/pull/421#discussion_r711279804

wagoodman (commented):
Can we add a case that tests this behavior in https://github.com/anchore/grype/blob/main/grype/matcher/msrc/matcher_test.go ? That would make the feature need more apparent

grype/matcher/common/distro_matchers.go
```diff
@@ -32,7 +35,12 @@ func FindMatchesByPackageDistro(store vulnerability.ProviderByDistro, d *distro.
 		// if the constraint it met, then the given package has the vulnerability
 		isPackageVulnerable, err := vuln.Constraint.Satisfied(verObj)
 		if err != nil {
-			return nil, fmt.Errorf("distro matcher failed to check constraint='%s' version='%s': %w", vuln.Constraint, verObj, err)
+			var e *version.NonFatalConstraintError
+			if errors.As(err, &e) {
+				log.Warn(e.Error())
```
luhring, line 40: nit: Typically in Go, you don't need to call `.Error()` on errors. When passing errors to functions that take `interface{}`, like for formatting strings, `.Error()` will be called automatically if needed. So you can just pass the error itself (`e` in this case).

grype/version/constraint.go
```diff
@@ -40,3 +40,13 @@ func MustGetConstraint(constStr string, format Format) Constraint {
 	}
 	return constraint
 }
+
+type NonFatalConstraintError struct {
```
luhring, line 44: Requesting change: It would be helpful to other devs if we added a doc comment for this error, explaining why it exists / how it should be used.

grype/version/kb_contraint.go
```diff
@@ -37,14 +37,14 @@ func (c kbConstraint) supported(format Format) bool {
...
+		return false, &NonFatalConstraintError{
+			constraint: c,
+			version:    version,
+			message:    "Unexpected data in DB: Empty raw version constraint.",
 		}
+	} else if version == nil {
 		return true, nil
 	}
```
luhring, line 49: nit: Since you're returning in the previous `if` block, you don't need the `else` here, so you could express this check as a distinct thought. For example:

```suggestion
	}
	
	if version == nil {
		return true, nil
	}
```

grype/version/kb_contraint_test.go
```diff
@@ -1,28 +1,54 @@
...
+			if test.shouldErr {
+				if test.errorAssertion != nil {
+					test.errorAssertion(t, err)
+				} else {
+					assert.Error(t, err)
+				}
+			} else {
+				assert.NoError(t, err, "unexpected error from kbConstraint.Satisfied: %v", err)
+			}
+			assert.Equal(t, test.satisfied, isSatisfied, "unexpected constraint check result: expected %+v, got %+v", test.satisfied, isSatisfied)
```
luhring, line 51: nit: The `assert.Equal` function should show a reasonable helpful explanation of what was expected vs. actual on its own, so you typically don't need to restate those values in the error message. But occasionally the `assert` functions output text that's not as helpful, so if you found that for this case, I think it makes sense to leave what you have.

luhring (changes requested):
Great work. Most comments are just `nit`, but I added one hard request for change. Let me know what you think!

luhring (approved):
LGTM!

Issue #419, opened 2021-09-16
False positive on Centos/Rhel openjdk package


dpippenger:
**What happened**:
Grype appears to be unable to accurately evaluate redhat's openjdk version strings resulting in false positives on the openjdk packages. Sometimes even showing the fixed version is the exact version installed already.

```
NAME                         INSTALLED                FIXED-IN                 VULNERABILITY        SEVERITY 
java-1.8.0-openjdk-headless  1:1.8.0.302.b08-0.el7_9  1:1.8.0.51-1.b16.el7_1   CVE-2015-4733        Critical  
java-1.8.0-openjdk-headless  1:1.8.0.302.b08-0.el7_9  1:1.8.0.51-1.b16.el7_1   CVE-2015-4748        Medium    
java-1.8.0-openjdk-headless  1:1.8.0.302.b08-0.el7_9  1:1.8.0.292.b10-1.el7_9  CVE-2021-2163        Medium    
java-1.8.0-openjdk-headless  1:1.8.0.302.b08-0.el7_9  1:1.8.0.65-2.b17.el7_1   CVE-2015-4911        Medium    
java-1.8.0-openjdk-headless  1:1.8.0.302.b08-0.el7_9  1:1.8.0.302.b08-0.el7_9  CVE-2021-2388        High      
java-1.8.0-openjdk-headless  1:1.8.0.302.b08-0.el7_9  1:1.8.0.302.b08-0.el7_9  CVE-2021-2369        Medium    
java-1.8.0-openjdk-headless  1:1.8.0.302.b08-0.el7_9  1:1.8.0.302.b08-0.el7_9  CVE-2021-2341        Medium    

```

**What you expected to happen**:
Report accurate CVE data on redhat openjdk

**How to reproduce it (as minimally and precisely as possible)**:
Build this minimal Dockerfile and analyze it with grype with default settings.

```
FROM centos:7
RUN yum install -y java-1.8.0-openjdk-headless
```

**Anything else we need to know?**:
Same issue present in other variations of containers using redhat openjdk such as running...
`grype registry.access.redhat.com/redhat-openjdk-18/openjdk18-openshift:latest
`

**Environment**:
```
$ grype version
Application:          grype
Version:              0.19.0
BuildDate:            2021-09-15T14:17:43Z
GitCommit:            7c082c2162fa36f9fd0cd2b34dfcb86348036275
GitTreeState:         clean
Platform:             linux/amd64
GoVersion:            go1.16.8
Compiler:             gc
[...]

PR #424, opened 2021-09-22
Bugfixes + Integration test for sbom input vs grype library comparison

Vijay-P:
There is one failure in the tests right now, and that should get fixed by https://github.com/anchore/grype/pull/425.
This PR adds an integration test which compares the results of running grype against an image when the input comes from a syft sbom vs when the input comes from syft as a library.
The need for these tests comes from having discovered that the code that unmarshals the syft json was missing a handler for alpine metadata, and the code that transforms the metadata from syft as a library into grype's sbom format was missing the epoch field in RPM metadata.
The hope and intent of this test is to help developers catch scenarios where differences in sbom translation between both inputs cause vulnerability matches to get dropped.
The PR also includes bugfixes for the issues mentioned above.

Thanks to @wagoodman for pairing on initial implementation.

fixes: https://github.com/anchore/grype/issues/419

### DIFF
--- a/.bouncer.yaml
+++ b/.bouncer.yaml
@@ -6,4 +6,4 @@ permit:
   - ISC
 ignore-packages:
   # packageurl-go is released under the MIT license located in the root of the repo at /mit.LICENSE
-  - github.com/package-url/packageurl-go
+  - github.com/anchore/packageurl-go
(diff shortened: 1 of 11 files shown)


luhring (approved):
Looks great!

Issue #417, opened 2021-09-15
Grype raise error: "failed to catalog: failed to parse CPE"


phocheng:
**What happened**:
I used syft to get bom file. then use command "grype sbom:a.txt -o json", it exits with below error:
[0000] ERROR failed to catalog: failed to parse CPE="cpe:2.3:a:http://jcp_org/en/jsr/detail?id=173:WstxBundleActivator:4.4.1:*:*:*:*:*:*:*": unbind formatted string: unquoted '?' inside the FSB fragment "cpe:2.3:a:http://jcp_org/en/jsr/detail?id=173:WstxBundleActivator:4.4.1:*:*:*:*:*:*:*" (true, 38)

but if I directly use command "grype <image_tag> -o json", it can return correct result.

**What you expected to happen**:
can generate expected vul file

**How to reproduce it (as minimally and precisely as possible)**:
1. copy attachment(a.txt) to any directory.  [a.txt](https://github.com/anchore/grype/files/7166081/a.txt)
2. go to that directory 
3. execute "grype sbom:a.txt -o json"

% grype sbom:a.bom -o json                                                                               
 ⠙ Vulnerability DB        [checking for update]
[0000] ERROR failed to catalog: failed to parse CPE="cpe:2.3:a:http://jcp-org/en/jsr/detail?id=173:WstxBundleActivator:4.4.1:*:*:*:*:*:*:*": unbind formatted string: unquoted '?' inside the FSB fragment "cpe:2.3:a:http://jcp-org/en/jsr/detail?id=173:WstxBundleActivator:4.4.1:*:*:*:*:*:*:*" (true, 38)

**Anything else we need to know?**:
None

**Environment**:
- Output of `grype version`: 0.17.0 & 0.18.0
- OS (e.g: `cat /etc/os-release` or similar): both MacOS Big Sur(11.6) and Ubuntu 20.04

luhring:
Hi @phocheng, thanks for the report. This might actually be a bug with [Syft](https://github.com/anchore/syft)'s CPE generation...

In case it's helpful to others, I was able to isolate the artifacts that have these suspicious CPEs with this `jq` command:

```
cat ./a.txt | jq '.artifacts | map(select(any(.cpes[]; contains("http://jcp-org/en/jsr/detail"))))'
```

luhring:
@wagoodman Can you take a peek at this when you get a chance?

nightfurys:
[...]

PR #425, opened 2021-09-22
Allow CPE parsing failures

luhring:
Partially addresses #417.

When Grype reads in a user-supplied SBOM from Syft, it needs to convert CPE strings into CPE objects. When malformed CPE strings are supplied, Grype errored out and provided no useful results.

This PR adjusts the CPE parsing logic (within Syft JSON parsing) to log a warning instead of returning an error in the scenario where a user-supplied CPE string cannot be parsed successfully, allowing the logic to continue to other CPEs and other packages that don't have any parsing issues.

**Note:** We should also make an adjustment in Syft to avoid the creation of malformed CPE strings.

### DIFF
--- a/grype/cpe/cpe.go
+++ b/grype/cpe/cpe.go
@@ -1,19 +1,22 @@
 package cpe
 
 import (
+	"github.com/anchore/grype/internal/log"
 	"github.com/anchore/syft/syft/pkg"
 )
 
 func NewSlice(cpeStrs ...string) ([]pkg.CPE, error) {
-	ret := make([]pkg.CPE, len(cpeStrs))
-	for idx, c := range cpeStrs {
+	var cpes []pkg.CPE
+	for _, c := range cpeStrs {
 		value, err := pkg.NewCPE(c)
 		if err != nil {
-			return nil, err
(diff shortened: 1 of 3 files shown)


grype/cpe/cpe.go
```diff
@@ -1,19 +1,22 @@
...
 
 func NewSlice(cpeStrs ...string) ([]pkg.CPE, error) {
-	ret := make([]pkg.CPE, len(cpeStrs))
-	for idx, c := range cpeStrs {
+	var cpes []pkg.CPE
+	for _, c := range cpeStrs {
 		value, err := pkg.NewCPE(c)
 		if err != nil {
-			return nil, err
+			log.Warnf("unable to hydrate CPE for string %q, omitting from result CPE slice: %v", c, err)
```
kzantow, line 13: maybe "hydrate" -> "parse" ? or less verbose: "Excluding invalid CPE %q: %v"?

kzantow (approved):
Just one suggestion about logging verbosity, otherwise LGTM

grype/cpe/cpe.go
luhring: @kzantow I like that! I'll create a PR for this

PR #428, opened 2021-09-24
Add announcement for KubeCon meetup

luhring:
Note: Also removes the warning about `zsh` completion, since we now support zsh.


kzantow (approved):
👍

PR #432, opened 2021-09-28
Add default matcher (language + CPE matching)

wagoodman:
Today if you input an SBOM into grype and there is a package with an unknown type then no vulnerability matching is done for that package. This PR adds a "stock" matcher (note: "default" is a reserve word in go); now any package will at least pass through this new stock matcher, which does CPE and language based matching.


spiffcs (approved):
PAIRED LGTM

PR #434, opened 2021-09-28
Add data-driven language matching

wagoodman:
Pulls in the upstream grype-db change for allow for using a "good" default language namespace guess when searching for vulnerabilities for potentially supported languages.

Additionally this PR fixes version comparisons for the fuzzy version constraint checks where one side includes a `v` prefix while the other side does not. This is a common case for the `docker` golang package since docker changed the version scheme after the first few versions.


grype/version/fuzzy_constraint.go
```diff
@@ -181,3 +183,7 @@ func leftPad(s string, n int) string {
 	sb.WriteString(s)
 	return sb.String()
 }
+
+func stripLeadingV(ver string) string {
```
luhring, line 187: I love breaking this out 👍

luhring (approved):
Looks good!

(I too have an outstanding Grype PR if you haven't seen it yet 😃 )

Issue #267, opened 2021-03-31
Unable to invoke grype as an external process since 0.8.0


wlfshmn:
**What happened**:

I'm invoking grype among other tools from some java code, which has worked up to an including 0.7.0 (from at least 0.2.0), but stopped working suddenly in 0.8.0

Grype execution on the command-line still works as intended.

**What you expected to happen**:

I would expect a normal execution and the related output, instead the grpye process appears to hang indefinitely.

**How to reproduce it (as minimally and precisely as possible)**:

The following minimal java code demonstrated the issue

```
import java.io.IOException;

public class GrypeExecutionTest {

  public static void main(String[] args) {
    try {

      ProcessBuilder builder = new ProcessBuilder("/usr/local/bin/grype-0.7.0/grype", "-o", "json", "alpine:3.7"); // works
//        ProcessBuilder builder = new ProcessBuilder("/usr/local/bin/grype-0.8.0/grype", "-o", "json", "alpine:3.7"); // does not work
//        ProcessBuilder builder = new ProcessBuilder("/usr/local/bin/grype-0.9.0/grype", "-o", "json", "alpine:3.7"); // does not work
//        ProcessBuilder builder = new ProcessBuilder("/bin/bash", "-c", "/usr/local/bin/grype-0.8.0/grype -o json alpine:3.7"); // shell based invocation that don't work either (post 0.7)
//        ProcessBuilder builder = new ProcessBuilder("/bin/bash", "-c", "/usr/local/bin/grype-0.8.0/grype -o json alpine:3.7 2>&1 > output.json"); // shell based invocation with redirect that don't work either (post 0.7)

      builder.inheritIO();
      Process process = builder.start();

      process.waitFor();

    } catch (IOException | InterruptedException e) {
      e.printStackTrace();
    }
  }
}
```

**Anything else we need to know?**:

I'm suspecting grype has started to expect something from the environment that is provided on an interactive shell but isn't provided when executed as a subprocess.

I've tested this both in environments with and without http/https proxies, and they do not appear to be at issue

**Environment**:

```
Application:          grype
Version:              0.7.0
BuildDate:            2021-01-28T14:03:23Z
GitCommit:            8344b8f0d3f61729cf0845c08b31f26103e21231
GitTreeState:         clean
Platform:             linux/amd64
GoVersion:            go1.14.14
Compiler:             gc
Supported DB Schema:  1
```
```
Application:          grype
Version:              0.8.0
[...]

PR #440, opened 2021-10-02
Retrieve target from directory sbom types in addition to image types

### LINKED ISSUE #298: Cannot handle Syft SBOM for directory scans
**What happened**:
Grype wasn't able to handle input of a Syft SBOM JSON document, when the Syft scan was of a _directory_ (as opposed to an image).

Grype should be able to process directory SBOMs just as well as image SBOMs.

```shell
# Create the SBOM
syft -o json dir:./some/directory > ./dir.syft.json

# Reproduce via STDIN
cat ./dir.syft.json | grype
# [0000] ERROR failed to catalog: unable to process input for scanning: ''

# Reproduce via SBOM scheme
grype sbom:./dir.syft.json
# [0000] ERROR failed to catalog: unable to process input for scanning: 'sbom:./dir.syft.json'
```

This seems to be reproducible with any recent version of Syft, and with Grype as of v0.8.0. Image scans seem unaffected by this.

**Environment**:
- Output of `grype version`:
```
Application:          grype
Version:              0.10.2
BuildDate:            2021-04-14T22:19:15Z
GitCommit:            48f12f7b7aac0ff5c3d30a790d23088281aa3aef
GitTreeState:         clean
GoVersion:            go1.16.3
Compiler:             gc
Supported DB Schema:  2
```
macOS 11.2.3 (20D91)

dakaneye:
Partially Fixes #298 


spiffcs:
@dakaneye I tried running your PR locally to see if it fixed the highlighted issue. 

```
syft -o json ./ > test.json // generate directory output from grype

cat test.json | grype 
[0000] ERROR failed to catalog: unable to process input for scanning: ''

grype sbomb:./test.json 
[0000] ERROR failed to catalog: unable to process input for scanning: 'sbomb:./test.json'
```

Do you have an example on your end where this worked?
Screen Shot 2021-10-04 at 9 24 04 AM

wagoodman:
@spiffcs heads up: the screen shot has `sbomb` instead of `sbom` as an input.

spiffcs:
Ahhh - That would explain why I should go make more coffee this AM. Apologies please disregard that test.

spiffcs:
Confirmed this works using `sbom:./<input>`. 

Do we also want to try and include the `STDIN` path in this PR?
```
cat ./dir.syft.json | grype
```

luhring:
> Confirmed this works using `sbom:./<input>`.
> 
> Do we also want to try and include the `STDIN` path in this PR?
> 
> ```
> cat ./dir.syft.json | grype
> ```

Possibly — although I agree with @dakaneye's [comment](https://github.com/anchore/grype/issues/298#issuecomment-932726227) that the broken piping might be a separate issue: #267. I plan to spend some time looking into #267, so I'm comfortable with this PR fixing just the `sbom:` method.

spiffcs (approved):

luhring:
It'd be nice to have an automated test for what was fixed in this PR, though... 🤔

spiffcs:
> > Confirmed this works using `sbom:./<input>`.
> > Do we also want to try and include the `STDIN` path in this PR?
> > ```
> > cat ./dir.syft.json | grype
> > ```
> 
> Possibly — although I agree with @dakaneye's [comment](https://github.com/anchore/grype/issues/298#issuecomment-932726227) that the broken piping might be a separate issue: #267. I plan to spend some time looking into #267, so I'm comfortable with this PR fixing just the `sbom:` method.

Approved

luhring (changes requested):
@dakaneye Do you think we could add a test to catch this case? Maybe in https://github.com/anchore/grype/blob/main/test/cli/sbom_input_test.go ?

wagoodman:
@luhring I added a test case to cover the SBOM ingest from a directory scan.

luhring (approved):
@dakaneye @wagoodman teamwork makes the dream work! ❤️

PR #442, opened 2021-10-05
Add option to output vulnerability report to a file

### LINKED ISSUE #207: add an option to output the report into a file rather than redirecting the output
I would like to have an option to output the report (json) into a file with an option in the commad line like
grype "dir:/xxx" -o json -f report.json
**Why is this needed**:
it's easier with this option to integrate grype in python code rather than running and getting the output with subprocess
**Additional context**:

wagoodman:
This PR adds a new `--file` option that writes the report bytes out to the given file path consistent with https://github.com/anchore/syft/pull/530 .

Closes #207


cmd/report_writer.go
```diff
@@ -0,0 +1,29 @@
...
+		if err != nil {
+			return nil, nop, fmt.Errorf("unable to create report file: %w", err)
+		}
+		return reportFile, func() error {
+			if !appConfig.Quiet {
+				fmt.Printf("Report written to %q\n", path)
+			}
+			return reportFile.Close()
+		}, nil
+	}
```
luhring, line 28: very small nit: This is a good example of code that's well structured, but could be more readable if vertical spacing were leveraged more. Blank lines to separate subsections of thoughts can be really nice for readers looking to get a sense of structure, particular with regard to control flow — e.g., using blank lines to make `return` lines more clear.

luhring (approved):
Looks good!

cmd/report_writer.go
wagoodman: https://github.com/anchore/syft/pull/536 and https://github.com/anchore/grype/pull/443

PR #443, opened 2021-10-06
Update linter + fix whitespace

wagoodman:
Follow up from https://github.com/anchore/grype/pull/442#discussion_r722612680 . While I was at it I updated our linter.

### DIFF
--- a/internal/config/application.go
+++ b/internal/config/application.go
@@ -184,7 +184,6 @@ func (cfg Application) String() string {
 }
 
 // readConfig attempts to read the given config path from disk or discover an alternate store location
-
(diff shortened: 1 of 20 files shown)


internal/config/application.go
```diff
@@ -184,7 +184,7 @@ func (cfg Application) String() string {
 }
 
 // readConfig attempts to read the given config path from disk or discover an alternate store location
-// nolint:funlen
+
```
luhring, line 187: (same two cents as in https://github.com/anchore/syft/pull/536#discussion_r723178891)

luhring (approved):

PR #448, opened 2021-10-08
175 ignore no fix

spiffcs:
Add `--only-fixed` option to root command. Grype will now exit with status code `0` when passing this option if vulnerabilities are detected but have no upstream resolution.


spiffcs:
Thanks @kzantow! Updated with your comments

PR #454, opened 2021-10-12
175 patch - allow ignore not fixed to work independently of configured rules

spiffcs:

### Things to discuss on this PR before we merge the patch:
With the new logic `removing the need for rules` the flag behaves as intended:
```
go run main.go docker:gradle:6.8.3-jdk --only-fixed -v -o json > test.json
│[0009]  INFO identified distro: ubuntu 20.04 from-lib=syft
│[0009]  INFO cataloging image from-lib=syft
│[0012]  INFO Ignoring 54 matches due to user-provided ignore rules
```

The output above was with no rules and default config.

With this PR `Unknown`, `Fixed`, and `Wont-Fix` are still in the JSON/template output, but `not-fixed` is filtered out to be in the `ignoredMatches` field.

CycloneDX and Table have not been updated to reflect this change for the new flag OR for rules.

@luhring Do you want the `table.NewPresenter(matches, packages, metadataProvider)` to also be updated with the ignoredMatches?

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -34,9 +34,9 @@ import (
 var persistentOpts = config.CliOnlyOptions{}
 
 var ignoreNonFixedMatches = []match.IgnoreRule{
-	{FixedState: string(grypeDb.NotFixedState)},
-	{FixedState: string(grypeDb.WontFixState)},
-	{FixedState: string(grypeDb.UnknownFixState)},
+	{FixState: string(grypeDb.NotFixedState)},
+	{FixState: string(grypeDb.WontFixState)},
+	{FixState: string(grypeDb.UnknownFixState)},
 }
 
 var (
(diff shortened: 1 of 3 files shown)


kzantow:
> > There was an issue that when ignores were not specified, this code path was not hit, no? I don't see anything rectifying that (I may be wrong!)
> 
> ```
> go run main.go docker:gradle:6.8.3-jdk --only-fixed -v -o json | jq '.ignoredMatches | length' 
> [0009]  INFO identified distro: ubuntu 20.04 from-lib=syft
> [0009]  INFO cataloging image from-lib=syft
> [0012]  INFO Ignoring 133 matches due to user-provided ignore rules
> -> 133
> ```
> 
> It's checking to ignore these matches now. The issue was that `len(rules)` was 0 which has been removed as a condition for the `fixed` logic to run.

outside of this PR?

PR #494, opened 2021-11-07
Support for private certificate authorities during DB curation

### LINKED ISSUE #493: Custom CA support for db.update-url
When overriding [db.update-url](https://github.com/anchore/grype/blob/00aa7d452348cdd1a94b25a78d0d04c6ff3fff6d/internal/config/database.go#L14) I would like the option to specify a custom CA certificate.

**Why is this needed**:
The update-url might not present a certificate from a well-known CA, 
the CA might not be in the host's trust store,
and/or a self-signed cert might be used in a development environment.

**Additional context**:
To run grype in an airgapped (or restricted egress) environment I would like to deploy an internal "database mirror" and configure grype to download updates from that location. For local testing I would likely use a self signed cert. For production environments I might not have control of the node's CA trust store.

For development, an `insecure-skip-tls-verify` might be sufficient, but in real environments a secure connection is still necessary.

luhring:
Closes #493 

**Objective:** Add a config option to let users provide a custom CA (certificate authority) certificate that's used in the validation of TLS connections while Grype queries the remote database service. (similar to the `--cacert` option for `curl`)

**Tasks:**

- [x] Add ability to inject custom `http.Client` instances into Grype's `file.Getter` implementation used for DB curation
- [x] Add failing test for `file.Getter`, where the failure is caused by TLS validation for a remote cert signed by an untrusted CA
- [x] Fix failing test by using the private CA cert in the test's `http.Client`
- [x] Create new config item in `db` config to leverage new custom CA cert option for `file.Getter`

**Note:** I've made a few assumptions in this implementation, such as that we want to continue to use github.com/hashicorp/go-getter, and that we're okay with checking in test data for X.509 certificates into the repository. Feel free to challenge these assumptions, happy to adapt!


grype/db/curator_test.go
```diff
@@ -120,11 +124,54 @@ func TestCuratorDownload(t *testing.T) {
 			if string(actual) != contents {
 				t.Fatalf("bad contents: %+v", string(actual))
 			}
-
 		})
 	}
 }
 
+func TestCuratorDownload_PrivateCA(t *testing.T) {
```
luhring, line 131: This test name doesn't match the method it's testing. I'll adjust.

luhring:
@wagoodman I just noticed that access of the remote DB backend (S3) is divided between Grype and GrypeDB — they each have their own `curator` packages, and they each have their own way of invoking `github.com/hashicorp/go-getter`: Grype abstracts this with an internal `file.Getter` interface, whereas GrypeDB calls the library directly.

I wanted to get your thoughts on the reasons for this design before proceeding much further with changes. In the current this state, this complicates the implementation of changing how Grype handles its interaction with the DB backend.

grype/db/curator.go
```diff
@@ -64,6 +76,10 @@ func (c *Curator) GetStore() (*reader.Reader, error) {
 	return s, err
 }
 
+func (c Curator) SupportedSchema() int {
```
wagoodman, line 79: I'll remove this

grype/db/curator.go
```diff
@@ -312,3 +328,49 @@ func (c *Curator) activate(dbDirPath string) error {
...
+	// parse the listing file
+	listing, err := curation.NewListingFromFile(c.fs, tempFile.Name())
+	if err != nil {
+		return curation.Listing{}, err
+	}
+	return listing, nil
+}
+
+func defaultHTTPClient(fs afero.Fs, caCertPath string) (*http.Client, error) {
+	httpClient := cleanhttp.DefaultClient()
```
wagoodman, line 360: getter test should use this as well

grype/db/curator_test.go
```diff
@@ -59,18 +66,118 @@ func (g *testGetter) GetToDir(dst, src string, _ ...*progress.Manual) error {
...
+	tests := []struct {
+		name    string
+		hasCert bool
+	}{
+		{
+			name:    "no custom cert should use default system root certs",
+			hasCert: false,
+		},
+		{
+			name:    "use single custom cert",
```
wagoodman, line 94: + `should`

grype/db/test-fixtures/tls/README.md
```diff
@@ -0,0 +1,22 @@
+# TLS test utils
+
+Note: Makefile, server.crt, and server.key are used in automated testing, the remaining files are for convenience in manual verification.
```
wagoodman, line 3: note the need for python3

grype/db/test-fixtures/tls/Makefile
```diff
@@ -0,0 +1,45 @@
+all: clean serve
+
+.PHONY: serve
+serve: www/listing.json www/db.tar.gz server.crt
+	python serve.py
```
wagoodman, line 5: use python3 explicitly

grype/db/test-fixtures/tls/Makefile
```diff
@@ -0,0 +1,45 @@
...
+	./generate-x509-cert-pair.sh
+
+www:
+	mkdir -p www
+
+listing.json:
+	curl -L -O https://toolbox-data.anchore.io/grype/databases/listing.json
+
+www/listing.json www/db.tar.gz: www listing.json
+	$(eval location=$(shell python listing.py))
```
wagoodman, line 34: use python3 explicitly

internal/string_helpers.go
```diff
@@ -0,0 +1,25 @@
+package internal
+
+import "strings"
+
+// HasAnyOfSuffixes returns an indication if the given string has any of the given suffixes.
+func HasAnyOfSuffixes(input string, prefixes ...string) bool {
+	for _, prefix := range prefixes {
+		if strings.HasSuffix(input, prefix) {
```
wagoodman, line 8: rename var to suffix

grype/db/curator.go
wagoodman: unfortunately the getter tests don't create their own HTTP Client, they are retrieved from the httptest server object (`server.Client()`). So there isn't a nice/easy way to use this default client in testing.

wagoodman (approved):
We reviewed the PR synchronously and agreed on approving 👍

PR #498, opened 2021-11-15
Bump syft version => v0.30.1

spiffcs:
## Update grype to use syft v0.30.1
- Add a new exception for OR license for `github.com/spdx/tools-golang`
- update syft version in go.mod
- run `go mod tidy` and generate new go.sum
- update `syft_sbom_provider` to use new signature for generating SBOM from the library
- update integration tests to use new `Encode` `Decode` pattern
- add `PhpComposerPkg` to exception list for integration test

### DIFF
--- a/grype/pkg/syft_sbom_provider.go
+++ b/grype/pkg/syft_sbom_provider.go
@@ -24,7 +24,7 @@ func syftSBOMProvider(userInput string) ([]Package, Context, error) {
 		return nil, Context{}, err
 	}
 
-	catalog, srcMetadata, theDistro, _, formatOption, err := syft.Decode(reader)
+	sbom, formatOption, err := syft.Decode(reader)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


jonasagx (approved):

grype/pkg/syft_sbom_provider.go
```diff
@@ -24,17 +24,17 @@ func syftSBOMProvider(userInput string) ([]Package, Context, error) {
 		return nil, Context{}, err
 	}
 
-	catalog, srcMetadata, theDistro, _, formatOption, err := syft.Decode(reader)
+	sbom, formatOption, err := syft.Decode(reader)
```
nightfurys, line 27: I'm guessing the syft update was to return a data structure instead of individual components

nightfurys (approved):
Looks good to me!

PR #499, opened 2021-11-15
update to newest grype-db

spiffcs:


spiffcs:
I'll update the tests here

PR #500, opened 2021-11-16
Support vulnerability matching for Rocky Linux

nightfurys:
- Update grype-db dependency for the distro-feed namespace mapping
- Add test to verify the above mapping

Signed-off-by: Swathi Gangisetty <swathi@anchore.com>


spiffcs (approved):

spiffcs:
Nice addition for Rocky Linux

PR #506, opened 2021-11-22
Add db list command

wagoodman:
Adds the ability to list entries discovered available DBs from the configured update URL (`db.update-url`)"
```
❯ syft db list
Built:    2021-11-22 08:13:50 +0000 UTC
URL:      https://toolbox-data.anchore.io/grype/databases/vulnerability-db_v3_2021-11-22T08:13:50Z.tar.gz
Checksum: sha256:b45e70819ab48d469cf4bbdb8a1fb41673f3465730885c49a1779bcdda5e6357

Built:    2021-11-21 08:15:44 +0000 UTC
URL:      https://toolbox-data.anchore.io/grype/databases/vulnerability-db_v3_2021-11-21T08:15:44Z.tar.gz
Checksum: sha256:c402d01fa909a3fa85a5c6733ef27a3a51a9105b6c62b9152adbd24c08358911

...

Built:    2021-06-02 10:24:36 +0000 UTC
URL:      https://toolbox-data.anchore.io/grype/databases/vulnerability-db_v3_2021-06-02T10:24:36Z.tar.gz
Checksum: sha256:50bdd79010dafbaa08d11cd62c3e9b4c6c86182012bd4f6dbad0df662bfe0c26

Built:    2021-06-01 20:56:41 +0000 UTC
URL:      https://toolbox-data.anchore.io/grype/databases/vulnerability-db_v3_2021-06-01T20:56:41Z.tar.gz
Checksum: sha256:9721e633543bb663814fad30a6f84f8f5de2fedcaaaea42547490ffb18e1f360

171 databases available for schema 3
```

Supports the following formats (via `-o`):
- `text`: the format demoed above
- `raw`: the entire listing file from the `db.update-url`
- `json`: the subsection of the listing file pertaining to the current supported schema version

### DIFF
--- a/cmd/db_list.go
+++ b/cmd/db_list.go
@@ -39,8 +39,7 @@ func runDBListCmd(_ *cobra.Command, _ []string) error {
 	available, exists := listing.Available[supportedSchema]
 
 	if len(available) == 0 || !exists {
-		fmt.Printf("No databases available for the current schema (%d)\n", supportedSchema)
-		return nil
+		return stderrPrintLnf("No databases available for the current schema (%d)", supportedSchema)
 	}
 
 	switch dbListOutputFormat {
--- a/cmd/db_list.go
+++ b/cmd/db_list.go
@@ -1,0 +1,77 @@
+package cmd
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+
+	"github.com/anchore/grype/grype/db"
+	"github.com/spf13/cobra"
+)
+
+var dbListOutputFormat string
+
+var dbListCmd = &cobra.Command{
+	Use:   "list",
+	Short: "list all DBs available according to the listing URL",
+	Args:  cobra.ExactArgs(0),
+	RunE:  runDBListCmd,
+}
+
+func init() {
+	dbListCmd.Flags().StringVarP(&dbListOutputFormat, "output", "o", "text", "format to display results (available=[text, raw, json])")
+
+	dbCmd.AddCommand(dbListCmd)
+}
+
+func runDBListCmd(_ *cobra.Command, _ []string) error {
+	dbCurator, err := db.NewCurator(appConfig.DB.ToCuratorConfig())
+	if err != nil {
+		return err
+	}
+
+	listing, err := dbCurator.ListingFromURL()
+	if err != nil {
+		return err
+	}
+
+	supportedSchema := dbCurator.SupportedSchema()
+	available, exists := listing.Available[supportedSchema]
+
+	if len(available) == 0 || !exists {
+		fmt.Printf("No databases available for the current schema (%d)\n", supportedSchema)
+		return nil
+	}
+
+	switch dbListOutputFormat {
+	case "text":
+		// summarize each listing entry for the current DB schema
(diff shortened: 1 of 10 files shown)


cmd/db_list.go
```diff
@@ -0,0 +1,77 @@
...
+	listing, err := dbCurator.ListingFromURL()
+	if err != nil {
+		return err
+	}
+
+	supportedSchema := dbCurator.SupportedSchema()
+	available, exists := listing.Available[supportedSchema]
+
+	if len(available) == 0 || !exists {
+		fmt.Printf("No databases available for the current schema (%d)\n", supportedSchema)
```
luhring, line 42: nit: this may be subjective, but this seems like the kind of message that should be routed to stderr. What do you think?

cmd/db_list.go
```diff
@@ -0,0 +1,77 @@
...
+			return fmt.Errorf("failed to db listing information: %+v", err)
+		}
+	case "raw":
+		// show the entire listing file
+		enc := json.NewEncoder(os.Stdout)
+		enc.SetEscapeHTML(false)
+		enc.SetIndent("", " ")
+		if err := enc.Encode(&listing); err != nil {
+			return fmt.Errorf("failed to db listing information: %+v", err)
+		}
```
luhring, line 71: heads up — This may boil down to a naming concern.... When I saw the format's name and description:

>raw: the entire listing file from the db.update-url

I was expecting this to be a raw (i.e. never parsed) representation of the file content retrieved for the listing URL. I could see this being very useful as a utility command for folks wanting to obtain a raw listing file w/o using the `curl` approach or worrying about JSON data structure (or even that the listing file is still being represented as JSON).

My suggestion would be to do one of the following: (1) adjust the implementation to show the raw bytes of the listing file, or (2) rename the format option to something besides "raw", to match the existing implementation (e.g. "all").

Curious for your thoughts here!

luhring (commented):
Very useful! Nice addition! 👏 

I think we should also document this functionality in the README, perhaps in the ["Managing Grype's database"](https://github.com/anchore/grype#managing-grypes-database) section.

cmd/db_list.go
wagoodman: I'm a bit torn --yes, in the sense that does align with my opinion. But I wouldn't update it just for this command, I would also update all other commands (only the root command has this preference too, all of the others print to stdout  only).

cmd/db_list.go
wagoodman: I see your point, however, after lots of thought I can't seem to think of a better name. This seems to be the best candidate so far (anyone have a better name?) and the description documents the specific behavior.

luhring: Let's think more on this over time, and treat this as a non-blocking conversation!

luhring (approved):

cmd/db_list.go
wagoodman: added a helper and started using that in all aux commands

wagoodman:
@luhring I've updated the readme as suggested with the new `db list` command, the existing `db import`, and more context about air-gapped usage of these commands.

PR #507, opened 2021-11-22
Ignore partybus unsubscribe errors

wagoodman:
There is no need to log bus unsubscribe errors (e.g. "unable to find subscription to unsubscribe" errors) as this is not actionable.


cmd/event_loop.go
```diff
@@ -14,7 +14,6 @@ import (
 // eventLoop listens to worker errors (from execution path), worker events (from a partybus subscription), and
 // signal interrupts. Is responsible for handling each event relative to a given UI an to coordinate eventing until
 // an eventual graceful exit.
-// nolint:gocognit
```
luhring, line 17: 🙌

luhring (approved):

PR #514, opened 2021-12-02
Adding AlmaLinux OS Support

srbala:
Necessary changes to provide AlmaLinux OS Support


wagoodman (approved):
Thanks for your contribution, @srbala !

PR #518, opened 2021-12-03
Fix release quality gate section

wagoodman:
Problem introduced from refactor: https://github.com/anchore/grype/pull/512

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -82,7 +82,7 @@ jobs:
 
 
   release:
-    needs: [ wait-for-checks ]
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -82,7 +82,7 @@ jobs:
 
 
   release:
-    needs: [ wait-for-checks ]
+    needs: [ quality-gate ]
```
spiffcs, line 85: Relevant name can be sourced from line 15 of this file. Approved

spiffcs (approved):

PR #547, opened 2021-12-16
Fix panic in FromCatalog function

### LINKED ISSUE #548: Panic when a package has been added to a catalog multiple times
When Grype is using a catalog that's had a package added more than once, Grype panics in [`FromCatalog`](https://github.com/anchore/grype/blob/0781fb028b56dbb4cc9ce1c0cb8988753e6f3b76/grype/pkg/package.go#L90-L96).

This is because Grype is adding items to the `result` slice by **index**, and it's trusting that the slice's length (currently set to the return value of `catalog.PackageCount()`) is large enough to accommodate any index value `i` that come from ranging over `catalog.Sorted()`.

This is not a safe assumption, since it's possible for `catalog.Sorted()` to return more items than the return value of `catalog.PackageCount()` indicates.

This issue was first discovered here: https://github.com/anchore/grype/issues/538#issuecomment-994890110 by @avelezxv

Example panic output:

```
panic: runtime error: index out of range [1216] with length 1216

goroutine 14 [running]:
github.com/anchore/grype/grype/pkg.FromCatalog(0xc004d16450, 0x12e4cf0, 0x8, 0xc004d16450)
        /Users/runner/work/grype/grype/grype/pkg/package.go:93 +0x272
github.com/anchore/grype/grype/pkg.syftProvider(0x7ffefe2e3b51, 0x10, 0x12e4cf0, 0x8, 0xc000c1b700, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /Users/runner/work/grype/grype/grype/pkg/syft_provider.go:25 +0x15a
github.com/anchore/grype/grype/pkg.Provide(0x7ffefe2e3b51, 0x10, 0x12e4cf0, 0x8, 0xc000c1b700, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /Users/runner/work/grype/grype/grype/pkg/provider.go:20 +0x115
github.com/anchore/grype/cmd.startWorker.func1.2(0xc0005623f0, 0x7ffefe2e3b51, 0x10, 0xc00056f458, 0xc00043bdf0, 0xc0002928b0, 0xc0001a21e0, 0xc0005623e7)
        /Users/runner/work/grype/grype/cmd/root.go:254 +0x105
created by github.com/anchore/grype/cmd.startWorker.func1
        /Users/runner/work/grype/grype/cmd/root.go:251 +0x35f
```

luhring:
Fixes #548 

This PR prevents a panic in Grype that can occur in `pkg.FromCatalog`.

Since it's currently possible for `catalog.Sorted()` to return more items than the return value of `catalog.PackageCount()` indicates, this PR adjusts the loop such that:

- An `index out of range` panic is no longer possible, since slice items aren't being addressed by index.
- The **benefit** of avoiding unnecessary allocations for the `result` slice's underlying array is **preserved** for the case where the return value of `catalog.PackageCount()` is greater than or equal to `len(catalog.Sorted())` (which is the expected case).

### DIFF
--- a/grype/pkg/package.go
+++ b/grype/pkg/package.go
@@ -88,9 +88,9 @@ func New(p pkg.Package) Package {
 }
 
 func FromCatalog(catalog *pkg.Catalog) []Package {
-	var result = make([]Package, catalog.PackageCount())
-	for i, p := range catalog.Sorted() {
-		result[i] = New(p)
+	result := make([]Package, 0, catalog.PackageCount())
+	for _, p := range catalog.Sorted() {
+		result = append(result, New(p))
 	}
 	return result
 }
--- a/grype/pkg/package_test.go
+++ b/grype/pkg/package_test.go
@@ -3,6 +3,8 @@ package pkg
 import (
 	"testing"
 
+	"github.com/anchore/syft/syft/source"
+
 	"github.com/anchore/syft/syft/file"
 	syftPkg "github.com/anchore/syft/syft/pkg"
 	"github.com/scylladb/go-set"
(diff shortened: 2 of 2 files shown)


wagoodman (commented):
You mentioned that PackageCount() is returning a different number than what you get from Sorted() , this seems to be a notable problem as well. (I'll take a closer look shortly)

wagoodman (approved):

Issue #389, opened 2021-08-19
Can we support path filtering for file system scanning？


song787:
**Why is this needed**:
Some paths do not need to be scanned, so they need to be excluded

Thanks

luhring:
@song787 Good suggestion! We've talked about this, and this was even one possible solution to another issue (#183).

So we can better understand, what's the specific use case you have that would be addressed by this feature? (E.g. what kinds of paths don't need to be scanned?)

song787:
@luhring ok, the case I'm currently facing is this: 
I put grype in a container of the k8s cluster and scan the host for vulnerabilities through the container. However, the error of permission problem occurs when scanning the `/proc` directory, as shown in the code. This problem occurred in the process of syft building the file tree of userinput. I tried to run the container in privileged mode, but it didn't work. At present, there is no way to solve it, so I consider whether I can mask the `/proc` directory through path filtering. Do you have any solutions? Thanks!

`/opt # ./grype2 dir:/opt/hostpath/proc  `
` ⠹ Vulnerability DB        ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━  [409 kB / 57 MB]`
` ⠹ Indexing /opt/hostpath/proc [file: /opt/hostpath/proc/30541/task/30803/net/dev_snmp6/veth785453e1]`
`[0004] ERROR failed to catalog: unable to determine resolver while cataloging packages: unable to index filesystem path="/opt/hostpath/proc": unable to access path="/opt/hostpath/proc/309/net": readdirent: invalid argument
`

luhring:
@song787 Which version of Grype are you using?

I ask because we made a [change to not look at directories like `/proc`](https://github.com/anchore/syft/pull/442) a while back. This was incorporated into Syft v0.19.0, which was brought into Grype as of Grype v0.14.0.

song787:
@luhring My current version is Grype v0.15.0.
Have you blocked the `/proc` directory in v0.15.0?
ok, I just tested it. The `/proc` directory has indeed been masked by local scanning on the host. However, when you put Grype in the container and scan the host through the mount directory, Grype will still scan the `/proc` directory of the host.

luhring:
Thanks for the info! Just so I understand the use case: You're running Grype in a container, in which you've mounted `/proc` from the host as `/proc` in the container, and you're having Grype perform a directory scan?

Also, are you able to share the exact Grype command you're running?

[...]

PR #551, opened 2021-12-16
Add --exclude flag

kzantow:
This addresses #389 after the corresponding Syft PR (https://github.com/anchore/syft/pull/695) has been merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-<p style="text-align: center">
+<p align="center">
     <img alt="Grype logo" src="https://user-images.githubusercontent.com/5199289/136855393-d0a9eef9-ccf1-4e2b-9d7c-7aad16a567e5.png" width="234">
 </p>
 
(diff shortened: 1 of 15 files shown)


grype/pkg/provider.go
```diff
@@ -4,18 +4,68 @@ import (
...
+
 	"github.com/anchore/stereoscope/pkg/image"
 	"github.com/anchore/syft/syft/source"
 )
 
 var errDoesNotProvide = fmt.Errorf("cannot provide packages from the given source")
 
 // Provide a set of packages and context metadata describing where they were sourced from.
-func Provide(userInput string, scopeOpt source.Scope, registryOptions *image.RegistryOptions) ([]Package, Context, error) {
+func Provide(userInput string, scopeOpt source.Scope, registryOptions *image.RegistryOptions, exclusions ...string) ([]Package, Context, error) {
```
luhring, line 16: nit: It's not clear to me why it helps to make `exclusions` a variadic parameter instead of a `[]slice` (I don't see any code using this to specify exclusions using anything else besides a slice of strings).

luhring: (this comment can apply all instances of `exclusions ...string` in this diff)

grype/pkg/provider.go
```diff
@@ -4,18 +4,68 @@ import (
...
+			}
+		}
 		return packages, ctx, err
 	}
 
-	return syftProvider(userInput, scopeOpt, registryOptions)
+	return syftProvider(userInput, scopeOpt, registryOptions, exclusions...)
+}
+
+func filterPackageExclusions(packages []Package, exclusions []string) ([]Package, error) {
```
luhring, line 31: nit: I see that this code is exercised via new integration tests, but the logic is complex enough here that unit tests for this specific function directly might not be a bad idea

README.md
```diff
@@ -1,5 +1,5 @@
 <p align="center">
-    <img src="https://user-images.githubusercontent.com/5199289/136855393-d0a9eef9-ccf1-4e2b-9d7c-7aad16a567e5.png" width="234">
+    <img alt="Grype logo" src="https://user-images.githubusercontent.com/5199289/136855393-d0a9eef9-ccf1-4e2b-9d7c-7aad16a567e5.png" width="234">
```
luhring, line 2: Nice!

luhring (approved):
Looks good to me — just a couple of nit comments.

grype/pkg/provider.go
kzantow: This is true. it was some earlier iteration in Syft that made me think this was a good idea; I only wish I had changed Syft 🤦

grype/pkg/provider.go
kzantow: Added a number of new tests for the helper functions

wagoodman (approved):
nice work!

Issue #552, opened 2021-12-16
CVE-2021-44228 false positives


dsyer:
Scanned an image for a Spring Boot application (e.g. Petclinic https://github.com/spring-projects/spring-petclinic). It should be negative for CVE-2021-44228 since it only depends on log4j-api (not log4j-core). So either the vulnerability database is wrong, or the matching is buggy. An image that contains only log4j-api or log4j-slf4j-impl or log4j-to-slf4j is not vulnerable.

luhring:
Hi @dsyer, thanks for the issue!

You mention you're seeing this behavior for a scan of an image — is there an easy way for us to get a hold of that image? I don't see a Dockerfile in that repo, is there an image already published somewhere?

hughpv:
Here's an example I found on Docker hub - just happened to be the most recently uploaded image that matched on "spring boot": `sidharthpradhan/springbootdemo`

Included in the grype output is the following:

```
log4j-api             2.14.1                          CVE-2021-44228       Critical
log4j-to-slf4j        2.14.1                          CVE-2021-44228       Critical
```

But when I inspect the JAR file inside the container, log4j-core is not there:

```
root@b69a6685d5bd:/# unzip -lv websocket-demo-0.0.1-SNAPSHOT.jar  | grep log4j
   17762  Stored    17762   0% 2021-03-06 22:13 e603665f  BOOT-INF/lib/log4j-to-slf4j-2.14.1.jar
  300365  Stored   300365   0% 2021-03-06 22:11 ebd5bc63  BOOT-INF/lib/log4j-api-2.14.1.jar
```

luhring:
Wow, awesome, thanks @hughpv!

dsyer:
You can generate an image with `mvn spring-boot:build-image`. Use that if you want to try anything new.

tsunamaru:
If possible, please leave the current behavior as option, configurable somewhere (we are using grype for bulk scanning our images and extracting all log4j related stuff).

Tsuesun:
> If possible, please leave the current behavior as option, configurable somewhere (we are using grype for bulk scanning our images and extracting all log4j related stuff).

Why not use https://github.com/anchore/syft for that?

knweiss:
FWIW: You could also test this [log4j-samples](https://github.com/mergebase/log4j-samples) collection.

> Samples of log4j library versions to help log4j scanners / detectors improve their accuracy for detecting CVE-2021-45046 and CVE-2021-44228.

[...]

PR #559, opened 2021-12-21
Ignore explicit list of log4j false positive matches

kzantow:
This PR causes Grype to ignore known false positive log4j matches, fixes: https://github.com/anchore/grype/issues/552

TODO:

- [x] add tests
- [x] add more tests

### DIFF
--- a/grype/logger/logger.go
+++ b/grype/logger/logger.go
@@ -9,5 +9,4 @@ type Logger interface {
 	Info(args ...interface{})
 	Debugf(format string, args ...interface{})
 	Debug(args ...interface{})
-	IsDebugEnabled() bool
 }
--- a/grype/logger/logger.go
+++ b/grype/logger/logger.go
@@ -9,4 +9,5 @@ type Logger interface {
 	Info(args ...interface{})
 	Debugf(format string, args ...interface{})
 	Debug(args ...interface{})
+	IsDebugEnabled() bool
 }
--- a/grype/match/explicit_ignores.go
+++ b/grype/match/explicit_ignores.go
@@ -42,9 +42,9 @@ func ApplyExplicitIgnoreRules(matches Matches) Matches {
 	if log.IsDebugEnabled() {
 		removed := ""
 		for _, i := range ignored {
-			removed += i.Match.Vulnerability.ID + ":" + i.Package.PURL + ","
+			removed += "    - " + i.Match.Vulnerability.ID + ":" + i.Package.PURL + "\n"
 		}
-		log.Debugf("Removed explicit matches: %s", removed)
+		log.Debugf("Removed explicit matches: %s\n", removed)
 	}
 	return matches
 }
--- a/grype/match/explicit_ignores.go
+++ b/grype/match/explicit_ignores.go
@@ -5,29 +5,29 @@ import "github.com/anchore/grype/internal/log"
 var explicitIgnoreRules []IgnoreRule
 
 func init() {
-	type ignore struct {
-		typ   string
-		vulns []string
-		pkgs  []string
+	type ignoreValues struct {
+		typ             string
+		vulnerabilities []string
+		packages        []string
 	}
 
-	// Based on https://github.com/anchore/grype/issues/552, which includes a reference to the
-	// https://github.com/mergebase/log4j-samples collection, we want to filter these explicitly:
-	var explicitIgnores = []ignore{
+	var explicitIgnores = []ignoreValues{
+		// Based on https://github.com/anchore/grype/issues/552, which includes a reference to the
+		// https://github.com/mergebase/log4j-samples collection, we want to filter these explicitly:
 		{"java-archive",
 			[]string{"CVE-2021-44228", "CVE-2021-45046", "GHSA-jfh8-c2jp-5v3q", "GHSA-7rjr-3q55-vv33"},
 			[]string{"log4j-api", "log4j-slf4j-impl", "log4j-to-slf4j", "log4j-1.2-api", "log4j-detector", "log4j-over-slf4j", "slf4j-log4j12"},
 		},
 	}
 
-	for _, ignoreValues := range explicitIgnores {
-		for _, vuln := range ignoreValues.vulns {
-			for _, pkg := range ignoreValues.pkgs {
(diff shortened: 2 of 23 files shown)


grype/match/explicit_ignores.go
```diff
@@ -0,0 +1,42 @@
...
+					},
+				})
+			}
+		}
+	}
+}
+
+func ApplyExplicitIgnoreRules(matches Matches) Matches {
+	matches, ignored := ApplyIgnoreRules(matches, explicitIgnoreRules)
+	log.Debugf("ignoring explicit matches: %+v", ignored)
```
wagoodman, line 40: a few of questions/observations:
1. do we want to log the matches we ignored at all? If this list is about correcting known bad matches I feel that we don't need to expose this in logging.
2. (if we keep the log entry) this log message could be very verbose in a line-length sense for even one entry, so if we leave the log entry probably needs to be summarized.
3. (if we keep the log entry) the verbiage of "ignore" is already somewhat reserved by the `ignore` configurable (could be confusing while debugging about what got applied ignore-rule-wise vs these hard coded corrections).

grype/match/explicit_ignores.go
```diff
@@ -0,0 +1,42 @@
+package match
+
+import "github.com/anchore/grype/internal/log"
+
+type Ignore struct {
```
wagoodman, line 5: since this is a helper type that utilizes mind reducing the scopes where this can be used? This can be done in a couple ways:
1. unexporting `Ignore` by renaming to `ignore`
2. further reduce the scope of both the `Ingore` type and `explicitIgnoreRules` variables by moving their definitions to within the `init` function

grype/match/explicit_ignores.go
kzantow: > correcting known bad matches

if, for some reason, a user expects to see `GHSA-jfh8-c2jp-5v3q` when they have `log4j-api` (or some other match in the future), we can say: "we believe this to be a false positive. run with `-d` for an explanation", no?

grype/logger/logger.go
```diff
@@ -9,4 +9,5 @@ type Logger interface {
 	Info(args ...interface{})
 	Debugf(format string, args ...interface{})
 	Debug(args ...interface{})
+	IsDebugEnabled() bool
```
kzantow, line 12: having these types of flags is typical in log4j :trollface:

grype/match/explicit_ignores.go
kzantow: Based on our discussion, it sounds like there's no problem with debug logging this with the changes I made: improved the error message (also to not use the word "ignore") and output one line per exclusion instead of a potentially large data structure

grype/match/explicit_ignores.go
```diff
@@ -0,0 +1,50 @@
...
+			}
+		}
+	}
+}
+
+// ApplyExplicitIgnoreRules This will filter out matches that are defined above, in the
+// explicitIgnores list
+func ApplyExplicitIgnoreRules(matches Matches) Matches {
+	matches, ignored := ApplyIgnoreRules(matches, explicitIgnoreRules)
+	if log.IsDebugEnabled() {
```
wagoodman, line 42: why check if debug is enabled? I don't think there is a performance concern we need to protect against here. (We should be free to log to Debug (or any level) without checking if it's necessary first)

grype/match/explicit_ignores.go
```diff
@@ -0,0 +1,50 @@
...
+		removed := ""
+		for _, i := range ignored {
+			removed += "    - " + i.Match.Vulnerability.ID + ":" + i.Package.PURL + "\n"
```
wagoodman, line 45: super nit: Today we tend to show nested information in the logging output in a tree-like way (where each line is a separate log line too):

```
[0002] DEBUG found 10 vulnerabilities for pkg=Pkg(type=deb, name=libc-bin, version=2.31-0ubuntu9.2)
[0002] DEBUG   ├── vuln="CVE-2020-6096" type="Exact-Indirect Match" searchedBy=...
[0002] DEBUG   ├── vuln="CVE-2021-3326" type="Exact-Indirect Match" searchedBy=...
[0002] DEBUG   ├── vuln="CVE-2016-10228" type="Exact-Indirect Match" searchedBy=...
[0002] DEBUG   ├── vuln="CVE-2021-35942" type="Exact-Indirect Match" searchedBy=...
[0002] DEBUG   ├── vuln="CVE-2021-33574" type="Exact-Indirect Match" searchedBy=...
[0002] DEBUG   ├── vuln="CVE-2021-38604" type="Exact-Indirect Match" searchedBy=...
[0002] DEBUG   ├── vuln="CVE-2020-27618" type="Exact-Indirect Match" searchedBy=...
[0002] DEBUG   ├── vuln="CVE-2021-27645" type="Exact-Indirect Match" searchedBy=...
[0002] DEBUG   ├── vuln="CVE-2020-29562" type="Exact-Indirect Match" searchedBy=...
[0002] DEBUG   └── vuln="CVE-2019-25013" type="Exact-Indirect Match" searchedBy=...
```

it would be nice to keep consistent here for nested logging info.

grype/logger/logger.go
wagoodman: Tagging onto https://github.com/anchore/grype/pull/559/files#r774058215 , it would be ideal to not expand the logging interface unless we really need to (say to add another logging level). Are we certain we need this information? The larger we make this interface the harder it is to plug in other loggers or have other projects interop with the logger here (we even held off for a long time before adding the warn log level to the interface!).

grype/match/explicit_ignores.go
kzantow: done! 
```
[0001] DEBUG Removed 12 explicit vulnerability matches:
[0001] DEBUG   ├── CVE-2021-44228 : pkg:maven/org.apache.logging.log4j/log4j-slf4j-impl@2.13.2
[0001] DEBUG   ├── CVE-2021-45046 : pkg:maven/org.apache.logging.log4j/log4j-slf4j-impl@2.13.2
...
```

wagoodman (approved):
💯  --thanks for the conversations on where this all finally lands and getting this in 🙌

Issue #560, opened 2021-12-22
MacOS install fails using GNU coreutils version of cp


zcrisler:
**What happened**:

Failed to install `grype` to the current directory on darwin/amd64:

```
/tmp/grype $ curl -sSfL https://raw.githubusercontent.com/anchore/grype/main/install.sh | sh -s -- -b .
anchore/grype info checking GitHub for latest tag
anchore/grype info found version: 0.27.3 for v0.27.3/darwin/amd64
install: cannot stat '/var/folders/yn/znb8ps494g16w59p5tljm6v40000gq/T/tmp.FqeCbV6ADi/grype': No such file or directory
```

Running again with `-x` gives a more detailed trace:

```
$ curl -sSfL https://raw.githubusercontent.com/anchore/grype/main/install.sh | sh -s -- -b . -x
+ getopts 'b:dh?x' arg
+ shift 3
+ TAG=
+ get_binaries
+ case "$PLATFORM" in
+ BINARIES=grype
+ tag_to_version
+ '[' -z '' ']'
+ log_info 'checking GitHub for latest tag'
+ log_priority 6
+ test -z 6
+ '[' 6 -le 6 ']'
++ log_prefix
++ echo anchore/grype
++ log_tag 6
++ case $1 in
++ echo info
+ echoerr anchore/grype info 'checking GitHub for latest tag'
+ echo anchore/grype info 'checking GitHub for latest tag'
anchore/grype info checking GitHub for latest tag
++ github_release anchore/grype ''
++ owner_repo=anchore/grype
++ version=
++ test -z ''
++ version=latest
++ giturl=https://github.com/anchore/grype/releases/latest
+++ http_copy https://github.com/anchore/grype/releases/latest Accept:application/json
++++ mktemp
+++ tmp=/var/folders/yn/znb8ps494g16w59p5tljm6v40000gq/T/tmp.WmjbOgeqGe
+++ http_download /var/folders/yn/znb8ps494g16w59p5tljm6v40000gq/T/tmp.WmjbOgeqGe https://github.com/anchore/grype/releases/latest Accept:application/json
+++ log_debug 'http_download https://github.com/anchore/grype/releases/latest'
+++ log_priority 7
+++ test -z 7
+++ '[' 7 -le 6 ']'
+++ return 0
+++ is_command curl
+++ command -v curl
[...]

PR #566, opened 2021-12-22
update so install.sh is compatible with coreutils 

spiffcs:
See #560 for details and investigation.

This adds a `.` following the path in the `cp` command under `extract_from_dmg` to enhance compatibility with `coreutils`


luhring (approved):
Nice! 🎉  **Should we make this change in Syft, too?**

spiffcs:
> Nice! 🎉 **Should we make this change in Syft, too?**

Yep! I'll file it over in Syft as well.

PR #587, opened 2022-01-11
Port grype-db to grype

wagoodman:
This PR ports all functionality from grype-db into the grype repo, making grype the source of truth for all schema versions. Grype itself will still only use a single schema version (the latest supported version), however, all historical expressions of each schema version will be stored here.

This was done to remove a module cycle between grype and grype-db (specifically in the namespace code for v1, v2 and v3 packages. There does not appear to be a good approach to eliminate the cycle via type decomposition, so this is the next best approach. This has the advantage of no longer needing the grype-db repo, which helps with dependency coordination some. 

The grype-db `v*/writer` packages contain code that depends on CGO, thus, cannot be used by grype (directly or indirectly). 

The structural changes to grype approximate to:
```
grype
├── ...
├── db
│   ├── curator.go
│   ├── curator_test.go
│   ├── listing.go                # migrated from grype-db
│   ├── listing_entry.go          # migrated from grype-db
│   ├── listing_test.go           # migrated from grype-db
│   ├── metadata.go               # migrated from grype-db
│   ├── metadata_test.go          # migrated from grype-db
│   ├── status.go
│   ├── test-fixtures
│   ├── v1/                         # migrated from grype-db
│   ├── v2/                         # migrated from grype-db
│   ├── v3/                         # migrated from grype-db
│   ├── vulnerability_metadata_provider.go                 # migrated from grype/vulnerability
│   ├── vulnerability_provider.go                          # migrated from grype/vulnerability
│   ├── vulnerability_provider_mocks_test.go               # migrated from grype/vulnerability
│   └── vulnerability_provider_test.go                     # migrated from grype/vulnerability
├── ...
└── vulnerability/       # removed all provider implementations and moved to db package
```


kzantow (approved):
I think this is a bit large to review in depth, but having a look at a few things:

* existing tests aren't removed
* grype-db tests are ported
* go.mod / go.sum indeed have grype-db removed

... coupled with our reasonably good test coverage passing makes me happy with this change.

PR #597, opened 2022-01-19
upgrade syft to v0.36.0

spiffcs:


jonasagx (approved):
🚀

wagoodman (changes requested):
there are several go.sum updates that don't look right

wagoodman (approved):

Issue #395, opened 2021-08-23
Allow for ingestion of SPDX SBOM documents as input


wagoodman:
Today we allow for the syft JSON document as an SBOM input. It would be ideal to additionally interop with tools that produce SPDX documents and allow that as input into grype for vulnerability scanning.

wagoodman:
Could be closely related to https://github.com/anchore/syft/issues/400 and may have architectural changes/concerns. (we should try and tackle SBOM ingestion for syft and grype together)

luhring:
Implementation idea (RFC): In addition to ingesting SPDX in general, we've mentioned the notion of _piping_ SPDX into Grype (or Syft, per https://github.com/anchore/grype/issues/395#issuecomment-903769157). This could get tricky if we accept multiple formats (e.g. Syft + SPDX) and have to try to detect what we're receiving in the byte stream (not impossible, though).

Many tools have a convention of using `-` to refer to a command's `stdin`. We could use this in combination with our "schemes" system to allow the user to instruct the tool explicitly on how to interpret the inbound bytes.

E.g.: `cat my-sbom.spdx | grype spdx:- ...`

luhring:
This will be unblocked once https://github.com/anchore/syft/issues/556 is closed

wagoodman:
This same comment applies https://github.com/anchore/grype/issues/481#issuecomment-996972824

kzantow:
An associated Syft PR can be found here: https://github.com/anchore/syft/pull/738

PR #602, opened 2022-01-25
Add ability to merge matches

wagoodman:
Primarily this PR adds the ability to merge two matches that have the same vulnerability-package pair, but the matching details may be different (what was searched by, what attribute was matched on, the matcher that found the result, etc). In the case of merging the list of `match.Details` is combined and one match object is kept.

This PR additionally makes the following adjustments:
- updates tests fixtures that use package objects to have package IDs
- updates the `match.NewMatches()` constructor to take an initial variadic list of matches
- renames/moves the `matchers.common` package to `search` which makes usage less awkward and more semantically correct (the matchers use search utilities, not common utilities... e.g. `common.FindMatchesByPackageLanguage(...)` --> `search.ByPackageLanguage(...)`)
- adds `search.ByCriteria` to reduce the code for each matcher object. This turns the old approach:
```golang 
	var matches = make([]match.Match, 0)
	langMatches, err := common.FindMatchesByPackageLanguage(store, p.Language, p, m.Type())
	if err != nil {
		return nil, err
	}
	matches = append(matches, langMatches...)

	cpeMatches, err := common.FindMatchesByPackageCPE(store, p, m.Type())
	if err != nil {
		return nil, err
	}
	matches = append(matches, cpeMatches...)
        return matches, nil
```
into:
```golang
       return search.ByCriteria(store, d, p, m.Type(), search.ByLanguage, search.ByCPE)
```

This work is loosely related to #395


cmd/root_test.go
```diff
@@ -37,19 +39,24 @@ func (d *mockMetadataStore) GetVulnerabilityMetadata(id, recordSource string) (*
 
 func TestAboveAllowableSeverity(t *testing.T) {
 	thePkg := pkg.Package{
+		ID:      pkg.ID(uuid.NewString()),
```
spiffcs, line 42: Nice update to include new syft ID format

grype/match/type.go
```diff
@@ -1,37 +1,26 @@
...
+import (
+	"github.com/anchore/grype/grype/pkg"
+)
 
 const (
-	UnknownMatchType Type = iota
-	ExactDirectMatch
-	ExactIndirectMatch
-	FuzzyMatch
+	ExactDirectMatch   Type = "exact-direct-match"
```
spiffcs, line 8: This is a much cleaner expression for the enums

grype/matcher/ruby/matcher.go
```diff
@@ -20,18 +20,6 @@ func (m *Matcher) Type() match.MatcherType {
...
-	matches = append(matches, langMatches...)
-
-	cpeMatches, err := common.FindMatchesByPackageCPE(store, p, m.Type())
-	if err != nil {
-		return nil, err
-	}
-	matches = append(matches, cpeMatches...)
-	return matches, nil
+func (m *Matcher) Match(store vulnerability.Provider, d *distro.Distro, p pkg.Package) ([]match.Match, error) {
+	return search.ByCriteria(store, d, p, m.Type(), search.CommonCriteria...)
```
spiffcs, line 24: Nice work condensing this. It also makes it easier to consume as a library going forward.

PR #606, opened 2022-01-26
Support for SBOMs with incomplete linux distribution or CPE information

kzantow:
This adds the ability to use non-Syft SBOM documents as input. This comes with a number of challenges regarding missing data: for one thing, CPEs are quite possibly not included, so there is a flag to auto-generate basic CPEs (`--add-cpes-if-none`). Some matches only occur based on the linux distro, so there is another flag to specify this (e.g. `--distro alpine:3.10`). SPDX files generated by Syft, however, will typically have CPEs and PURLs with distro information and not require these flags. CycloneDX files generated by Syft will need these flags until we've implemented multi-CPE output (using `syft:` or, more ideally `spdx:` properties)


luhring (commented):
Love this concept! Added a few thoughts for how we conceptualize what's happening for the user surface area...

cmd/root.go
```diff
@@ -304,6 +321,44 @@ func startWorker(userInput string, failOnSeverity *vulnerability.Severity) <-cha
...
+		context.Distro = &linux.Release{
+			PrettyName: d,
+			Name:       d,
+			ID:         d,
+			IDLike: []string{
+				d,
+			},
+			Version:   v,
+			VersionID: v,
+			HomeURL:   "",
```
wagoodman, line 343: nit: empty fields could be left out

grype/pkg/package.go
```diff
@@ -56,11 +57,24 @@ func New(p pkg.Package) Package {
...
+func FromCatalog(catalog *pkg.Catalog, config ProviderConfig) []Package {
 	result := make([]Package, 0, catalog.PackageCount())
+	missingCPEs := false
 	for _, p := range catalog.Sorted() {
+		if len(p.CPEs) == 0 {
+			// For SPDX (or any format, really) we may have no CPEs
+			if config.GenerateMissingCPEs {
+				p.CPEs = cpe.Generate(p)
+			} else {
+				log.Debugf("No CPEs for package: %s", p)
```
wagoodman, line 69: nit: lowercase for consistency
```suggestion
				log.Debugf("no CPEs for package: %s", p)
```

grype/pkg/package.go
```diff
@@ -56,11 +57,24 @@ func New(p pkg.Package) Package {
...
+				missingCPEs = true
+			}
+		}
 		result = append(result, New(p))
 	}
+	if missingCPEs {
+		log.Warnf("Some package(s) are missing CPEs. This may result in missing vulnerabilities. You may autogenerate these using: --add-cpes-if-none")
```
wagoodman, line 76: nit: 
```suggestion
		log.Warnf("some package(s) are missing CPEs. This may result in missing vulnerabilities. You may autogenerate these using: --add-cpes-if-none")
```

wagoodman (commented):
We should add README.md updates for the new `--distro` and `--add-cpes-if-none` functionality

internal/config/application.go
```diff
@@ -150,7 +152,7 @@ func (cfg *Application) parseLogLevelOption() error {
 		case v >= 2:
 			cfg.Log.LevelOpt = logrus.DebugLevel
 		default:
-			cfg.Log.LevelOpt = logrus.ErrorLevel
+			cfg.Log.LevelOpt = logrus.WarnLevel
```
jonasagx, line 155: The PR description doesn't mention this default log level change, should we default to warning in syft too?

jonasagx (approved):
LGTM -- left one question on log level

internal/config/application.go
kzantow: The problem is by default warning messages do not show up, but I do not believe the messages I want a user to see are errors... it's a case where a user _should_ see something, but it hasn't necessarily prevented the process from running. I think logging warn and higher messages is fairly typical.

jonasagx: What about using the TUI to present that information?

Since this is the global log setting, all other warning messages will be shown, which might be a problem.

wagoodman (commented):
Did we want to introduce a `grype db namespaces` or `grype db list-namespaces` command to help users understand what values can be supplied to distro?

wagoodman (approved):
Great addition --consider https://github.com/anchore/grype/pull/606#pullrequestreview-899524876 to be non-blocking

kzantow:
@wagoodman yes, I'll add an issue for a follow-on at some point with `grype db list-namespaces` or whatever we decide

PR #607, opened 2022-01-26
Abstract upstream package before matching

### LINKED ISSUE #395: Allow for ingestion of SPDX SBOM documents as input

wagoodman:
This PR makes the following adjustments:
- ~attempts to extract upstream OS package information from pURLs when rich metadata is not available, which aids in package indirect matching for APK, RPM, and DPKG package types.~ based on PR review, this is being moved further back in processing into syft, probably in https://github.com/anchore/syft/pull/738.
- adds the concept of an "upstream package" with a name and optional version. Any package can now indicate one or more source packages via `pkg.Package.Upstreams[]`. 
- Source package information is extracted from metadata in advance of the matching process. This is better for two reasons: 1) we can remove multiple `pkg.*Metadata` structs since they can be represented as `pkg.UpstreamPackage` with no loss of information, and 2) simplifies the matching process for matchers that do matching by source-package indirection by sharing similar code (since they now share the same `UpstreamPackage` abstraction).
- removes the use of the Syft `pkg.MetadataType` in favor of a new Grype `pkg.Metadata` type. This further enforces the bound context between Syft and Grype and in this case is better since the `MetadataType` hints at the shape of the `pkg.Package.Metadata` field, which no longer has any shared struct definitions with Syft.
- exposes `MetadataType` in the json output

Fixes #395


grype/pkg/package.go
```diff
@@ -2,102 +2,257 @@ package pkg
...
+		case pkg.RpmPkg:
+			m, u := rpmdbDataFromPURL(p.PURL)
+			upstreams = u
+			if m != nil {
+				metadata = *m
+				metadataType = RpmdbMetadataType
 			}
-		} else {
-			log.Warnf("unable to extract Java metadata for %s", p)
 		}
```
kzantow, line 112: I'm not really sure this is the best location to deal with incoming PURLs. I've added some bits that attempt to create a Syft data model, including the metadata from the PURLs. This way I'd think Grype would be able to use a more consistent data model to populate the upstreams? See: https://github.com/anchore/syft/pull/738/files#diff-b60239c3dd2d41cfef2107e92599d7426bd918d4232e5e8273da3b05d9349bedR296

grype/pkg/metadata.go
```diff
@@ -0,0 +1,12 @@
+package pkg
+
+// MetadataType represents the data shape stored within pkg.Package.Metadata.
+type MetadataType string
+
+const (
+	// this is the full set of data shapes that can be represented within the pkg.Package.Metadata field
+
+	UnknownMetadataType MetadataType = "UnknownMetadata"
```
kzantow, line 9: Should this be just serialized during output based on the actual metadata data type?

grype/pkg/upstream_package.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/anchore/syft/syft/pkg"
+	"github.com/scylladb/go-set/strset"
+)
+
+type UpstreamPackage struct {
+	Name    string // the package name
+	Version string // the version of the package
+}
+
+func UpstreamPackages(p Package) <-chan Package {
```
kzantow, line 15: Is there a reason for this being `<-chan Package` instead of, say, `[]Package`?

kzantow (commented):
Overall I don't see anything wrong here but a few general questions:

1) Is there a reason to use `MetadataType` instead of relying on the Go types, like using:
```
switch metadata.(type) {
case JavaMetadata:
...
}
```
This tripped me up a bit and I had to debug a small amount to realize just setting the metadata was not enough. I could not really understand why a variable is used to denote a type when the type information is already available (and could be serialized/deserialized without issue based on the Go type, I suspect).

2) Would it be better to keep all the PURL parsing in the Syft decoding process? It seems wrong that we might generate a format and _not_ be able to decode it in a meaningful manner but then handle it in Grype. That said, I don't really see a problem having some extra handling to just try to handle whatever we've been given. For Syft, SPDX, CycloneDX and any other supported formats, I would hope we're decoding these properly and won't need it, though.

3) Is the "upstream" abstraction shared across enough matchers that it should be top-level data? It seems as though it is specific to Apk, Dpkg, and Rpm matchers but not the others. Looking at [the dataFromPkg function](https://github.com/anchore/grype/pull/607/files#diff-c1eeca4b68c3cef87a8131f596f5e598d5b9a56704624ff94afd7db799e4193aR74), it seems like it's really all just different types of metadata much like it was factored before. Might the upstreams list just replace the specific metadata types for those matchers instead of being a top-level thing? Or keep each of the metadata types but add a list of upstreams so the functions to deal with them can remain consistent for the aforementioned 3 matchers?

grype/pkg/package.go
```diff
@@ -2,102 +2,257 @@ package pkg
...
 	"github.com/anchore/syft/syft/source"
 )
 
+// the source-rpm field has something akin to "util-linux-ng-2.17.2-12.28.el6_9.2.src.rpm"
+// in which case the pattern will extract out the following values for the named capture groups:
+//		name = "util-linux-ng"
+//		version = "2.17.2" (or, if there's an epoch, we'd expect a value like "4:2.17.2")
+//		release = "12.28.el6_9.2"
+//		arch = "src"
+var rpmPackageNamePattern = regexp.MustCompile(`^(?P<name>.*)-(?P<version>.*)-(?P<release>.*)\.(?P<arch>[a-zA-Z][^.]+)(\.rpm)$`)
```
kzantow, line 21: It seems like there are a lot of type specific things added here; maybe this should predominantly be factored out to the subpackages under [the matcher package](https://github.com/anchore/grype/tree/main/grype/matcher) or something similar?

grype/pkg/package.go
wagoodman: you're right --syft is the better spot for this logic. I'll take out the specifics about deriving metadata values from pURL but keep the upstream package generation 👍

grype/pkg/metadata.go
wagoodman: That's how this originally worked, yes. I decided to make a more firm division of responsibilities here between syft and grype since the data shapes represented by these constants are different depending which application being referred to.

grype/pkg/upstream_package.go
wagoodman: There is no strong reason for this --no. The current approach is more like a python generator, but I can change it to return a static slice too 👍

grype/pkg/package.go
wagoodman: > It seems like there are a lot of type specific things added here

Indeed, this was intentional. The observation was that by moving type specific handling as much as possible upstream in processing (in this case always during package creation) downstream processing becomes simplified.

The reason for the type-specific logic is to craft the `UpstreamPackage` (which is agnostic to the package type) as early as possible. Today (before this PR) we push a lot of that processing into matcher sub-packages, which was causing lots of type assertions to get specific upstream package info out in order to do a fairly standard DB lookup. It became apparent that making the Upstream Package agnostic to the package type (information is inferred from the parent package) made it easier to do this logic when we are creating the Grype Package based off of the Syft package just after SBOM decoding or Syft cataloging (thereby making downstream processing in the matchers simpler). 

I don't think I'd be able to put the rpm-specific logic for parsing into the RPM matcher package since that would introduce a package cycle. However, I could move this logic into another new package, but it isn't clear what that new package's purpose (thus name) would be other than something like `rpm-utils`. For that reason I left this logic unexported in the `pkg` package (closest to where it's used).

wagoodman:
> 2. Would it be better to keep all the PURL parsing in the Syft decoding process?...

absolutely, I'll make that update (good call)

> 3. ... Is the "upstream" abstraction shared across enough matchers that it should be top-level data? It seems as though it is specific to Apk, Dpkg, and Rpm matchers but not the others ...

It's true that this abstraction is not universal to all package types in grype, however, it is only these ecosystems (+nvd) where grype cares about looking for upstream packages. It may be one day that we add matching for additional upstream packages for other matchers, but we just haven't done so yet. 

> 3. Looking at the dataFromPkg function, it seems like it's really all just different types of metadata much like it was factored before. 

Yes, however, the upstreams processing is new and metadata processing for DPKG and APK was removed. This functionality was decomposed into a separate function was all.

> 3. ...Might the upstreams list just replace the specific metadata types for those matchers instead of being a top-level thing?... 

That's exactly what this change aims to do. This PR already removes all of the package metadata's where possible (DPKG and APK metadata structs were flat out deleted). I could have gone further and removed the metadata for RPMdb, I can give that a shot and see what shakes out? I think this would mean a Syft update (your PR would be a good spot, I can add a commit?)

kzantow (approved):
LGTM - I had a thought I'll bring up elsewhere about handling the metadata across things in a more uniform manner, I think this moves toward that goal 👍

kzantow (approved):
LGTM

PR #608, opened 2022-01-26
Update README.md

fjammes:
sudo is required to write to `/usr/local/bin`

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ If you encounter an issue, please [let us know using the issue tracker](https://
 curl -sSfL https://raw.githubusercontent.com/anchore/grype/main/install.sh | sh -s -- -b /usr/local/bin
 ```
 
-...or, you can specify a release version and destination directory for the installation:
+You can also choose another destination directory and release version for the installation. The destination directory doesn't need to be `/usr/local/bin`, it just needs to be a location found in the user's PATH and writable by the user that's installing Grype.
 
 ```
 curl -sSfL https://raw.githubusercontent.com/anchore/grype/main/install.sh | sh -s -- -b <DESTINATION_DIR> <RELEASE_VERSION>


luhring (changes requested):
This shouldn't require sudo access, the user just needs write access to whatever the destination folder is.

IIUC, this change shouldn't be needed.

fjammes:
Please be aware that on most Linux distributions, the default is that `/usr/local/bin` is only writable by root, and this is a good security practice not to change this. So the initial command you propose to install `grype` will fail on 99% Linux systems and use will have to hack the command (like I has to do), change the default permissions on `/usr/local/bin` or set a new install directory. I think this is not user-friendly. So, if you don't want to use `sudo`, I propose you at least set an other destination than `/usr/local/bin` (maybe `$HOME/bin` and then add it to $PATH) in the documentation.

luhring:
>please be aware that on most Linux distributions, the default is that /usr/local/bin is only writable by root, and this is a good security practice not to change this. So the initial command you propose to install grype will fail on 99% Linux systems and use will have to hack the command (like I has to do), change the default permissions on /usr/local/bin or set a new install directory. I think this is not user-friendly. So, if you don't want to use sudo, I propose you at least set an other destination than /usr/local/bin (maybe $HOME/bin and then add it to $PATH) in the documentation.

Hi @fjammes, thanks for the explanation. And apologies for the frustration this caused.

As you point out, the destination location doesn't actually need to be `/usr/local/bin`. This is a value that the user can choose — we've just found that `/usr/local/bin` works 99% of the time (including for our Linux users).

Whatever value the user chooses, it needs to be writable by the current user, and it should also be in the intended user's `PATH` (as you mentioned).

Immediately below the line you quoted, the README shows an alternative approach to running the installation script, where the destination directory and release version are values that the user can choose:

```
curl -sSfL https://raw.githubusercontent.com/anchore/grype/main/install.sh | sh -s -- -b <DESTINATION_DIR> <RELEASE_VERSION>
```

I agree with your point that these instructions should "just work" for everyone, this just turns out to be difficult to do 100% of the time.

A large concern I have with adding `sudo` in particular is that it creates a security risk for our users that isn't necessary. As maintainers of a security tool, we're very sensitive about this (and hopefully this is understandable 😃 ). Users have been installing Grype on Linux systems for over a year now, and we haven't seen this issue reported before.

>I propose you at least set an other destination than /usr/local/bin (maybe $HOME/bin and then add it to $PATH) in the documentation.

I like this idea — or at least some form of it. Since our default instructions have been working for folks, but not everyone, I could see it being worthwhile to be more explicit that the installation path doesn't need to be `/usr/local/bin`, it just needs to be a location found in the user's `PATH`, and writable by the user that's installing Grype.

Let me know if you're interested in adjusting this PR to make this change. And if not, I'm happy to take it on.

fjammes:
@luhring , thanks so much for your detailed answer, there was no frustation on my side ;-) ! In my understanding, users that have successfully installed `grype` in  `/usr/local/bin` were certainly using `root` account. It looks to me it is not a good security practice to delegate write access to `/usr/local/bin` to a non-root account  (`sudo` seems a safer an more trackable way IMHO). I have a new proposal based on your comment. Does it work for you?

luhring:
Thanks @fjammes! This looks pretty good, I just pushed one suggested change to the branch, the goal of which is to avoid obscuring the default recommendation command (since this command does work for the vast majority of our users), so users can find this command as easily as possible. Then, right below, we adjust the language to include the notice to users that the destination path can be adjusted within certain constraints.

If this looks good to you, we can proceed toward merging.

The only other item to address would be the first commit (06bcd0c) needing to be **signed** per our [CONTRIBUTING.md](https://github.com/anchore/grype/blob/main/CONTRIBUTING.md). Let me know if I can help with this.

fjammes:
@luhring , LGTM, thanks so much. However signing and existing commit without have to do a force push does not look obvious to me. Could you please help with this?

fjammes:
Hi @luhring ,

I have signed the commit, so I propose you merge the PR?

Regards,

wagoodman (approved):

luhring:
There's a signature verification problem, caused by a commit having two authors but only one signature. (Only one signature is _possible_ for a commit, so that means it needs to have a single author, who's the same person as the signer.)

To expedite the solution to this, I can attribute these commits to myself and sign them, but that would steal credit from you @fjammes. I'll wait to do this until later in the day, so that in the meantime, if you'd like to retain the authorship, you just need to remove me as the author of any commits, ensure all commits are signed off, and then sign your commits again. One way or another, we'll get this PR merged by end of today.

Sorry for all the trouble here.

PR #630, opened 2022-02-11
Normalize release assets and refactor install.sh

wagoodman:
This brings grype up to parity with syft from https://github.com/anchore/syft/pull/765 and https://github.com/anchore/syft/pull/789 (and a few others, but those are the main PRs).

The same benefits from those PRs apply here:
- today the release assets do not match the snapshot assets (the signing process creates new assets unnecessarily), making it difficult to test the install.sh against a snapshot. This PR makes the release assets the same as the snapshot assets.
- install.sh today is not under test, this PR puts them under test such that changes to the goreleaser builds can be checked against a branch snapshot build to ensure that they will function as expected for all assets generated.

Fixes anchore/project-toolbox#113


luhring (approved):
Approving w/ the assumption that this work lines up with the work done in Syft — so let me know if that's a flawed assumption! 😀

wagoodman:
not flawed @luhring  , lots of copy-paste 👍

PR #635, opened 2022-02-15
Bump syft to v0.38.0 for release

wagoodman:
Includes partially the features from https://github.com/anchore/syft/releases/tag/v0.38.0


kzantow (approved):

kzantow:
(You probably know, but looks like a couple snapshots need updating)

wagoodman:
indeed --though in the same vein as https://github.com/anchore/syft/pull/823/files, I'm electing to keep the original behavior (show empty collections and not `null`, so updated the presenter logic)

PR #642, opened 2022-02-24
chore(reproducibility): add buildid= and trimpath

developer-guy:
Signed-off-by: Batuhan Apaydın <batuhan.apaydin@trendyol.com>


wagoodman:
For reference about what this is doing:
- https://stackoverflow.com/questions/63831540/removing-module-path-in-trace-in-go
- https://github.com/golang/go/issues/34186
- https://github.com/golang/go/issues/33772#issuecomment-528176001

wagoodman (approved):
Nice addition  🙌

wagoodman:
@developer-guy out of curiosity, why the change to explicitly specify GOPATH in the workflow files?

developer-guy:
I saw it from the documentation: 
https://goreleaser.com/customization/build/

wagoodman:
I see the reason for `trimpath`, but what I'm referring to is:
- https://github.com/anchore/grype/pull/642/files#diff-e426ed45842837026e10e66af23d9c7077e89eacbe6958ce7cb991130ad05adaR143
- https://github.com/anchore/grype/pull/642/files#diff-f3c08cd4e56926b1732a8183a4eb36057b86d9503161db0021350279b18144a7R184

Where there was a change to explicitly set `GOPATH: /home/runner/go`. I'm confused as to why this is needed --can you elaborate?

developer-guy:
> I see the reason for `trimpath`, but what I'm referring to is:
> 
> * [#642 (files)](https://github.com/anchore/grype/pull/642/files#diff-e426ed45842837026e10e66af23d9c7077e89eacbe6958ce7cb991130ad05adaR143)
> * [#642 (files)](https://github.com/anchore/grype/pull/642/files#diff-f3c08cd4e56926b1732a8183a4eb36057b86d9503161db0021350279b18144a7R184)
> 
> Where there was a change to explicitly set `GOPATH: /home/runner/go`. I'm confused as to why this is needed --can you elaborate?

to be able to use it in .goreleaser.yml via .Env, I think

developer-guy:
https://github.com/hashicorp/vault-csi-provider/pull/143

wagoodman:
> I've replaced with ${{ env.GOPATH }} this one.

Right, but doesn't this do nothing?
```
        env:
           GOPATH: ${{ env.GOPATH }}
```

... since this is setting an environment variable based off of the current environment variable value of the same name? Are these 'GOPATH' changes necessary?

developer-guy:
I reverted GOPATH changes, let's what will happen 😮

developer-guy:
```
   ⨯ release failed after 96.69s error=failed to build for linux_amd64: exit status 2: # github.com/anchore/grype
open main: no such file or directory
```

@wagoodman :(

wagoodman (changes requested):
We shouldn't need to set the GOPATH explicitly for this change. Additionally I think we should be using the build `trimpath` flag and not the gcflag `trimpath=<some-path>`. I think this would be the equivalent change:

```yaml
  - id: linux-build
    binary: grype
    goos:
      - linux
    goarch:
      - amd64
      - arm64
    # set the modified timestamp on the output binary to the git timestamp to ensure a reproducible build
    mod_timestamp: &build-timestamp '{{ .CommitTimestamp }}'
    env: &build-env
      - CGO_ENABLED=0
    flags: &build-flags
      - -trimpath
    ldflags: &build-ldflags |
      -buildid=
      -w
      -s
      -extldflags '-static'
      -X github.com/anchore/grype/internal/version.version={{.Version}}
      -X github.com/anchore/grype/internal/version.syftVersion={{.Env.SYFT_VERSION}}
      -X github.com/anchore/grype/internal/version.gitCommit={{.Commit}}
      -X github.com/anchore/grype/internal/version.buildDate={{.CommitDate}}
      -X github.com/anchore/grype/internal/version.gitDescription={{.Summary}}
```

I made one extra update:
```
-X github.com/anchore/grype/internal/version.buildDate={{.CommitDate}}
```
... so the commit date is referenced.

Note: this would not get us 100% to reproducible builds, but I'm not certain what's left to close the gap.

developer-guy:
I think we should get build date via https://reproducible-builds.org/docs/source-date-epoch/

developer-guy:
seems everything is fine @wagoodman, thanks a ton 🙋🏻‍♂️

wagoodman:
@developer-guy  friendly nudge on https://github.com/anchore/grype/pull/642#issuecomment-1060717671 (also 1:1 with https://github.com/anchore/syft/pull/847)

wagoodman:
I'm going to close this as stale, but please reach out on a new issue if you wanted to chat further about this.

PR #647, opened 2022-02-26
Add gosimports linter

wagoodman:
Today we use the `goimports` linter to check and fix imports preferences, however, there is a downside that IDEs tend to add extra whitespace when automatically adding import statements, a case that `goimports` ignores (see https://github.com/golang/go/issues/20818). This PR replaces `goimports` with `gosimports` which is 1:1 the same tool with the exception for this specific case. 

Additionally this PR incorporates @kzantow s suggestion to additionally group `anchore` imports last 🙌 .

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -19,7 +19,6 @@ linters:
     - gocritic
     - gocyclo
     - gofmt
-    - goimports
     - goprintffuncname
     - gosec
     - gosimple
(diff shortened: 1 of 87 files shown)


jonasagx (approved):

kzantow (approved):
Looks great! `-local` really does the trick IMO

PR #654, opened 2022-03-03
Add SARIF presenter option

### LINKED ISSUE #304: SARIF report template

I see https://github.com/anchore/scan-action supports a SARIF report.
However, instead of using the `grype --template` option, it creates it in the code.

Is there any plan to create a template for use by grype?

**Why is this needed**:

Being able to create a SARIF report no mater how you use grype.

**Additional context**:

kzantow:
This fixes #304 it will also be the basis of a fix for https://github.com/anchore/scan-action/issues/60, https://github.com/anchore/scan-action/issues/61, https://github.com/anchore/scan-action/issues/98 and https://github.com/anchore/scan-action/issues/116


grype/presenter/sarif/presenter.go
```diff
@@ -0,0 +1,348 @@
+package sarif
+
+import (
+	"fmt"
+	"io"
+	"strings"
+
+	s "github.com/owenrumney/go-sarif/sarif"
```
wagoodman, line 8: `s` isn't a great import alias, why not stick with `sarif`?

kzantow: `sarif` conflicts with the package name 😬

kzantow: Oddly, I found it made things pretty quick to work with this way but if you feel strongly it should be something else, I could change it to something maybe `sarifLib` or happy to hear other suggestions. or we _could_ rename the presenter something like `sarifv1` or whatever.

grype/presenter/sarif/presenter_test.go
```diff
@@ -0,0 +1,252 @@
...
+	"github.com/anchore/grype/grype/match"
+	"github.com/anchore/grype/grype/pkg"
+	"github.com/anchore/grype/grype/presenter/models"
+	"github.com/anchore/grype/grype/vulnerability"
+	"github.com/anchore/stereoscope/pkg/imagetest"
+	syftPkg "github.com/anchore/syft/syft/pkg"
+	"github.com/anchore/syft/syft/source"
+)
+
+var update = flag.Bool("update", true, "update the *.golden files for json presenters")
```
kzantow, line 22: I'll update the golden files and flip this when the rest gets reviewed and it's ready to merge

grype/presenter/sarif/presenter.go
```diff
@@ -0,0 +1,348 @@
...
+			return "high"
+		case vulnerability.MediumSeverity:
+			return "medium"
+		}
+	}
+	return "low"
+}
+
+// securitySeverityValue GitHub uses a numeric severity value to determine whether things are critical, high, etc.
+func (pres *Presenter) securitySeverityValue(m match.Match) string {
```
luhring, line 241: Are these numbers expected to correspond to the CVSS scale? The number range looks similar so I wanted to check, in case we should handle this especially for when CVSS data is present

grype/presenter/sarif/presenter.go
```diff
@@ -0,0 +1,348 @@
...
+	v := version.FromBuild().Version
+	if v == "[not provided]" {
+		// Need a semver to pass the MS SARIF validator
+		v = "0.0.0-dev"
+	}
+
+	doc.AddRun(&sarif.Run{
+		Tool: sarif.Tool{
+			Driver: &sarif.ToolComponent{
+				Name:           "Anchore Grype Scan",
```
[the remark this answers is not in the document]
kzantow: It is a "three word or less" description of the tool which generated the sarif report; happy to name this something different

grype/presenter/sarif/presenter.go
kzantow: This corresponds to the GitHub security-severity property (as noted in the comment 😄)
> Recommended. A string representing a score that indicates the level of severity, between 0.0 and 10.0, for security queries (tags includes security). This, with the properties.precision property, determines whether the results are displayed by default on GitHub so that the results with the highest security-severity, and highest precision are shown first. Code scanning translates numerical scores as follows: over 9.0 is critical, 7.0 to 8.9 is high, 4.0 to 6.9 is medium and 3.9 or less is low.

Although there is a typo, it should say "9.0 and greater" or similar

grype/presenter/sarif/presenter.go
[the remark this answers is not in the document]
luhring: If it's the tool name, then I'd say just "Grype"? I haven't heard us refer to it as "Anchore Grype Scan"

grype/presenter/sarif/presenter.go
kzantow: Added a little more info to the comment, mentioning the `security-severity` property

grype/presenter/sarif/presenter.go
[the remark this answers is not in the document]
kzantow: Much better idea - Grype it is! (this _was_ `Anchore Container Vulnerability Report (T0)` in the `scan-action`)

grype/presenter/sarif/presenter.go
luhring: So actually these numbers correspond **exactly** with CVSS v3.0 ratings. (See this if you're unfamiliar with these ratings: https://nvd.nist.gov/vuln-metrics/cvss.)

I'm wondering if it makes sense to use our CVSS data when we have it, to preserve some precision about the assigned severity? Just a thought

kzantow: Ah yeah, makes sense; I'll also add this context about why we are returning CVSS if we have it :)

grype/presenter/sarif/presenter.go
```diff
@@ -0,0 +1,390 @@
...
+func (pres *Presenter) toSarifReport() (*sarif.Report, error) {
+	doc, err := sarif.New(sarif.Version210)
+	if err != nil {
+		return nil, err
+	}
+
+	v := version.FromBuild().Version
+	if v == "[not provided]" {
+		// Need a semver to pass the MS SARIF validator
+		v = "0.0.0-dev"
```
wagoodman, line 56: does version usually start with `v`? I can't remember (didn't check)

grype/presenter/sarif/presenter.go
```diff
@@ -0,0 +1,390 @@
...
+			return "high"
+		case vulnerability.MediumSeverity:
+			return "medium"
+		}
+	}
+	return "low"
+}
+
+// cvssScore attempts to get the best CVSS score that our vulnerability data contains
+func (pres *Presenter) cvssScore(v vulnerability.Vulnerability) float64 {
```
wagoodman, line 239: it would be ideal to have a test explicitly for this (there's a lot packed in here)

grype/presenter/sarif/presenter.go
```diff
@@ -0,0 +1,390 @@
...
+	var all []*vulnerability.Metadata
+	for _, related := range v.RelatedVulnerabilities {
+		meta, _ := pres.metadataProvider.GetMetadata(related.ID, related.Namespace)
```
wagoodman, line 242: ignoring an error here --we shouldn't trust any metadata here if there is an error

grype/presenter/sarif/presenter.go
```diff
@@ -0,0 +1,390 @@
...
+	score := -1.0
+
+	// first check vendor-specific entries
+	for _, m := range all {
+		if m.Namespace == "nvd" {
+			continue
+		}
+		for _, cvss := range m.Cvss {
+			if cvss.Metrics.BaseScore > score {
+				score = cvss.Metrics.BaseScore
```
wagoodman, line 255: break when found?

grype/presenter/sarif/presenter.go
```diff
@@ -0,0 +1,390 @@
...
+
+	if score > 0 {
+		return score
+	}
+
+	// next, check nvd entries
+	for _, m := range all {
+		for _, cvss := range m.Cvss {
+			if cvss.Metrics.BaseScore > score {
+				score = cvss.Metrics.BaseScore
```
wagoodman, line 268: break when found?

grype/presenter/sarif/presenter.go
kzantow: gets the highest score, not sure if this is correct behavior

grype/presenter/sarif/presenter.go
kzantow: good question. a `v#.#` would not be semver, so in fact this would be bad to return and i might need to check it

kzantow: Looks like it does not include a `v`, which is good. `grype -vv` outputs:
```
[0000]  INFO grype version: 0.33.1
```
... which is using the same `version.FromBuild().Version` effectively

grype/presenter/sarif/presenter.go
wagoodman: nvm --good call on the highest score extraction

grype/presenter/sarif/presenter.go
wagoodman: nvm --good call on the highest score extraction

wagoodman (approved):

PR #657, opened 2022-03-06
Add additional DB archive decompressors

wagoodman:
This PR does a couple things:
- restores the original decompressors list to match that of `hashicorp/go-getter` used in the `grype db update` execution path
- adds more decompressors to the `grype db import` path

This paves the way for replacing `gzip` DB archives for `zstd` DB archives in the future.

### DIFF
--- a/internal/file/getter.go
+++ b/internal/file/getter.go
@@ -13,15 +13,9 @@ import (
 )
 
 var (
-	archiveExtensions   []string
+	archiveExtensions   = getterDecompressorNames()
 	ErrNonArchiveSource = fmt.Errorf("non-archive sources are not supported for directory destinations")
 )
-
-func init() {
-	for name := range getter.Decompressors {
-		archiveExtensions = append(archiveExtensions, name)
-	}
-}
 
 type Getter interface {
 	// GetFile downloads the give URL into the given path. The URL must reference a single file.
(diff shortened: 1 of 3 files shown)


internal/file/getter.go
```diff
@@ -13,15 +13,16 @@ import (
 )
 
 var (
-	archiveExtensions = []string{
-		".tar",
-		".tar.gz",
-		".tgz",
-		".zip",
-	}
+	archiveExtensions   []string
```
luhring, line 16: nit: It could help readers to understand how `archiveExtensions`'s value is set a bit more easily if we assign the value immediately here as we declare `archiveExtensions` instead of by using an `init()` function. (With this, instead of grabbing the values from `getter.Decompressors` in `init()`, we'd do this logic in a helper function.)

E.g.:

```go
var (
    archiveExtensions = decompressorNames()
    // ...
)
```

luhring (approved):
Looks good! Excited about the performance improvement described in https://github.com/anchore/grype-db-builder/pull/176...

wagoodman:
@luhring same! We'll need to hold off until a new DB schema bump and tie archive formats to DB versions themselves since not all released versions of grype will be able to unarchive these additional formats.

PR #666, opened 2022-03-14
Add platform selection

wagoodman:
Adds `--platform` option to the root commands, allowing someone to specify values similar to `docker pull`, for example:

- os only override: `windows`
- architecture override: `arm64`
- combination: `linux/arm`
- architecture variants: `arm/v8`
- all options: `windows/arm/v6`

Additionally if you attempt to specify `--platform` on non-image sources you will receive an error.


kzantow (approved):

kzantow:
Might add a blurb in the README, at least the big all-config-option-yaml section; fine as is to me, though.

PR #671, opened 2022-03-16
Fix nil pointer access in SARIF output

kzantow:
Got a SIGSEGV when outputting SARIF in one unexplained case; this should prevent that from happening.


luhring (changes requested):
If this is fixing a bug, we need to add a test that proves that this fix works

luhring (approved):
Nice work on the test — thanks!

Issue #623, opened 2022-02-06
Template errors don't lead to non-zero exit status


luhring:
### Steps to reproduce

```console
$ echo 'not a valid Go template {{.' > my.template
$ grype alpine:3.10 -o template -t ./my.template
 ✔ Vulnerability DB        [no update available]
 ✔ Parsed image
 ✔ Cataloged packages      [14 packages]
 ✔ Scanned image           [25 vulnerabilities]
[0001] ERROR unable to show grype-vulnerability-scanning-finished event: unable to show vulnerability report: unable to parse template: template: ./my.template:2: unclosed action started at ./my.template:1
$ echo $?
0
```

### Expectation

Any problems that warrant "error"-level log entries should cause the application to exit non-zero.

Akasurde:
@luhring I raised a PR https://github.com/anchore/grype/pull/674 which might resolve this issue. Can you please take a look? Thanks,


```console
# echo 'not a valid Go template {{.' > a.tmpl
# go run main.go alpine:3.10 -o template -t ./a.tmpl

1 error occurred:
	* unable to parse template: template: a.tmpl:2: unclosed action started at a.tmpl:1

exit status 1
# echo $?
1
```

Issue #625, opened 2022-02-06
Issues with Grype's handling of template output for invalid templates


luhring:
I'm seeing a couple of problems with Grype's template output for invalid user templates. These problems may be related and have a common cause under the hood, so I'm reporting them together here. Feel free to separate these out!

Also, these are related to, but different from, #623.

### Problem 1

Invalid template causes "ERROR" log entry with **ETUI** output, but "WARN" log entry with **logger** output.

```console
$ echo 'not a valid Go template {{.' > my.template
$
$ ### ETUI output
$ grype alpine:3.10 -o template -t ./my.template
 ✔ Vulnerability DB        [no update available]
 ✔ Parsed image
 ✔ Cataloged packages      [14 packages]
 ✔ Scanned image           [25 vulnerabilities]
[0001] ERROR unable to show grype-vulnerability-scanning-finished event: unable to show vulnerability report: unable to parse template: template: ./my.template:2: unclosed action started at ./my.template:1
$
$ ### Logger output
$ grype alpine:3.10 -o template -t ./my.template -v
[0000]  INFO grype version: 0.32.0
[0001]  INFO identified distro: Alpine Linux v3.10 from-lib=syft
[0001]  INFO cataloging image from-lib=syft
[0001]  WARN unable to show catalog image finished event: unable to show vulnerability report: unable to parse template: template: ./my.template:2: unclosed action started at ./my.template:1
```

Note that the final logged events in the two scenarios are not exactly identical.

### Problem 2

When piping an SBOM into Grype via stdin, if the user-supplied template is invalid, the user gets **no output at all** — unless they use `-v` or `-vv`.

When using `-v` or `-vv`, the user does see the same "WARN" event as shown in **Problem 1**.

```console
$ echo 'not a valid Go template {{.' > my.template
$
$ syft alpine:3.10 -o json -q > alpine-3:10.syft.json
$
$ ### default scenario
$ cat ./alpine-3:10.syft.json | grype -o template -t ./my.template
$ # ^^^ no output!
$
$ ### using `-v` or `-vv`
$ cat ./alpine-3:10.syft.json | grype -o template -t ./my.template -v
[0000]  INFO grype version: 0.32.0
[0000]  WARN found package with empty ID while adding to the catalog: Pkg(name="alpine-baselayout" version="3.1.2-r0" type="apk" id="") from-lib=syft
[...]

PR #674, opened 2022-03-17
template: Check sanity for template file

Akasurde:
* Check if template file exists
* Check if template file is readable
* Check if template file is parsable using text\template

Signed-off-by: Abhijeet Kasurde <akasurde@redhat.com>

### DIFF
--- a/grype/presenter/config.go
+++ b/grype/presenter/config.go
@@ -1,6 +1,11 @@
 package presenter
 
-import "fmt"
+import (
+	"errors"
+	"fmt"
+	"os"
+	"text/template"
+)
 
 // Config is the presenter domain's configuration data structure.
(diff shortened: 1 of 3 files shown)


wagoodman (approved):
nice improvement 👍

spiffcs:
I'll find some time to update the tests here so this can pass and we can merge it in

luhring:
I believe this fixes #623 and #625

spiffcs:
updated template tests asserting on new correct template - using the current template that was there through the error `no function toJson`

spiffcs (approved):
Approved now tests are passing

Akasurde:
@spiffcs @wagoodman Thanks for the review and merge.

PR #678, opened 2022-03-17
add initial support for embedded CycloneDX VEX documents

### LINKED ISSUE #591: Add support for cyclonedx 1.4 and VEX
**What would you like to be added**: CycloneDX 1.4 was released with added support for a common vulnerability exchange format. 

It would be great if grype could output its vulnerability reports in this format. This could also be helpful down the road as a standardized format to attach vulnerability data as intoto attestations. 

**Why is this needed**: This provides a well defined standard to output and parse vulnerability information. syft already supports Cyclonedx SBOMs and this could be a great counterpart for grype. 

**Additional context**:
 More details at 
https://cyclonedx.org/capabilities/vex/

https://github.com/CycloneDX/sbom-examples/blob/master/VEX/vex.json

samj1912:
Signed-off-by: Sambhav Kothari <skothari44@bloomberg.net>

Fixes #591 

This adds some initial support for CycloneDX VEX documents. There are a couple of things we will need to think through - 

- Currently this PR produces an embedded VEX document - i.e. we duplicate partial information from the components in the VEX document, add a bom-ref for each and then have the vulnerabilities link to the bom-refs in the same BOM. (We should also verify if my usage of links for components within the same bom is accurate or not, I could not find any example of embedded VEX files at https://github.com/CycloneDX/bom-examples/tree/master/VEX but it is supported according to the docs)
- We need to figure out a way to inject bom-refs in syft for components. Currently I use the PURL when available as a bom-ref and if it is not I use $type-$name-$version as the bom ref.
- We need to figure out a way to pass bom-refs from input sboms to grype so that it can reference it in the vex. In case an input is provided that does not contain a cyclonedx sbom and a non embedded vex is selected, we would need to output both a cyclonedx SBOM and a VEX from grype (this is one possible solution but we can explore other alternatives)
- Currently grype converts input formats into its internal schema which leads to some loss in information that is format specific (for eg. bom refs in cyclonedx's case) - we need to figure out some way to preserve such info if a bom is given as an input.
- VEX had a number of breaking changes from the previous vulnerability extension schema in cyclonedx. As a result, I have kept the previous extension output for b/w compat. See https://github.com/CycloneDX/specification/pull/91 for a full list of changes.
- Because of the lack of support for bom refs in spdx, we would need to figure out a way to generate a non-embedded vex for spdx sbom inputs or we may choose to error out entirely if the input is an spdx bom and the output is a non-embedded cyclonedx vex document until we figure out a way to have bom-refs for spdx documents.

Marking it as a draft while we resolve the above questions.

cc: @luhring , @wagoodman , @spiffcs , @kzantow , @stevespringett, @coderpatros, @nscuro


schema/cyclonedxvex/README.md
```diff
@@ -0,0 +1,5 @@
+# CycloneDX Schemas
+
+`grype` generates a CycloneDX BOm output with the vulnerability extension. This validation is similar to what is done in `syft`, validating output against CycloneDX schemas.
```
hectorj2f, line 3: typo: BOM*

grype/presenter/format.go
```diff
@@ -11,6 +11,8 @@ const (
 	cycloneDXFormat format = "cyclonedx"
 	sarifFormat     format = "sarif"
 	templateFormat  format = "template"
+	embeddedVEXJSON format = "embedded-cyclondex-vex-json"
```
hectorj2f, line 14: typo: `embedded-cyclonedx-vex-json`

grype/presenter/format.go
```diff
@@ -11,6 +11,8 @@ const (
...
+	embeddedVEXXML  format = "embedded-cyclondex-vex-xml"
```
hectorj2f, line 15: same as above

grype/presenter/cyclonedxvex/document.go
```diff
@@ -0,0 +1,84 @@
...
+
+		pkgMatches := matches.GetByPkgID(p.ID)
+
+		if len(pkgMatches) > 0 {
+			for _, m := range pkgMatches {
+				v, err := NewVulnerability(m, provider)
+				if err != nil {
+					return &cyclonedx.BOM{}, err
+				}
+				v.Affects = &[]cyclonedx.Affects{
```
hectorj2f, line 69: Question: should we add a default analysis element per vulnerability ? I'd assume you won't need to analyze any single vulnerability, but I might be wrong. Also, would it make sense to filter the vulnerability added to the VEX file per severity level e.g. severity > Medium ?  
```
      "analysis": {
        "state": "not_affected",        
        }
```

samj1912:
The first cut of this VEX PR is done. 

I have opted to go with an embedded VEX for now so that it is portable across input SBOMs, we can add support for external VEX once the upstream syft PR lands and we have more discussions on how we want to pass format specific data around in grype. (in this case the cdx bom's uuid)

I have also opted to generate a minimal list of components for the embedded VEX which use the same bom ref generation process as the sibling syft PR https://github.com/anchore/syft/pull/914

Finally from conversations with the cyclonedx core team, I have verified that in case of embedded vex documents, the target ref needs to be a bom ref and not an external link. I have put up a PR at https://github.com/CycloneDX/specification/pull/136 to document the same in the upstream spec.

grype/presenter/cyclonedxvex/document.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func getComponent(p pkg.Package) cyclonedx.Component {
+	bomRef := string(p.ID)
+	// try and parse the PURL if possible and append syft id to it, to make
+	// the purl unique in the BOM.
+	// TODO: In the future we may want to dedupe by PURL and combine components with
+	// the same PURL while preserving their unique metadata.
+	if parsedPURL, err := packageurl.FromString(p.PURL); err == nil {
+		parsedPURL.Qualifiers = append(parsedPURL.Qualifiers, packageurl.Qualifier{Key: "syft-id", Value: string(p.ID)})
```
wagoodman, line 63: there is no longer a guarantee that the input SBOM will be from syft, so having a `syft-id` key here could be misleading in some cases. (also, if we make a change here, we should probably head back to syft and match these approaches up... maybe simply `id` would suffice?)

grype/presenter/format.go
```diff
@@ -11,6 +11,8 @@ const (
 	cycloneDXFormat format = "cyclonedx"
 	sarifFormat     format = "sarif"
 	templateFormat  format = "template"
+	embeddedVEXJSON format = "embedded-cyclonedx-vex-json"
+	embeddedVEXXML  format = "embedded-cyclonedx-vex-xml"
```
wagoodman, line 15: these are accurate, though pretty verbose. I'll give a little more thought to something that is shorter / friendlier to input.

hectorj2f:
👋🏻  @samj1912, Do you need a hand with the changes ? We'd be glad to help :)

samj1912:
> 👋🏻 @samj1912, Do you need a hand with the changes ? We'd be glad to help :)

I think the PR is functionally complete. Waiting for @wagoodman's reply on better output format names. As for the syft-id, id comment, will address that soon but I think we should use something more specific than id. Maybe package-id? Thoughts?

hectorj2f:
@samj1912 that is great, I'll give it a try. It sounds good to me to use package ids.

samj1912:
Updated this PR to reflect the comments around syft-id v/s package-id.

grype/presenter/format.go
spiffcs: I don't have any suggestions for how these could be shorter. @wagoodman did you still have thoughts?

hectorj2f:
@samj1912 Thanks 👏🏻

PR #682, opened 2022-03-21
Correct issue with SARIF dir scan relative paths

kzantow:
SARIF is used for submitting to GitHub's security alerts. There is an issue with directory scans that results in relative paths being prefixed by `/`, resulting in the file "not found" in the repository when looking at the Code scanning alerts page.

### DIFF
--- a/grype/presenter/sarif/presenter.go
+++ b/grype/presenter/sarif/presenter.go
@@ -171,10 +171,11 @@ func (pres *Presenter) inputPath() string {
 
 // locationPath returns a path for the location
 func locationPath(l source.Location) string {
+	path := l.RealPath
 	if l.VirtualPath != "" {
-		return l.VirtualPath
+		path = l.VirtualPath
 	}
-	return l.RealPath
+	return strings.TrimPrefix(path, "./")
 }
 
 // locations the locations array is a single "physical" location with potentially multiple logical locations
--- a/grype/presenter/sarif/presenter.go
+++ b/grype/presenter/sarif/presenter.go
@@ -159,14 +159,14 @@ func (pres *Presenter) packagePath(p pkg.Package) string {
 
 // inputPath returns a friendlier relative path or absolute path depending on the input, not prefixed by . or ./
 func (pres *Presenter) inputPath() string {
-	if pres.srcMetadata != nil {
-		inputPath := strings.TrimPrefix(pres.srcMetadata.Path, "./")
-		if inputPath == "." {
-			return ""
-		}
-		return inputPath
+	if pres.srcMetadata == nil {
+		return ""
 	}
-	return ""
+	inputPath := strings.TrimPrefix(pres.srcMetadata.Path, "./")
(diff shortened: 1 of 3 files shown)


grype/presenter/sarif/presenter.go
```diff
@@ -151,34 +151,43 @@ func (pres *Presenter) helpText(m match.Match, link string) *sarif.MultiformatMe
...
+func (pres *Presenter) inputPath() string {
+	if pres.srcMetadata != nil {
+		inputPath := strings.TrimPrefix(pres.srcMetadata.Path, "./")
+		if inputPath == "." {
+			return ""
 		}
-		return packagePath
+		return inputPath
+	}
+	return ""
```
luhring, line 169: nit: This impl could be less indented overall if the check for `pres.srcMetadata == nil` came first

grype/presenter/sarif/presenter.go
```diff
@@ -151,34 +151,43 @@ func (pres *Presenter) helpText(m match.Match, link string) *sarif.MultiformatMe
...
-		}
-		if pres.srcMetadata.Scheme == source.DirectoryScheme {
-			packagePath = fmt.Sprintf("%s/%s", inputPath, packagePath)
+		return locationPath(p.Locations[0])
+	}
+	return pres.inputPath()
+}
+
+// inputPath returns a friendlier relative path or absolute path depending on the input, not prefixed by . or ./
+func (pres *Presenter) inputPath() string {
```
luhring, line 161: Should this expectation be asserted in a test of some sort?

grype/presenter/sarif/presenter_test.go
```diff
@@ -132,6 +131,103 @@ func createDirPresenter(t *testing.T) *Presenter {
 	return pres
 }
 
+func Test_locations(t *testing.T) {
```
luhring, line 134: I think this test is okay for now. ✅  If you're looking to make tests more idiomatic to Go in the future, here are a few thoughts for your consideration:

- It looks like this is testing two independent receiver methods of the `sarif.Presenter` type — `Presenter.inputPath()` and `Presenter.packagePath`. This ends up being a nice way to organize the tests, where the test name can reflect the receiver method name. So instead of `Test_locations`, you might have `Test_Presenter_inputPath` and `Test_Presenter_packagePath`.
- Table-driven tests tend to be a good fit when you have repeating patterns in test setups and assertions. So typically you'd start by defining a slice of test cases, which each test case including a name for the test case, the input data, and then the expected data. And then after that slice is defined, you just just need to define the setup and assertion implementation once, because that can be done inside a `for` loop that iterates over the test cases.

kzantow: Right, these tests just seemed to be every-so-slightly different, which is why I didn't go with the table-driven tests, and these things are quite interrelated which is why I did not separate them into different methods, since I'd end up testing the latter (packgePath) quite a lot with different inputPath permuations.

PR #685, opened 2022-03-22
Correct go.mod to enforce go 1.18

06kellyjac:
Same vein as https://github.com/anchore/syft/pull/897

Since grype now depends on debug/buildinfo go 1.18 is required to build grype and as such go.mod needs updating

> Go 1.18 Release Notes
> New debug/buildinfo package

<https://tip.golang.org/doc/go1.18>

Signed-off-by: 06kellyjac <jack@control-plane.io>

cc: @luhring


luhring (approved):
Thanks! I don't understand exactly why 1,074 lines of `go.sum` are being removed, but I can reproduce the same effect when I make this change locally. Between that, and CI passing, this is good enough for me!

🎉

06kellyjac:
It seems like 1.18 can be a bit smarter when pulling in transitive deps?

But yeah it was all from the `go mod tidy`

PR #686, opened 2022-03-22
Improve SARIF path handling and severity

kzantow:
In SARIF output, which is used by GitHub to display security findings, we need to best represent the files found during directory scans so they may match files within a repository, this corrects one case and adds a number of test cases for different Grype invocations: cwd, relative, and absolute paths.

Additionally, we don't really need a lot of precision for the severity, so we now limit this to 1 decimal place.


luhring (approved):
LGTM!

PR #689, opened 2022-03-23
Bump strset version to fix 386 builds

wagoodman:
Pull in fix from https://github.com/scylladb/go-set/commit/cc7b2070d91ebf40d233207b633e28f5bd8f03a5 to prevent compile issues on 32 bit architectures:

```bash
$ GOOS=linux GOARCH=386 go build  -o /tmp/out examples/basic.go
# github.com/scylladb/go-set/strset
../../.gvm/pkgsets/go1.18/global/pkg/mod/github.com/scylladb/go-set@v1.0.2/strset/strset.go:248:13: cannot use math.MaxInt64 (untyped int constant 9223372036854775807) as int value in assignment (overflows)
../../.gvm/pkgsets/go1.18/global/pkg/mod/github.com/scylladb/go-set@v1.0.2/strset/strset.go:255:16: math.MaxInt64 (untyped int constant 9223372036854775807) overflows int

$ go get -d github.com/scylladb/go-set@master
go: downloading github.com/scylladb/go-set v1.0.3-0.20200225121959-cc7b2070d91e
go: upgraded github.com/scylladb/go-set v1.0.2 => v1.0.3-0.20200225121959-cc7b2070d91e

$ GOOS=linux GOARCH=386 go build  -o /tmp/out examples/basic.go
# success!
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -34,7 +34,8 @@ require (
 	github.com/olekukonko/tablewriter v0.0.5
 	github.com/owenrumney/go-sarif v1.1.1
 	github.com/pkg/profile v1.6.0
-	github.com/scylladb/go-set v1.0.2
+	// pinned to pull in 386 arch fix: https://github.com/scylladb/go-set/commit/cc7b2070d91ebf40d233207b633e28f5bd8f03a5
+	github.com/scylladb/go-set v1.0.3-0.20200225121959-cc7b2070d91e
 	github.com/sergi/go-diff v1.2.0
 	github.com/sirupsen/logrus v1.8.1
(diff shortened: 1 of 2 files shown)


jonasagx (approved):
Do we have tests validating this?

wagoodman:
@jonasagx we don't officially support 32bit builds yet, when we do, we will have a test in the form of a snapshot build (which we already run on all PRs). When we're ready to support 32 bit then simply expanding the build architecture section for linux builds in the `.goreleaser.yaml` will do the trick.

jonasagx:
Nice! GoReleaser is awesome 😎

PR #694, opened 2022-03-24
Include package type in table output

sparrowt:
This helps avoid confusion between packages of the same name but different types.

Examples where I've hit this recently:
 - `tar` is a node package as well as a linux executable apk/rpm/deb
 - `msgpack` is a node package but also a python package
 - `jsonpointer` is also the name of both a node and a python package

In each case when I saw the vuln reported in grype output I started investigating the "wrong" one (call me unlucky!) and it took some digging to realise the issue.

The "type" is a succinct representation of _where_ Grype found this package, so hopefully including this should make things a lot clearer.

I'm aware of the templating functionality with `-o` however:

1. I've struggled to make a custom text-tabular format in order to simply add that column while retaining the nice alignment (I see `tablewriter` in `...table/presenter.go` but I don't think custom templates can make use of that via `...template/presenter.go`)
2. Even using `printf` with fixed-length padding (🤮) the ordering is problematic (#696)
3. This change hopefully makes an improvement which can benefit everyone by default.

### DIFF
--- a/grype/presenter/table/presenter.go
+++ b/grype/presenter/table/presenter.go
@@ -34,7 +34,7 @@ func NewPresenter(results match.Matches, packages []pkg.Package, metadataProvide
 func (pres *Presenter) Present(output io.Writer) error {
 	rows := make([][]string, 0)
 
-	columns := []string{"Name", "Type", "Installed", "Fixed-In", "Vulnerability", "Severity"}
+	columns := []string{"Name", "Installed", "Fixed-In", "Type", "Vulnerability", "Severity"}
 	for m := range pres.results.Enumerate() {
 		var severity string
 
@@ -55,7 +55,7 @@ func (pres *Presenter) Present(output io.Writer) error {
 			fixVersion = ""
 		}
 
-		rows = append(rows, []string{m.Package.Name, string(m.Package.Type), m.Package.Version, fixVersion, m.Vulnerability.ID, severity})
+		rows = append(rows, []string{m.Package.Name, m.Package.Version, fixVersion, string(m.Package.Type), m.Vulnerability.ID, severity})
 	}
 
 	if len(rows) == 0 {
--- a/grype/presenter/table/presenter.go
+++ b/grype/presenter/table/presenter.go
@@ -34,7 +34,7 @@ func NewPresenter(results match.Matches, packages []pkg.Package, metadataProvide
 func (pres *Presenter) Present(output io.Writer) error {
 	rows := make([][]string, 0)
 
-	columns := []string{"Name", "Installed", "Fixed-In", "Vulnerability", "Severity"}
+	columns := []string{"Name", "Type", "Installed", "Fixed-In", "Vulnerability", "Severity"}
 	for m := range pres.results.Enumerate() {
 		var severity string
 
@@ -55,7 +55,7 @@ func (pres *Presenter) Present(output io.Writer) error {
 			fixVersion = ""
 		}
 
-		rows = append(rows, []string{m.Package.Name, m.Package.Version, fixVersion, m.Vulnerability.ID, severity})
+		rows = append(rows, []string{m.Package.Name, string(m.Package.Type), m.Package.Version, fixVersion, m.Vulnerability.ID, severity})
 	}
 
 	if len(rows) == 0 {
--- a/grype/presenter/cyclonedx/presenter_test.go
+++ b/grype/presenter/cyclonedx/presenter_test.go
@@ -18,7 +18,7 @@ import (
 	"github.com/anchore/syft/syft/source"
 )
 
-var update = flag.Bool("update", false, "update the *.golden files for json presenters")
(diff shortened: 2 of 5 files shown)


grype/presenter/table/presenter.go
```diff
@@ -55,7 +55,7 @@ func (pres *Presenter) Present(output io.Writer) error {
 			fixVersion = ""
 		}
 
-		rows = append(rows, []string{m.Package.Name, m.Package.Version, fixVersion, m.Vulnerability.ID, severity})
+		rows = append(rows, []string{m.Package.Name, m.Package.Type, m.Package.Version, fixVersion, m.Vulnerability.ID, severity})
```
luhring, line 58: Note: `m.Package.Type` will have to be converted to a `string` type for this to compile successfully.

luhring:
This is intriguing. I have a few thoughts so far. 🤔 

The table view is the default view that we show to all Grype users. We seldom change it. We want it to "just work". Adding a new column to the table view gives the user more information, but it also risks the table starting to "wrap", where it becomes unclear which values are part of which column. So we like to be sensitive to the potential "width" that results from adding new columns.

>In each case when I saw the vuln reported in grype output I started investigating the "wrong" one (call me unlucky!) and it took some digging to realise the issue.

The table view is intended to be just a summary. For full visibility into the vulnerability scan results, we recommend using a different format based on the user's needs, such as via `-o json`.

Having said that, I definitely like the idea of adding "type" to the table view. We show it in Syft's table view. As you point out, this qualifier for a package ends up being extremely handy, and I could see it being a pain to need the JSON output just to understand which "type" a particular package is.

grype/presenter/table/presenter.go
sparrowt: Ah yes, sorry not to have tried it locally already - I've hit some issues during dev setup and (foolishly!) thought I might get away with a nice simple change... but have now resolved these, got a working local setup & pushed a fix for this 👍

sparrowt:
Thanks @luhring, yeah I appreciate that this is a very visible change so requires careful thought - if it's something you folks are happy with though, I think it's quite a good usefulness:space ratio and would save digging through verbose JSON output.

Now that I'm _aware_ of the potential for confusion I can of course go looking in the JSON output (or build my own wrapper for use in CI to display it automatically) - but including it by default would definitely have helped my past self as a n00b here.

Looking at the width increase, for a decent chunk (>40%) of the [types](https://github.com/anchore/syft/blob/47ea91086879c3e57cf399a19a8aa2a3d0d71340/syft/pkg/type.go#L10-L22) it's going to be an increase of just 6 chars (`TYPE  ` being longer than `apk`/`rpm` etc.) though of course it could be up to a 16 char increase (e.g. `jenkins-plugin  ` = 14 + 2 padding spaces). On average the increase is _9.5 characters_ (making a slightly nonsense assumption of a uniform distribution of different syft `Type` options...). Notably, unlike some other fields one might consider adding, there is a hard limit to the maximum width increase this column could add given the fixed set of possible strings.

Sampling a bunch of random grype outputs from different CI jobs I use, my (doubtless unrepresentative!) sample ranged from 58 to 78 characters wide, so even at the upper end there (78 + 16 maximum increase) it's not getting over the width of the [README demo gif](https://user-images.githubusercontent.com/590471/90276236-9868f300-de31-11ea-8068-4268b6b68529.gif) (which is, I'm sure, a very important measuring stick 😉)

Jokes aside, I'd propose that this increase would be worth it & not likely to cause wrapping in the majority of scenarios, but of course the decision lies with ye excellent folks who maintain this tremendously useful project (for which thank you!)

spiffcs:
Thanks for the other cleanups also on the update flags. I'm going to LGTM this based on @sparrowt last comment on his analysis of the width changes.

Waiting on a second from @wagoodman to see if this is ok to go in.

grype/presenter/table/presenter.go
```diff
@@ -34,7 +34,7 @@ func NewPresenter(results match.Matches, packages []pkg.Package, metadataProvide
 func (pres *Presenter) Present(output io.Writer) error {
 	rows := make([][]string, 0)
 
-	columns := []string{"Name", "Installed", "Fixed-In", "Vulnerability", "Severity"}
+	columns := []string{"Name", "Type", "Installed", "Fixed-In", "Vulnerability", "Severity"}
```
wagoodman, line 37: open question @sparrowt @luhring @spiffcs : should we try to be consistent with syft, which is `name, version, type ...` thus change this to `name, installed, fixed-in, type ...`,... or not?

spiffcs: Yea I think consistency here between the tools is good and then we are good to merge this PR.

sparrowt: Okey doke, moved in 14177cc

wagoodman (approved):
Thanks for polishing the output!

adriens:
:clap:

luhring:
Just following up to say that I'm really liking this extra context in the output when I run Grype!!! 🙏

PR #699, opened 2022-03-30
Add byMatchName custom function for custom template users

spiffcs:
## Custom template functions `byName`

Add a `byName` custom function for custom template consumers.

Example template:
```
TYPE   NAME                 INSTALLED    VULN                 SEVERITY     FIXED?   FIXED IN     UPSTREAM PACKAGE
{{- range .Matches | byMatchName}}
{{printf "%-6s %-20s %-12s %-20s %-12s %-8s %-12s %s" .Artifact.Type .Artifact.Name .Artifact.Version .Vulnerability.ID .Vulnerability.Severity .Vulnerability.Fix.State (.Vulnerability.Fix.Versions | join ", ") (.Artifact.Upstreams | join ", ")}}
{{- end}}
```

Note the `|` for `Matches` to be sorted.

Using the above template run:
```
grype alpine:3.14.1 -o template -t mycustom.tmpl
```

Output should now be able to be sorted by name for custom templates:
Screen Shot 2022-03-30 at 11 51 09 AM

### DIFF
--- a/grype/presenter/template/presenter.go
+++ b/grype/presenter/template/presenter.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"os"
 	"reflect"
+	"sort"
 	"text/template"
 
 	"github.com/Masterminds/sprig/v3"
@@ -83,6 +84,20 @@ var funcMap = func() template.FuncMap {
 		}
 
 		return 0
+	}
+	f["byName"] = func(collection interface{}) interface{} {
+		t := reflect.TypeOf(collection)
+		if t.Kind() != reflect.Slice {
+			return collection
+		}
+
+		matches, ok := collection.([]match.Match)
+		if !ok {
+			return collection
+		}
+
+		sort.Sort(match.ByName(matches))
+		return matches
 	}
 	return f
 }()
--- a/grype/presenter/template/presenter.go
+++ b/grype/presenter/template/presenter.go
@@ -91,12 +91,12 @@ var funcMap = func() template.FuncMap {
 			return collection
 		}
 
-		matches, ok := collection.([]match.Match)
+		matches, ok := collection.([]models.Match)
 		if !ok {
 			return collection
 		}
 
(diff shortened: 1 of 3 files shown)


grype/presenter/template/presenter.go
```diff
@@ -84,5 +85,19 @@ var funcMap = func() template.FuncMap {
 
 		return 0
 	}
+	f["byName"] = func(collection interface{}) interface{} {
+		t := reflect.TypeOf(collection)
+		if t.Kind() != reflect.Slice {
+			return collection
+		}
```
kzantow, line 92: Is this necessary, given the next check?

kzantow (approved):
One small nit, not sure the extra code is necessary

grype/presenter/template/presenter.go
```diff
@@ -84,5 +85,19 @@ var funcMap = func() template.FuncMap {
 
 		return 0
 	}
+	f["byName"] = func(collection interface{}) interface{} {
```
wagoodman, line 88: since this function name can be used anywhere in the template, and the function is only intended for matches, we should probably make the name more explicit to convey proper usage.

Something like `byMatchName` or `sortByMatchName`, so there is a hint that if you aren't using this on matches that you're probably using it incorrectly.

grype/presenter/template/presenter.go
spiffcs: I think you need this check before doing the next check since if it was NOT a slice then the next check would panic

grype/presenter/template/presenter.go
spiffcs: `byMatchName` I think reads correctly here based on other templating languages. I'll update it to that.

grype/presenter/template/presenter.go
kzantow: It won't panic because of the form `var, ok := thing.(cast)`

PR #703, opened 2022-03-31
Migrate LocationSet and add Dart support

jonasagx:
Adapts grype to use LocationSet introduced in https://github.com/anchore/syft/pull/930
Updates tests to support Dart packages.

Signed-off-by: Jonas Galvão Xavier <jonas.agx@gmail.com>


test/integration/compare_sbom_input_vs_lib_test.go
```diff
@@ -44,14 +44,15 @@ func TestCompareSBOMInputToLibResults(t *testing.T) {
 
 	definedPkgTypes := strset.New()
 	for _, p := range syftPkg.AllPkgs {
-		definedPkgTypes.Add(string(p))
+		definedPkgTypes.Add(p)
 	}
-	// exceptions: rust, php, and msrc (kb) are not under test
+	// exceptions: rust, php, dart and msrc (kb) are not under test
```
jonasagx, line 49: Q: how we decide what goes in this list of exceptions?

wagoodman: good question. I think context matters is the short answer. MSRC is not officially supported, and rust/php does not have specific support other than the stock matcher. I think dart falls into the latter category as well. As we add more catalogers we should try to eventually follow up with equivalent matcher support in grype.

wagoodman (approved):
Thanks for picking this one up 🙌

Issue #644, opened 2022-02-24
Read attestation file, validate attestation, produce vulnerability report


spiffcs:
## What would you like to be added:
Grype should be able to consume SBOM inputs from verified image attestations.

### Context:
Attestations represent authenticated metadata about a set of software artifacts. They are normally intended for consumption by automated policy engines. They are stored as `layers` on `manifests` of a given image `digest`

Key attestation terms:
- [Envelope](https://github.com/in-toto/attestation/blob/main/spec/README.md#envelope): Handles authentication and serialization
- [Statement](https://github.com/in-toto/attestation/blob/main/spec/README.md#statement): Binds the attestation to a particular subject and unambiguously identifies the types of the predicate.
- [Predicate](https://github.com/in-toto/attestation/blob/main/spec/README.md#predicate): Contains arbitrary metadata about the subject, with a type-specific schema
- [Bundle](https://github.com/in-toto/attestation/blob/main/spec/bundle.md): Defines a method of grouping multiple attestations together

Example:
The following commands use [crane](https://github.com/google/go-containerregistry/tree/main/cmd/crane) to help inspect a sample image.

Additional reading can be found here on how to navigate around using the OCI specification:
https://github.com/google/go-containerregistry/blob/main/pkg/v1/mutate/README.md
https://github.com/google/go-containerregistry/blob/main/pkg/v1/remote/README.md
https://github.com/google/go-containerregistry/blob/main/pkg/v1/stream/README.md
https://github.com/google/go-containerregistry/blob/main/pkg/v1/tarball/README.md
https://github.com/google/go-containerregistry/blob/main/pkg/v1/remote/transport/README.md

This command will get you the manifest for a given image's latest tag
```
crane manifest caphill4/attest:latest | sha256sum 
031edd78299a1f03443826ac7b74df2e97f10469078f6d49949aebe5ccfb019f
```

You can then use that output with the following command to get you the details about that manifest
```
crane manifest caphill4/attest:sha256-031edd78299a1f03443826ac7b74df2e97f10469078f6d49949aebe5ccfb019f.att | jq .  

{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.manifest.v1+json",
  "config": {
    "mediaType": "application/vnd.oci.image.config.v1+json",
    "size": 778,
    "digest": "sha256:ac4e91a48dbb594b4fdfea9d5927acf77ef473c1b97a186c3f96d894e1a46a63"
  },
  "layers": [
    {
[...]

PR #706, opened 2022-04-01
Consume attestation files

jonasagx:
Closes: https://github.com/anchore/grype/issues/644 by consuming attestations files. Scanning attestation attached to images will be tackled later.

## output with verification
image

## output without verification
image

Please try it out:
```sh
# build local grype
go build main.go
# consume implicit attestation
./main agx-alpine-hello.att.json --key cosign.pub
# consume explicit attestation
./main att:agx-alpine-hello.att.json --key cosign.pub
```

## Design decisions
- Grype is implementing attestation verification and decoding independent of Syft, we might move them to Syft as library features used by Grype.


luhring (changes requested):
I left some inline review comments for the code itself, but I'll also include some overall comments from looking at the feature from a user's perspective. I'll number them here so that it's a bit easier to refer to these individual comments later...

- [ ] I don't see any documentation for this new feature. We should explain how this new functionality works to users in the project's README.

- [x] It looks like piping an attestation envelope into Grype (via STDIN) doesn't work. Just like with SBOMs, users should be able to run a command like `cat ./attestation/json | grype --key ./cosign.pub` and have it behave as if the attestation path was provided as a CLI arg to Grype. We should also put this workflow under test.

- [ ] The Grype binary size increases significantly with this change — from `31 MB` to `85 MB` (almost a 3x increase). It wasn't clear to me from the description if this was known and/or expected. I know we saw a binary size increase in Syft when we introduced the attestation functionality, but I wasn't sure if the impact to Grype was known to be the exact same or if there was some mitigation opportunity available to us. (For reference, it looks like the latest Syft binary is at `67 MB`).

- [ ] I'm wondering if we considered an implementation where SBOM attestation logic lives solely within Syft and then is exported to consumers, like Grype. At first glance, this approach would seem analogous to how we handle SBOM creation and SBOM consumption between Syft and Grype, respectively. (To be clear, I don't see this as a necessity for the Grype implementation, I just want to understand the thinking.)

- [ ] Should Grype warn (or even **error out**) when the user provides a `--key` argument for an input that's **not an attestation**? I'm thinking about security expectations... Right now, if a key is provided but the input is not an attestation, Grype will continue to scan the target as usual. I think it'd be reasonable to assume that if the user has provided a `--key` argument, they are expecting a cryptographic verification to performed, and a successful verification to be required as a condition to a successful vulnerability scan. We should try to avoid scenarios where the user believes they can trust the input when really we're not sure if they can trust the input.

- [ ] When I run Grype with `att:./SOMETHING`, where the "SOMETHING" is **not an attestation**, I think the error message shown to users could be more direct about what the problem is. Right now, I see `failed to catalog: invalid payloadType  on envelope. Expected application/vnd.in-toto+json` (and actually, with an extra space after `payloadType` now that I look at the raw output again). I'm wondering if we could just say something that communicates that the input is not an attestation, even though the user told Grype to interpret the input as an attestation (using `att:`). What are your thoughts?

- [ ] When I run Grype on an attestation, but I **don't** provide the `--key` flag, I think the error message could be more direct, and could tell me that I didn't provide a public key for the verification. Right now, the error message is: `failed to catalog: failed to verify attestation signature: failed to load public key: read .: is a directory`.

jonasagx:
Thanks for the feedback @luhring -- I am working on the actionable items. Items 3 and 4 on your list are design discussions we can have over zoom.

internal/config/application.go
```diff
@@ -48,6 +48,7 @@ type Application struct {
 	FailOn              string                  `yaml:"fail-on-severity" json:"fail-on-severity" mapstructure:"fail-on-severity"`
 	FailOnSeverity      *vulnerability.Severity `yaml:"-" json:"-"`
 	Registry            registry                `yaml:"registry" json:"registry" mapstructure:"registry"`
+	Key                 string                  `yaml:"key" json:"key" mapstructure:"key"`
```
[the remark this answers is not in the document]
jonasagx: I renamed it to `AttestationKey` for now.

grype/pkg/syft_sbom_provider.go
```diff
@@ -23,62 +34,73 @@ func (e errEmptySBOM) Error() string {
...
-	if isPossibleSBOM(userInput) {
-		sbom, err := openSbom(userInput)
+		return sbomFile, nil
+	case isPossibleSBOM(userInput):
+		sbomFile, err := openSbom(userInput)
 		if err == nil {
-			return sbom, nil
+			return sbomFile, nil
 		}
+		log.Warnf("failed openning input file: %v", err)
```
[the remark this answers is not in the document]
jonasagx: It was meant for local debugging, I just dropped it

grype/pkg/syft_sbom_provider.go
```diff
@@ -23,62 +34,73 @@ func (e errEmptySBOM) Error() string {
...
 
 	// no usable SBOM is available
 	return nil, errDoesNotProvide
 }
 
-// sbomHasContent returns a bool indicating whether the given SBOM file has data that could possibly be utilized in
+// fileHasContent returns a bool indicating whether the given SBOM file has data that could possibly be utilized in
 // downstream processing.
-func sbomHasContent(sbom *os.File) bool {
+func fileHasContent(sbom *os.File) bool {
```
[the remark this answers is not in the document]
jonasagx: yes, done

grype/pkg/syft_sbom_provider.go
```diff
@@ -109,34 +131,114 @@ func stdinReader() io.Reader {
...
+
+func verifyAttestationSignature(env *ssldsse.Envelope, key string) error {
+	pubKey, err := signature.PublicKeyFromKeyRef(context.TODO(), key)
+	if err != nil {
+		return fmt.Errorf("failed to load public key: %w", err)
+	}
+	pkcs11Key, ok := pubKey.(*pkcs11key.Key)
+	if ok {
+		defer pkcs11Key.Close()
+	}
```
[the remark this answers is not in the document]
jonasagx: I thought they were closing the file, turns out the close function is not defined.
Cosign does that for some reason: https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/verify/verify.go#L125-L128

grype/pkg/syft_sbom_provider.go
```diff
@@ -23,62 +34,73 @@ func (e errEmptySBOM) Error() string {
...
+func getSBOM(userInput string, config ProviderConfig) (*sbom.SBOM, error) {
+	reader, err := getSBOMReader(userInput, config)
+	if err != nil {
+		return nil, err
+	}
+
+	s, format, err := syft.Decode(reader)
+	if err != nil && err.Error() == "unable to identify format" && !explicitlySpecifyAttestation(userInput) {
+		return getSBOM(fmt.Sprintf("att:%s", userInput), config)
+	}
```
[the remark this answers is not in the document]
jonasagx: I refactored that function

grype/pkg/syft_sbom_provider.go
```diff
@@ -23,67 +34,120 @@ func (e errEmptySBOM) Error() string {
 }
 
 func syftSBOMProvider(userInput string, config ProviderConfig) ([]Package, Context, error) {
-	reader, err := getSBOMReader(userInput)
+	s, err := getSBOM(userInput, config)
```
spiffcs, line 37: Nice! I like that this is simpler without the extra step. A good candidate we can look at to what we might want to see eventually exported from syft since we had to write our own function here.

jonasagx:
@luhring on binary size:
Syft's size at v0.44.0 is 70MB. 
The snapshot of Grype (Darwin/amd64) for this PR is 71MB.

README.md
```diff
@@ -123,6 +123,27 @@ use the `--distro <distro>:<version>` flag. A full example is:
 grype --add-cpes-if-none --distro alpine:3.10 sbom:some-apline-3.10.spdx.json
 ```
 
+### Scan attestations
+Grype can scan SBOMs from attestations, as long as they are encoded as part of (in-toto envelope)[https://github.com/in-toto/attestation/blob/main/spec/README.md#envelope], that is the format used by Syft.
```
spiffcs, line 127: nit:
```suggestion
Grype can scan SBOMs from attestations, as long as they are encoded as part of and [in-toto envelope](https://github.com/in-toto/attestation/blob/main/spec/README.md#envelope).
```

This should clean up the link. Let me know if you think it reads better by dropping the end statement

cmd/root.go
```diff
@@ -60,7 +60,8 @@ You can also explicitly specify the scheme to use:
     {{.appName}} dir:path/to/yourproject                read directly from a path on disk (any directory)
     {{.appName}} sbom:path/to/syft.json                 read Syft JSON from path on disk
     {{.appName}} registry:yourrepo/yourimage:tag        pull image directly from a registry (no container runtime required)
-
+    {{.appName}} att:attestation.json --key cosign.pub  extract and scan SBOM from attestation file. Checkout Syft to attest your SBOMs
+    {{.appName}} attestation.json --key cosign.pub			the scheme att: is optional
```
spiffcs, line 64: nit: instead of informing the user that the scheme is optional I would use the above help as an example and say something like `explicitly use the input as an attestation`

cmd/root.go
```diff
@@ -153,6 +154,11 @@ func setRootFlags(flags *pflag.FlagSet) {
 		"platform", "", "",
 		"an optional platform specifier for container image sources (e.g. 'linux/arm64', 'linux/arm64/v8', 'arm64', 'linux')",
 	)
+
+	flags.String(
+		"key", "",
```
spiffcs, line 159: I think we want the default here to be "cosign.key" since that's the default that syft uses as well.
```suggestion
		"key", "cosign.key",
```

spiffcs (commented):
Nice Job on the PR! Let's sync this afternoon about steps going forward that unblock this. I still need to read and understand the other PR comments for what else was request, but I think we're really close here!

grype/pkg/syft_sbom_provider.go
```diff
@@ -23,67 +34,120 @@ func (e errEmptySBOM) Error() string {
...
 		if err != nil {
-			return nil, fmt.Errorf("unable to use specified SBOM: %w", err)
+			return nil, nil, err
 		}
+		r, err := getSBOMFromAttestation(f, config)
+		return r, newInputInfo("att", "att"), err
 
-		if !sbomHasContent(sbom) {
-			return nil, errEmptySBOM{filepath}
+	case isPossibleAttestation(userInput):
```
spiffcs, line 110: We talked about this on the call but just wanted to capture this for others to see. I think this can return true if an sbom had some description or comment field with the mime type in it.

I think we want to make sure we account for cases where `isPossibleAttestation` and `isPossibleSBOM` both return true. 

If both return true then only one of the resulting workflow will not return an error so we can try both and decide which one to return to the user. If both return an error saying we tried both

test/cli/utils_test.go
```diff
@@ -120,19 +122,9 @@ func repoRoot(tb testing.TB) string {
...
-
-	_, err = io.Copy(stdin, file)
-	if err != nil {
-		tb.Fatal(err)
-	}
-	err = stdin.Close()
-	if err != nil {
-		tb.Fatal(err)
-	}
+	b, err := io.ReadAll(file)
```
spiffcs, line 125: Was this erroring before? What was the reason for the change in this file?

spiffcs:
Small comments added and I think this needs a rebase or merge into master

test/cli/utils_test.go
jonasagx: this approach errors when the input file is > 30k, I suspect the stdin buffer got full making the test hang for ever.

spiffcs:
@jonasagx this looks really good! Nice job on the PR and thanks for responding to the feedback. 

All my changes have been addressed. Waiting for @luhrings feedback. Also maybe we can all sync on the concerns about the size?

spiffcs:
Per meeting comments - We need to update the error for when an attestation prefix is passed on a file that is a regular `JSON` sbom. The test should also be updated to check the error text.

grype/pkg/syft_sbom_provider.go
```diff
@@ -23,67 +34,126 @@ func (e errEmptySBOM) Error() string {
...
+	}
+}
+
+type inputInfo struct {
+	ContentType string
+	Scheme      string
+}
+
+func getSBOM(userInput string, config ProviderConfig) (*sbom.SBOM, error) {
+	reader, info, err := getSBOMReader(userInput, config)
```
spiffcs, line 61: I think we don't want to change the signature here. `getSBOMReader` should not need to return the extra info here . `getSBOMReader` can be given to the agency to determine which error it returns based on the input discovered.

grype/pkg/syft_sbom_provider_test.go
```diff
@@ -21,6 +23,180 @@ func must(c pkg.CPE, e error) pkg.CPE {
...
+func assertAs(expected string) assert.ErrorAssertionFunc {
+	return func(t assert.TestingT, err error, i ...interface{}) bool {
+		return assert.ErrorContains(t, errors.New(expected), err.Error())
+	}
+}
+
+func TestDecodeStdin(t *testing.T) {
+	tests := []struct {
+		Name    string
+		Fixture string
```
spiffcs, line 35: Per meeting comments `fixture` => `input`

grype/pkg/syft_sbom_provider_test.go
```diff
@@ -21,6 +23,180 @@ func must(c pkg.CPE, e error) pkg.CPE {
...
+func TestParseAttestation(t *testing.T) {
+	tests := []struct {
+		Name    string
+		Fixture string
+		Key     string
+		WantErr assert.ErrorAssertionFunc
+		PkgsLen int
+	}{
+		{
+			Name:    "happy path with schema",
```
spiffcs, line 106: replace all instances of mistake `schema` => `scheme`

README.md
```diff
@@ -123,6 +123,27 @@ use the `--distro <distro>:<version>` flag. A full example is:
 grype --add-cpes-if-none --distro alpine:3.10 sbom:some-apline-3.10.spdx.json
 ```
 
+### Scan attestations
+Grype can scan SBOMs from attestations, as long as they are encoded as part of [in-toto envelope](https://github.com/in-toto/attestation/blob/main/spec/README.md#envelope).
```
spiffcs, line 127: nit: I think we remove comma here

spiffcs: Apologies if this came from my earlier suggestion

spiffcs:
Fast follow after this PR on validating predicate type in case of modification between attestation generation and attestation consumption

README.md
spiffcs: Comment from meeting - Add information to feature session for this PR

spiffcs: Comment from meeting - update scheme section to talk about new source and how it is a more secure version of the grype workflow

spiffcs: Comment from meeting - update Scan attestations section Examples with full connective story of commands

spiffcs:
From Meeting: Update TUI message to show verification rather than bury it in the info section

internal/config/application.go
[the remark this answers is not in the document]
luhring: Thought: should we update to `AttestationVerificationKey` or something like that? We're considering adding attestation creation down the road, where we'd need a signing key, too. At that point, I could see `AttestationKey` being ambiguous as to its intended use... WDYT?

grype/pkg/syft_sbom_provider.go
```diff
@@ -23,62 +34,73 @@ func (e errEmptySBOM) Error() string {
...
-		Source: &sbom.Source,
-		Distro: sbom.Artifacts.LinuxDistribution,
-	}, nil
+	return s, nil
 }
 
-func getSBOMReader(userInput string) (io.Reader, error) {
-	if userInput == "" {
+func getSBOMReader(userInput string, config ProviderConfig) (io.Reader, error) {
+	switch {
```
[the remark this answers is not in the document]
luhring: @jonasagx Ping on this one as still needing an answer

internal/config/application.go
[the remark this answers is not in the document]
luhring: A related question: Should `Attestation` be its own config section?

grype/pkg/syft_sbom_provider.go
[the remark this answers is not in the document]
jonasagx: With attestations two condition were added, and the conditions are all mutually exclusive so it seems easier to read them as a switch/case block.

internal/config/application.go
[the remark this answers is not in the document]
jonasagx: 1. The field name currently is `AttestationKey`, which doesn't differentiate whether it is for public our private key, so we can use the same field for verification _and_ creation of an attestation. This is similar to how [cosign handles](https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/attest.go#L67) the `--key` parameter.
2. Adding an attestation section sounds like a good idea, but I for the use cases of this PR the only field needed is a key. We can include it now and expand later as well. Keyless workflows might be the biggest contributor here -- cc @spiffcs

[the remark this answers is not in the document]
luhring: When it comes to code, I think we should be careful not to use the same data field to store something that's not sensitive (like a public key) as something that is sensitive (like a private key). The consequence of getting these two keys mixed up could be very, very bad.

>so we can use the same field for verification and creation of an attestation

I don't think I follow... How would this work? If Grype was consuming an SBOM attestation and producing a vulnerability scan attestation at the same time, we'd use the same key parameter for both the public verification key and the private signing key?

ui/event_handlers.go
```diff
@@ -131,3 +131,51 @@ func (r *Handler) VulnerabilityScanningStartedHandler(ctx context.Context, fr *f
...
+	if err != nil {
+		return err
+	}
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+
+		spin := color.Green.Sprint(completedStatus)
+		title := tileFormat.Sprint("Skipped attestation proof")
```
jonasagx, line 175: @luhring WDYT of this message?

ui/event_handlers.go
```diff
@@ -25,7 +25,7 @@ const completedStatus = "✔"                // "●"
 const tileFormat = color.Bold
 
 var auxInfoFormat = color.HEX("#777777")
-var statusTitleTemplate = fmt.Sprintf(" %%s %%-%ds ", syftUI.StatusTitleColumn)
+var statusTitleTemplate = fmt.Sprintf(" %%s %%-%ds ", syftUI.StatusTitleColumn+4)
```
jonasagx, line 28: @wagoodman is this ok? Background: increased it so new (longer) messages get aligned.

internal/config/application.go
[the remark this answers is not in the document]
jonasagx: I see, you're right. I will rename the config field to `Attestation.VerificationKey`, since it has its own section now.

ui/event_handlers.go
luhring: @jonasagx I'd say "verification" instead of "proof"

ui/event_handlers.go
```diff
@@ -131,3 +131,51 @@ func (r *Handler) VulnerabilityScanningStartedHandler(ctx context.Context, fr *f
...
+		sig = k.Sig.Sig
+	}
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+
+		spin := color.Green.Sprint(completedStatus)
+		title := tileFormat.Sprint("Attestation verified")
+		auxInfo := auxInfoFormat.Sprintf("[signature %s]", sig)
```
luhring, line 157: I don't think we need to output the signature itself to the TUI

ui/event_handlers.go
```diff
@@ -131,3 +131,51 @@ func (r *Handler) VulnerabilityScanningStartedHandler(ctx context.Context, fr *f
...
+		return err
+	}
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+
+		spin := color.Green.Sprint(completedStatus)
+		title := tileFormat.Sprint("Skipped attestation proof")
+		auxInfo := auxInfoFormat.Sprint("[no key was used to verify content]")
```
luhring, line 176: I don't think we need this secondary message

ui/event_handlers.go
jonasagx: please ignore

spiffcs (approved):
Approved with new changes thanks @jonasagx - Looking for @luhring review and to remove his request for changes here

luhring (approved):

PR #714, opened 2022-04-08
Add java.Matcher configuration to includes maven upstream sha1 query

spiffcs:
## Add Grype Matcher configuration for upstream maven query based on SHA1

Syft is now able to generate `sha1` digests for discovered java archives. These digests are included along with the java archive metadata output.
https://github.com/anchore/syft/pull/941

This PR adds the ability for the user to configure the `java.Matcher` to query the maven upstream repository for additional data that may not be present at time of SBOM generation. 

Specifically, sboms of java archives are sometimes limited in their details when `pom.xml` or `pom.properties` are not present at the time of analysis. Querying `maven` given an archives sha1 digest value allows users to add fidelity to their vulnerability report when matching against data not relying on CPE generation (GHSA or GL Advisory Database).

It should be noted that a NEW package is not generated for the grype output. Matches found with this method are assigned directly to the java package that presented the sha1 for search. No duplication should be caused from doing this additional query.

The `.grype.yaml` configuration has been updated to include the following block:
```
external-sources:
  enable: true
  maven:
    search-maven-upstream: true
    base-url: "https://search.maven.org/solrsearch/select"
```

### Features added

- Matcher Config options at the app config level
- Maven upstream package search with vulnerability comparison

### TODO

- [ ] Test Updates

### Demo
Disabled - 606 vulnerabilities reported:
Screen Shot 2022-04-11 at 3 25 58 PM

Enabled - 682 vulnerabilities reported:
Screen Shot 2022-04-11 at 3 26 07 PM

### Yardstick Analysis vs current release and branch:
This shows all of the additional vulnerabilities captured when enabling upstream matching:

Screen Shot 2022-04-11 at 6 01 37 PM

None of these findings appear to be FP and can all be accurately attributed to the image.

What's interesting here is that after running yardstick the new branch shows `22` new vulnerabilities reported. The output of the respective commands (current release vs branch) shows a difference of `76` when comparing the `Scanned image           [xxx vulnerabilities]` field (606 vs 682). This disparity has been found to be related to additional `relatedVulnerabiities` being populated as a result of enabling the upstream source. When searching by sha and matching with `GHSA` is also will generate a new `relatedVulnerability` for the `NVD source`. New vulnerabilities PLUS packages that can now be found via the GHSA and not just CPE account for this increase in the output.

I have found that the branch results will overwrite `exact-direct-match` with `exact-indirect-match` if a vulnerability is found upstream via the checksum search. 


grype/matcher/java/matcher.go
```diff
@@ -9,7 +16,104 @@ import (
...
+	// artifacts might have the same SHA-1 digests.
+	// e.g. "javax.servlet:jstl" and "jstl:jstl"
+	docs := res.Response.Docs
+	sort.Slice(docs, func(i, j int) bool {
+		return docs[i].ID < docs[j].ID
+	})
+	d := docs[0]
+
+	return &pkg.Package{
+		Name:     strings.Join([]string{d.GroupID, d.ArtifactID}, ":"),
```
kzantow, line 94: `fmt.Sprintf("%s:%s", d.GroupID, d.ArtifactID)` might be more consistent with the rest of the codebase?

grype/matcher/java/matcher.go
```diff
@@ -9,7 +16,104 @@ import (
...
+		return nil, fmt.Errorf("status %s from %s", resp.Status, req.URL.String())
+	}
+
+	var res mavenAPIResponse
+	if err = json.NewDecoder(resp.Body).Decode(&res); err != nil {
+		return nil, fmt.Errorf("json decode error: %w", err)
+	}
+
+	if len(res.Response.Docs) == 0 {
+		return nil, fmt.Errorf("digest %s: %w", sha1, errors.New("no artifact found"))
```
kzantow, line 82: Is this case an error? Will this cause something to fail downstream or just exclude additional stuff from this lookup?

grype/matcher/java/matcher.go
```diff
@@ -9,7 +16,104 @@ import (
...
+	}
+
+	// artifacts might have the same SHA-1 digests.
```
kzantow, line 85: Right, this is a good point; would it make sense to somehow check if this matches the original package name in some way?

grype/matcher/java/matcher.go
```diff
@@ -21,5 +125,44 @@ func (m *Matcher) Type() match.MatcherType {
 }
 
 func (m *Matcher) Match(store vulnerability.Provider, d *distro.Distro, p pkg.Package) ([]match.Match, error) {
-	return search.ByCriteria(store, d, p, m.Type(), search.CommonCriteria...)
+	matches := make([]match.Match, 0)
```
kzantow, line 128: `var matches []match.Match`?

grype/lib.go
```diff
@@ -29,11 +29,13 @@ func FindVulnerabilities(provider vulnerability.Provider, userImageStr string, s
...
-	return FindVulnerabilitiesForPackage(provider, context.Distro, packages...), context, packages, nil
+	matchers := matcher.NewDefaultMatchers(matcher.Config{})
+
+	return FindVulnerabilitiesForPackage(provider, context.Distro, packages, matchers), context, packages, nil
 }
 
-func FindVulnerabilitiesForPackage(provider vulnerability.Provider, d *linux.Release, packages ...pkg.Package) match.Matches {
-	return matcher.FindMatches(provider, d, packages...)
+func FindVulnerabilitiesForPackage(provider vulnerability.Provider, d *linux.Release, packages []pkg.Package, matchers []matcher.Matcher) match.Matches {
+	return matcher.FindMatches(provider, d, packages, matchers)
```
kzantow, line 38: This seems odd to me: a `matcher` and `matchers`

grype/lib.go
```diff
@@ -29,11 +29,13 @@ func FindVulnerabilities(provider vulnerability.Provider, userImageStr string, s
 		return match.Matches{}, pkg.Context{}, nil, err
 	}
 
-	return FindVulnerabilitiesForPackage(provider, context.Distro, packages...), context, packages, nil
+	matchers := matcher.NewDefaultMatchers(matcher.Config{})
```
kzantow, line 32: Maybe this `FindVulnerabilities` should be moved to the tests somehow? It looks like it's only used in tests and if anyone was using grype as a lib, it might be confusing why they can't configure java

grype/matcher/java/matcher.go
```diff
@@ -21,5 +125,44 @@ func (m *Matcher) Type() match.MatcherType {
...
+	if m.SearchMavenUpstream {
+		upstreamMatches, err := m.matchUpstreamMavenPackages(store, p)
+		if err != nil {
+			return nil, fmt.Errorf("failed to match upstream maven data")
```
kzantow, line 132: Following up on an earlier comment, it looks like a single SHA that isn't found will skip what could be valid matches without the maven data since it isn't executing the stuff below?

grype/matcher/matchers.go
```diff
@@ -24,50 +24,30 @@ import (
 	syftPkg "github.com/anchore/syft/syft/pkg"
 )
 
-var controllerInstance controller
```
kzantow, line 27: 🎉 MEGA-👍 for getting rid of global state and the init() function

grype/matcher/matchers.go
```diff
@@ -81,9 +61,26 @@ func (c *controller) trackMatcher() (*progress.Manual, *progress.Manual) {
 	return &packagesProcessed, &vulnerabilitiesDiscovered
 }
 
-func (c *controller) findMatches(provider vulnerability.Provider, release *linux.Release, packages ...pkg.Package) match.Matches {
+func newMatcherIndex(matchers []Matcher) map[syftPkg.Type][]Matcher {
+	matcherIndex := make(map[syftPkg.Type][]Matcher)
+	for _, m := range matchers {
+		for _, t := range m.PackageTypes() {
+			if _, ok := matcherIndex[t]; ok {
```
kzantow, line 68: should this be `!ok`, it seems like it's going to make a new matcher array only if it finds one? maybe just remove this altogether, i think the append will do the right thing and you'll get `nil` if it's not found instead of a panic or something

grype/matcher/matchers.go
```diff
@@ -81,9 +61,26 @@ func (c *controller) trackMatcher() (*progress.Manual, *progress.Manual) {
 	return &packagesProcessed, &vulnerabilitiesDiscovered
 }
 
-func (c *controller) findMatches(provider vulnerability.Provider, release *linux.Release, packages ...pkg.Package) match.Matches {
+func newMatcherIndex(matchers []Matcher) map[syftPkg.Type][]Matcher {
```
kzantow, line 64: I wonder if it would be better to do something like:

```
type Matchers map[syftPkg.Type][]Matcher
```
and then instead of having this function, just return the `Matchers` type like `func NewDefaultMatchers(mc Config) Matchers {` above?

kzantow (commented):
Only other thing I'll note at the moment is that we'll probably want to update the README with this one. Removing global state is MEGA-👍, though!

grype/lib.go
spiffcs: I don't want to use this pull request to change the current behavior of the tests. I can follow up on this comment in a second PR that refactors the tests, but I want to err on the side of less changes for now so that this change is isolated by itself without touching too many things.

grype/matcher/java/matcher.go
```diff
@@ -9,7 +16,104 @@ import (
 	syftPkg "github.com/anchore/syft/syft/pkg"
 )
 
+const (
+	sha1Query = `1:"%s"`
```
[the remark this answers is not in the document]
spiffcs: I think I'm going to keep it here for now as a const to highlight how we're doing the maven queries. 

There might be a future where we include other query paths to lookup a package.  Keeping this easy to discover helps future readers find where the initial query is being executed.

Also giving it a title as `sha1Query` makes it easier to come back and see what's happening. Since this is happening in a very specific place this string is I'd rather it be self-documented than a magic inline query string.

grype/matcher/java/matcher.go
spiffcs: I think it's correct to return the error here, but I do need to update the caller so that it can continue on error if no artifact was found

grype/matcher/matchers.go
spiffcs: This is taken directly from the old controller code:
https://github.com/anchore/grype/blob/e77a6c8d63fe020d8f82c334947e2592a0567905/grype/matcher/controller.go#L57-L69

I'll double-check it to see if if should refactor it

grype/pkg/java_metadata.go
```diff
@@ -1,8 +1,14 @@
...
-	PomGroupID    string `json:"pomGroupID"`
-	ManifestName  string `json:"manifestName"`
+	VirtualPath    string   `json:"virtualPath"`
+	PomArtifactID  string   `json:"pomArtifactID"`
+	PomGroupID     string   `json:"pomGroupID"`
+	ManifestName   string   `json:"manifestName"`
+	ArchiveDigests []Digest `json:"archiveDigests"`
+}
+
+type Digest struct {
```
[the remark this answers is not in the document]
spiffcs: There is already an exported `syft.JavaMetadata` but `grype` also makes its own distinction there. Should we update both? I was just trying to follow along with what we already have here.

internal/config/datasources.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/spf13/viper"
+
+	"github.com/anchore/grype/grype/matcher/java"
+)
+
+const (
+	defaultMavenBaseURL = "https://search.maven.org/solrsearch/select"
+)
+
+type externalSources struct {
```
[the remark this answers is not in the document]
spiffcs: "It seems if there are additional external sources we add, having them all mixed together here might not be ideal."

I think that's when we do the refactor. I think since this is the first PR with anything akin to external sources it's ok to have it under one file and then reorganize it as we learn more about the shape that this feature will take. I don't want to organize it a certain way and make choices for future authors that could presume a more advanced config design. 

"I think having external sources be an all-or-nothing thing (e.g. external-sources.enable"
This was specifically asked for to have an on off master switch.

If that's enabled THEN we check the individual configs.

internal/config/datasources.go
```diff
@@ -0,0 +1,39 @@
...
+	Enable bool  `yaml:"enable" json:"enable" mapstructure:"enable"`
+	Maven  maven `yaml:"maven" json:"maven" mapsructure:"maven"`
+}
+
+type maven struct {
+	SearchMavenUpstream bool   `yaml:"search-maven-upstream" json:"search_maven_upstream" mapstructure:"search-maven-upstream"`
```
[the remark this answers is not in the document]
spiffcs: Definitely, an autocompletion setting is off here. I'll update it again so it's consistent with the rest of the codebase.

grype/lib.go
spiffcs: FindMatches uses a `provider`, `distro`, and `matchers` to return a list of matches from `packages`

It reads ok to me but I'm open to suggestions on the refactor.

grype/matcher/matchers.go
spiffcs: From a library standpoint, I think it's good to have the list returned. We've chosen to use this list to build the `syftPkg.Type` map for our own lookup purposes, but we should not exclude users from being able to also run `NewDefaultMatchers()` and use them in different ways
```
for _, m := range matchers {
  m.SomeFunctionality
}
```

spiffcs: Also, I don't know how I feel creating an exported custom type that is a coupling of the two tools together in this case.

grype/pkg/java_metadata.go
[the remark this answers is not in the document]
spiffcs: @wagoodman @luhring @jonasagx thoughts here?

internal/config/datasources.go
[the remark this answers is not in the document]
spiffcs: @wagoodman @luhring @jonasagx thoughts here?

grype/matcher/java/matcher.go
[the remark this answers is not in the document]
spiffcs: @wagoodman @luhring @jonasagx thoughts here?

grype/lib.go
spiffcs: I tried moving this to the tests and the more I move this around, the more I think it's for a later PR where we make our intentions clear about grype as a library. I'd rather keep things closer to main, for this change and am grateful we still have a function available that allows our tests to validate grype behavior outside of this change with no matchers configured.

grype/matcher/matchers.go
spiffcs: @wagoodman @luhring @jonasagx thoughts here?

grype/matcher/java/matcher.go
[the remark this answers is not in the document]
jonasagx: `sha1Query` will always be called as `fmt.Sprintf(sha1Query, sha1)`, maybe make that a function wrapping the query construction

grype/pkg/java_metadata.go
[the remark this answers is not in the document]
kzantow: I think it's fine to duplicate this as you say many of the structures between syft and grype are more-or-less duplicated such that grype has its own data model. Only other thought is it's in the `grype/pkg` as `Digest` but in `java_metadata.go`, maybe move to its own `digest.go` or something?

internal/config/datasources.go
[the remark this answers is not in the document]
kzantow: > This was specifically asked for to have an on off master switch.
👍 

Agree solving for all the cases is unnecessary until there are more than one.

grype/matcher/matchers.go
kzantow: Agree fine as is and it does become a little odd to export something based on the syft data model, given the current way the grype data model works.

grype/lib.go
kzantow: Agree it is somewhat of a separate concern. The reason I brought it up was because this change does impact the meaning of the function to a degree. This was definitely not a blocker.

grype/pkg/java_metadata.go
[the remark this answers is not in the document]
wagoodman: no blocking comment here: I could go either way here, I have a slight pref to using the syft definition here, but admit that my preference is unmotivated

internal/config/datasources.go
```diff
@@ -0,0 +1,39 @@
...
+	BaseURL             string `yaml:"base-url" json:"base-url" mapstructure:"base-url"`
+}
+
+func (cfg externalSources) loadDefaultValues(v *viper.Viper) {
+	v.SetDefault("external-sources.enable", false)
+	v.SetDefault("external-sources.maven.search-maven-upstream", false)
```
wagoodman, line 25: nit: since we have the master switch (which is great) we could have the maven search option default to on

internal/config/datasources.go
wagoodman, line 19: nit: the full config path stutters: `external-sources.maven.search-maven-upstream`
maybe rename the field to `search-upstream`? Or more descriptive to the what or how of the operation, such as `search-by-sha1` or `find-artifact-group-id` (not saying these suggestions are necessarily better, but wanted to give an idea)

internal/config/datasources.go
kzantow: I strongly +1 this

internal/config/datasources.go
spiffcs: Yea nice suggestion. I'll update the config here.

spiffcs:
TODO before merge and thanks for the 🟢 @wagoodman and @kzantow :
- config update to not stutter
- default updates
- test behavior correction

kzantow:
@spiffcs I'd like to see two more things:
* make sure the appropriate environment variables work
* make sure the README is updated

spiffcs:
Still debugging these env variables

PR #721, opened 2022-04-12
Update json field values to be similar across codebase

spiffcs:
Some JSON fields in the output would have `snakecase` as their default tag. This leads to output errors in `jq` and other parsing tools like the one shown below.

Screen Shot 2022-04-11 at 6 34 23 PM

This PR attempts to move those fields in line with the `camelCase` values displayed across fields that can be parsed by automatic tooling.

### DIFF
--- a/grype/match/ignore.go
+++ b/grype/match/ignore.go
@@ -18,7 +18,7 @@ type IgnoredMatch struct {
 // rule to apply.
 type IgnoreRule struct {
 	Vulnerability string            `yaml:"vulnerability" json:"vulnerability" mapstructure:"vulnerability"`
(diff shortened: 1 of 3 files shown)


spiffcs:
Added platform team to make sure this change doesn't cause undue scheme changes that might affect upstream consumers.

kzantow (commented):
This looks like a great change 👍 But also possibly breaking, is that a concern? Is there a schema version to bump? And would it be good to figure out some CI check so that all json names are camel case / or at least not snake case?

spiffcs:
Looks like we already have some CLI checks. I'll go back and update those to pass and see if anything needs to be added. Also, I think it's best to merge this only after a sign-off from @nightfurys or @Vijay-P just to be totally sure any consumers of this document are not getting a breaking change out of the blue.

PR #722, opened 2022-04-13
Add workflow for automatic PR for new Syft releases

kzantow:
This PR adds a workflow that will automatically create a PR to update Syft if the released version of Syft is higher than what is specified in Grype. This will allow us to:
* More easily bump the Syft version if that's the only thing needed for a Grype release
* More quickly identify things that need to be fixed if upgrading the Syft release causes build or tests to fail

NOTE: this may have some added noise in the case we commit `syft@some-hash` instead of a release tag.

Unfortunately, we can't even test this until it gets merged into `main`...

### DIFF
--- a/.github/workflows/update-syft-release.yml
+++ b/.github/workflows/update-syft-release.yml
@@ -23,7 +23,6 @@ jobs:
 
       - run: |
           LATEST_VERSION=$(curl "https://api.github.com/repos/anchore/syft/releases/latest" 2>/dev/null | jq -r '.tag_name')
-          echo "export const VERSION = \"$LATEST_VERSION\";" > src/SyftVersion.ts
 
           # update go.mod
           go get github.com/anchore/syft@$LATEST_VERSION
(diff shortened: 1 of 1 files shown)


.github/workflows/update-syft-release.yml
```diff
@@ -0,0 +1,50 @@
...
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-go@v2
+        with:
+          go-version: ${{ env.GO_VERSION }}
+          stable: ${{ env.GO_STABLE_VERSION }}
+
+      - run: |
+          LATEST_VERSION=$(curl "https://api.github.com/repos/anchore/syft/releases/latest" 2>/dev/null | jq -r '.tag_name')
```
spiffcs, line 25: Same comment added on the syft PR.

This is a really good change. Are there ways where we can make sure `LATEST_VERSION` is different from the current version and then act rather than always act and create a PR?

kzantow: This is sorta already handled, I think. Basically when you run `go get x`, if it's the same version that exists, there won't be any changes in the go.mod/sum. If there aren't any changes in the repo, the `peter-evans/create-pull-request` action won't create a PR.

kzantow:
I think it's worth giving this a try, if it becomes annoying or isn't helpful we can just delete it!

PR #725, opened 2022-04-13
reduce log level for warning so not in default output for upstream matcher

spiffcs:
The warning was too noisy for the default output. Reduce to `Debug` so users can view via `-vv`


kzantow (approved):
🎉

Issue #743, opened 2022-05-04
Support comparisons between pseudo version numbers for Go modules


westonsteimel:
**What would you like to be added**:
 
Grype should support comparisons on [Go module pseudo version numbers](https://go.dev/doc/modules/version-numbers#pseudo-version-number).

**Why is this needed**:

Grype currently misses potential vulnerability matches when pseudo version numbers are used.  An example is https://github.com/anchore/syft/issues/972 where Grype is unable to detect https://github.com/advisories/GHSA-8c26-wmh5-6g9v because the `affected` and `fix` versions both use pseudo version numbers.

**Additional context**:

Useful info provided by @wagoodman:

Current version comparison logic lives at https://github.com/anchore/grype/tree/main/grype/version

I think in this particular case we need to write a go version comparison object to be able to look at the "other" fields in a go version (the date and maybe the git commit) and be able to answer if they occur "before" or "after" another version

It seems possible (though unproven) that semver and the date may be "good enough" to consider for our comparisons

willmurphyscode:
Trying to generate a list of potential GHSAs that we might be wrong about, for better comparison:

```
❯ sqlite3 --column --header ~/Library/Caches/grype/db/5/vulnerability.db 'select id, package_name, version_constraint, namespace from vulnerability where namespace like "%go%" order by RANDOM() limit 2000;' | rg '0.0.0-\d{14}-[a-f0-9]{12}'
GHSA-ffhg-7mh4-33c4  golang.org/x/crypto                                                             <0.0.0-20200220183623-bac4c82f6975                                        github:language:go
GHSA-4r8x-2p26-976p  github.com/elazarl/goproxy                                                      <0.0.0-20230731152917-f99041a5c027                                        github:language:go
GHSA-jpf8-h7h7-3ppm  github.com/whyrusleeping/tar-utils                                              <0.0.0-20201201191210-20a61371de5b                                        github:language:go
GHSA-r5c5-pr8j-pfp7  golang.org/x/crypto                                                             <0.0.0-20190320223903-b7391e95e576                                        github:language:go
GHSA-fcf9-6fv2-fc5v  golang.org/x/net                                                                <0.0.0-20180921000356-2f5d2388922f                                        github:language:go
GHSA-5796-p3m6-9qj4  github.com/cloudfoundry/gorouter                                                <0.0.0-20191101214924-b1b5c44e050f                                        github:language:go
[...]

PR #745, opened 2022-05-12
add case for pseudo versions

spiffcs:
### Draft
Adding cases to try and solve why grype did not detect https://github.com/advisories/GHSA-8c26-wmh5-6g9v.

See: #743 for more details.


spiffcs:
@westonsteimel I'm digging into this issue now to see if I can find why this was a miss earlier in the week

spiffcs:
Closing this until we see it again in the wild as adding the tests showed that Grype is detecting these correctly.

westonsteimel:
@spiffcs, so this may be because the constraint type in grype db isn't `semantic`, it's `unknown` so it'll end up using the fuzzy logic

westonsteimel:
Looks like everything from the GitHub feed except for python end up with `unknown` as the `version_format`.  Python ends up with `python` as the format

westonsteimel:
Hmm, yeah so the fuzzy matcher logic appears to work as well when I try some testcases, but grype still doesn't find matches

PR #759, opened 2022-05-25
README: add MacPorts install info

herbygillot:
https://ports.macports.org/port/grype/


jonasagx (approved):
Thank you for doing this 🎉

Issue #762, opened 2022-05-26
grype fixed version cyclonedxjson


cpendery:
**What would you like to be added**:

Add a fixed version for grype. It makes sense to me to add in under the properties of a vulnerability. 

```json
{
  "name": "grype:fixed_versions",
  "value": "v0.1.2,v01.3"
}
```

**Why is this needed**:

It is given in the `-o json` output and not in the cyclonedx json output

**Additional context**:
I have an issue here https://github.com/CycloneDX/cyclonedx-property-taxonomy/issues/12 to add a grype namespace. Please feel free to comment on it for context or if you want to reuse the syft namespace

nicoskiy:
^ I do not see this enhancement in v0.104.1 version of grype on the --output cyclonedx-json format. 
I do not see the properties field at all in the vulnerabilities section of the cyclonedx json output

nicoskiy:
Example command
`grype --output cyclonedx-json python@sha256:01f36278bcca9fb4a5bdda7edc85be44218190a9371f642d8f852f36ad9fa23d`

Example output in vulnerabilities section:

```
{
      "bom-ref": "urn:uuid:37ab927c-c344-434b-98d8-870ca1b1c763",
      "id": "CVE-2025-64720",
      "source": {
        "name": "debian-distro-debian-13",
        "url": "https://security-tracker.debian.org/tracker/CVE-2025-64720"
      },
      "references": [
        {
          "id": "CVE-2025-64720",
          "source": {
            "name": "debian-distro-debian-13",
            "url": "https://security-tracker.debian.org/tracker/CVE-2025-64720"
          }
        }
      ],
      "ratings": [
        {
          "score": 7.1,
          "severity": "high",
          "method": "CVSSv31",
          "vector": "CVSS:3.1/AV:N/AC:L/PR:N/UI:R/S:U/C:L/I:N/A:H"
        },
        {
          "source": {
            "name": "FIRST",
            "url": "https://www.first.org/epss/"
          },
          "score": 0.00033,
          "method": "other"
        }
      ],
[...]

PR #763, opened 2022-05-26
fix: add fixed versions to cyclonedxjson output

cpendery:
## :memo: Description
Add fixed version to cyclonedxjson output

closes https://github.com/anchore/grype/issues/762


grype/presenter/cyclonedxvex/vulnerability_test.go
```diff
@@ -144,3 +145,59 @@ func TestNewVulnerability_AlwaysIncludesSeverity(t *testing.T) {
...
+					},
+				},
+				Package: pkg.Package{},
+				Details: nil,
+			},
+			metadataProvider: &metadataProvider{},
+			expected:         &[]cyclonedx.Property{},
+		},
+		{
+			name: "No known fixed version",
```
spiffcs, line 172: The test name has been repeated here - Did you mean for this to be the case where a fixed version is shown?

grype/presenter/cyclonedxvex/vulnerability.go
```diff
@@ -92,11 +93,23 @@ func NewVulnerability(m match.Match, p vulnerability.MetadataProvider) (cycloned
...
 	}
 
 	return v, nil
 }
 
+func makeProperties(fix vulnerability.Fix) *[]cyclonedx.Property {
+	properties := []cyclonedx.Property{}
+	if fix.State == v3.FixedState {
+		properties = append(properties, cyclonedx.Property{
+			Name:  "grype:fixed_versions",
```
spiffcs, line 106: Is this the name of the property? cc @wagoodman

spiffcs (commented):
Small test edit and cc @wagoodman for 👀 if he's good with the property name.

LGTM - besides the test name!

spiffcs:
This LGTM - I do want to build on this work since we had this filed today:
https://github.com/anchore/grype/issues/796

The next step here is getting grype updated to output 1.4 and adding some validation around that. I think we're outputting invalid 1.2 at this point so an overhaul before the next release is needed to bring us up to speed on the cyclonedx side for grype.

spiffcs (approved):

Issue #767, opened 2022-05-27
"Matcher failed to parse version" when scanning a Ruby project using bundler 2.2.0 or newer


nicbet:
**What happened**:
grype does not recognize Ruby gems with an architecture suffix specified in `Gemfile.lock`.

```
[0008]  WARN could not match by package language (package=Pkg(type=gem, name=nokogiri, version=1.13.1-x86_64-darwin, upstreams=0)): matcher failed to parse version pkg="nokogiri" ver="1.13.1-x86_64-darwin": unable to crate semver obj: Malformed version: 1.13.1-x86_64-darwin
[0008]  WARN could not match by package CPE (package=Pkg(type=gem, name=nokogiri, version=1.13.1-x86_64-darwin, upstreams=0)): matcher failed to parse version pkg="nokogiri" ver="1.13.1-x86_64-darwin": unable to crate semver obj: Malformed version: 1.13.1-x86_64-darwin
[0008]  WARN could not match by package language (package=Pkg(type=gem, name=nokogiri, version=1.13.1-x86_64-linux, upstreams=0)): matcher failed to parse version pkg="nokogiri" ver="1.13.1-x86_64-linux": unable to crate semver obj: Malformed version: 1.13.1-x86_64-linux
[0008]  WARN could not match by package CPE (package=Pkg(type=gem, name=nokogiri, version=1.13.1-x86_64-linux, upstreams=0)): matcher failed to parse version pkg="nokogiri" ver="1.13.1-x86_64-linux": unable to crate semver obj: Malformed version: 1.13.1-x86_64-linux
```

as a result it reports a fixed gem as vulnerable:

```
nokogiri              1.13.6-arm64-darwin  1.13.6    gem   GHSA-xh29-r2w5-wx8m  High
```

**What you expected to happen**:

Grype can successfully parse `Gemfile.lock` packages that contain the architecture suffixes introduced with Bundler 2.2.0 in December 2020.

Example `Gemfile.lock`:

```
GEM
  remote: https://rubygems.org/
  specs:
    nokogiri (1.13.1)
      mini_portile2 (~> 2.7.0)
      racc (~> 1.4)
    nokogiri (1.13.1-x86_64-darwin)
      racc (~> 1.4)
    nokogiri (1.13.1-x86_64-linux)
      racc (~> 1.4)

PLATFORMS
  ruby
  x86_64-darwin-19
  x86_64-linux-musl

RUBY VERSION
   ruby 2.7.5p203

BUNDLED WITH
[...]

PR #776, opened 2022-06-06
ignore gemfile rich version for semVer comparison

jonasagx:
This PR extracts semVer from `Gemfile.lock`'s new format, that may include arch and distro, so ruby package versions can be compared semantically. The raw version is preserved.

Fix: #767

Signed-off-by: Jonas Xavier <jonasx@anchore.com>


grype/version/gemfile_constraint.go
```diff
@@ -0,0 +1,69 @@
...
+	"fmt"
+	"strings"
+
+	hashiVer "github.com/anchore/go-version"
+)
+
+// bundler >= 2.2.0 adds metadata suffixes to each package version
+// these values are removed to create clean/normalized semver value,
+// while the orignal value is preserved in raw.
+var gemfileNormalizer = strings.NewReplacer("-x86_64", "", "-darwin", "", "-linux", "", "-x86", "", ".alpha", "-alpha", ".beta", "-beta", ".rc", "-rc")
```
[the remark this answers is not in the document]
jonasagx: Yes, I was looking for that. The bundler slack folks channel told me they don't have specs, but pointed me on a good direction: `gem help platform`.

jonasagx:
> A higher-level question: is it valid in the gem ecosystem to ignore OS and architecture for comparison? or should they never be equal?

I believe yes to the first question, because the comparable semantic version represents the code state, which we can use for CPEs and pURLs. Any extra data appended is about the platform it is meant to run on[1]. Bundler cares about installing dependencies and documenting that in `Gemfile.lock`[2].

[1] -- https://bundler.io/v2.3/man/gemfile.5.html#PLATFORMS 
[2] -- https://stackoverflow.com/a/71190958

grype/version/gemfile_constraint_test.go
```diff
@@ -0,0 +1,40 @@
+package version
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGemfileVersionSemantic(t *testing.T) {
+	tests := []testCase{
```
jonasagx, line 10: Hey @wagoodman what do you think of these test cases?

grype/version/semantic_constraint.go
```diff
@@ -14,6 +14,7 @@ var normalizer = strings.NewReplacer(".alpha", "-alpha", ".beta", "-beta", ".rc"
 type semanticConstraint struct {
 	raw        string
 	constraint hashiVer.Constraints
+	checker    func(hashiVer.Constraints, *Version) (bool, error)
```
wagoodman, line 17: this allows for others to monkey patch behavior onto the object. While this works I think there are better ways of achieving this (I think struct embedding with method overriding).

grype/version/semantic_constraint.go
```diff
@@ -50,14 +59,11 @@ func (c semanticConstraint) Satisfied(version *Version) (bool, error) {
 		return true, nil
 	}
 
-	if !c.supported(version.Format) {
-		return false, fmt.Errorf("(semantic) unsupported format: %s", version.Format)
+	if c.checker == nil {
+		return false, fmt.Errorf("check function is undefined")
```
wagoodman, line 63: is an undefined check function an error? also this error isn't very specific about what is wrong (from a user perspective).

grype/version/gemfile_version_test.go
```diff
@@ -0,0 +1,93 @@
+package version
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_newGemfileVersion(t *testing.T) {
```
wagoodman, line 10: this should additionally cover cases with valid prereleases (e.g. `-alpha` or `-beta-2`) and metadata (e.g. `+build.12`) as well as permutations of these.

grype/version/gemfile_version.go
```diff
@@ -0,0 +1,59 @@
...
+// the underscore. Also, we can't sort based on arch and OS in a way that make sense
+// for versions. SemVer is a characteristic of the code, not which arch OS it runs on.
+//
+// CPU is the most structured value present in gemfile.lock versions, we use it
+// to split the version info in half, the first half has semVer, and
+// the second half has arch and OS which we ignore.
+func extractSemVer(raw string) string {
+	lower := strings.ToLower(raw)
+
+	cpus := []string{"-x86", "-x86_64", "-universal", "-arm", "-armv5", "-armv6", "-armv7"}
```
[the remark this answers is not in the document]
wagoodman: it would be ideal to also decouple the suffixes so you can specify a more generic list like `-arm\d*` so we don't need to updated the list as new architecture versions are introduced.

grype/version/gemfile_constraint_test.go
```diff
@@ -0,0 +1,49 @@
...
+		{version: "1.2.0-x86_64-linux", constraint: "< 1.2.1", satisfied: true},
+		// https://semver.org/#spec-item-11
+		{version: "1.2.0-alpha-x86-linux", constraint: "<1.2.0", satisfied: true},
+		{version: "1.2.0-alpha-1-x86-linux", constraint: "<1.2.0", satisfied: true},
+		{version: "1.2.0-alpha-1-x86-linux+meta", constraint: "<1.2.0", satisfied: true},
+		{version: "1.2.0-alpha-1-x86-linux+meta", constraint: ">1.1.0", satisfied: true},
+		{version: "1.2.0-alpha-1-arm-linux+meta", constraint: ">1.1.0", satisfied: true},
+		{version: "1.0.0-alpha-a.b-c-somethinglong+build.1-aef.1-its-okay", constraint: "<1.0.0", satisfied: true},
+		{version: "1.2.3----RC-SNAPSHOT.12.9.1--.12+788", constraint: "> 1.0.0", satisfied: true},
+		{version: "1.2.3----RC-SNAPSHOT.12.9.1--.12+788-armv7-darwin", constraint: "< 1.2.3", satisfied: true},
```
wagoodman, line 37: does ruby provide this os/arch info in the prerelease section of the semver? or sometimes the metadata section (like what's in this example)?

grype/version/gemfile_version.go
```diff
@@ -0,0 +1,59 @@
...
+// after the semvVer, for example, 12.2.1-alpha-x86_64-darwin-8, where `2.2.1-alpha`
+// is a valid and comparable semVer, and `x86_64-darwin-8` is not a semVer due to
+// the underscore. Also, we can't sort based on arch and OS in a way that make sense
+// for versions. SemVer is a characteristic of the code, not which arch OS it runs on.
+//
+// CPU is the most structured value present in gemfile.lock versions, we use it
+// to split the version info in half, the first half has semVer, and
+// the second half has arch and OS which we ignore.
+func extractSemVer(raw string) string {
+	lower := strings.ToLower(raw)
```
wagoodman, line 25: do we need to convert the entire version to lowercase? will this make a difference in comparison?

grype/version/gemfile_version.go
```diff
@@ -0,0 +1,59 @@
...
+		return nil, fmt.Errorf("unable to crate gemfile version obj: %w", err)
+	}
+	return &gemfileVersion{
+		raw:    raw,
+		semVer: semVer,
+	}, nil
+}
+
+func (g *gemfileVersion) Compare(other *Version) (int, error) {
+	if other.Format != GemfileFormat {
```
wagoodman, line 51: should we allow comparisons with objects that are semver? this will help make certain that records from grype-db will be comparable even though they are not know to be gem packages.

grype/version/gemfile_constraint_test.go
jonasagx: They don't specify that, I added such cases to make we can handle them, seems worth a comment in the code.

wagoodman (commented):
~should we be making a change in the `anchore/go-version` to relax the version validation regex to allow for `_` to appear in prerelease and metadata sections (even though this is against spec)?~

~I feel that we should consider this, as it would give a better chance of comparing with a semver object instead of the fuzzy comparison object in cases where there isn't enough information.~

update: @jonasagx, we can explore this another time (disregard my comment here / no action needed)

grype/version/gemfile_constraint_test.go
```diff
@@ -0,0 +1,40 @@
...
+		{version: "1.2.0-x86_64-linux", constraint: "= 1.2.0", satisfied: true},
+		{version: "1.2.0-x86_64-linux", constraint: "< 1.2.1", satisfied: true},
+		// https://semver.org/#spec-item-11
+		{version: "1.2.0-alpha-x86-linux", constraint: "<1.2.0", satisfied: true},
+		{version: "1.2.0-alpha-1-x86-linux", constraint: "<1.2.0", satisfied: true},
+		{version: "1.2.0-alpha-1-x86-linux+meta", constraint: "<1.2.0", satisfied: true},
+		{version: "1.2.0-alpha-1-x86-linux+meta", constraint: ">1.1.0", satisfied: true},
+		{version: "1.2.0-alpha-1-arm-linux+meta", constraint: ">1.1.0", satisfied: true},
+		{version: "1.0.0-alpha-a.b-c-somethinglong+build.1-aef.1-its-okay", constraint: "<1.0.0", satisfied: true},
+		{version: "1.2.3----RC-SNAPSHOT.12.9.1--.12+788", constraint: "> 1.0.0", satisfied: true},
```
wagoodman, line 36: this technically isn't a clause-11 case

grype/version/gemfile_constraint_test.go
wagoodman: is there code in the rubygem repos to help pin this down?

grype/version/gemfile_version.go
jonasagx: Yes, added support for that.

grype/version/gemfile_constraint_test.go
jonasagx: Fixed order on tests.

grype/version/semantic_constraint.go
jonasagx: In that case it would, because the zero value for a function var is nil, it would panic. I moved away from that with struct embedding and method overriding.

grype/version/gemfile_version.go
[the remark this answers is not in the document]
jonasagx: I agree, by keeping just "arm" as a split token we can cover that case, and it is simpler than a regex.

grype/version/gemfile_version.go
jonasagx: The arch case is consistent, so no need to lower it. The case would make a different version object cause and comparison. I removed the lowercase step.

jonasagx: In constraint comparison that wouldn't make difference.

grype/version/semantic_constraint.go
wagoodman: > In that case it would, because the zero value for a function var is nil, it would panic.

correct, and though I'm not endorsing this path at all, I was asking is returning an error the right thing to do or should we patch a more sane default when this is found:
```golang
if c.checker == nil {
		c.checker = func(...){}
}
```
preventing the panic and not stopping the constraint checks (unless we really do need this check).

grype/version/gemfile_constraint_test.go
wagoodman: this seems to hint that they ignore the existence of metadata, always adding it to the end of the version https://github.com/rubygems/rubygems/blob/32af6bdbef756487b91ec4668f5b8dd2fd25114d/bundler/spec/support/artifice/compact_index.rb#L12-L16 , however, I haven't checked the path that `version` takes (is it guarenteed to never have metadata?)

grype/version/gemfile_version.go
```diff
@@ -0,0 +1,66 @@
...
+// Bunlder's code: https://github.com/rubygems/rubygems/blob/2070231bf0c7c4654bbc2e4c08882bf414840360/bundler/spec/install/gemfile/platform_spec.rb offers more info on possible architecture values, for example `mswin32` may appead without arch.
+//
+// Spec for pre-release info: https://github.com/rubygems/rubygems/blob/2070231bf0c7c4654bbc2e4c08882bf414840360/bundler/spec/install/gemfile/path_spec.rb#L186
+//
+// CPU/arch is the most structured value present in gemfile.lock versions, we use it
+// to split the version info in half, the first half has semVer, and
+// the second half has arch and OS which we ignore.
+// When there is no arch we split the version string with: {java, delvik, mswin32}
+func extractSemVer(raw string) string {
+	platforms := []string{"x86", "x86_64", "universal", "arm", "java", "dalvik", "x64", "powerpc", "sparc", "mswin32"}
```
wagoodman, line 30: nit: don't need `x86_64` if you already have `x86`

however, should `mswin32` be relaxed to `mswin`?

grype/version/gemfile_version.go
```diff
@@ -0,0 +1,66 @@
...
+		vals := strings.SplitN(raw, dash+p, 2)
+		if len(vals) == 2 {
+			return vals[0]
+		}
+	}
+
+	return raw
+}
+
+func newGemfileVersion(raw string) (*gemfileVersion, error) {
```
wagoodman, line 42: could we simplify this further by returning `*semanticVersion` (which is cleaned) thus allowing us to drop the need for a `gemfileVersion` and `gemfileConstraint` entirely? (that would eliminate a lot of questions about how this gemfile version objects should interact with semver objects)

grype/search/language_test.go
```diff
@@ -36,7 +36,7 @@ func (pr *mockLanguageProvider) stub() {
 				Namespace:  "github:ruby",
```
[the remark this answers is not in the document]
jonasagx: On this PR page you might have to expand the code up to see it.
```go
{
	Constraint: version.MustGetConstraint("< 3.7.6", version.SemanticFormat),
	ID:         "CVE-2017-fake-1",
	Namespace:  "github:ruby",
}
```

grype/version/gemfile_version.go
jonasagx: Yes, done. I checked and gem version objects fit semver and gemfile constraints.

grype/version/format.go
```diff
@@ -26,6 +27,7 @@ var formatStr = []string{
 	"RPM",
 	"Python",
 	"KB",
+	"Gemfile",
```
wagoodman, line 30: nit: I think this would be more accurate as `Gem`

grype/search/language_test.go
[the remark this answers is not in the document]
wagoodman: this is the right spot for adding the test case, and I see that you update the mock data 👍 

the next step is to update the test such that the new mock data can be positively found (and also make certain that it won't match on negative cases). The entry you changed is meant to be the negative case in the `TestFindMatchesByPackageLanguage` test case, you'll need to add another mock entry that will act as the negative case and write another test (or change the current one) such as to match on the entry with the `GemfileFormat`)

grype/version/gemfile_constraint_test.go
wagoodman: I think so 🤔

grype/version/semantic_constraint.go
```diff
@@ -35,7 +35,7 @@ func newSemanticConstraint(constStr string) (semanticConstraint, error) {
 }
 
 func (c semanticConstraint) supported(format Format) bool {
-	return format == SemanticFormat
+	return format == SemanticFormat || format == GemfileFormat
```
wagoodman, line 38: since this feels a bit odd, it would be worth adding a comment here as to why there is an ecosystem exception made a the generic semver method.

grype/version/gemfile_version.go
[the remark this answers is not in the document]
wagoodman: Right, I'm saying that this can be simplified down to
```diff 
-	cpus := []string{"-x86", "-x86_64", "-universal", "-arm", "-armv5", "-armv6", "-armv7"}
+	cpus := []string{"-x86", "-universal", "-arm"}
```

grype/version/gemfile_version.go
wagoodman: this looks a lot simpler ❤️

grype/version/gemfile_version.go
wagoodman: I didn't review the `anchore/go-version` lib to see what the sensitive factors to comparisons are, but I think there is a risk that `1.0.3-alpha` and `1.0.3-ALPHA` will not be considered equal --I think it's good that the `ToLower` has been removed in the current state of the PR for this reason. 

However, if there is a concern of casing, the existing `extractSemVer` could be refactored to split in a case-insensitive way, but always return the substring with no casing changes.

To be clear, I'm not asking for this change (no need to add this to the PR).

wagoodman (approved):
looks great!

Issue #240, opened 2021-02-25
Be clear about version and data staleness 


wagoodman:
This is capturing the output from an internal conversation.

1. Today we do not fail a scan if the DB is stale (hasn't been updated within some "acceptable" time period). It would be ideal to allow for a **configurable** option (that is on by default) which controls the exit code to always be 1 if the result was made with a stale DB.
2. We should have a way to notify users of DB staleness warnings and other useful messages.
3. ~Question: Should we allow for downloading OLD DBs where the schemas are deprecated? Should this show a warning?~ (we haven't implemented a deprecation mechanism yet)
4. Should the application version banner show the existing version compared with the new version available? That may help with illustrating the need to update versions to users.

jonasagx:
I can look into this.

PR #785, opened 2022-06-13
add db staleness check

jonasagx:
* Add staleness check to vuln db. Execution fails if db is 5 or more days old.  Configurable via `grype.yaml` and enabled by default. 
* Warn user that db is stale even when staleness check is disabled.
* Add version banner comparing application version with new version available illustrating  to users how far behind their local grype is.

FIx: #240 

Note on staleness field type: it uses go's time duration notation. Anchore provides daily updates. However users might use grype with a different database and different release schedules. Go's duration notation has flexible granularity to different use cases. It also helps maintaining grype's config file consistent over time.

Signed-off-by: Jonas Xavier <jonasx@anchore.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -401,7 +401,7 @@ You can set the cache directory path using the environment variable `GRYPE_DB_CA
 
 #### Data staleness
 
-Grype needs up-to-date vulnerability information to provide good matches. By default, it will fail execution if the local database was not updated in the last 5 days. The data staleness check is configurable via the field `staleness-threshold`, under `db`. It uses golang's time duration syntax. Set it to 0 to disable staleness check.
+Grype needs up-to-date vulnerability information to provide accurate matches. By default, it will fail execution if the local database was not built in the last 5 days. The data staleness check is configurable via the field `max-allowed-db-age`, under `db`. It uses [golang's time duration syntax](https://pkg.go.dev/time#ParseDuration). Set it to 0 to disable staleness check.
 
 #### Offline and air-gapped environments
 
@@ -588,11 +588,11 @@ db:
   # same as GRYPE_DB_UPDATE_URL env var
   update-url: "https://toolbox-data.anchore.io/grype/databases/listing.json"
 
-  # if this period has passed since the
-  # last db update and `now` then db data is stale.
-  # Default threshold of 120h (or five days)
-  # Set it to 0 (zero) to disable staleness check
-  staleness-threshold: "120h"
+  # Max allowed age for vulnerability database,
+  # age being the time since it was built
+  # Default max age is 120h (or five days)
+  # Set it to 0 (zero) to disable check
+  max-allowed-db-age: "120h"
 
 search:
 
--- a/README.md
+++ b/README.md
@@ -401,7 +401,7 @@ You can set the cache directory path using the environment variable `GRYPE_DB_CA
 
 #### Data staleness
 
-Grype needs up-to-date vulnerability information to provide accurate matches. By default, it will fail execution if the local database was not built in the last 5 days. The data staleness check is configurable via the field `max-allowed-db-age`, under `db`. It uses [golang's time duration syntax](https://pkg.go.dev/time#ParseDuration). Set it to 0 to disable staleness check.
+Grype needs up-to-date vulnerability information to provide accurate matches. By default, it will fail execution if the local database was not built in the last 5 days. The data staleness check is configurable via the field `max-allowed-built-age`, under `db`. It uses [golang's time duration syntax](https://pkg.go.dev/time#ParseDuration). Set it to 0 to disable staleness check.
 
 #### Offline and air-gapped environments
 
@@ -592,7 +592,7 @@ db:
   # age being the time since it was built
   # Default max age is 120h (or five days)
   # Set it to 0 (zero) to disable check
-  max-allowed-db-age: "120h"
+  max-allowed-built-age: "120h"
 
 search:
 
--- a/README.md
+++ b/README.md
@@ -401,7 +401,7 @@ You can set the cache directory path using the environment variable `GRYPE_DB_CA
 
 #### Data staleness
 
-Grype needs up-to-date vulnerability information to provide accurate matches. By default, it will fail execution if the local database was not built in the last 5 days. The data staleness check is configurable via the field `max-allowed-built-age`, under `db`. It uses [golang's time duration syntax](https://pkg.go.dev/time#ParseDuration). Set it to 0 to disable staleness check.
+Grype needs up-to-date vulnerability information to provide accurate matches. By default, it will fail execution if the local database was not built in the last 5 days. The data staleness check is configurable via the field `max-allowed-built-age` and `validate-age`, under `db`. It uses [golang's time duration syntax](https://pkg.go.dev/time#ParseDuration). Set `validate-age` to `false` to disable staleness check.
 
 #### Offline and air-gapped environments
 
@@ -588,10 +588,13 @@ db:
   # same as GRYPE_DB_UPDATE_URL env var
   update-url: "https://toolbox-data.anchore.io/grype/databases/listing.json"
 
+  # it ensures db build is no older than the max-allowed-built-age
+  # set to false to disable check
+  validate-age: true
+
   # Max allowed age for vulnerability database,
   # age being the time since it was built
   # Default max age is 120h (or five days)
-  # Set it to 0 (zero) to disable check
   max-allowed-built-age: "120h"
 
 search:
--- a/README.md
+++ b/README.md
@@ -399,6 +399,10 @@ By default, the database is cached on the local filesystem in the directory `$XD
 
 You can set the cache directory path using the environment variable `GRYPE_DB_CACHE_DIR`.
 
+#### Data staleness
+
+Grype needs up-to-date vulnerability information to provide good matches. By default, it will fail execution if the local database was not updated in the last 5 days. The data staleness check is configurable via the field `staleness-threshold`, under `db`. It uses golang's time duration syntax. Set it to 0 to disable staleness check.
+
 #### Offline and air-gapped environments
 
 By default, Grype checks for a new database on every run, by making a network call over the Internet. You can tell Grype not to perform this check by setting the environment variable `GRYPE_DB_AUTO_UPDATE` to `false`.
@@ -584,6 +588,11 @@ db:
   # same as GRYPE_DB_UPDATE_URL env var
   update-url: "https://toolbox-data.anchore.io/grype/databases/listing.json"
 
+  # if this period has passed since the
+  # last db update and `now` then db data is stale.
+  # Default threshold of 120h (or five days)
+  # Set it to 0 (zero) to disable staleness check
+  staleness-threshold: "120h"
 
 search:
 
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -270,7 +270,7 @@ func startWorker(userInput string, failOnSeverity *vulnerability.Severity) <-cha
 				log.Errorf(err.Error())
 			}
 			if isAvailable {
-				log.Infof("New version of %s is available: %s", internal.ApplicationName, newVersion)
+				log.Infof("You're currently running %s version %s and a new version is available: %s", internal.ApplicationName, version.FromBuild().Version, newVersion)
 
 				bus.Publish(partybus.Event{
 					Type:  event.AppUpdateAvailable,
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -270,7 +270,7 @@ func startWorker(userInput string, failOnSeverity *vulnerability.Severity) <-cha
 				log.Errorf(err.Error())
 			}
 			if isAvailable {
-				log.Infof("new version of %s is available: %s (current version is %s)", internal.ApplicationName, newVersion, version.FromBuild().Version)
+				log.Infof("new version of %s is available: %s (currently running: %s)", internal.ApplicationName, newVersion, version.FromBuild().Version)
 
 				bus.Publish(partybus.Event{
 					Type:  event.AppUpdateAvailable,
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -270,7 +270,7 @@ func startWorker(userInput string, failOnSeverity *vulnerability.Severity) <-cha
 				log.Errorf(err.Error())
 			}
 			if isAvailable {
-				log.Infof("You're currently running %s version %s and a new version is available: %s", internal.ApplicationName, version.FromBuild().Version, newVersion)
+				log.Infof("New version of % available: %s. Currently running: %s", internal.ApplicationName, newVersion, version.FromBuild().Version)
 
 				bus.Publish(partybus.Event{
 					Type:  event.AppUpdateAvailable,
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -270,7 +270,7 @@ func startWorker(userInput string, failOnSeverity *vulnerability.Severity) <-cha
 				log.Errorf(err.Error())
 			}
 			if isAvailable {
-				log.Infof("new version of %s available: %s (current version is %s)", internal.ApplicationName, newVersion, version.FromBuild().Version)
+				log.Infof("new version of %s is available: %s (current version is %s)", internal.ApplicationName, newVersion, version.FromBuild().Version)
 
 				bus.Publish(partybus.Event{
 					Type:  event.AppUpdateAvailable,
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -270,7 +270,7 @@ func startWorker(userInput string, failOnSeverity *vulnerability.Severity) <-cha
 				log.Errorf(err.Error())
 			}
 			if isAvailable {
-				log.Infof("New version of % available: %s. Currently running: %s", internal.ApplicationName, newVersion, version.FromBuild().Version)
+				log.Infof("new version of %s available: %s (current version is %s)", internal.ApplicationName, newVersion, version.FromBuild().Version)
 
 				bus.Publish(partybus.Event{
 					Type:  event.AppUpdateAvailable,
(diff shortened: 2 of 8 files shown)


grype/db/curator.go
```diff
@@ -267,6 +278,24 @@ func (c *Curator) download(listing *ListingEntry, downloadProgress *progress.Man
 	return tempDir, nil
 }
 
+func (c *Curator) validateStaleness(m *Metadata) error {
+	if m == nil {
+		return nil
+	}
+
+	age := time.Since(m.Built).Round(time.Hour)
```
luhring, line 286: What is the motivation for rounding here?

grype/db/curator.go
```diff
@@ -267,6 +278,24 @@ func (c *Curator) download(listing *ListingEntry, downloadProgress *progress.Man
...
+	// dataStalenessLimit == 0 means to not error on staleness
+	if c.dataStalenessLimit > 0 && age > c.dataStalenessLimit {
+		return fmt.Errorf("%w: last updated %s ago (> threshold %s)", ErrDataIsStale, age, c.dataStalenessLimit)
+	}
+
+	if age > DefaultStalenessThreshold {
+		log.Warnf("your vuln db was updated %s ago and it might generate wrong results. To update run: grype db update", age)
```
luhring, line 293: This message could be improved to be more clear and idiomatic to our log messages:

1. The problem isn't when the database was updated, it's when it was built.
2. I'd say "inaccurate" instead of something as simplified as "wrong".
3. We typically don't say "your vuln db". Instead we tend to say "the vulnerability database".

cmd/root.go
```diff
@@ -270,7 +270,7 @@ func startWorker(userInput string, failOnSeverity *vulnerability.Severity) <-cha
 				log.Errorf(err.Error())
 			}
 			if isAvailable {
-				log.Infof("New version of %s is available: %s", internal.ApplicationName, newVersion)
+				log.Infof("You're currently running %s version %s and a new version is available: %s", internal.ApplicationName, version.FromBuild().Version, newVersion)
```
luhring, line 273: (Same comment as the other instance of this message)

README.md
```diff
@@ -399,6 +399,10 @@ By default, the database is cached on the local filesystem in the directory `$XD
 
 You can set the cache directory path using the environment variable `GRYPE_DB_CACHE_DIR`.
 
+#### Data staleness
+
+Grype needs up-to-date vulnerability information to provide good matches. By default, it will fail execution if the local database was not updated in the last 5 days. The data staleness check is configurable via the field `staleness-threshold`, under `db`. It uses golang's time duration syntax. Set it to 0 to disable staleness check.
```
luhring, line 404: I'd say "accurate matches" instead of "good matches".

luhring, line 404: >It uses golang's time duration syntax

It'd be helpful to link to where this is described, in case the user isn't familiar with this syntax.

grype/db/curator.go
```diff
@@ -28,11 +30,18 @@ const (
...
+
+	DefaultStalenessThreshold = time.Hour * 24 * 5
+)
+
 type Config struct {
 	DBRootDir           string
 	ListingURL          string
 	CACert              string
 	ValidateByHashOnGet bool
+	StalenessThreshold  time.Duration
```
luhring, line 44: It's not immediately obvious what "staleness threshold" means. We should be exact here. We want this value to represent the max time allowed since the database was built.

internal/config/database.go
```diff
@@ -11,11 +12,12 @@ import (
...
-	UpdateURL             string `yaml:"update-url" json:"update-url" mapstructure:"update-url"`
-	CACert                string `yaml:"ca-cert" json:"ca-cert" mapstructure:"ca-cert"`
-	AutoUpdate            bool   `yaml:"auto-update" json:"auto-update" mapstructure:"auto-update"`
-	ValidateByHashOnStart bool   `yaml:"validate-by-hash-on-start" json:"validate-by-hash-on-start" mapstructure:"validate-by-hash-on-start"`
+	Dir                   string        `yaml:"cache-dir" json:"cache-dir" mapstructure:"cache-dir"`
+	UpdateURL             string        `yaml:"update-url" json:"update-url" mapstructure:"update-url"`
+	CACert                string        `yaml:"ca-cert" json:"ca-cert" mapstructure:"ca-cert"`
+	AutoUpdate            bool          `yaml:"auto-update" json:"auto-update" mapstructure:"auto-update"`
+	ValidateByHashOnStart bool          `yaml:"validate-by-hash-on-start" json:"validate-by-hash-on-start" mapstructure:"validate-by-hash-on-start"`
+	StalenessThreshold    time.Duration `yaml:"staleness-threshold" json:"staleness-threshold" mapstructure:"staleness-threshold"`
```
luhring, line 20: It's not immediately obvious what "staleness threshold" means. We should be exact here. We want this value to represent the max time allowed since the database was built.

luhring (changes requested):
Thanks for getting this started! I left a few notes on clarity, logic, and wording. Let me know if you have any questions.

grype/db/curator.go
jonasagx: I added a code comment explaining it; 

tl;dr time comparison has fractional seconds of precision which makes the error message noisy. Rounding it to hour provides enough context for action.

grype/db/curator.go
```diff
@@ -62,6 +72,7 @@ func NewCurator(cfg Config) (Curator, error) {
 		dbPath:              path.Join(dbDir, FileName),
 		listingURL:          cfg.ListingURL,
 		validateByHashOnGet: cfg.ValidateByHashOnGet,
+		dataStalenessLimit:  cfg.StalenessThreshold,
```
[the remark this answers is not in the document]
jonasagx: Agreed, I renamed the fields accordingly.

grype/db/curator_test.go
```diff
@@ -301,3 +301,60 @@ func TestCuratorDBPathHasSchemaVersion(t *testing.T) {
 	assert.Equal(t, path.Join(dbRootPath, strconv.Itoa(cur.targetSchema)), cur.dbDir, "unexpected dir")
 	assert.Contains(t, cur.dbPath, path.Join(dbRootPath, strconv.Itoa(cur.targetSchema)), "unexpected path")
 }
+
+func assertAs(expected string) assert.ErrorAssertionFunc {
+	return func(t assert.TestingT, err error, i ...interface{}) bool {
+		return assert.ErrorContains(t, err, expected)
+	}
+}
```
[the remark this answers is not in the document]
jonasagx: That makes sense. After some code and testing refactoring this helper became unnecessary.

grype/db/curator_test.go
```diff
@@ -301,3 +301,60 @@ func TestCuratorDBPathHasSchemaVersion(t *testing.T) {
...
+				staleLimist: 2 * time.Hour,
+				md:          &Metadata{Built: time.Now()},
+			},
+			wantErr: assert.NoError,
+		},
+		{
+			name: "stale-data",
+			fields: fields{
+				staleLimist: time.Hour,
+				md:          &Metadata{Built: time.Now().Add(-4 * time.Hour)},
```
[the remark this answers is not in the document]
jonasagx: In go `time.Sub` returns the duration between two times, while `time.Add` adds a duration to a time. Here `time.Time` is made from a given duration and why `time.Add` is necessary.

internal/ui/etui_event_handlers.go
```diff
@@ -28,7 +29,7 @@ func handleAppUpdateAvailable(_ context.Context, fr *frame.Frame, event partybus
 		return err
 	}
 
-	message := color.Magenta.Sprintf("New version of %s is available: %s", internal.ApplicationName, newVersion)
+	message := color.Magenta.Sprintf("You're currently running %s version %s and a new version is available: %s", internal.ApplicationName, version.FromBuild().Version, newVersion)
```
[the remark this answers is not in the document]
jonasagx: Yes, it reads better with that suggestion, ty.

grype/db/curator.go
```diff
@@ -267,6 +278,24 @@ func (c *Curator) download(listing *ListingEntry, downloadProgress *progress.Man
 	return tempDir, nil
 }
 
+func (c *Curator) validateStaleness(m *Metadata) error {
```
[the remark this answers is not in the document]
jonasagx: Sure, I added code comments to it.

grype/db/curator.go
jonasagx: Thank you for the language clarifications, I applied the suggestions.

cmd/root.go
jonasagx: Yes, it reads better with the above suggestion, ty.

README.md
jonasagx: Agreed, I added the link

internal/config/database.go
jonasagx: I agree, I renamed it to `MaxAllowedDBAge`, which I think is more explicit and direct. Please let me know of any possible redundancy.

grype/db/curator.go
```diff
@@ -267,6 +278,24 @@ func (c *Curator) download(listing *ListingEntry, downloadProgress *progress.Man
...
+	if m == nil {
+		return nil
+	}
+
+	age := time.Since(m.Built).Round(time.Hour)
+	// dataStalenessLimit == 0 means to not error on staleness
+	if c.dataStalenessLimit > 0 && age > c.dataStalenessLimit {
+		return fmt.Errorf("%w: last updated %s ago (> threshold %s)", ErrDataIsStale, age, c.dataStalenessLimit)
```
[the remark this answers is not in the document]
jonasagx: I dropped `ErrDataIsStale` (it was used only here anyways) in favor of creating it locally and added context information: `fmt.Errorf("the vulnerability database was built %s ago (> max allowed %s)", age, c.maxAllowedDBAge)`

README.md
```diff
@@ -584,6 +588,11 @@ db:
   # same as GRYPE_DB_UPDATE_URL env var
   update-url: "https://toolbox-data.anchore.io/grype/databases/listing.json"
 
+  # Max allowed age for vulnerability database,
+  # age being the time since it was built
+  # Default max age is 120h (or five days)
+  # Set it to 0 (zero) to disable check
+  max-allowed-db-age: "120h"
```
wagoodman, line 595: Can we additionally allow for `d` for days for config input? I realize that go `time.ParseDuration` doesn't handle this, but it's a little awkward to configure high values.

I've got similar comments around the output that we are indicating, for example:
```
$ GRYPE_DB_AUTO_UPDATE=false grype alpine:latest
1 error occurred:
	* failed to load vulnerability db: vulnerability database is invalid (run db update to correct): the vulnerability database was built 539h0m0s ago (> max allowed 120h0m0s)
```

Something like `539h0m0s` should get converted to `22 days` (or similar) since this message is intended for the user. Consider using something like https://github.com/dustin/go-humanize#times to help with showing human readable values.

wagoodman, line 595: since this is already under the `DB` section, to prevent a stutter renaming to `max-allowed-age` I think makes sense -- what's your take?

wagoodman:
Are we intending to block the `db import` path? That's the behavior I'm seeing:
```
$ grype db import ./ReallyOldDB.tar.gz
unable to import vulnerability database: the vulnerability database was built 539h0m0s ago (> max allowed 120h0m0s)
```

my opinion is that we should allow the user to import any DB they'd like (since the importing doesn't affect results) but when we run grype to get vulnerability results we should warn in that case (when the data is used).

README.md
jonasagx: There is no general way to parse a duration like `1d` because days don't have consistent length. More on https://github.com/golang/go/issues/11473#issuecomment-116930849

I am in favor of human friendly outputs 👍🏼

jonasagx: I changed it to `max-allowed-built-age`.

README.md
```diff
@@ -584,6 +588,11 @@ db:
   # same as GRYPE_DB_UPDATE_URL env var
   update-url: "https://toolbox-data.anchore.io/grype/databases/listing.json"
 
+  # Max allowed age for vulnerability database,
```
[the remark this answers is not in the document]
jonasagx: I added `validate-age` config field 👍🏼

jonasagx:
> I'm seeing two warnings instead of one when setting a large value and disabling the DB updating
I fixed that.

Other comment:
> my opinion is that we should allow the user to import any DB they'd like (since the importing doesn't affect results) but when we run grype to get vulnerability results we should warn in that case (when the data is used).
I agree with you and changed to only check staleness before a vuln scan.

README.md
jonasagx: I've added https://github.com/hako/durafmt since its implementation doesn't assume local timezone (like https://github.com/dustin/go-humanize#times does). We have to compare duration to UTC; the db built value is in UTC.

README.md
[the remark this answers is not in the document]
luhring: Should that new field get added to this config documentation here?

cmd/root.go
```diff
@@ -270,7 +270,7 @@ func startWorker(userInput string, failOnSeverity *vulnerability.Severity) <-cha
 				log.Errorf(err.Error())
 			}
 			if isAvailable {
-				log.Infof("New version of %s is available: %s", internal.ApplicationName, newVersion)
+				log.Infof("new version of %s is available: %s (current version is %s)", internal.ApplicationName, newVersion, version.FromBuild().Version)
```
luhring, line 273: nit: I think "current version" is slightly ambiguous here. We've used "current version" in conversation before to refer to the _latest_ released version. I'd suggest something like `"<version> currently installed"`

README.md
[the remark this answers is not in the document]
luhring: Also, does this new config value _replace_ the "set to 0 to disable" mechanism?

grype/db/curator.go
```diff
@@ -28,11 +30,18 @@ const (
 	FileName = grypeDB.VulnerabilityStoreFileName
 )
 
+var (
+	ErrDataIsStale = errors.New("data is stale")
+
+	DefaultStalenessThreshold = time.Hour * 24 * 5
```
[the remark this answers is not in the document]
luhring: (@jonasagx looks like this comment might have been missed — this question is about exporting)

internal/ui/etui_event_handlers.go
```diff
@@ -28,7 +29,7 @@ func handleAppUpdateAvailable(_ context.Context, fr *frame.Frame, event partybus
 		return err
 	}
 
-	message := color.Magenta.Sprintf("New version of %s is available: %s", internal.ApplicationName, newVersion)
+	message := color.Magenta.Sprintf("New version of %s is available: %s (current version is %s)", internal.ApplicationName, newVersion, version.FromBuild().Version)
```
luhring, line 32: (see comment in other instance of this message. Also, should we consider factoring out this message into a shared value?)

grype/db/curator.go
```diff
@@ -28,11 +30,18 @@ const (
 	FileName = grypeDB.VulnerabilityStoreFileName
 )
 
+var (
+	ErrDataIsStale = errors.New("data is stale")
```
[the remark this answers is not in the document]
luhring: nit: I'd suggest a slight wording tweak -> `(max allowed age is %s)`

grype/db/curator_test.go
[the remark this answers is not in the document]
luhring: Good call!

README.md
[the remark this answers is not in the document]
wagoodman: it probably should be mutually exclusive, good catch

internal/ui/etui_event_handlers.go
wagoodman: my take, it's better as-is (not being a shared constant); the ETUI and logging are different-enough of concerns to not need a single source of truth for this message (and they have their own rules for how something is displayed). For instances, in logging the sentence should start with a lower case but in the ETUI it should be capitalized.

grype/db/curator.go
```diff
@@ -214,7 +226,12 @@ func (c *Curator) UpdateTo(listing *ListingEntry, downloadProgress, importProgre
 
 // Validate checks the current database to ensure file integrity and if it can be used by this version of the application.
 func (c *Curator) Validate() error {
-	return c.validate(c.dbDir)
+	metadata, err := c.validate(c.dbDir)
+	if err != nil {
+		return err
+	}
+
+	return c.validateStaleness(metadata)
```
wagoodman, line 234: it seems odd that `validateStaleness` is not part of the `validate` call -- if there is a specific reason it should probably be added as a comment here. Also if `validate` isn't covering all validations it should probably be renamed to the scope that it's responsible for, such as `validateIntegrity` or similar.

grype/db/curator.go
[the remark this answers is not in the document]
jonasagx: @luhring `DefaultMaxAllowedBuiltAge` is exported for use in two packages: `grype/db/curator` and `internal/config/db`

grype/db/curator.go
jonasagx: Rounding still happens but through a time formatting library capable of printing longer durations in a more friendly way, like `22 days` instead of `528h`

cmd/root.go
jonasagx: 👍🏼  I changed it to `new version of %s is available: %s (currently running: %s)"` so the two versions are not next to each other helping visually.

grype/db/curator.go
jonasagx: Yes, I renamed it to `validateIntegrity`. That is the relevant check while importing a database for example.

grype/db/curator.go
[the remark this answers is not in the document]
jonasagx: 👍🏼  I changed it to `(max allowed age is %s)`

grype/db/curator.go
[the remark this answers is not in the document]
jonasagx: You're right, it _was_ used by both, but now only referenced in `internal/config`. I moved it there

luhring (approved):
Looks good, thanks @jonasagx!

Issue #757, opened 2022-05-25
Request vulnerability data by a single cpe string


BrianPfitz:
**What would you like to be added**:
Hey guys, first of all thanks for this great tool! I really like working with Syft and Grype.
However, I would find it very practical to be able to query the Grype database with individual cpes (or with a list of several cpes). This is currently not possible, is it? Is there a possible workaround for this?

**Why is this needed**:
After an initial scan, I would also like to check the version of a cpe in which the vulnerability found (according to Grype) was fixed. This would be possible without any problems with the above-mentioned feature. In addition, the extensive Grype database could be used much more extensively.

spiffcs:
Thanks @BrianPfitz for the enhancement suggestion! Just to get clarity here - you want something like `grype xxx <CPE1> <CPE2>` to get full details from the DB about what was found?

Current workaround would be to do something like getting the DB from sqlite and querying via the cli or a DB program like `DB Browser Sqlite`:
<img width="1792" alt="Screen Shot 2022-05-25 at 3 15 16 PM" src="https://user-images.githubusercontent.com/32073428/170350314-d9767e29-1a54-4e8e-80ce-c77bb368afb6.png">

BrianPfitz:
Hey @spiffcs, thanks for your answer!
Yes, exactly! Just as you can run ```grype sbom:path/to/syft.json```, it would be nice if you could also run 
```grype cpe:path/to/cpe_list.csv```. 
In principle, the query for vulnerabilities should be based on the CPEs and less on the countless other pieces of information contained in an SBOM, right? If so, this "feature" could be implemented relatively straightforwardly.

spiffcs:
Awesome! I'll talk with the team here and we can prioritize getting a look at adding this. Feel free to take a stab at writing a small PR to if you think it's straightforward =)

BrianPfitz:
Hello @spiffcs,
Were you able to discuss my proposal in your meeting? What are your opinions on this?

In the meantime, I have looked over the code from time to time and have to conclude that it is unfortunately not as simple as I first thought. This is because everything is closely linked to Syft and the creation of an SBoM. Unfortunately, you can hardly reuse a function only with a CPE (and without locations, for example). 
I think I will search the database for the CPEs for now.

spiffcs:
[...]

Issue #516, opened 2021-12-02
Accept simple package list as input


wagoodman:
Today we accept syft JSON as input and in the near future will allow for SPDX and CycloneDX. However, we can lower the bar further by allowing a simple CSV file list of package-version pairs.

```
$ cat input.csv
my-package,v1.2.3
my-other-package,v1.0.0

$ grype ./input.csv
 ✔ Vulnerability DB        [updated]
 ✔ Loaded image
 ✔ Parsed image
 ✔ Cataloged packages      [2 packages]
 ✔ Scanned SBOM           [1 vulnerabilities]

NAME                INSTALLED                 FIXED-IN          VULNERABILITY     SEVERITY
my-package        v1.2.3                      v2.0              CVE-2019-18276    Low
```

luhring:
I love this.... ❤️ 

Offhand ideas:

1. Maybe we allow for specifying the **ecosystem**, too (e.g. "npm") — that can go a long way to reduce false positives
2. If the data structure were kept simple enough (e.g. just a tuple of scalars), we could potential support a large number of _file_ formats without much code (e.g. CSV, TSV, JSON, YAML, pretty much anything)

PR #795, opened 2022-06-19
feat: simple input case to request vulnerability data via purl

cpendery:
## 📝 Description
Enables users to pass in a `.csv` of CPEs, with an optional second item in each row of a pURL, to quickly match against the vulnerability database without requiring a complete sbom. This uses the same matches as Grype would use when processing an SBOM in order to have the results be consistent event though it may under report actual CPE matches

Closes https://github.com/anchore/grype/issues/757
Closes https://github.com/anchore/grype/issues/516


cmd/root.go
```diff
@@ -63,6 +63,7 @@ You can also explicitly specify the scheme to use:
     {{.appName}} sbom:path/to/syft.json                 read Syft JSON from path on disk
     {{.appName}} registry:yourrepo/yourimage:tag        pull image directly from a registry (no container runtime required)
     {{.appName}} att:attestation.json --key cosign.pub  explicitly use the input as an attestation
+    {{.appName}} csv:path/to/yourcsv                    read a CSV of CPE,PURL from a path on disk
```
spiffcs, line 66: Thanks for updating the documentation!

spiffcs (commented):
Small nit on the var delcaration, but @wagoodman my only question on this PR is for you.

There is limited data that is available on the PURL side of this kind of input. For example, if a package does not have certain metadata associated with it then grype will not know which matcher to use to provide the most accurate result.

Are we ok introducing this kind of input where incomplete package information is considered a best-effort at providing a vulnerability report?

wagoodman:
This also looks like it closes https://github.com/anchore/grype/issues/516

grype/pkg/csv_provider.go
```diff
@@ -0,0 +1,132 @@
...
+	csvReader := csv.NewReader(reader)
+	records, err := csvReader.ReadAll()
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse csv: %w", err)
+	}
+
+	packages := []Package{}
+
+	for _, row := range records {
+		rawCpe := strings.TrimSpace(row[0])
```
wagoodman, line 56: question: do we need CPEs? That is, could we accept a list of only pURLs?

cpendery: We definitely don't need a cpe, as we can generate a cpe set using the information from a purl. Would you rather allow a cpe as optional or only allow purl?

cpendery:
> ~This also looks like it closes https://github.com/anchore/grype/issues/516~
> 
> 
> 
> Actually, it looks like it almost covers this case. I'm curious what it would take to close the gap to cover "simple" input as well. The answer could be as simple as "pURLs are better and should be minimally used".

This could be changed to prioritize a purl. Since purl's include ecosystems, they provide everything you need to match and generate the cpes for.

Though distro would still be information needed for some matches though I think passing it in the through the cli flag should still be fine `--platform <string>`

cmd/root.go
spiffcs: @wagoodman we might have talked about this offline, but what should this be renamed too if we don't want to fully support `cve` parsing as a grype input?

cpendery:
> Functionality on this is amazing thank you again @cpendery.
> 
> 
> 
> The only comment I had is where we don't outright say we support `csv` since that might imply you can set arbitrary header/column information for an input. 
> 
> 
> 
> Requesting input from @wagoodman since I think he had some ideas on what this feature's input surface might ultimately look like

Sounds good to me. I agree `csv` is too broad. I'll wait on that review before changing it to be mandatory purl, optional cpe

spiffcs (approved):

Issue #771, opened 2022-06-03
`grype` flags CVE from unrelated package with same name


wbadart:
**What happened**:
Running `grype dir:.` in the root of a Svelte project I'm working on reports

```
...
sorcery   0.10.0               npm   CVE-2020-11052  Critical
...
```

**What you expected to happen**:
On further investigation, it seems that [CVE-2020-11052](https://cve.circl.lu/cve/CVE-2020-11052) was an issue with the [`sorcery` ruby gem](https://rubygems.org/gems/sorcery), not the [`sorcery` npm package](https://www.npmjs.com/package/sorcery). `grype` successfully identified my `sorcery` as an npm package, so I guess I expected it to be able to disambiguate it from the ruby gem and its associated CVE.

**How to reproduce it (as minimally and precisely as possible)**:
I used `npm` 8.5.5 in a blank directory for the below:

```shell
$ npm init --yes
...

$ npm install sorcery@0.10.0
added 20 packages, and audited 21 packages in 794ms

1 package is looking for funding
  run `npm fund` for details

found 0 vulnerabilities

$ grype dir:.
 ✔ Vulnerability DB        [no update available]
 ✔ Indexed .
 ✔ Cataloged packages      [20 packages]
 ✔ Scanned image           [1 vulnerabilities]

NAME     INSTALLED  FIXED-IN  TYPE  VULNERABILITY   SEVERITY
sorcery  0.10.0               npm   CVE-2020-11052  Critical
```

**Anything else we need to know?**:

**Environment**:
- Output of `grype version`:
    ```
    Application:          grype
    Version:              0.38.0
    Syft Version:         v0.46.2
    BuildDate:            2022-05-23T14:41:50Z
    GitCommit:            06d28dad9f7e7d9aa65fc16d45c6ce785826664c
    GitDescription:       v0.38.0
    Platform:             darwin/amd64
    GoVersion:            go1.18.2
    Compiler:             gc
    Supported DB Schema:  3
    ```
- OS (e.g: `cat /etc/os-release` or similar):
    - macOS Monterey v12.4

Thanks for your consideration!

Issue #445, opened 2021-10-06
false positive for CVE-2021-29940 for npm through


thediveo:
**What happened**:

running grype on github.com/thediveo/lxkns results in CVE-2021-29940 for component through. However,  CVE-2021-29940 refers to a Rust component, but the component found is a nodejs npm package.

**What you expected to happen**:

No false positive for  CVE-2021-29940 when using npm through package.

**Environment**:
- grype build from 2dd4131
- go1.17.1
- linux/arm64

luhring:
Good catch!

>The search says rust ? why is js picking it up ?

The vulnerability record's CPE has `rust` for its `target_sw` field value. The `target_sw` field for the artifact here is set to `*` — and `*` matches any value.

My two cents is that we should focus more on our use of `*` for the `vendor` field.

TheDiveO:
> My two cents is that we should focus more on our use of `*` for the `vendor` field.
Judging from my ignore file I can second that. It contains several unfiled false positives. Do you want them in a batch here?

luhring:
That'd be great, @TheDiveO, thanks. Whatever method is easiest for reporting. The more data we get about this particular cause of FPs, the better we can handle it.

Issue #390, opened 2021-08-19
Filter CPE matches by target SW to reduce FPs


wagoodman:
If a CPE search results in a match against a CPE that has target SW specified, then check the package type and/or language against the value to see if it has any known incompatibilities (e.g. python package language matching against a vulnerability with a CPE that has a target SW of `nodejs` is probably wrong, so the match should be filtered out).

It cannot be assumed that target SW is specified; if not specified this filtering should be skipped.

Also, there are some odd permutations that may indeed be valid, for example, a java package wrapping a javascript package. It may be entirely valid for the vulnerability to have a CPE for `nodejs` even if the package is a `java` type. This needs to be explored more --is this contrived? or is this case prevalent?

westonsteimel:
I think this makes a lot of sense.  There is at least an example of that scenario happening at https://github.com/anchore/grype/issues/445#issuecomment-939705285

I guess you could make it configurable?

westonsteimel:
Another thing which might potentially be useful (and should probably be configurable on individual ecosystem matcher level) could be if we find an NVD match and that CVE has a GHSA for another ecosystem (and the eocsystem we're matching for is supported by GitHub) we filter it out.  Hopefully that makes sense but if not I can try and explain it better.

PR #810, opened 2022-06-27
feat: add filtering support using target software field in cpe

cpendery:
## ⛰️ Background
In terms of addressing the concerns from the issue, I processed all the cpes, curtesy of @westonsteimel's [cpe repo](https://github.com/westonsteimel/nvd-data-analysis/tree/main/data/cpe), and identified the ~10k different possible product|vendor|target_sw pairings. I manually evaluated 200 of these from the set of those supported by Syft ~1500 (13%) and saw no language wrapping where filtering would cause a bad mismatch. [Work here](https://gist.github.com/cpendery/82aadba917c2dd380cf9de4ed5743899).

## 📝 Description
Add's support for filtering out vulnerabilities where every cpe's target software does not match the package's language and has the loosest possible filtering by including all target software's that Syft doesn't support regardless of if a match is found or not to reduce missed vulnerabilities. 

Closes https://github.com/anchore/grype/issues/390
Closes https://github.com/anchore/grype/issues/445
Closes https://github.com/anchore/grype/issues/771
Closes https://github.com/anchore/grype/issues/445

### DIFF
--- a/grype/search/cpe.go
+++ b/grype/search/cpe.go
@@ -82,6 +82,8 @@ func ByPackageCPE(store vulnerability.ProviderByCPE, p pkg.Package, upstreamMatc
 			return nil, fmt.Errorf("unable to filter cpe-related vulnerabilities: %w", err)
 		}
 
+		applicableVulns = onlyVulnerableTargets(p, allPkgVulns)
+
 		// for each vulnerability record found, check the version constraint. If the constraint is satisfied
 		// relative to the current version information from the CPE (or the package) then the given package
 		// is vulnerable.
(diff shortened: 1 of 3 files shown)


westonsteimel:
What happens here if there are multiple CPE's for an entry with different target software components?  We'd want to make sure a match isn't excluded if any of the software targets match the current package.  I haven't tried that scenario yet, but perhaps we can add a test case for that?

cpendery:
> What happens here if there are multiple CPE's for an entry with different target software components? We'd want to make sure a match isn't excluded if any of the software targets match the current package. I haven't tried that scenario yet, but perhaps we can add a test case for that?

It'll only exclude a match if every item in the cpe set isn't a match. I can add a test case for that

PR #825, opened 2022-07-12
add DBCloser. Clients can aviod db connection leak if vulnerability db is loaded many times

artsv79:
add DBCloser interface to LoadVulnerabilityDB() function. Clients can avoid db connection leak if vulnerability db is loaded many times.
Important for tools like https://github.com/ckotzbauer/vulnerability-operator which loads vulnerability db periodically during its runtime cycle.

Signed-off-by: Sergey Artamonov <artsv79@gmail.com>

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -284,6 +284,7 @@ func startWorker(userInput string, failOnSeverity *vulnerability.Severity) <-cha
 
 		var store *store.Store
 		var status *db.Status
+		var dbCloser *db.Closer
 		var packages []pkg.Package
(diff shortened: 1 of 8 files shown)


spiffcs:
Thanks for the cleanup on this @artsv79!

artsv79:
@spiffcs Could you please create a tag on the latest `main` so I could refer to this bugfixed revision from my project ?

spiffcs:
hey @artsv79 - we're working through a few more things before tagging the next version - does doing

```
```

PR #838, opened 2022-07-21
update grype to use syft v0.52.0

spiffcs:
Add integration test coverage for Haskell detection and bump syft to latest version including newest cataloger


test/integration/compare_sbom_input_vs_lib_test.go
```diff
@@ -68,6 +68,7 @@ func TestCompareSBOMInputToLibResults(t *testing.T) {
 		string(syftPkg.ConanPkg),
 		string(syftPkg.PortagePkg),
 		string(syftPkg.CocoapodsPkg),
+		string(syftPkg.HackagePkg),
```
spiffcs, line 71: We're going to update this test in a future PR so that we get more coverage for the newly supported detection over in syft

Issue #837, opened 2022-07-21
Grype execution fails with timeout when downloading vulnerability db on high cpu usage


awca22:
**What happened**:
During our ci pipelines, we have noticed that when we run grype to scan a container image it sometimes fails when there is a high CPU usage on our pod or on the node that the pod is running. The failure comes from a timeout when it tries to download the vulnerability db.

Is there any requirement on how much resources does grype need to have reserved for this task?
The limits we have are these:
`      resources:
        limits:
          cpu: '1'
          memory: 2Gi
        requests:
          cpu: 500m
          memory: 1Gi`

**What you expected to happen**:
The database gets downloaded normally and the execution runs fine.

**How to reproduce it (as minimally and precisely as possible)**:
In any Kubernetes pod, just add some CPU load with something similar to : 
`stress-ng --cpu 8 --vm 10 --vm-bytes 2G --timeout 60000s` 

After this try to run a scan: 
`time grype ubuntu:latest -vv`

After some minutes you will see the following line:
[0030] ERROR failed to fetch latest version: Get "https://toolbox-data.anchore.io/grype/releases/latest/VERSION": net/http: TLS handshake timeout

This takes around 
real	6m20.532s
user	0m3.045s
sys	1m48.307s

In our case, we download it from our artifactory registry.

**Anything else we need to know?**:
When there is no load, the execution is fine and takes less time:
real	0m50.315s
user	0m11.424s
sys	0m6.151s

**Environment**:
- Output of `grype version`:
          root:/# grype version
          Application:          grype
          Version:              0.43.0
          Syft Version:         v0.51.0
          BuildDate:            2022-07-18T21:07:27Z
          GitCommit:            addbd07b4f8cbfea283e10ffbd30e4cc6bfbfc0a
          GitDescription:       v0.43.0
          Platform:             linux/amd64
          GoVersion:            go1.18.3
          Compiler:             gc
          Supported DB Schema:  4

- OS (e.g: `cat /etc/os-release` or similar):
        PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
        NAME="Debian GNU/Linux"
        VERSION_ID="11"
        VERSION="11 (bullseye)"
        VERSION_CODENAME=bullseye
        ID=debian
        HOME_URL="https://www.debian.org/"
        SUPPORT_URL="https://www.debian.org/support"
        BUG_REPORT_URL="https://bugs.debian.org/"

cpendery:
[...]

PR #841, opened 2022-07-23
[Example] ci: add grype image bundled with db for ci

cpendery:
Just an example idea https://github.com/anchore/grype/issues/837

Issues:
- doesn't tag with the version, giturl, or commit inside using goreleaser
- adds additional goreleaser config / would also need added to current goreleaser setup
- hard to handle if you want pegged version ex. 34.7, need to republish every image as part of the re-release job

### DIFF
--- a/.github/workflows/republish-ci-container.yml
+++ b/.github/workflows/republish-ci-container.yml
@@ -1,0 +1,19 @@
+name: "Republish grype:ci Container"
+on:
+  repository_dispatch:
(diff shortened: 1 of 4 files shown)


nigelgbanks:
Will `grype` not be re-publishing the image with the database included is this intended as only an example?

cpendery:
This was just an example. I'm going to chat with the team about how they want to move forward with this to find a good solution for the issue

PR #843, opened 2022-07-25
add env variables and keychain for GHCR publish

spiffcs:


kzantow (approved):
🎉

Issue #754, opened 2022-05-20
ux: db: update: append more information about the next update


Dentrax:
**What would you like to be added**:

We can enrich the following subcommand by adding some details about the next update. For example, current db version, next db version, released datetime, is compatible, digest, etc.

```
$ grype db check
Update available!
```

**Why is this needed**:

For better UX and knowledge.

**Additional context**:

cc @developer-guy

luhring:
Great idea!

jonasagx:
For reference, here is the debug logs from grype currently when an update is available:
```
$ grype db check -vv
[0000] DEBUG checking for available database updates
[0000] DEBUG found database update candidate: Listing(url=https://toolbox-data.anchore.io/grype/databases/vulnerability-db_v3_2022-05-26T08:16:35Z_0210c940e39bcc249c7b.tar.gz)
[0000] DEBUG existing database (2022-05-25 08:16:22 +0000 UTC) is older than candidate update (2022-05-26 08:16:35 +0000 UTC), using update...
[0000] DEBUG database update available: Listing(url=https://toolbox-data.anchore.io/grype/databases/vulnerability-db_v3_2022-05-26T08:16:35Z_0210c940e39bcc249c7b.tar.gz)
Update available!
```

That info comes from [this struct](https://github.com/jonasagx/grype/blob/10c360449864cc91115b23fc267c5ca2273639fc/grype/db/listing_entry.go#L17-L25) representing an update candidate.

PR #853, opened 2022-08-02
feat: enrich db check cmd feedback

desenna:
Signed-off-by: Michael de Senna <desenna@gmail.com>

Closes https://github.com/anchore/grype/issues/754

### DIFF
--- a/cmd/db_check.go
+++ b/cmd/db_check.go
@@ -35,9 +35,13 @@ func runDBCheckCmd(_ *cobra.Command, _ []string) error {
 	}
 
 	fmt.Println("Update available!")
-	fmt.Printf("Current DB version %d was built on %s\n", currentDBMetadata.Version, currentDBMetadata.Built.String())
-	fmt.Printf("Update DB version %d was built on %s\n", updateDBEntry.Version, updateDBEntry.Built.String())
-	fmt.Printf("Update DB URL: %s\n", updateDBEntry.URL.String())
+
+	if currentDBMetadata != nil {
+		fmt.Printf("Current DB version %d was built on %s\n", currentDBMetadata.Version, currentDBMetadata.Built.String())
+	}
+
+	fmt.Printf("Updated DB version %d was built on %s\n", updateDBEntry.Version, updateDBEntry.Built.String())
+	fmt.Printf("Updated DB URL: %s\n", updateDBEntry.URL.String())
 	fmt.Println("You can run 'grype db update' to update to the latest db")
 
 	return nil
--- a/cmd/db_check.go
+++ b/cmd/db_check.go
@@ -25,7 +25,7 @@ func runDBCheckCmd(_ *cobra.Command, _ []string) error {
 		return err
 	}
 
-	updateAvailable, _, err := dbCurator.IsUpdateAvailable()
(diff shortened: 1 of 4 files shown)


spiffcs:
Thanks for the PR @desenna - I'll give this a look - just kicked off our tests

cmd/db_check.go
```diff
@@ -34,5 +34,10 @@ func runDBCheckCmd(_ *cobra.Command, _ []string) error {
 		return stderrPrintLnf("No update available")
 	}
 
-	return stderrPrintLnf("Update available!")
+	fmt.Println("Update available!")
+	fmt.Printf("Current DB version %d was built on %s\n", currentDBMetadata.Version, currentDBMetadata.Built.String())
+	fmt.Printf("Update DB version %d was built on %s\n", updateDBEntry.Version, updateDBEntry.Built.String())
+	fmt.Printf("Update DB URL: %s\n", updateDBEntry.URL.String())
```
spiffcs, line 40: I think just one more line here might be good for the user.

`You can run 'grype db update' to update to the latest db`

grype/db/curator.go
```diff
@@ -143,7 +143,7 @@ func (c *Curator) Update() (bool, error) {
 	defer downloadProgress.SetCompleted()
 	defer importProgress.SetCompleted()
 
-	updateAvailable, updateEntry, err := c.IsUpdateAvailable()
+	updateAvailable, _, updateEntry, err := c.IsUpdateAvailable()
```
spiffcs, line 146: I think we might be able to use the new `metadata` object below this line

`downloading new vulnerability DB` <-- we can add `log.Debug` level logs with the more verbose information included in the check command output

spiffcs (commented):
Small comments - great enhancment! Thanks again for taking the time to improve this command.

grype/db/curator.go
desenna: Since `info` logs were already present, I added `metadata` info to one of them to stay consistent with the log level instead of using `log.Debug`... hope that's alright. If you feel strongly about making it debug though, just let me know and I'll fix. Thanks for reviewing!

spiffcs (approved):

spiffcs:
I'll add a commit today that cleans up these tests

spiffcs:
Locally when I pulled this down the CLI and integration tests passed:

I ran `make cli` and `make validate-cyclonedx-schema`
```
PASS
ok      github.com/anchore/grype/test/cli       112.975s
```

```
xmllint --noout --schema ./cyclonedx.xsd bom.xml
bom.xml validates
```

I pushed a commit that busts the integration cache to see if it is a cache problem

spiffcs:
Just confirmed locally the nil pointer will show up when running `grype db update` with the changes. Investigating now to see how we can project against this.

spiffcs:
@desenna Thanks again for the PR.

I think I'm going to make one more change here based on my read of #754.

It looks like the user wants the details to be included in the default out as well. Currently `-v` or `-vv` would need to be passed to view the info/debug logs.

I'll update https://github.com/anchore/grype/blob/main/cmd/db_update.go so we can surface this new data.

spiffcs (approved):

Dentrax:
This is cool! Thank you for tackled this up! 🚀

PR #854, opened 2022-08-02
Ensure database access is readonly

kzantow:
If running multiple instances of Grype, it is possible to get errors due to the way the database is opened, for example, running:

```shell
for i in `seq 1 15`
do
go run main.go alpine:latest >> output.log 2>&1 &      
done
```

This PR opens the database in readonly mode, which prevents these types of errors.

### DIFF
--- a/grype/db/internal/gormadapter/open.go
+++ b/grype/db/internal/gormadapter/open.go
@@ -9,7 +9,7 @@ import (
 	"github.com/anchore/sqlite"
 )
 
-var connectStatements = []string{
+var writerStatements = []string{
 	// performance improvements (note: will result in lost data on write interruptions).
 	// on my box it reduces the time to write from 10 minutes to 10 seconds (with ~1GB memory utilization spikes)
 	`PRAGMA synchronous = OFF`,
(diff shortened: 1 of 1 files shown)


grype/db/internal/gormadapter/open.go
```diff
@@ -9,16 +9,16 @@ import (
...
 	// performance improvements (note: will result in lost data on write interruptions).
 	// on my box it reduces the time to write from 10 minutes to 10 seconds (with ~1GB memory utilization spikes)
 	`PRAGMA synchronous = OFF`,
 	`PRAGMA journal_mode = MEMORY`,
 }
 
 // Open a new connection to a sqlite3 database file
-func Open(path string, overwrite bool) (*gorm.DB, error) {
-	if overwrite {
+func Open(path string, write bool) (*gorm.DB, error) {
```
spiffcs, line 20: Nice! 

Since we're changing the signature naming from `overwrite` => `write` do we want to modify that in the callers as well? I think overwrite as a variable name is used a couple of calls up.

Otherwise LGTM

spiffcs (approved):
Approve with a small comment on naming!

I checked the callers and we're using `false` for `write` for the main path of grype access when running the app. We're using `true` for `write` during our integration tests so we can manipulate entries.

Also, I checked the gorm library and there was no way I could find to query the returned `dbObj` on what connection string was used to initialize it.  Asserting that `&immutable=1` is being used seemed hard as a unit test to add, but I think we have pretty good coverage from our current integration suite. 

Maybe we can add a test that we expect to error when trying to do a write when constructing a db object with `false` as write?

Issue #855, opened 2022-08-02
Grype slow on parallel execution


cpendery:
**What happened**:
When I run Grype in parallel on a database, I'm seeing that up to 50% of the runtime is spending waiting on locks via pprof. I don't think it should be spending this long querying especially when the cli can only perform read only actions.

**What you expected to happen**:
I'd expect almost no time spending locking and unlocking the database

**How to reproduce it (as minimally and precisely as possible)**:
Run grype in parallel while using pprof on the process

**Anything else we need to know?**:
I think this would be an easy fix. Maybe adding in a readonly flag to the config which is on by default that appends `?cache=shared&mode=ro&_journal_mode=WAL` to the Grype db file path in order to set it as shared, read only, and disable locking by using Write-Ahead Logging based on connection string options from https://github.com/mattn/go-sqlite3

**Environment**:
- Output of `grype version`: `0.44.0`
- OS (e.g: `cat /etc/os-release` or similar): MacOS

spiffcs:
We just merged:
https://github.com/anchore/grype/pull/854

This adds `connStr += "&immutable=1"` to the conn string it `READONLY`

We can also add the improvement here before the next release so that reads are the most efficient they can be.

cpendery:
Awesome! Thanks for the quick response @spiffcs and sorry I didn't see that earlier

PR #857, opened 2022-08-02
additional readOptions added  per 855

spiffcs:
Solves #855 

The only one I don't see here is `_journal_mode=WAL`
https://www.sqlite.org/uri.html#uri_filenames_in_sqlite

https://www.sqlite.org/c3ref/open.html <-- Immutable ref
https://www.sqlite.org/wal.html <-- WAL ref
https://www.sqlite.org/sharedcache.html#sqlite_shared_cache_mode <-- Cache ref
https://www.sqlite.org/uri.html#uri_filenames_in_sqlite <-- mode=ro ref

```
immutable=1
cache=shared
mode=ro
_journal_mode=WAL
```

### DIFF
--- a/grype/db/internal/gormadapter/open.go
+++ b/grype/db/internal/gormadapter/open.go
@@ -16,10 +16,6 @@ var writerStatements = []string{
 	`PRAGMA journal_mode = MEMORY`,
 }
 
-var readerStatments = []string{
-	`PRAGMA journal_mode = WAL`,
-}
-
 var readOptions = []string{
 	"&immutable=1",
 	"&cache=shared",
@@ -52,13 +48,6 @@ func Open(path string, write bool) (*gorm.DB, error) {
 
 	if write {
 		for _, sqlStmt := range writerStatements {
-			dbObj.Exec(sqlStmt)
-			if dbObj.Error != nil {
-				return nil, fmt.Errorf("unable to execute (%s): %w", sqlStmt, dbObj.Error)
-			}
-		}
-	} else {
-		for _, sqlStmt := range readerStatments {
 			dbObj.Exec(sqlStmt)
 			if dbObj.Error != nil {
 				return nil, fmt.Errorf("unable to execute (%s): %w", sqlStmt, dbObj.Error)
--- a/grype/db/internal/gormadapter/open.go
+++ b/grype/db/internal/gormadapter/open.go
@@ -16,6 +16,13 @@ var writerStatements = []string{
 	`PRAGMA journal_mode = MEMORY`,
 }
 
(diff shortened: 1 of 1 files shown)


spiffcs:
@kzantow serendipity! @cpendery filed #855 which seems to dovetail with your new PR

Is there any additional testing you did to confirm your PR besides running the unit/integration test?

kzantow (approved):
👍

kzantow:
NOTE: I don't think WAL is needed, if we're opening the database in read only mode; a quick read makes me think this is a performance improvement for writes. @spiffcs

grype/db/internal/gormadapter/open.go
```diff
@@ -16,6 +16,13 @@ var writerStatements = []string{
 	`PRAGMA journal_mode = MEMORY`,
 }
 
+var readOptions = []string{
+	"&immutable=1",
+	"&cache=shared",
+	"&mode=ro",
+	"&_journal_mode=WAL",
```
wagoodman: I agree with the conversation out in comments that `WAL` shouldn't help us out from a readers perspective --we should probably remove this option.

grype/db/internal/gormadapter/open.go
```diff
@@ -29,7 +36,10 @@ func Open(path string, write bool) (*gorm.DB, error) {
 	}
 
 	if !write {
-		connStr += "&immutable=1"
+		// &immutable=1&cache=shared&mode=ro&_journal_mode=WAL
+		for _, o := range readOptions {
```
wagoodman, line 39: nit: could we remove the `&` from the list elements up above and add them in explicitly here?

wagoodman (approved):

grype/db/internal/gormadapter/open.go
```diff
@@ -29,7 +35,10 @@ func Open(path string, write bool) (*gorm.DB, error) {
 	}
 
 	if !write {
-		connStr += "&immutable=1"
+		// &immutable=1&cache=shared&mode=ro&_journal_mode=WAL
```
kzantow, line 38: nit: `_journal_mode=WAL` is no longer used

cpendery:
> NOTE: I don't think WAL is needed, if we're opening the database in read only mode; a quick read makes me think this is a performance improvement for writes. @spiffcs 
> 
> 
> 
> Also from the WAL docs:
> 
> > WAL might be very slightly slower (perhaps 1% or 2% slower) than the traditional rollback-journal approach in applications that do mostly reads and seldom write.
> 
> 

Thanks for the catch @kzantow. That's my bad

PR #858, opened 2022-08-02
Add test for running Grype concurrently

kzantow:
This PR adds a test to run Grype concurrently and validate there are no database locking or other related issues.


kzantow:
Closing this for now, as the problem appears to have been fixed via manual testing, but this test was not able to replicate the problem and therefore does not serve as a good validation of the fix.

PR #862, opened 2022-08-03
move docker release into separate release workflow

spiffcs:

This PR removes Grype's docker assets from the main `.gorelease.yaml` configuration.

### Context:

Currently, the release process needs to run on the `mac-os` github runner. This constraint is necessary because the binary assets published for darwin need to be signed and notarized on an xcode-compatible system.

This presents challenges in the form of installing and wrangling the dependencies for docker builds to be possible on the `mac-os` runner. Grype previously used `https://github.com/docker-practice/actions-setup-docker` to meet this challenge. 

Recently the action has taken 15-21 minutes to get a working docker installation up and running. In some cases, it would fail outright and kill any potential release that had been triggered for publishing.

This change removes the need for the docker dependency on the `mac-os` runner, thus removing `actions-setup-docker` from our release workflow.

With this change a release will be processed as follows:
- Release commit is tagged which triggers the quality gate 
- If the quality gate passes the `make release` command is run on a `mac-os` runner (no longer using any docker dependencies)
- If the release is published successfully then the docker assets are triggered to be built and pushed to their respective registries

### Changes
This PR introduces a separate `.goreleaser_docker` configuration that is used to build and publish images after a GitHub release has been published. The new block being used is labeled as `release-docker-assets` and can be re-triggered without having to re-run the `quality gate` or `release` steps should it fail independently.

The Makefile has been updated to add `snapshot-docker-assets` and `release-docker-assets`. These are analogous to their `snapshot` and `release` counterparts for the main `.goreleaser.yaml` file.


kzantow (approved):
👍

Issue #848, opened 2022-07-27
Grype update command


willyw0nka:
**What would you like to be added**:
A new command ```grype update```
**Why is this needed**:
It would make the update process a lot easier. Grype already knows if you are running an outdated version.
**Additional context**:
Example output
```bash
$ grype <image_name>
New version of grype is available: 0.44.0 (currently running: 0.41.0)
You can update grype to the latest version running 'grype update'x
...
$ grype update
Updating grype...
$ grype version
Application:          grype
Version:              0.44.0
Syft Version:         v0.52.0
BuildDate:            2022-07-25T16:28:13Z
GitCommit:            a6ec8f11be5fdc116dc7bbfebdc03bfc13cd2005
GitDescription:       v0.44.0
Platform:             linux/amd64
GoVersion:            go1.18.4
Compiler:             gc
Supported DB Schema:  4
```

spiffcs:
👋 Awesome! Thanks, @willyw0nka. This sounds like a great command to add and is a good first issue. Feel free to tackle it if you want to try and add a command to the project otherwise we'll try and find time to get it out into the next release.

adriens:
:heavy_plus_sign: :+1: :smile_cat:

willyw0nka:
Thanks! I'll give it a try myself when I find some time, probably after my vacations

wagoodman:
I see there is a closed PR with this issue. It seems like one of the leading reasons for closing is that it duplicates logic about releasing to multiple places. I do generally agree with this point -- I think I'm in favor of closing this issue if we do not ever intend on adding this kind of functionality. We can always revisit this in the future.

PR #874, opened 2022-08-11
Added grype update command

willyw0nka:
Added `grype update` command #848. When running `grype update` latest version available is downloaded and grype binary is replaced.

I just learned to write golang so any code update or review is welcome. Not sure how we should test this feature.


spiffcs:
Thanks @willyw0nka! This looks great. Kicked off the ci and will get around to getting you comments.

spiffcs:
@willyw0nka! 

I ran the actions and it looks like our Static Analysis tool picked up some suggested changes. Since you said you're new to the language these are great to look through to see some changes that can be made/updated. 

Normally I can just add a commit that fixes these but didn't want to miss the chance to give you more information about your Golang journey.

cmd/update.go
```diff
@@ -0,0 +1,197 @@
...
+	if err != nil {
+		defer os.Remove(f.Name())
+		panic(err)
+	}
+
+	var downloadURLTemplate = "https://github.com/%s/%s/releases/download/v%s/grype_%s_%s_%s.%s"
+	var downloadURL = fmt.Sprintf(downloadURLTemplate, "anchore", "grype", version, version, runtime.GOOS, runtime.GOARCH, extension)
+	log.Infof("Downloading v%s", version)
+	fmt.Printf("Downloading v%s...\n", version)
+	resp, err := http.Get(downloadURL) //nolint
```
willyw0nka, line 186: Added `//nolint` here to address `G107: Url provided to HTTP request as taint input` (https://securego.io/docs/rules/g107.html). Is there a more elegant way to resolve this?

cmd/update.go
```diff
@@ -52,6 +49,9 @@ func extractBinary(fileName string) string {
 		panic(err)
 	}
 	tempDir, err := os.MkdirTemp("", "e_grype")
+	if err != nil {
+		panic(err)
```
spiffcs, line 53: We definitely don't want panics to be in the user's path here consider returning the error up the call stack and modifying the function to `extractBinary(fileName string) (string, error)`

cmd/update.go
```diff
@@ -0,0 +1,197 @@
...
+
+func init() {
+	rootCmd.AddCommand(updateCmd)
+}
+
+func runUpdate(_ *cobra.Command, _ []string) error {
+	fmt.Println("Checking for a newer version...")
+	newerVersionAvailable, desiredVersion, err := checkLatestVersion()
+	if err != nil {
+		panic(err)
```
spiffcs, line 33: `panic` is rarely used as the correct error path in Golang code. It's only reserved for very niche cases where a program has no where else to go and needs to be extremely verbose about the error. 

Consider returning the error here

cmd/update.go
```diff
@@ -0,0 +1,197 @@
...
+	updatedGrypePath := fmt.Sprintf("%s/grype", tempDir)
+	err = os.Chmod(updatedGrypePath, 0777)
+	if err != nil {
+		panic(err)
+	}
+	return updatedGrypePath
+}
+
+func replaceBinary(fileName string) {
+	fmt.Println("Applying update...")
```
spiffcs, line 72: nit: try not to mix and match the `log/fmt` packages here for grype

consider using the TUI that's been established and used through the other commands

cmd/update.go
```diff
@@ -0,0 +1,197 @@
...
+	fmt.Println("Checking for a newer version...")
+	newerVersionAvailable, desiredVersion, err := checkLatestVersion()
+	if err != nil {
+		panic(err)
+	}
+	if newerVersionAvailable {
+		compressed := downloadCompressed(desiredVersion)
+		filePath := extractBinary(compressed)
+		replaceBinary(filePath)
+	} else {
```
spiffcs, line 39: consider doing a return after `replaceBinary` - this removes the extra `else` block and the default case for this function can be informing the user the newest version is already installed

cmd/update.go
```diff
@@ -0,0 +1,197 @@
...
+		extension = "zip"
+	}
+	f, err := os.CreateTemp("", "grype")
+	log.Debugf("Created temporary file %s", f.Name())
+	if err != nil {
+		defer os.Remove(f.Name())
+		panic(err)
+	}
+
+	var downloadURLTemplate = "https://github.com/%s/%s/releases/download/v%s/grype_%s_%s_%s.%s"
```
spiffcs, line 182: Consider lifting these declarations to the top of the file and removing the formatting statements for nondynamic string like `anchore` and `grype`

cmd/update.go
spiffcs: Just adding context here to make sure users know why this would not be considered a tainted input. In this case the user has control over some parts of the URL string but would have to hijack the full domain for the redirect to happen.

Consider also formatting the string and using https://pkg.go.dev/net/url#Parse to parse the input into a valid URL struct

spiffcs (commented):
Initial comments - Thanks again for the PR!

Please make sure to remove all instances of `panic` and return errors up the call stack modifying signatures where needed.

Also, take a look at the other commands and see how they utilize the `TUI` to format and express certain outputs. It would probably be better to express the update process as part of that flow than just `fmt.Println` calls. I'll look about adding tests here to prove out the replace/extract code since I think there could be some subtle bugs there.

willyw0nka:
Hi @spiffcs, after reviewing your comments and checking how the other commands were implemented now `grype update` uses the standard output format.

Output on stage `download grype` could be improved adopting `grype db update` download tracking progress bar output style, currently a spinner is shown while the newer grype version is downloaded.

This improvement was not included on this last commit due to the code duplication that it implies, mainly on `grype/db/curator.go`. Take a look at the current command output, if a more polished result is desired I could work on adapting the download tracking progress bar for `grype update`.

willyw0nka:
Hi @spiffcs :wave:! did you find some time to review the pull request update? I would love to recieve feedback on the update  :smile:

spiffcs:
@willyw0nka so sorry for missing this I'll review today

spiffcs:
Sorry, let me try and resolve this conflict so we can at least get the CI jobs running again. Apologies for the delay on this. I've been heads down over in syft and context switching at the moment is a little costly.

spiffcs:
Did some cleanup and merged into main - will look again in the AM and ask the team about merging

spiffcs:
I've played a bit more with this recently and don't think we want to increase the release surface area with this command.

This current implementation couples a lot of the update process with the current release mechanism. It duplicates a lot of the logic across the codebase. If any of those were to update then this command would break which gives me the impression that we need to use the linked issue to discuss/come up with a design that attempts to single source of truth more of this logic.

How can we get an update command that helps the user without duplicating or interfering with other methods of installation (package manager, brew, nix, etc)?

PR #917, opened 2022-09-08
Add support for scanning RPM files

kzantow:
This PR is a follow-on to https://github.com/anchore/syft/pull/1188, which adds support for scanning RPM files. This aligns the naming from `Rpmdb` to `Rpm` in a number of places.

Closes #570 

NOTE: scanning RPM files for vulnerabilities may not work as desired, as there is no Linux distribution identified. In the case there are RPMs identified, it may be required to specify a distribution, which requires knowing which _version_ to use:
```
% grype ~/Downloads/rpm 
 ✔ Vulnerability DB        [no update available]
 ✔ Indexed /Users/kzantow/Downloads/rpm 
 ✔ Cataloged packages      [1 packages]
 ✔ Scanned image           [0 vulnerabilities]

[0000]  WARN Unable to determine the OS distribution. This may result in missing vulnerabilities. You may specify a distro using: --distro <distro>:<version>
No vulnerabilities found

% grype ~/Downloads/rpm --distro centos:6
 ✔ Vulnerability DB        [no update available]
 ✔ Indexed /Users/kzantow/Downloads/rpm 
 ✔ Cataloged packages      [1 packages]
 ✔ Scanned image           [1 vulnerabilities]

NAME        INSTALLED  FIXED-IN     TYPE  VULNERABILITY  SEVERITY 
audit-libs  2.4.5      (won't fix)  rpm   CVE-2015-5186  Medium    
```

PR #920, opened 2022-09-12
Revert "Add support for scanning RPM files (#917)"

westonsteimel:
This reverts commit ba73ab362a72b8ed5dc9d25fa86ba64f5eeec44b.

This seems to have introduced some CGO dependencies into Grype and is currently causing the daily grype database publish pipeline to fail.


robcresswell (approved):

westonsteimel:
We've merged a change in yardstick which should hopefully fix this issue with the db build

PR #944, opened 2022-10-03
implement v5 db schema to support improved matching between rpm appstream modules

westonsteimel:
Adds support for a `package_qualifiers` column to allow evaluating package matches to vulnerabilities based on more than just version constraints.  Currently adds an `rpm-modularity` qualifier in order to support matching to correct app stream module in order to reduce false positives within rpm-based distro ecosystems.  In order to prevent an increase in false positive matches for previous versions of grype using the v4 schema, this change (along with the vulnerability source driver parser updates) requires bumping the schema to v5.


.github/workflows/validations.yaml
```diff
@@ -10,6 +10,7 @@ env:
   GO_VERSION: "1.18.x"
   GO_STABLE_VERSION: true
   PYTHON_VERSION: "3.10"
+  GRYPE_DB_UPDATE_URL: "https://toolbox-data.anchore.io/grype/staging-databases/listing.json"
```
wagoodman, line 13: reminder before merging: remove this

grype/db/v5/pkg/qualifier/from_json.go
```diff
@@ -0,0 +1,39 @@
...
+func FromJSON(data []byte) ([]Qualifier, error) {
+	var records []map[string]interface{}
+	if err := json.Unmarshal(data, &records); err != nil {
+		return nil, err
+	}
+
+	var qualifiers []Qualifier
+
+	for _, r := range records {
+		if k, ok := r["kind"]; ok {
```
wagoodman, line 21: nit/question: do we want to warn when there is no `kind` key found?

grype/db/v5/pkg/qualifier/rpmmodularity/qualifier.go
```diff
@@ -0,0 +1,21 @@
...
+	Kind   string `json:"kind" mapstructure:"kind"`                         // Kind of qualifier
+	Module string `json:"module,omitempty" mapstructure:"module,omitempty"` // Modularity label
+}
+
+func (q Qualifier) Parse() qualifier.Qualifier {
+	return rpmmodularity.NewRpmModularityQualifier(q.Module)
+}
+
+func (q Qualifier) String() string {
+	return fmt.Sprintf("kind: %s, module: %s", q.Kind, q.Module)
```
wagoodman, line 20: nit
```suggestion
	return fmt.Sprintf("kind: %s, module: %q", q.Kind, q.Module)
```

grype/pkg/qualifier/rpmmodularity/qualifier.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"github.com/anchore/grype/grype/pkg"
+	"github.com/anchore/grype/grype/pkg/qualifier"
+)
+
+type rpmModularity struct {
+	module string
+}
+
+func NewRpmModularityQualifier(module string) qualifier.Qualifier {
```
wagoodman, line 14: nit: from a naming/organization point of view, this should be either:
- renamed to `New()` so that callers use `rpmmodularity.New()`, or
- be moved to the `qualifier` package, so that callers call `qualifier.NewRpmModularityQualifier()`

I can see the value of keeping the underlying interface separate from the capabilities, so the first option seems like the right call at a surface glance.

grype/search/language.go
```diff
@@ -20,7 +20,12 @@ func ByPackageLanguage(store vulnerability.ProviderByLanguage, p pkg.Package, up
 		return nil, fmt.Errorf("matcher failed to fetch language=%q pkg=%q: %w", p.Language, p.Name, err)
 	}
 
-	applicableVulns, err := onlyVulnerableVersions(verObj, allPkgVulns)
+	applicableVulns, err := onlyQualifiedPackages(p, allPkgVulns)
+	if err != nil {
+		return nil, fmt.Errorf("unable to filter language-related vulnerabilities: %w", err)
+	}
+
+	applicableVulns, err = onlyVulnerableVersions(verObj, applicableVulns)
```
wagoodman, line 28: nit: it would be nice to add a TODO here that states we want to port over this check to be a qualifier (and removed)

test/grype-test-config.yaml
```diff
@@ -1 +1,4 @@
-check-for-app-update: false
\ No newline at end of file
+check-for-app-update: false
+
+db:
+  update-url: https://toolbox-data.anchore.io/grype/staging-databases/listing.json
```
wagoodman, line 4: just a reminder before merge: remove this

test/integration/diff_test.go
```diff
@@ -18,8 +18,8 @@ import (
 var update = flag.Bool("update", false, "update the *.golden files for diff presenter")
 
 const (
-	baseURL   = "https://toolbox-data.anchore.io/grype/databases/vulnerability-db_v4_2022-07-05T08:18:22Z_39868af44fc51829a7c9.tar.gz"
-	targetURL = "https://toolbox-data.anchore.io/grype/databases/vulnerability-db_v4_2022-07-06T08:16:42Z_c840f17244dea46d0c07.tar.gz"
+	baseURL   = "https://toolbox-data.anchore.io/grype/staging-databases/vulnerability-db_v5_2022-10-14T08:22:01Z_69c99aa5917dea969f2d.tar.gz"
```
wagoodman, line 21: not in scope for this PR: we need to rework this test to not be dependent on hosted DBs like this. We're thinking about deprecating beyond a time window in the future, which means this test would start to fail after 90days (or whatever window we go with). 

We could capture this as an OCI artifact, but there is also the problem that this is coupled to the current DB schema version... should we change this? or is this assumption OK? In which case, we just need a nice way to continually update the fixture when we bump the schema version.

wagoodman (approved):
Note: merging this will put grype in an unreleasable state until there is a V5 DB built and published.

I've added a few nits and reminder comments, but overall looks great! 🥇

wagoodman (approved):

PR #961, opened 2022-10-20
feat: consider well-known false-positive generating CPE target SW components in match filtering logic

westonsteimel:
Enhances the CPE target software component match filtering logic to consider ecosystems which aren't currently supported by syft cataloging but are well-known sources of false-positives. This currently adds support for filtering various permutations of `wordpress`, `joomla`, and `drupal`

Signed-off-by: Weston Steimel <weston.steimel@anchore.com>

### DIFF
--- a/grype/search/only_vulnerable_targets.go
+++ b/grype/search/only_vulnerable_targets.go
@@ -8,6 +8,29 @@ import (
 	syftPkg "github.com/anchore/syft/syft/pkg"
 )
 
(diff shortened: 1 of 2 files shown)


westonsteimel:
You can see we drop a few false positives from the quality gate tests at https://github.com/anchore/grype/actions/runs/3380662750/jobs/5613691124#step:9:920

kzantow (approved):
👍

Issue #964, opened 2022-10-22
Not applying Alpine secdb data correctly for "edge"


luhring:
### Summary

Grype does not handle apk matching correctly for "edge" versions of Alpine.

When scanning a container image that uses an "edge" version of Alpine (i.e. a version beyond the last official release), Grype does not apply **any** secdb fixes to its matching (when it _should_ apply fixes from the _"edge"_ secdb data, that is, from `https://secdb.alpinelinux.org/edge/`).

This behavior results in **false positives**.

### Example

We can look at the `alpine:edge` image on Docker Hub. Here's the distro information from Syft:

```console
$ syft -q registry:alpine:edge -o json | jq '.distro'
{
  "prettyName": "Alpine Linux edge",
  "name": "Alpine Linux",
  "id": "alpine",
  "versionID": "3.17_alpha20220715",
  "homeURL": "https://alpinelinux.org/",
  "bugReportURL": "https://gitlab.alpinelinux.org/alpine/aports/-/issues"
}
```

We know this image uses an "edge" version of Alpine, because the latest release of Alpine (today) is `3.16`, yet this image is using `3.17_alpha20220715`.

Here's what Grype reports for this image:

```console
$ grype -q registry:alpine:edge
NAME        INSTALLED   FIXED-IN  TYPE  VULNERABILITY   SEVERITY
busybox     1.35.0-r18            apk   CVE-2022-28391  High
ssl_client  1.35.0-r18            apk   CVE-2022-28391  High
zlib        1.2.12-r1             apk   CVE-2022-37434  Critical
zlib        1.2.12-r1             apk   CVE-2018-25032  High
```

For this example we'll look at the **_first_** reported vulnerability, `busybox` being affected by `CVE-2022-28391` (although you can follow these steps for the **_fourth_** reported vulnerability (CVE-2018-25032), too).

We can look up the secfixes data for busybox:

```console
$ curl -s https://secdb.alpinelinux.org/edge/main.json | jq '.packages[] | select(.pkg.name == "busybox")'
...
      "1.35.0-r7": [
        "ALPINE-13661",
        "CVE-2022-28391"
      ]
...
```

This shows that for busybox, the vulnerability `CVE-2022-28391` was resolved as of `1.35.0-r7`. The version of busybox installed in the image is `1.35.0-r18`, which is beyond the fix version and this not affected by CVE-2022-28391, but it's still showing up in Grype's output.

Feel free to ping me if you have any questions!

spiffcs:
[...]

PR #965, opened 2022-10-24
fix: update case for alpine:edge correct vuln feed

spiffcs:
Partial Fix for #964 

## Summary
When scanning `alpine:edge` with grype the incorrect vulnerability namespace would be used causing false positives that would already be fixed in the upstream source:

https://security.alpinelinux.org/branch/edge-main
https://gitlab.alpinelinux.org/ariadne/secfixes-tracker

This fix treats nil distro version segments (a product of a version string that could not be parsed), when distro is of type `alpine`, as the signal for returning the `alpine:edge` namespace.

### Notes

Please read the related issue for upstream recommendations on reducing the false positives produced by this PR. 

### DIFF
--- a/grype/db/v5/namespace/index.go
+++ b/grype/db/v5/namespace/index.go
@@ -2,6 +2,7 @@ package namespace
 
 import (
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/anchore/grype/grype/db/v5/namespace/cpe"
@@ -11,6 +12,8 @@ import (
 	"github.com/anchore/grype/internal/log"
 	syftPkg "github.com/anchore/syft/syft/pkg"
 )
+
+var alpineVersionRegularExpression = regexp.MustCompile(`^(\d+)\.(\d+)(?:\.(\d+))?(?:-r(\d+))?$`)
 
 type Index struct {
 	all         []Namespace
@@ -112,6 +115,23 @@ func (i *Index) NamespacesForDistro(d *grypeDistro.Distro) []*distro.Namespace {
 		case grypeDistro.CentOS, grypeDistro.RedHat, grypeDistro.Fedora, grypeDistro.RockyLinux, grypeDistro.AlmaLinux, grypeDistro.Gentoo:
 			// TODO: there is no mapping of fedora version to RHEL latest version (only the name)
 			distroKey = fmt.Sprintf("%s:%d", strings.ToLower(string(grypeDistro.RedHat)), versionSegments[0])
+			if v, ok := i.byDistroKey[distroKey]; ok {
+				return v
+			}
+		case grypeDistro.Alpine:
+			// Check if the alpine raw version matches x.y.z
+			if alpineVersionRegularExpression.Match([]byte(d.RawVersion)) {
+				// Get the first two version components
+				distroKey = fmt.Sprintf("%s:%d.%d", strings.ToLower(d.Type.String()), versionSegments[0], versionSegments[1])
+				if v, ok := i.byDistroKey[distroKey]; ok {
+					return v
+				}
+			}
+
+			// If the version does not match x.y.z then it is edge
+			// In this case it would have - or _ alpha,beta,etc
+			// https://github.com/anchore/grype/issues/964#issuecomment-1290888755
+			distroKey := fmt.Sprintf("%s:%s", strings.ToLower(d.Type.String()), "edge")
 			if v, ok := i.byDistroKey[distroKey]; ok {
 				return v
 			}
--- a/grype/db/v5/namespace/index.go
+++ b/grype/db/v5/namespace/index.go
@@ -119,20 +119,8 @@ func (i *Index) NamespacesForDistro(d *grypeDistro.Distro) []*distro.Namespace {
 				return v
 			}
 		case grypeDistro.Alpine:
-			// Check if the alpine raw version matches x.y.z
-			if alpineVersionRegularExpression.Match([]byte(d.RawVersion)) {
-				// Get the first two version components
-				distroKey = fmt.Sprintf("%s:%d.%d", strings.ToLower(d.Type.String()), versionSegments[0], versionSegments[1])
-				if v, ok := i.byDistroKey[distroKey]; ok {
-					return v
-				}
-			}
-
-			// If the version does not match x.y.z then it is edge
-			// In this case it would have - or _ alpha,beta,etc
-			// https://github.com/anchore/grype/issues/964#issuecomment-1290888755
-			distroKey := fmt.Sprintf("%s:%s", strings.ToLower(d.Type.String()), "edge")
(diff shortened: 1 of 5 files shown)


grype/db/v5/namespace/index.go
```diff
@@ -118,6 +118,15 @@ func (i *Index) NamespacesForDistro(d *grypeDistro.Distro) []*distro.Namespace {
 		}
 	}
 
+	// Fall back to alpine:edge if no version segments found
+	// alpine:edge is labeled as alpine-x.x_alphaYYYYMMDD
+	if versionSegments == nil && d.Type.String() == "alpine" {
```
wagoodman, line 123: I think there is a `grypeDistro.Alpine` to compare against instead of the string (or does that not work in this case?)

grype/db/v5/namespace/index.go
```diff
@@ -118,6 +118,15 @@ func (i *Index) NamespacesForDistro(d *grypeDistro.Distro) []*distro.Namespace {
...
+		distroKey := fmt.Sprintf("%s:%s", strings.ToLower(d.Type.String()), "edge")
```
wagoodman, line 124: should we check to see if the raw version has something like `alpha` or `beta`? (from https://github.com/anchore/grype/blob/main/grype/distro/distro.go#L16)

grype/db/v5/namespace/index_test.go
```diff
@@ -137,124 +138,155 @@ func TestIndex_NamespacesForDistro(t *testing.T) {
...
 		{
+			name:   "alpine minor version matches minor version namespace",
 			distro: newDistro(t, osDistro.Alpine, "3.16", []string{}),
 			namespaces: []*distro.Namespace{
 				distro.NewNamespace("alpine", osDistro.Alpine, "3.16"),
 			},
 		},
 		{
+			name:   "alpine edge version matches edge namespace",
+			distro: &osDistro.Distro{Type: osDistro.Alpine, Version: nil, RawVersion: "3.17_alpha20221002", IDLike: []string{"alpine"}},
```
wagoodman, line 161: we should also add a case that covers the possibility if they start changing their versions to be semver compliant (or close) as an extra precaution (e.g. `3.17-alpha20221002` or `3.17.0-alpha20221002`). I realize this is unlikely, but knowing the namespace logic is resilient in cases like these is useful since the data can change out from underneath us (and the alternative is a code change when that happens).

luhring:
Cross-posting this update for visibility 😃  https://github.com/anchore/grype/issues/964#issuecomment-1290888755

grype/db/v5/namespace/index.go
luhring: Re: checking the version, my suggestion here would be to find a way to test this value to see if it matches the `x.y.z` format mentioned in [this comment](https://github.com/anchore/grype/issues/964#issuecomment-1290888755). The logic right now is still just a bit off, because there can be cases where the version is something like `1.2.3-beta`, so `versionSegments` would not be `nil`, and this code won't use edge when it actually should use edge.

spiffcs: Updated and added the case for the distro to match the version against a valid regex for `x.y.z` falling back to edge if `x.y.z` is unmatched.

grype/db/v5/namespace/index.go
spiffcs: Updated this with the new test cases to match against `x.y.z` and then falling back to edge so no need to check for `alpha` or `beta` in this case

grype/db/v5/namespace/index_test.go
spiffcs: Test cases were added for this possibility to check for namespace resilience.

spiffcs:
@luhring and @kzantow this one has been updated. Nice find @luhring on the possibility of `versionSegments` - I've updated the logic and test cases to account for this.

grype/db/v5/namespace/index.go
```diff
@@ -12,6 +13,8 @@ import (
 	syftPkg "github.com/anchore/syft/syft/pkg"
 )
 
+var alpineVersionRegularExpression = regexp.MustCompile(`^(\d+)\.(\d+)(?:\.(\d+))?(?:-r(\d+))?$`)
```
luhring, line 16: I think this regex should be changed to `^(\d+)\.(\d+)\.(\d+)$`. As written the regex will match version strings beyond the `x.y.z` rule, such as `3.16` and `1.2.3-r4`.

grype/db/v5/namespace/index_test.go
```diff
@@ -137,124 +138,179 @@ func TestIndex_NamespacesForDistro(t *testing.T) {
...
 			distro: newDistro(t, osDistro.Alpine, "3.15.4", []string{"alpine"}),
 			namespaces: []*distro.Namespace{
 				distro.NewNamespace("alpine", osDistro.Alpine, "3.15"),
 			},
 		},
 		{
+			name:   "alpine minor version matches minor version namespace",
 			distro: newDistro(t, osDistro.Alpine, "3.16", []string{}),
 			namespaces: []*distro.Namespace{
 				distro.NewNamespace("alpine", osDistro.Alpine, "3.16"),
```
luhring, line 156: This test case should expect the "edge" namespace, not "3.16"

grype/db/v5/namespace/index.go
```diff
@@ -12,6 +13,8 @@ import (
...
+
```
luhring, line 17: For **lines 100-103** _(GitHub isn't letting me comment on those lines for some reason, sorry!)_:

If the Alpine version string is something like `3.16.0-r4`, we'll fall into the conditional case in these lines and potentially return the wrong namespace. The version string `3.16.0-r4` should map to `alpine:edge`, but instead the code will see that `len(versionSegments) == 3` and that the `alpine:3.16` namespace is present in the index, so I believe it will incorrectly return `alpine:3.16` in that scenario.

luhring: It might be useful to add a test case to catch this scenario

luhring (commented):
This is getting close! Added some comments around a theme - attempting to clarify that the `x.y.z` rule means literally 3 digits, and anything else should map to the edge data. Hope this is helpful 😃

grype/db/v5/namespace/index.go
spiffcs: ^ Ahh yea - I thought that `x.y.z` would also include `x.y` as a derivative. I'll update this to be more strict and add a case for the r4

spiffcs:
Thanks for the follow-up @luhring and the catch about the minor vs patch namespace selection:
- I added a case for `rc4`
- I updated `alpine` to be a priority and checked first with the regex you suggested so that we don't fall into the more general logic of the other distros

luhring:
Awesome! This behavior looks correct to me 👍

kzantow (approved):

Issue #887, opened 2022-08-23
Show all vulnerabilities, even suppressed 


tafli:
**What would you like to be added**:
We're using Grype together with Syft in our CI Pipeline to scan built images. We also using a suppression file to mitigate problem. 

However I haven't found a way to show all unfiltered vulnerabilities of a scan. So I like to see even the suppressed vulnerabilities in the output list. Maybe a `(suppressed)` should be added after the name.

**Why is this needed**:
When using Grype in our CI-Pipeline with suppression, we don't see what is suppressed, without looking at the suppression file.

freedom-isnotanarchy:
(1) Good enhancement. 
(2) I wonder if this (below), would be preferable ?
`NAME                            INSTALLED      FIXED-IN                                  TYPE          VULNERABILITY        SEVERITY
geronimo-activation_1.1_spec    1.1                                                      java-archive  CVE-2011-5034        High
geronimo-activation_1.1_spec    1.1                                                      java-archive  CVE-2008-0732        Low
geronimo-javamail_1.4_mail      1.8.4                                                    java-archive  CVE-2011-5034        High
geronimo-javamail_1.4_mail      1.8.4                                                    java-archive  CVE-2008-0732        Low`

kzantow:
This sounds like a great addition! We would like the default behavior to remain the same but we absolutely accept PRs for things like this!

vimalpatel19:
Hi grype maintainers! I would like to contribute and assist with this.

kzantow:
@vimalpatel19 that would be great! I'd suggest maybe we think about adding `(suppressed)` to the `SEVERITY` column as opposed to installed or fixed-in.

vimalpatel19:
@kzantow Got it! Before I begin working on this, I just wanted to confirm the context for this: 

Instead of not adding the filtered out vulnerabilities to the output list, we will instead include them and append `-(suppressed)` to the value in the `SEVERITY` column for those filtered out vulnerabilities when the user passes in a new command line flag such as `-a` or `--all`, or something along those lines. Does that sound right?

kzantow:
@vimalpatel19 yeah, something like that... maybe the flag could be `--show-suppressed` or `--include-suppressed`?

freedom-isnotanarchy:
I see a certain symmetry below, with what Grype has today:
@vimalpatel19  ,  @kzantow 
-----
```
--include-suppressed
--exclude   stringArray
[...]

PR #966, opened 2022-10-24
Implemented new CLI flag: --show-suppressed

vimalpatel19:
Implemented new CLI flag (--include-suppressed) that allows for ignored/suppressed matches to be displayed in the table output format when that flag is provided.

Closes: #887

### DIFF
--- a/grype/presenter/config.go
+++ b/grype/presenter/config.go
@@ -11,13 +11,14 @@ import (
 
 // Config is the presenter domain's configuration data structure.
 type Config struct {
-	format           format
-	templateFilePath string
+	format            format
+	templateFilePath  string
+	includeSuppressed bool
 }
 
 // ValidatedConfig returns a new, validated presenter.Config. If a valid Config cannot be created using the given input,
 // an error is returned.
-func ValidatedConfig(output, outputTemplateFile string) (Config, error) {
+func ValidatedConfig(inclSuppressed bool, output, outputTemplateFile string) (Config, error) {
 	format := parse(output)
 
 	if format == unknownFormat {
@@ -58,6 +59,7 @@ func ValidatedConfig(output, outputTemplateFile string) (Config, error) {
 	}
 
 	return Config{
-		format: format,
+		format:            format,
+		includeSuppressed: inclSuppressed,
 	}, nil
 }
--- a/grype/presenter/config.go
+++ b/grype/presenter/config.go
@@ -11,14 +11,14 @@ import (
 
 // Config is the presenter domain's configuration data structure.
 type Config struct {
-	format            format
-	templateFilePath  string
-	includeSuppressed bool
+	format           format
+	templateFilePath string
+	showSuppressed   bool
 }
 
 // ValidatedConfig returns a new, validated presenter.Config. If a valid Config cannot be created using the given input,
 // an error is returned.
-func ValidatedConfig(inclSuppressed bool, output, outputTemplateFile string) (Config, error) {
+func ValidatedConfig(output, outputTemplateFile string, showSuppressed bool) (Config, error) {
 	format := parse(output)
 
 	if format == unknownFormat {
@@ -59,7 +59,7 @@ func ValidatedConfig(inclSuppressed bool, output, outputTemplateFile string) (Co
 	}
 
 	return Config{
-		format:            format,
-		includeSuppressed: inclSuppressed,
+		format:         format,
+		showSuppressed: showSuppressed,
 	}, nil
 }
--- a/grype/presenter/table/presenter.go
+++ b/grype/presenter/table/presenter.go
@@ -15,7 +15,7 @@ import (
 )
 
 const (
-	AppendSuppressed = "-(suppressed)"
+	appendSuppressed = " (suppressed)"
 )
 
 // Presenter is a generic struct for holding fields needed for reporting
@@ -24,17 +24,15 @@ type Presenter struct {
 	ignoredMatches   []match.IgnoredMatch
 	packages         []pkg.Package
 	metadataProvider vulnerability.MetadataProvider
-	inclSuppressed   bool
 }
 
 // NewPresenter is a *Presenter constructor
-func NewPresenter(results match.Matches, ignoredMatches []match.IgnoredMatch, packages []pkg.Package, metadataProvider vulnerability.MetadataProvider, inclSuppressed bool) *Presenter {
+func NewPresenter(results match.Matches, packages []pkg.Package, metadataProvider vulnerability.MetadataProvider, ignoredMatches []match.IgnoredMatch) *Presenter {
 	return &Presenter{
 		results:          results,
 		ignoredMatches:   ignoredMatches,
 		packages:         packages,
 		metadataProvider: metadataProvider,
-		inclSuppressed:   inclSuppressed,
 	}
 }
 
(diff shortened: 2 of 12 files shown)


kzantow (commented):
I've had a cursory look through this change but I have a question: do these suppressed vulnerabilities need to be included in the JSON or just the table view? If it's just the latter, maybe a better approach would be to not modify the data structure, but just print out both normal and suppressed vulns together?

freedom-isnotanarchy:
I can't seem to find the Feature Request, for this.  I know I cared ... and a search does not find that Feature Request.
Might you happen to have that URL, handy ?

kzantow:
@freedom-isnotanarchy it's linked in the description and on the right in the Development section

vimalpatel19:
@kzantow the suppressed vulnerabilities should be displayed in the JSON and template output formats. The question I had was do we want to also display the suppressed vulnerabilities in the just the table format, or do we want to attempt to display them in the rest of the output formats as well (such as cyclonedx, sarif, etc.)?

 If it is just for the table format, I can proceed as you suggested and not modify Match data structure. Otherwise if we need to include suppressed vulnerabilities for each of the remaining formats that don't currently do that, I think it might be better to keep the modified Match data structure as opposed to passing in an additional parameter (containing the list of suppressed vulnerabilities) to the NewPresenter method for each of the output formats here: https://github.com/anchore/grype/blob/main/grype/presenter/presenter.go.

freedom-isnotanarchy:
> @freedom-isnotanarchy it's linked in the description and on the right in the Development section
pr_linked_to_issue

kzantow:
@vimalpatel19 these are already included in JSON output, I'm pretty sure, though in a different spot, so I think we should just add these to the table output when the flag is set. 👍

vimalpatel19:
Yes, the JSON output format does have both the matching and suppressed vulnerabilities in the output. I will proceed with just the table output format with the approach you had discussed previously above.

vimalpatel19:
@kzantow can you review again when you get a please chance? Pushed updates based on your suggestions earlier.

grype/presenter/config.go
```diff
@@ -11,13 +11,14 @@ import (
...
-	templateFilePath string
+	format            format
+	templateFilePath  string
+	includeSuppressed bool
 }
 
 // ValidatedConfig returns a new, validated presenter.Config. If a valid Config cannot be created using the given input,
 // an error is returned.
-func ValidatedConfig(output, outputTemplateFile string) (Config, error) {
+func ValidatedConfig(inclSuppressed bool, output, outputTemplateFile string) (Config, error) {
```
kzantow, line 21: nit: I'd add this flag as the last parameter and write it out fully as `includeSuppressed`

grype/presenter/table/presenter.go
```diff
@@ -14,19 +14,27 @@ import (
...
 	results          match.Matches
+	ignoredMatches   []match.IgnoredMatch
 	packages         []pkg.Package
 	metadataProvider vulnerability.MetadataProvider
+	inclSuppressed   bool
 }
 
 // NewPresenter is a *Presenter constructor
-func NewPresenter(results match.Matches, packages []pkg.Package, metadataProvider vulnerability.MetadataProvider) *Presenter {
+func NewPresenter(results match.Matches, ignoredMatches []match.IgnoredMatch, packages []pkg.Package, metadataProvider vulnerability.MetadataProvider, inclSuppressed bool) *Presenter {
```
kzantow, line 31: nit: we should probably leave `results match.Matches, packages []pkg.Package, metadataProvider vulnerability.MetadataProvider` as the first 3 parameters, since the other 2 are related and would make a bit more sense to be grouped together.

I wonder if we could simplify this a little so we just pass in `ignoredMatches []match.IgnoredMatch` if `presenterConfig.includeSuppressed` is true so we don't need to track the boolean to display them in multiple places. It would change [this line](https://github.com/anchore/grype/pull/966/files#diff-e12805a26e8031bd6f10e235003454ad4ebd52a6af6c78c740df26193e41d28dR30) to something like:

```go
if presenterConfig.includeSuppressed {
    return table.NewPresenter(matches, packages, metadataProvider, ignoredMatches)
}
return table.NewPresenter(matches, packages, metadataProvider, nil)
```

grype/presenter/table/presenter.go
```diff
@@ -35,27 +43,26 @@ func (pres *Presenter) Present(output io.Writer) error {
...
 		}
+		rows = append(rows, row)
+	}
 
-		if metadata != nil {
-			severity = metadata.Severity
-		}
+	// Generate rows for suppressed vulnerabilities
+	if pres.inclSuppressed {
+		for _, m := range pres.ignoredMatches {
```
kzantow, line 58: ... continuing from the previous thought of simplifying the presenter to just include the `ignoredMatches`, the `for _, m := range pres.ignoredMatches {` would just automatically skip if it's `nil`

grype/presenter/table/presenter.go
```diff
@@ -14,19 +14,27 @@ import (
 	"github.com/anchore/grype/grype/vulnerability"
 )
 
+const (
+	AppendSuppressed = "-(suppressed)"
```
kzantow, line 18: This isn't used outside this package, is it? In which case, it should not be exported (in other words, lowercase; `appendSuppressed`)

Maybe we should change this to ` (Suppressed)`? 

So output would look like: `Low (Suppressed)`?

...or lowercased, is also probably fine, `Low (suppressed)`, but probably replace the dash with a space.

grype/presenter/table/presenter.go
```diff
@@ -113,3 +120,30 @@ func removeDuplicateRows(items [][]string) [][]string {
 	}
 	return result
 }
+
+func createRow(m match.Match, suppressed bool, metadataProvider vulnerability.MetadataProvider) ([]string, error) {
```
kzantow, line 124: For consistency, I'd probably add `suppressed bool` as the last parameter, but I wonder if this would be better as `severitySuffix string`, and pass in the `AppendSuppressed` variable.

kzantow (commented):
Sorry for the delay here, I left some more specific feedback, but a couple general comments:

* I feel the parameters dealing with showing suppressed vulnerabilities should be the last parameters in every function call for consistency, especially since it's the most recent featured added
* after pondering this a bit more, I think `--show-suppressed` is probably a better name for this parameter, as `--include-suppressed` does not affect failure codes and such, which seems a bit misleading

Happy to hear thoughts on this!

grype/presenter/table/presenter.go
vimalpatel19: Sounds good, I can make those updates

grype/presenter/table/presenter.go
vimalpatel19: Yes, I can make the additional parameter the last one and change it from being a boolean to the suffix string.

freedom-isnotanarchy: May I ask, is the expected output, to look like this ?
```
NAME                            INSTALLED            FIXED-IN   TYPE          VULNERABILITY   SEVERITY
geronimo-activation_1.1_spec    1.1 suppressed)     -          java-archive  CVE-2011-5034   High
geronimo-activation_1.1_spec    1.1 (suppressed)     2.0        java-archive  CVE-2008-0732   Low
geronimo-javamail_1.4_mail      1.8.4 (suppressed)   _          java-archive  CVE-2011-5034   High
geronimo-javamail_1.4_mail      1.8.4 (suppressed)   _          java-archive  CVE-2008-0732   Low
```

kzantow: No, this should be appended to the severity at the end.

kzantow (approved):
Thanks for the contribution @vimalpatel19, this looks great!

kzantow:
Hi @vimalpatel19 -- it looks like there's just one tiny lint issue: https://github.com/anchore/grype/actions/runs/3370373110/jobs/5592852609#step:8:20 

I'd be fine with just adding a `//nolint:funlen` at the top of the function for this one. To validate it fixes the problem you can run `make lint` locally.

vimalpatel19:
@kzantow addressed the lint issue for function length.

PR #978, opened 2022-11-03
chore: update syft to v0.60.3

spiffcs:
## Summary
Consume https://github.com/anchore/syft/pull/1312 as the latest syft.

Current syft v0.60.2 has a small bug regarding PURL generation that makes some matching inconsistent for `TestCompareSBOMInputToLibResults`

The test images found for that integration will produce different results if grype is updated to consume `v0.60.2`.

🚧 Merge only when v0.60.3 has been released 🚧 

Note: the logger has not been fully refactored. Only the syft wrapper has been updated for compatibility. A follow-up PR will fully port grype over to the newest logger separately.


cmd/cmd.go
```diff
@@ -73,9 +76,16 @@ func initLogging() {
 	grype.SetLogger(logWrapper)
 
 	// add a structured field to all loggers of dependencies
-	syft.SetLogger(&logger.LogrusNestedLogger{
-		Logger: logWrapper.Logger.WithField("from-lib", "syft"),
-	})
+	syftLoggerCfg := logrus.Config{
+		EnableConsole: enableConsole,
+		FileLocation:  appConfig.Log.FileLocation,
```
kzantow, line 81: Is there a possibility having 2 loggers with the same file will end up overwriting or failing to open the file?

cmd/cmd.go
```diff
@@ -11,6 +11,8 @@ import (
 	"github.com/spf13/viper"
 	"github.com/wagoodman/go-partybus"
 
+	iface "github.com/anchore/go-logger"
```
kzantow, line 14: nit: maybe something with `log` or `logger` in the name? like `loggerAPI`? to make it a tiny bit more clear below

kzantow (approved):
Only question about having 2 different loggers with the same file reference

cmd/cmd.go
spiffcs: Removed until we can get a PR that get's both syft and grype onto the same logger

cmd/cmd.go
spiffcs: Updated to `anchoreLogger` since `log`, `logger`, and other names were taken

spiffcs: Updated to anchoreLogger since log, logger, and other names were taken

PR #988, opened 2022-11-10
Update token permissions to be read-only

### LINKED ISSUE #984: Update token permissions to be read-only
As part of the OSSF Scorecard, we have identified a few areas to improve. See the initial comment [here](https://github.com/anchore/grype/pull/929#issuecomment-1295379392).

We should update token permissions to be read-only in every place we can.

joycebrum:
Closes #984 

I've made the changes (which were quite simples) to set to each workflow the default permission to read-only, any questions let me know.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -14,6 +14,9 @@ on:
   schedule:
     - cron: '0 14 * * 4'
 
+permissions:
+  contents: read
+
 jobs:
   analyze:
     name: Analyze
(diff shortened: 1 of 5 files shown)


spiffcs (approved):

joycebrum:
> Thanks so much @joycebrum! I'm going to check out the new Codeql documentation since the action is doing a `PUT` and is now `403` when running on merge. I'll follow up and link this PR.

Sure, problably it would need a permission to write in one of its steps since now the default is read-only, let me know if I can be of any help

PR #989, opened 2022-11-11
Add SECURITY.md

### LINKED ISSUE #982: Add SECURITY.md
As part of the OSSF Scorecard, we have identified a few areas to improve. See the initial comment [here](https://github.com/anchore/grype/pull/929#issuecomment-1295379392).

We should add a SECURITY.md security policy document.

joycebrum:
Closes #982 

I've created the document following a template, we should personalize with an email or the best way for your team to receive security vulnerability reports.

Also I've defined it as if the only version supported with security updates would always be the latest ones, but if you have a list or range of versions you give that kind of support we can use the "table of content" template.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -100,3 +100,8 @@ $ make integration
 ## Document your changes
 
 When proposed changes are modifying user-facing functionality or output, it is expected the PR will include updates to the documentation as well.
+
+
+## Security Vulnerabilities
+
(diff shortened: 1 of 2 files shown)


spiffcs:
Thanks for the extra effort in defining which releases are supported! I think that `latest` is a good start considering the size of our team, but I'll add it to the list of things to harden before we do `v1.0.0` as far as our support window plan

spiffcs:
Email Added - 👍  :shipit:

PR #1024, opened 2022-11-30
fix: Exclude binary packages that have overlap by file ownership relationship

kzantow:
This PR excludes binary packages found which are identified as children in an `ownership-by-file-overlap` relationship -- e.g. a binary file that was installed with APK.


test/quality/.python-version
```diff
@@ -0,0 +1 @@
+3.10.7
```
kzantow, line 1: I'm not convinced this is a good thing to check in -- what if the user has `3.10.6` or some other functional `3.10` version, just not `3.10.7`?

PR #1038, opened 2022-12-13
feat: swap custom cyclone-dx model for cyclone-dx library

### LINKED ISSUE #796: OWASP dependency track is not listing vulnerabilities (cyclone dx format) from grype , syft is working however
**What happened**:
OWASP dependency track is not listing vulnerabilities (cyclone dx format) from grype , syft is working however . Grype cyclonedx sbom only listing components. 

List vulnerabilties correctly so that various dashboard tools like depenendcy track can enlist vulnerabilties. just syft generated SBOM (cyclonedx format. )
genrate SBOM cyclonedx from both of the tools (syft and grype) and play them in depenedncy track and you will see the issue
use it for OWASP DVWA project. generate sbom for entire all layers of docker image. 
syft and grype xml generation is not consistent. 

**Environment**: MAC, docker compose, OWASP DVWA
- Output of `grype version`: xml
- OS (e.g: `cat /etc/os-release` or similar): MAC OS

### LINKED ISSUE #951: Include Syft's cyclonedx component properties in Grype output
Syft's cyclonedx output was improved in https://github.com/anchore/syft/pull/710. 
I would like Grype to offer the same level of detail as Syft.

**Why is this needed**:
The current cyclonedx or vex outputs don't contain full information to aid in triage. I need to cross-reference grype cyclonedx against syft json to find additional metadata (e.g. file path containing a go module, which syft encodes as `properties`).

**Additional context**:
Syft component encoding:
https://github.com/anchore/syft/blob/main/syft/formats/common/cyclonedxhelpers/component.go#L30

Grype:
https://github.com/anchore/grype/blob/main/grype/presenter/cyclonedx/document.go#L48
https://github.com/anchore/grype/blob/main/grype/presenter/cyclonedxvex/document.go#L67

spiffcs:
## Summary

Grype currently produces CYCLONE-DX SBOM that are not compliant with the cyclone-dx tooling libraries. Rather than write the logic in two places, this PR moves grype to use syft's formatting functions as a library to produce valid CYCLONE-DX SBOM components along with the discovered vulnerabilities. 

For more context on why this is an issue see: 
Fixes #796
Fixes #951 

Testing note: `image-simple` was unused under the cyclone-dx test-fixures so was removed in favor of the snapshot presenter testing.

### Breaking Changes
- Grype no longer makes a distinction between cyclonedx-schema and the cyclonedx-vex-schema. See [here](https://github.com/CycloneDX/specification/tree/master/schema) for schema definitions that only make a distinction between the json and xml schema.
- Given the above, format options for grype have been simplified. Grype now does `cyclonedx` and `cyclonedx-xml`. The options for `embedded-cyclonedx-vex-json` and `embedded-cyclonedx-vex-xml` have been removed.
- Grype no longer outputs CYCLONE-DX 1.2. Vulnerabilities are no longer nested under components as a consequence of this change.
- Updated schema definitions have been added to `schema/cyclonedx/`. These have been sourced from the above schema link

Comparison chart:
```
OLD									New
embedded-cyclonedx-vex-json ---> 	cyclonedx
embedded-cyclonedx-vex-xml--->		cyclonedx-xml

Format changes
cyclonedx				  ----> 	now outputs 1.4 json ---- was 1.2 XML
```
### To consider
Take a look at the two following outputs. The version of grype is `0.54.0`
```
grype -o embedded-cyclonedx-vex-json alpine:latest > vuln_old.json 
```

This command uses this branch:
```
go run main.go -o cyclonedx alpine:latest > vuln.json
```

The above commands should both output `cyclonedx-1.4-json` documents:

Note: in the above example we're seeing binary cataloger overlaps with package cataloger matching so that's probably a PR we want in separately before the next release so we can match on the correct package.

Notice how alpine shows `1.35.0-r29` vs binary cataloger which is showing `1.35.0`, the latter of which is vulnerable.

### Highlights from the above output comparrisons
- Metadata component now has a bomref
- component data contains expanded set of fields including but not limited to `publisher, description, properties, and references`
- Vulnerabilities now come with a bom ref field
- Vulnerabilities now come with a reference field
- Empty analysis and properties fields have been dropped


grype/presenter/cyclonedx/test-fixtures/snapshot/TestCycloneDxPresenterDir.golden
```diff
@@ -1,83 +1,85 @@
...
-      <name>/some/path</name>
-      <version></version>
-    </component>
   </metadata>
   <components>
-    <component type="library">
+    <component bom-ref="76bd1479d016ce8f" type="library">
       <name>package-1</name>
       <version>1.1.1</version>
-      <v:vulnerabilities>
```
spiffcs, line 21: Vulnerabilities are no longer embedded in components for 1.4 -- see latest schema definitions in PR description

grype/presenter/cyclonedx/vulnerability.go
```diff
@@ -5,141 +5,164 @@ import (
 	"strconv"
```
spiffcs, line 5: @spiffcs review this file one more time to double check nothing was removed

grype/presenter/cyclonedx/presenter.go
```diff
@@ -16,38 +22,49 @@ type Presenter struct {
...
 	return &Presenter{
-		results:          results,
-		packages:         packages,
-		metadataProvider: metadataProvider,
-		srcMetadata:      srcMetadata,
+		results:          pb.Matches,
+		packages:         pb.Packages,
+		metadataProvider: pb.MetadataProvider,
+		srcMetadata:      pb.Context.Source,
+		sbom:             pb.SBOM,
```
kzantow, line 36: nit: if we're going to use a `PresenterBundle` should the formats just have a reference to it instead of copying all/most of the properties individually? (same comment for all presenters, which might just reduce a bit of code)

grype/presenter/presenter.go
```diff
@@ -3,17 +3,14 @@ package presenter
 import (
 	"io"
 
-	cdx "github.com/CycloneDX/cyclonedx-go"
+	cyclonedxLib "github.com/CycloneDX/cyclonedx-go"
```
kzantow, line 6: nit: if we're refactoring this anyway, it might be nice to move this reference into the `github.com/anchore/grype/grype/presenter/cyclonedx` package and instead expose 2 functions from there: `cyclonedx. NewJSONPresenter` and `cyclonedx.NewXMLPresenter` or similar

kzantow (approved):
LGTM 👍 and the MEGA-⚓ of the day for -5k LoC!

spiffcs:
@wagoodman do you want me to invert the new output options here?

`cyclonedx` --> returns `xml`
`cyclonedx-json` --> returns `json` <-- New json option

Should we keep the `embedded-xx` options and deprecate them?

grype/presenter/cyclonedx/presenter.go
spiffcs: I've made a note to refactor this in its own separate PR so it's clear that this is an optimization and not part of the general upgrade to the syft library.

cmd/root.go
```diff
@@ -378,9 +385,20 @@ func startWorker(userInput string, failOnSeverity *vulnerability.Severity) <-cha
 			}
 		}
 
+		pb := models.PresenterBundle{
```
wagoodman, line 388: nit: is `models.PresenterConfig` taken? (that would be a more consistent name)

grype/presenter/format.go
```diff
@@ -9,10 +9,9 @@ const (
 	jsonFormat      format = "json"
 	tableFormat     format = "table"
 	cycloneDXFormat format = "cyclonedx"
+	cycloneDXXML    format = "cyclonedx-xml"
 	sarifFormat     format = "sarif"
 	templateFormat  format = "template"
-	embeddedVEXJSON format = "embedded-cyclonedx-vex-json"
```
wagoodman, line 14: I think we should keep these names (allow users to still input them) and add a deprecation warning with some hints on the right version to use. We should also open an issue to remove these options and add it under a grype 1.0 milestone.

grype/presenter/format.go
```diff
@@ -9,10 +9,9 @@ const (
 	jsonFormat      format = "json"
 	tableFormat     format = "table"
 	cycloneDXFormat format = "cyclonedx"
```
wagoodman, line 11: I don't think we should make a breaking change on the user surface area yet. That is, `cyclonedx` should still produce the XML format and probably add a `cyclonedx-json` as well as `cyclonedx-xml` (as done here). Then we can add deprecation warnings around `cyclonedx` usage mentioning that the behavior will change at grype 1.0 and make the swap then. (which means we should add an issue and capture under a grype 1.0 milestone)

grype/presenter/format.go
spiffcs: Names have been kept and added as part of a deprecation scheme

grype/presenter/format.go
spiffcs: +1 -- I like being v1.0 at the moment because we can make choices like this while doing a graceful deprecation of things that didn't work. I'll follow the above and we'll put it on the `TODO` pile when we go to release v1.0

PR #1046, opened 2022-12-23
fix: orient by cve merging

kzantow:
This PR corrects an issue when orienting results by CVE that may result in related vulnerabilities and CPEs being lost in output during the match merge process. This was also the cause of an intermittent test failure, which _should_ now be resolved.

### DIFF
--- a/grype/match/match.go
+++ b/grype/match/match.go
@@ -85,6 +85,10 @@ func (m *Match) Merge(other Match) error {
 
 	// retain all unique CPEs for consistent output
 	m.Vulnerability.CPEs = cpe.Merge(m.Vulnerability.CPEs, other.Vulnerability.CPEs)
+	if m.Vulnerability.CPEs == nil {
+		// ensure we always have a non-nil slice
+		m.Vulnerability.CPEs = []cpe.CPE{}
+	}
 
 	return nil
 }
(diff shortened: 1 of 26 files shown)


grype/vulnerability_matcher_test.go
```diff
@@ -552,11 +553,17 @@ func TestVulnerabilityMatcher_FindMatches(t *testing.T) {
...
 							mustCPE("cpe:2.3:*:activerecord:activerecord:*:*:*:*:*:rails:*:*"),
 						},
 						PackageQualifiers: []qualifier.Qualifier{},
 						Advisories:        []vulnerability.Advisory{},
+						RelatedVulnerabilities: []vulnerability.Reference{
+							{
+								ID:        "GHSA-2014-fake-3",
+								Namespace: "github:language:ruby",
+							},
+						},
```
kzantow, line 566: This is actually one of the non-CPE changes with this PR: additional fields will be merged.

grype/match/match.go
```diff
@@ -42,17 +44,56 @@ func (m *Match) Merge(other Match) error {
 		return ErrCannotMerge
 	}
 
+	// there are cases related vulnerabilities are synthetic, for example when
```
kzantow, line 47: The changes below here are the main gist of this PR unrelated to the CPE refactoring

wagoodman (approved):
thanks for taking the time to also clean up the `must(pkg.NewCPE...` calls that have been piling up 🙌

PR #1062, opened 2023-01-20
Add protobuf FPs to default ignore list

### LINKED ISSUE #558: False positive CVE-2015-5237 for protobuf-go
**What happened**:
Our go.mod dependency of `google.golang.org/protobuf@v1.27.1` (a.k.a https://github.com/protocolbuffers/protobuf-go) is detected by syft/grype as `cpe:2.3:a:google:protobuf:v1.27.1:*:*:*:*:*:*:*` and grype matches CVE-2015-5237 based on `cpe:2.3:a:google:protobuf:*:*:*:*:*:*:*:* <= 3.1.0` but 1.27.1 is the latest version.

The CVE links to https://github.com/protocolbuffers/protobuf/issues/760 which is a different project (protobuf language spec vs the golang runtime).

Distinguish between google.golang.org/protobuf and github.com/protocolbuffers/protobuf

```
mkdir -p /tmp/gomod/
echo '
module foo
go 1.17
require google.golang.org/protobuf v1.27.1' > /tmp/gomod/go.mod
grype dir:/tmp/gomod
```

https://github.com/golang/vuln/blob/master/triaged-cve-list marks CVE-2015-5237 as a false postive - but I'm not sure what package/version that refers to.

**Environment**:
- Output of `grype version`:
```
Application:          grype
Version:              0.27.3
Syft Version:         v0.33.0
BuildDate:            2021-12-16T15:11:16Z
GitCommit:            4f964c4ee26ad01a80b8bcffb6bf23c0afb71d09
GitTreeState:         clean
GoVersion:            go1.16.12
Compiler:             gc
Supported DB Schema:  3
```
MacOS Big Sur 11.6.2

luhring:
Fixes #558 

Let me know if there's a better way to cancel these out in the code. I saw this method used for some log4j FPs so I continued that approach here.

One unfortunate drawback to this approach is the UI becomes a bit confusing:

```console
$ go run . anchore/grype   
 ✔ Parsed image            
 ✔ Cataloged packages      [212 packages]
 ✔ Scanned image           [2 vulnerabilities]
No vulnerabilities found
```

"2 vulnerabilities", and then "no vulnerabilities". Curious if anyone has thoughts to make this more straightforward.

### DIFF
--- a/grype/match/explicit_ignores.go
+++ b/grype/match/explicit_ignores.go
@@ -16,9 +16,16 @@ func init() {
 	var explicitIgnores = []ignoreValues{
 		// Based on https://github.com/anchore/grype/issues/552, which includes a reference to the
 		// https://github.com/mergebase/log4j-samples collection, we want to filter these explicitly:
-		{"java-archive",
-			[]string{"CVE-2021-44228", "CVE-2021-45046", "GHSA-jfh8-c2jp-5v3q", "GHSA-7rjr-3q55-vv33"},
-			[]string{"log4j-api", "log4j-slf4j-impl", "log4j-to-slf4j", "log4j-1.2-api", "log4j-detector", "log4j-over-slf4j", "slf4j-log4j12"},
+		{
+			typ:             "java-archive",
+			vulnerabilities: []string{"CVE-2021-44228", "CVE-2021-45046", "GHSA-jfh8-c2jp-5v3q", "GHSA-7rjr-3q55-vv33"},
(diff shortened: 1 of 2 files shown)


spiffcs (approved):

spiffcs:
Thanks @luhring! I think this is pretty straightforward. If this list keeps getting bigger we can revisit it. Before we merge I'll wait for others from the open source team to comment on if we can add a commit to reduce the UI confusion here

kzantow (approved):
Also LGTM -- the count mismatch is an existing problem, so doesn't need to be solved here, IMO.

luhring:
Thanks @spiffcs and @kzantow! Good to merge from my end then 😃

PR #1104, opened 2023-01-28
Fix for broken install script

### LINKED ISSUE #1102: Unable to install Grype
**What happened**:

Starting a few hours ago, it was noticed that installation of Grype is no longer working with the install.sh script.

No errors

`curl -sSfL https://raw.githubusercontent.com/anchore/grype/main/install.sh | sh -s -- -b /usr/local/bin
[info] checking github for the current release tag 
[error] unable to find tag='' 
[error] do not specify a version or select a valid version from https://github.com/anchore/grype/releases `

**Environment**:
- Output of `grype version`:
- OS (e.g: `cat /etc/os-release` or similar): 22.04.1 LTS (Jammy Jellyfish)

jdolitsky:
Resolves #1102 

See also https://github.com/orgs/community/discussions/45590
If the above issue is resolved, this change is not needed

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -10,7 +10,7 @@ INSTALL_SH_BASE_URL=https://raw.githubusercontent.com/${OWNER}/${PROJECT_NAME}
 PROGRAM_ARGS=$@
 
 # do not change the name of this parameter (this must always be backwards compatible)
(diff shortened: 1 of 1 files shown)


spiffcs:
@jdolitsky looks like this has been reverted:
https://github.com/orgs/community/discussions/45590#discussioncomment-4802887

I'm going to close this PR for now considering we're back to the correct json response state.

If there are other considerations or you think we should still merge let me know and I can reopen and review.

PR #1100, opened 2023-01-27
chore: prune cosign dependency for grype builds

spiffcs:
## Grype dependency prune

Increase grype build times by removing cosign library dependency

Users will now have to decode SBOM from attestations as part of a separate command and pipe the result into grype for vulnerability scanning

### Example workflow with this change:
```
COSIGN_EXPERIMENTAL=1 cosign verify-attestation caphill4/java-spdx-tools:latest | jq -r .payload | base64 --decode | jq -r .predicate.Data | grype

Verification for caphill4/java-spdx-tools:latest --
The following checks were performed on each of these signatures:
  - The cosign claims were validated
  - Existence of the claims in the transparency log was verified offline
  - Any certificates were verified against the Fulcio roots.
Certificate subject:  christopher.phillips@anchore.com
Certificate issuer URL:  https://github.com/login/oauth
NAME           INSTALLED     FIXED-IN  TYPE          VULNERABILITY        SEVERITY
commons-text   1.9                     java-archive  CVE-2022-42889       Critical
commons-text   1.9           1.10.0    java-archive  GHSA-599f-7c49-w659  Critical
guava          28.2-android            java-archive  CVE-2020-8908        Low
guava          28.2-android            java-archive  GHSA-5mg8-w23w-74h3  Low
protobuf-java  3.21.4                  java-archive  CVE-2022-3171        High
protobuf-java  3.21.4                  java-archive  CVE-2022-3509        High
protobuf-java  3.21.4                  java-archive  CVE-2022-3510        High
protobuf-java  3.21.4        3.21.7    java-archive  GHSA-4gg5-vx3j-xwc7  High
protobuf-java  3.21.4        3.21.7    java-archive  GHSA-g5ww-5jh7-63cx  High
protobuf-java  3.21.4        3.21.7    java-archive  GHSA-h4h5-3hr4-j3g2  Medium
snakeyaml      1.32                    java-archive  GHSA-mjmj-j48q-9wg2  High
woodstox-core  6.3.1         6.4.0     java-archive  GHSA-3f7h-mf4q-vrm4  Low
```

### Boyscout change
- Updated Golang version 18.x ==> 19.x
Signed-off-by: Christopher Phillips <christopher.phillips@anchore.com>

PR #1110, opened 2023-01-31
feat: upgrade golang to 1.19

bradleyjones:
Go 1.18 will become EOL with the upcoming release of Go 1.20


kzantow:
FWIW these are also bumped here: https://github.com/anchore/grype/pull/1100

bradleyjones:
Thanks @kzantow hadn’t seen that will close this one

PR #1147, opened 2023-02-28
Release Pipeline Upgrade

spiffcs:
## WIP ##


.github/actions/bootstrap/action.yaml
```diff
@@ -0,0 +1,99 @@
+name: "Bootstrap"
```
spiffcs, line 1: This uses the "Bootstrap" action that we also have built in syft. Note: this adds a few other actions including configuring python tooling an caches.

.github/workflows/release.yaml
```diff
@@ -161,46 +161,3 @@ jobs:
         with:
           name: artifacts
           path: dist/**/*
-
-  release-docker-assets:
```
spiffcs, line 165: This is now done in a single release

.github/workflows/validations.yaml
```diff
@@ -1,15 +1,11 @@
...
+
 on:
   workflow_dispatch:
+  pull_request:
   push:
     branches:
       - main
-  pull_request:
-
-env:
```
spiffcs, line 9: Moved to bootstrap action

.github/workflows/validations.yaml
```diff
@@ -20,138 +16,39 @@ jobs:
...
-        if: steps.tool-cache.outputs.cache-hit != 'true' || steps.go-cache.outputs.cache-hit != 'true'
-        run: make bootstrap
-
-      - name: Bootstrap CI environment dependencies
-        run: make ci-bootstrap
+      - uses: actions/checkout@v3
 
+      - name: Bootstrap environment
+        uses: ./.github/actions/bootstrap
+      # TODO: any test fixtures
```
wagoodman, line 36: anything needed here?

wagoodman:
we'll need to additionally verify if the correct CI secrets are available for grype as well (quill takes slightly different values than the old versions).

.github/workflows/validations.yaml
spiffcs: Nothing needed to be updated here - added this as a placeholder in case I had missed something when composing the bootstrap step - all good!

wagoodman (approved):

PR #1165, opened 2023-03-08
fix: correct APK CPE version comparison logic

westonsteimel:
Previously, the -r{buildindex} suffix of APK package versions were treated as pre-release versions per the fuzzy matcher logic; however, these should be treated as equivalent to the release version for the purposes of collecting CPE-based matches for APK packages.

We may want to make a similar change in syft to generate cleaner CPE versions for APK packages, but making the change in grype corrects behaviour for previously-generated SBOMs as well.

### DIFF
--- a/grype/matcher/apk/matcher.go
+++ b/grype/matcher/apk/matcher.go
@@ -2,7 +2,6 @@ package apk
 
 import (
 	"fmt"
-	"strings"
 
 	"github.com/anchore/grype/grype/distro"
 	"github.com/anchore/grype/grype/match"
@@ -46,8 +45,7 @@ func (m *Matcher) Match(store vulnerability.Provider, d *distro.Distro, p pkg.Pa
 
 func (m *Matcher) cpeMatchesWithoutSecDBFixes(store vulnerability.Provider, d *distro.Distro, p pkg.Package) ([]match.Match, error) {
 	// find CPE-indexed vulnerability matches specific to the given package name and version
-	cpePackage := cpeComparablePackage(p)
-	cpeMatches, err := search.ByPackageCPE(store, cpePackage, m.Type())
+	cpeMatches, err := search.ByPackageCPE(store, p, m.Type())
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


westonsteimel:
An example illustrating this is

```
grype registry:cgr.dev/chainguard/ruby:latest-20230308
 ✔ Parsed image
 ✔ Cataloged packages      [114 packages]
 ✔ Scanning image...       [12 vulnerabilities]
   ├── 1 critical, 6 high, 4 medium, 1 low, 0 negligible
   └── 0 fixed
NAME      INSTALLED  FIXED-IN  TYPE  VULNERABILITY   SEVERITY
find      0.1.1                gem   CVE-2015-10065  Critical
find      0.1.1                gem   CVE-2020-24550  Medium
readline  0.0.3                gem   CVE-2014-2524   Low
ruby-3.2  3.2.0-r6             apk   CVE-2023-22795  High
```

CVE-2023-22795 is fixed for ruby 3.2.0 and therefore shouldn't show up, but since the -r6 is causing it to be treated as a pre-release, it gets pulled in.

grype/matcher/apk/matcher.go
```diff
@@ -45,7 +46,8 @@ func (m *Matcher) Match(store vulnerability.Provider, d *distro.Distro, p pkg.Pa
 
 func (m *Matcher) cpeMatchesWithoutSecDBFixes(store vulnerability.Provider, d *distro.Distro, p pkg.Package) ([]match.Match, error) {
 	// find CPE-indexed vulnerability matches specific to the given package name and version
-	cpeMatches, err := search.ByPackageCPE(store, p, m.Type())
+	cpePackage := cpeComparablePackage(p)
+	cpeMatches, err := search.ByPackageCPE(store, cpePackage, m.Type())
```
wagoodman, line 50: we should be surfacing matches with the original package, the match details should have the necessary (mutated) details. This way we surface the vuln, the original package [from the SBOM as we understand it], and the match details showing what values were used in the search.

wagoodman (approved):

PR #1174, opened 2023-03-13
chore: bump vuln match quality dataset

westonsteimel:


kzantow (approved):

wagoodman:
I don't understand why the gitlab image is causing an issue here:
```
	* failed to construct source from user input "docker.io/gitlab/gitlab-ce:15.6.1-ce.0@sha256:04d4219d5dfb3acccc9997e50477c8d24b371387a95857e1ea8fc779e17a716c": could not fetch image "docker.io/gitlab/gitlab-ce:15.6.1-ce.0@sha256:04d4219d5dfb3acccc9997e50477c8d24b371387a95857e1ea8fc779e17a716c": unable to use DockerDaemon source: unable to save image to tar: write /tmp/stereoscope-592605836/docker-daemon-image-195173514/image.tar: no space left on device
```

I'm seeing that the cache is being cleaned up for each image quickly. Something that I think is odd though is why syft is running at all, it should be leveraging the vulnerability-match-label SBOM cache from it's OCI registry.

PR #1180, opened 2023-03-14
feat: disable CPE-based matching by default for javascript

westonsteimel:
CPE-based matching leads to a large amount of false positives, and disabling it by default for the javascript matcher seems like a reasonable first step since GitHub owns NPM and already has coverage of NPM-related security advisories in the Github Security Advisory dataset


kzantow (approved):
Still 👍

PR #1188, opened 2023-03-20
fix: by-cpe pivot by vuln metadata rather than vulnerability record

### LINKED ISSUE #1185: --by-cve takes a noticeable amount more time to complete
Using the `--by-cve` flag doesn't filter out the duplicates (and also takes 10x as long, from 15s to 2m38s). Both CVE-2020-9547 and GHSA-q93h-jc49-78gg are still in the output and only the GHSA-q93h-jc49-78gg one has a fixed version.
```
./grype version
Application:          grype
Version:              0.59.1
Syft Version:         v0.74.1
BuildDate:            2023-03-09T14:57:12Z
GitCommit:            29b646568901d1ef48a528cf35f67f3cead49c9f
GitDescription:       v0.59.1
Platform:             linux/amd64
GoVersion:            go1.19.6
Compiler:             gc
Supported DB Schema:  5
```

_Originally posted by @JipSogeti in https://github.com/anchore/grype/issues/236#issuecomment-1470015661_

westonsteimel:
This changes the `--by-cve` logic to only reorient the vulnerability metadata around NVD data rather than attempting to get the corresponding vulnerability table record.  This will address performance concerns since their was no index on `id, namespace` on the vulnerabilities table, but there is on the `vulnerability_metadata` table.  This also preserves the fixed in information from the actual match rather than losing it since the NVD data has no fix info.  It also correctly returns the CVE identifier in cases where there is more than one vulnerability record for a given `id, namespace` since that is the primary key on the `vulnerability_metadata` table.

Fixes #1185

### DIFF
--- a/grype/vulnerability_matcher.go
+++ b/grype/vulnerability_matcher.go
@@ -109,20 +109,13 @@ func (m *VulnerabilityMatcher) normalizeByCVE(match match.Match) match.Match {
 
 	ref := effectiveCVERecordRefs[0]
 
-	upstreamVulnRecords, err := m.Store.Get(ref.ID, ref.Namespace)
+	upstreamMetadata, err := m.Store.GetMetadata(ref.ID, ref.Namespace)
 	if err != nil {
-		log.Warnf("unable to fetch effective CVE record for id=%q namespace=%q : %v", ref.ID, ref.Namespace, err)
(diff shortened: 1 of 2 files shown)


kzantow (approved):
LGTM

Issue #1204, opened 2023-03-29
Need a way to always prefer registry over local Docker when scanning an image


luhring:
**What would you like to be added**:

It'd be great to have an easy way to avoid ever accidentally scanning a stale version of a container image, specifically because Grype retrieved the image from Docker instead of directly from the registry. Scanning stale images results in wasted time, pointless escalations/firedrills, etc.

Short of changing the default behavior in the tool itself, it'd be great to have a persistent fix available to the user, like setting a value in a config file on disk.

**Why is this needed**:

Today's behavior is a common footgun for folks. They intuitively type a command like `grype alpine:3.16`, or `grype my-registry.io/my-company/my-image`, and they don't realize until much later that what they've scanned is not reflective of the latest state of the image.

This happens primarily when scanning images that the user didn't build themselves. So any time you scan an official Docker Hub image, or an open source project's image, this is a mistake waiting to happen. Even for an organization's own image, it's increasingly common that images are built by CI instead of locally — in which case, defaulting to retrieving the image _from Docker_ just means introducing the risk that you're scanning a stale image.

**Additional context**:

IIRC, there's no strong reason why today Grype prefers Docker over the registry — I believe it was simply that the developers (myself included) added support for Docker _first_, and for the registry _second_.

This might be implicit — but it'd probably be a good UX for whatever solution is applied to Grype for this ticket to also be applied to _Syft_.

cc: @spiffcs — [link to community Slack discussion](https://anchorecommunity.slack.com/archives/C027JE5M345/p1680111360264169)

PR #1215, opened 2023-04-03
feat: default image source config option

spiffcs:
Follow up to https://github.com/anchore/syft/pull/1703

## Add config option to allow user to select the default image source location

https://github.com/anchore/grype/issues/1204 surfaces the need for allowing a user to express a preference over the `default-image-pull-source` to be used when building an SBOM for vulnerability scanning.

This adds a config option into grype to consume the new syft behavior.

There is also a small boy scout change that includes the `name` flag (it was part of the function signature being edited for pull source) which was introduced into syft recently. `name` is used to provide users an override for naming the source. `name` and `default-image-pull-source` have been added into the syft provider config.

### DIFF
--- a/README.md
+++ b/README.md
@@ -555,6 +555,7 @@ check-for-app-update: true
 
 # allows users to specify which image source should be used to generate the sbom
 # valid values are: registry, docker, podman
+# same as GRYPE_DEFAULT_IMAGE_PULL_SOURCE env var
 default-image-pull-source: ""
 
 # same as --name; set the name of the target being analyzed
--- a/README.md
+++ b/README.md
@@ -140,6 +140,12 @@ sbom:path/to/syft.json                 read Syft JSON from path on disk
 registry:yourrepo/yourimage:tag        pull image directly from a registry (no container runtime required)
 ```
 
+If an image source is not provided and cannot be detected from the given reference it is assumed the image should be pulled from the Docker daemon.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -547,6 +553,10 @@ Configuration options (example values are the default):
 # same as GRYPE_CHECK_FOR_APP_UPDATE env var
 check-for-app-update: true
 
+# allows users to specify which image source should be used to generate the sbom
+# valid values are: registry, docker, podman
```
kzantow, line 557: would be good t add the `# same as GRYPE_DEFAULT_IMAGE_PULL_SOURCE env var`

grype/pkg/syft_provider.go
```diff
@@ -11,7 +11,7 @@ func syftProvider(userInput string, config ProviderConfig) ([]Package, Context,
 		return nil, Context{}, nil, errDoesNotProvide
 	}
 
-	sourceInput, err := source.ParseInput(userInput, config.Platform)
+	sourceInput, err := source.ParseInputWithName(userInput, config.Platform, "", config.DefaultImagePullSource)
```
kzantow, line 14: is the name supposed to be included here instead of `""`?

spiffcs: Ohh no! Good catch - dropped that commit earlier to keep them small and didn't add it back in

PR #1247, opened 2023-04-20
Fail when grype cant check for db update

### LINKED ISSUE #310: "No vulnerability database update available" when actually the check for an update was unsuccessful
**What happened**:

Grype shows "No vulnerability database update available" when it _hasn't actually_ checked to see if a database update is available.

If Grype is unable (for any reason) to check to see if a database update is available, it should report the failure to the user and exit non-zero.

1. For the sake of a clean environment, start a new container in which we'll run Grype: `docker run --rm -it ubuntu:latest bash`
2. Install `curl` if it's not already present: `apt-get update && apt-get install -y curl`
3. Install `grype`: `curl -sSfL https://raw.githubusercontent.com/anchore/grype/main/install.sh | sh -s -- -b /usr/local/bin`
4. Edit `/etc/resolv.conf` so that the system relies on an incorrect address for its DNS server. For example: `echo "nameserver 8.8.8.9" > /etc/resolv.conf`
5. Tell Grype to check for a database update: `grype db update`

Grype will show `No vulnerability database update available`, and its exit code will be `0`.

But in fact, **_Grype doesn't even have a local database yet_**. The same behavior can be witnessed if Grype has an _outdated_ database, too.

There may exist other network-related failure modes that produce the same symptom, but this is the mode that I encountered today.

Also, **this impacts the security of the user** — because this output leads them to behave their Grype scans have considered the latest vulnerability data when this might not really be the case.

**Environment**:
- Output of `grype version`:

```
Application:          grype
Version:              0.11.0
BuildDate:            2021-04-22T21:51:14Z
GitCommit:            0a46e54aae837135f73d2ff18a84061189ee6f28
GitTreeState:         clean
Platform:             linux/amd64
GoVersion:            go1.16.3
Compiler:             gc
Supported DB Schema:  2
```

```
NAME="Ubuntu"
VERSION="20.04.2 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.2 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal
```

(I've also seen this issue on **macOS** 11.2.3)

shanedell:
Fail when grype cant check for db update

Closes #310

**Before change when running**:

```bash
docker run -v $PWD:/grype -w /grype -d --name grype-test --rm -it golang:1.19 tail -f /dev/null
docker exec -w /grype -it grype-test bash -c "go install"
docker exec -it grype-test bash -c 'echo "nameserver 8.8.8.9" > /etc/resolv.conf'
docker exec -w /grype -it grype-test bash -c "go run main.go db update"
```

It would output:

```bash
No vulnerability database update available
```

and exit with 0.

**After update when running**:

```bash
docker run -v $PWD:/grype -w /grype -d --name grype-test --rm -it golang:1.19 tail -f /dev/null
docker exec -w /grype -it grype-test bash -c "go install"
docker exec -it grype-test bash -c 'echo "nameserver 8.8.8.9" > /etc/resolv.conf'
docker exec -w /grype -it grype-test bash -c "go run main.go db update"
```

the output is

```bash
 ✔ Vulnerability DB        [checking for update]

No vulnerability database update available
1 error occurred:
        * unable to update vulnerability database: unable to update vulnerability database: unable to download listing: Get "https://toolbox-data.anchore.io/grype/databases/listing.json": dial tcp: lookup toolbox-data.anchore.io on 8.8.8.9:53: read udp 172.17.0.2:39128->8.8.8.9:53: i/o timeout

exit status 1
```

with exit code 1.


grype/db/curator.go
```diff
@@ -141,9 +141,9 @@ func (c *Curator) Update() (bool, error) {
 
 	updateAvailable, metadata, updateEntry, err := c.IsUpdateAvailable()
 	if err != nil {
-		// we want to continue if possible even if we can't check for an update
 		log.Warnf("unable to check for vulnerability database update")
 		log.Debugf("check for vulnerability update failed: %+v", err)
+		return false, fmt.Errorf("unable to update vulnerability database: %w", err)
```
kzantow, line 146: I think we may want to only return an error if there is _no database to use_ after this check

Shanedell: But wouldn't you still want return an error if unable to update the database? For example if I had an old outdated database locally and tried to update it but it failed to update, say from a dns error like in this example. If the error is not thrown the command would return that `No vulnerability database update available` which would be misleading. This would cause the user to believe their outdated database is up to date.

Shanedell: @kzantow Or would something like this work?

```go
stage.Current = "no update available"
updateAvailable, metadata, updateEntry, err := c.IsUpdateAvailable()

if err != nil {
    log.Warnf("unable to check for vulnerability database update")
    log.Debugf("check for vulnerability update failed: %+v", err)
    stage.Current = fmt.Sprintf("unable to update vulnerability database: %w", err)
}

...

if stage.Current != "no update available" {
    return false, errors.New(state.Current)
} else {
    return false, nil
}
```

?

sprathod369:
The "**Fail when grype cant check for db update**" issue continues to be sporadic and have been occuring across Grype versions. Recently, I'm facing a similar issue when tried using Grype v0.61.1. On my Ubuntu 18x VM it worked smoothly but it does not work on my RHEL 8x VM. All my VMs are on the same network.

Also when exploring Kubeclarity that integrates with Grype, a similar issue blocks me when I try to setup kubeclarity-grype-server. Below are my K8s logs.

```
root@xxxx:~/KubeClarity# kubectl logs pod/kubeclarity-kubeclarity-grype-server-7d969b865-9lxks --namespace kubeclarity
time="2023-05-25T11:12:48Z" level=fatal msg="Failed to start scanner: failed to load DB: unable to check for vulnerability database update: unable to download listing: Get \"https://toolbox-data.anchore.io/grype/databases/listing.json\": tls: failed to verify certificate: x509: certificate signed by unknown authority" func=main.run file="/build/grype-server/cmd/grype-server/main.go:53"
```

kzantow:
Hi all -- what if we modified the `grype db update` command to fail if it fails to download the database? For users where this is an issue, you could use 2 steps: 
```
grype db update
grype ...
```
Would this be an acceptable solution where grype itself will still be able to function if there is a slightly old db downloaded (it already complains if the db is too old)?

wagoodman:
Agreed with @kzantow we should not have an exit code of 1 in this case -- grype should work offline without error. We have a separate validator to not use stale DBs incase you're offline for too long (which is configurable). The issue is more about ensuring the UI doesn't lie to the user and say "there is no DB update available" when it didn't check. Instead the UI should really be updated to be clear that it was not able to check for an update and gracefully continue processing.

@shanedell did you want to update this PR to account for the UI update? if not, we can also close this PR instead.

willmurphyscode:
Wanted to add that https://github.com/anchore/grype/pull/1463 will make it so that `grype db check` will exit zero if the DB is known to be current, 100 if the DB is known to be outdated, and 1 if it can't check?

Maybe `grype db check && grype ...` will do what you need after that PR is merged?

cmd/grype/cli/options/database.go
```diff
@@ -41,6 +42,7 @@ func DefaultDatabase(id clio.Identification) Database {
 		ValidateAge: true,
 		// After this period (5 days) the db data is considered stale
 		MaxAllowedBuiltAge:     defaultMaxDBAge,
+		RequireUpdateCheck:     true,
```
kzantow, line 45: If we default this to `false`, this PR as written would essentially have the same behavior as grype today with what I believe is a useful change that `grype db update` will fail if it is unable to determine if the db is up to date because the listing file failed.

cmd/grype/cli/commands/db_update.go
```diff
@@ -26,7 +26,9 @@ func DBUpdate(app clio.Application) *cobra.Command {
 }
 
 func runDBUpdate(opts options.Database) error {
-	dbCurator, err := db.NewCurator(opts.ToCuratorConfig())
+	cfg := opts.ToCuratorConfig()
+	cfg.RequireUpdateCheck = true // always consider an update check failure a db update failure
```
willmurphyscode, line 30: Do we want to move this to PreRunE so that the config is printed with this change in debug logging?

kzantow: It turns out the PreRunE didn't work. I think, though, that any `grype db <command>` which happens to run the db update should, by default, consider failure to check for an update to be an update failure, so moved the default configuration to the appropriate place. Note that this does _not_ currently apply to running the normal grype scan, only the `grype db` operations. This also allows users to disable this behavior if they _really_ want to, using an explicitly configured `GRYPE_DB_REQUIRE_UPDATE_CHECK=false`...

wagoodman (approved):

willmurphyscode (approved):

Issue #1170, opened 2023-03-10
Grype|Include Timestamp and Image Name to Reports


mike-19:
**What would you like to be added**:

Include data like the timestamp on when grype scan was run and also the image name to the report information. 

**Why is this needed**:
This is required to provide structural data for visualization as well evidence for the scan report is associated with the particular image and was executed on a particular date (sort it to the latest) - Evidence needs to be provided for perusals

spiffcs:
👋 Thanks for the issue @mike-19 - when you run `grype -o json <IMAGE>` I think you can see the image name under `source.target.userInput`

If the user input is something different then `source.target.tags` might also give you what you need for the image report.

As for timestamp I think that's an easy add.

Are there other formats you're looking for this update in? Were you specifically referencing a report/format type when filing the issue?

mike-19:
@spiffcs  Thanks for the follow-up; based on your comments, I went through the JSON, and I was able to get the image name 
`timestamp` of when the scan was run with timezone would be helpful 
I am not looking for any other formats

Shanedell:
Should this be closed? Looked like it was added in https://github.com/anchore/grype/commit/04c0bc1bcde0b28600acd720447243307aca93b0

kzantow:
@Shanedell good catch -- looks like this was completed with #1249

PR #1249, opened 2023-04-21
feat: add timestamp to json output (#1170)

jneate:
Adds timestamp into the JSON Output, timestamp is a new field inside Descriptor.

Unit tests were a little fiddly because the timestamp changes so first check the value is not nil and is in RFC 3339 format, then essentially use the normal assert.JSONEq to check the rest of the document(s).

### DIFF
--- a/grype/presenter/models/document.go
+++ b/grype/presenter/models/document.go
@@ -2,6 +2,7 @@ package models
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/anchore/grype/grype/match"
 	"github.com/anchore/grype/grype/pkg"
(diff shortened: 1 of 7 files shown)


grype/presenter/models/document.go
```diff
@@ -75,6 +81,7 @@ func NewDocument(packages []pkg.Package, context pkg.Context, matches match.Matc
 			Version:               version.FromBuild().Version,
 			Configuration:         appConfig,
 			VulnerabilityDBStatus: dbStatus,
+			Timestamp:             string(timestamp),
```
kzantow, line 84: If I could suggest an alternate method of testing: add a simple test that a valid timestamp is included from this `NewDocument` function, in `document_test.go`.

For the `presenter_test.go` tests, just add a `redact` function that removes the timestamp, something like this: 
```go
func redact(s string) string {
	return regexp.MustCompile(`"timestamp":\s*"[^"]+"`).
		ReplaceAllString(s, `"timestamp":""`)
}
```
and then the only change you need to make in the `presenter_test.go` is before the `assert.Equal`, add:
```go
actual = redact(actual)
```
... or something equivalent

jneate: Yeah that makes way more sense

kzantow (approved):
LGTM 👍

PR #1252, opened 2023-04-23
docs: add "cyclonedx-json" to output formats

HNKNTA:
The output format is already supported:
`supported formats are: [json table cyclonedx cyclonedx-json sarif template]`
CycloneDX 1.4 is also there:
https://github.com/anchore/grype/issues/591

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,10 +247,11 @@ The output format for Grype is configurable as well:
 grype <image> -o <format>
 ```
 
-Where the `format`s available are:
+Where the formats available are:
 
 - `table`: A columnar summary (default).
-- `cyclonedx`: An XML report conforming to the [CycloneDX 1.2](https://cyclonedx.org/) specification.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -247,10 +247,11 @@ The output format for Grype is configurable as well:
 grype <image> -o <format>
 ```
 
-Where the `format`s available are:
+Where the `formats` available are:
```
kzantow, line 250: nit: this should probably not be code-quoted at all

HNKNTA: Well, the `format` is mentioned as an arg, so it is worth to highlight it somehow. But the existing way it looks a bit weird for my taste, the `formats` looks better but make it less consistent IMO. 
So, should I get rid of quotes at all?

kzantow: I'd say to just get rid of the code quotes, as it's not exactly code or leave it as is -- it's just used in that example as _format_, not _formats_. Also, there's no `--format` option (it's `--output` / `-o`).

kzantow (approved):

PR #1274, opened 2023-05-05
docs: add config flag to configuration section (#1271)

### LINKED ISSUE #1271: Document command line flag for config file location
An extra custom place to search for `.grype.yaml` config file. it can be an environment variable for the path.

**Why is this needed**:
right now, the only places are static and in a ci/cd environment sometimes we want to make sure that the config file we specify is the one being used either it's in the shared directory or somewhere else maybe even a different config file for different steps (for a different ignore list of cve) etc

**Additional context**:
right now, the places grype searches for its configurations are. 
`
.grype.yaml 
.grype/config.yaml 
~/.grype.yaml 
<XDG_CONFIG_HOME>/grype/config.yaml
`

I am not sure if <XDG_CONFIG_HOME> is configurable or an env variable that we can change because if yes then that's exactly what we need.

jneate:
This resolves #1271

### DIFF
--- a/README.md
+++ b/README.md
@@ -540,12 +540,18 @@ They will also not be dependent on a docker daemon, (or some other runtime softw
 
 ## Configuration
 
-Configuration search paths:
+Default configuration search paths:
 
 - `.grype.yaml`
 - `.grype/config.yaml`
 - `~/.grype.yaml`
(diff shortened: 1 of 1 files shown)


kzantow (approved):
LGTM thanks @jneate

PR #1275, opened 2023-05-06
feat: include file location in table output (#1199)

### LINKED ISSUE #1199: Share which file is vulnerable when reporting an intra-file vulnerability in the default table output

In the default output format, for image scans (or really, any scan of a target that contains multiple files), when reporting a vulnerability for a dependency inside a file (e.g. a module compiled into a Go binary), **show which file has the vulnerability**.

**Why is this needed**:

Without this feature, the output row for these kinds of vulnerabilities is not very helpful.

If I'm trying to ensure my image has _as few vulnerabilities as possible_, and I'm using Grype to help me, I want to get to the point where I can **_do something_** about every given vulnerability as quickly as possible. Any factor that gets in my way of doing this is noticeable friction added to my process of getting the job done.

For a distro package, (e.g. `apk` type), remediation is obvious. From anywhere in the image (or Dockerfile, etc.) I can run `apk add -u pkgName` and I'm done.

But for cases like Go binaries, JARs, etc. — there is a missing layer of hierarchy in Grype's table output. I see the vulnerable package, and I know what image I scanned, but I'm missing a crucial piece of information: the file itself. So I need to re-run the scan with a more helpful output.

Here's an example:

```console
$ grype -q quay.io/cilium/hubble-ui-backend:v0.10.0
NAME                      INSTALLED                           FIXED-IN                           TYPE       VULNERABILITY        SEVERITY
github.com/cilium/cilium  v1.12.4                                                                go-module  CVE-2023-27593       Medium
github.com/cilium/cilium  v1.12.4                                                                go-module  CVE-2023-27594       High
github.com/cilium/cilium  v1.12.4                             1.12.8                             go-module  GHSA-4hc4-pgfx-3mrx  Medium
github.com/cilium/cilium  v1.12.4                             1.12.8                             go-module  GHSA-8fg8-jh2h-f2hc  Medium
golang.org/x/net          v0.4.0                              0.7.0                              go-module  GHSA-vvpx-j8f3-3w6h  High
golang.org/x/sys          v0.0.0-20210902050250-f475640dd07b  0.0.0-20220412211240-33da011f77ad  go-module  GHSA-p782-xgp4-8hr8  Medium
```

Which Go binary (or binaries, plural?) has the vulnerability? I have no idea. In the very best case, the image only has a single Go binary, and I have this knowledge in advance. But in all other cases, this Grype scan brought me no closer to a solution for those vulnerabilities.

**Additional context**:

The "default" piece of this ask might be more important than one would think. In practice, for better or for worse, I've seen lots of people sharing screenshots (or console copy/paste like the above example) of **_the table view_**. Of course, in some cases, folks are persisting a more machine-readable format, like the native JSON output or SARIF. But still, lots of times this isn't the case. And it would be great for folks on the receiving end of these snippets to more readily understand what Grype found.

cc: @kzantow

jneate:
This could hopefully resolve #1199 

I wasn't 100% sure which types to exclude so I started with Apk,Deb&Rpm - placed the boolean flag into a map in case they need to be driven by config in the future.

Regarding the tests, I just thought it'd be nice to have a non-distro package in the output so added a 3rd match in models_helpers and updated the other output snapshots. I didn't think there was any gain adding a 4th package which is why the CreateRow test just recycles the values from other objects instead of being created from the helper.

Attached some example output below from various runs.

Scanning: openjdk:17.0.1-jdk-oracle
image

Scanning: quay.io/cilium/hubble-ui-backend:v0.10.0
image

Scanning Directory:
image

Scanning Directory:
image


spiffcs:
This one is also going to go in after we get the UI for grype updated with the new https://github.com/anchore/clio library

tgerla:
Hi @jneate, thanks for this. In an effort to keep the table view as concise as possible, we are thinking it might make sense to create a "table-locations.tmpl" template that includes the locations of the vulnerability. What do you think about that, instead of changing the default table view?

kzantow:
Although it's not exactly the same as the table output, a couple options to fairly easily get the location are:
1) a template (based on https://github.com/anchore/grype/blob/main/templates/table.tmpl, but modified)
or
2) `grype explain --id <vulnerability-id>`
Would either of these suffice to get the location information you're looking for easily?

Issue #1243, opened 2023-04-19
Sprig functions are not defined


anetanel:
**What happened**:
Attempting to create a template that uses Sprig fundtions such as "now", "date" and "env".
Getting error message such as:
` unable to parse template: template: csv.tpl:2: function "now" not defined`
**What you expected to happen**:
The template should work and create a report.
**How to reproduce it (as minimally and precisely as possible)**:
use the following template:
{{- range .Matches }}
{{ now | date "2006-01-02" }} {{ .Vulnerability.Severity }}
{{- end }}

**Anything else we need to know?**:
This works on Trivy ;)
**Environment**:
- Output of `grype version`:
Application:          grype
Version:              0.61.0
Syft Version:         v0.76.0
BuildDate:            2023-04-04T15:11:17Z
GitCommit:            d8c0c0805b59659c4d6e49d6806a0eba11bdc2ee
GitDescription:       v0.61.0
Platform:             darwin/amd64
GoVersion:            go1.19.7
Compiler:             gc
Supported DB Schema:  5
- OS (e.g: `cat /etc/os-release` or similar):
macOS 13.3.1 (Intel)

tgerla:
Hi @anetanel, thanks for your report. I've done some testing and I have reproduced this issue. It looks like perhaps we aren't making all of the Sprig functions available to template authors. I think the problem is here:

https://github.com/anchore/grype/blob/ae2fe4f0632cec102da60188c5ac56a75b39fd53/grype/presenter/template/presenter.go#L80

I think `sprig.HermeticTxtFuncMap` needs to be changed to `sprig.GenericFuncMap`. I will review with the team and we will update this issue when we are able. Thanks again!

kzantow:
I think we could update the templating function to have this, as @tgerla mentioned -- thanks for the report @anetanel

wagoodman:
We've decided to roll back this change, using the safer hermetic functions over the generic sprig functions (#1288 ). This means that `now`,  `date` and `env` will no longer be available as functions to call in a grype template. However, if you are looking for a timestamp this was recently added to the grype JSON output (thus should be accessible in a template) https://github.com/anchore/grype/pull/1249 .

I'm going to close this as not planned for now, but feel free to lay out more thoughts/opinions about this here!

wagoodman:
[...]

PR #1273, opened 2023-05-05
feat: add non-hermetic sprig functions (#1243)

jneate:
This resolves #1243

PR #1282, opened 2023-05-08
Allow generic sprig functions

willmurphyscode:
Previously, functions that could return different results at different times, such as `now`, were not allowed in sprig templates. Allow them.

Fixes https://github.com/anchore/grype/issues/1243


willmurphyscode:
Closing in favor of https://github.com/anchore/grype/pull/1273

PR #1284, opened 2023-05-09
Colorize severity in table output

### LINKED ISSUE #225: Colorize severity in table output
It would be nice to display each severity label in its own color in the table output. Trivy does that well and it makes visually parsing the output just a little bit faster, with just a quick glance. I'd suggest something like low = blue, medium = yellow, high = red, critical = bold red. Happy to contribute a PR if you think this sounds interesting!

shanedell:
Colorize severity in table output

- ~Create flag "--no-color" to allow disabling the color. By default its enabled.~ Responds to the `NO_COLOR` and `CLICOLOR_FORCE` environment variables.
- default output behavior for the table presenter is to highlight severity in its color:
  - Critical -> Bold Red
  - High -> Red
  - Medium -> Yellow
  - Low -> Green
  - Negligible -> Blue
  - Note: Golang doesn't have all colors available. Also, doesn't seem to be able use hex codes properly.

Closes #225

**Note**:

When you look at https://github.com/olekukonko/tablewriter/blob/master/table_with_color.go#L63 there are only so many colors available. Through testing I couldn't take the color hex codes from https://github.com/anchore/grype-vscode/blob/main/src/ui/react-components/FindingsList.tsx#L148-L162 and make them work in Golang. So I went with the below colors based on ones available:

- Critical -> Bold Red
- High -> Red
- Medium -> Yellow
- Low -> Green
- Negligible -> Blue
- Others -> Non-colored

Open to other suggestions if anyone is able to point me to other ways of coloring the tablewriter.


spiffcs (approved):

spiffcs:
Here is a quick sample output to show what this looks like in the new table output.

I think it's a great change and 👍 
Screenshot 2023-05-30 at 11 11 07 AM

cc @wagoodman for a final comment. I thought we might want to inverse so `--no-color` is default and users could add `--color` to stay consistent with our current default, but after running it a couple times to test and look at the output I like the color =)

shanedell:
@spiffcs I fixed the merge conflicts but did you also want me to incorporate what @wagoodman mentions with `termenv` before merging? Or would that be a separate issue?

shanedell:
@spiffcs I added `termenv` to the PR to check if color output should be enabled by default or not

wagoodman:
actually, we're about to move to bubbletea which might already respond to NO_COLOR env vars. I'll try it out and post back here. In the meantime let's hold off before adding a switch for this until we have an answer.

wagoodman:
@spiffcs indeed, but there is still one more case to cover. There is a related issue in syft that shows the case https://github.com/anchore/syft/issues/1893 .. here's the offending line https://github.com/anchore/go-logger/blob/d65da42248673d1596c5c144a974db4a10d6af60/adapter/logrus/logger.go#L47

spiffcs:
Added the blocked label so that we can rebase this work on top of some WIP regarding a refactor of the UI library

wagoodman:
@shanedell I've updated the branch with the latest from main and also removed the CLI flag since this should already be determined from the underlying lipgloss lib. I still need to verify the correctness, but this at least gets the branch updated.

wagoodman (approved):

PR #1296, opened 2023-05-15
io/ioutil is deprecated

testwill:


Makefile
```diff
@@ -14,7 +14,7 @@ GOLANGCILINT_VERSION := v1.52.2
 GOSIMPORTS_VERSION := v0.3.8
 BOUNCER_VERSION := v0.4.0
 CHRONICLE_VERSION := v0.6.0
-GORELEASER_VERSION := v1.18.2
+GORELEASER_VERSION := v1.17.2
```
kzantow, line 17: This should probably not be part of a change to `io/ioutil` (it is _downgrading_ the goreleaser version)

grype/db/v1/store/store_test.go
```diff
@@ -460,13 +459,13 @@ func TestStore_MergeVulnerabilityMetadata(t *testing.T) {
 
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
-			dbTempDir, err := ioutil.TempDir("", "grype-db-test-store")
+			dbTempDir, err := os.CreateTemp("", "grype-db-test-store")
```
kzantow, line 462: All of the changes from `ioutil.TempDir` should probably be: `os.MkdirTemp`, if they weren't in testing code.

However, since all of these changes are in test code, any `ioutil.TempDir` call should be changed to: `t.TempDir()`. This will also simplify the code because no error handling is required by that call.

For the `ioutil.TempFile` changes, these could easily be replaced with: `path.Join(t.TempDir(), <file-name>)`, e.g. `dbTempDir := path.Join(t.TempDir(), "grype-db-test-store")`

Additionally, by switching to `t.TempDir()`, all these files will be cleaned up automatically, so the `os.RemoveAll` calls can and should be removed.

kzantow (changes requested):

wagoodman (approved):

PR #1299, opened 2023-05-15
Fix reading syft json from stdin by redirect

devfbe:
I figured out that running `cat syft.json | grype` works but `grype < syft.json` does not work. This happens, because the IsPipedInput method only checks if stdin is a pipe which will be false if stdin is fed by a redirect.

The go idiomatic way to fix this is by just checking if the file produced by stat has a size > 0.

Implemented this check, that will recognize stdin by redirect, in the `IsPipedInput()` method. Renamed the method to `IsStdinPipeOrRedirect()`.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -283,14 +283,14 @@ func rootExec(_ *cobra.Command, args []string) error {
 }
 
 func isVerbose() (result bool) {
-	isPipedInput, err := internal.IsPipedInput()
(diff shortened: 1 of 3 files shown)


wagoodman (approved):
nice improvement! We used to only use the size as an indication of piping, which is wrong (plenty of race conditions there). But I see the difference with the use case you highlighted 👍

PR #1310, opened 2023-05-22
bump syft to pre-release of v0.81.0

wagoodman:
Bump syft version used to v0.81.0, which includes:
- new license shape (not carried through in grype json output yet)
- new R package cataloger (no vulnerability data to match against this yet)
- ...+ more fixes


spiffcs (approved):

spiffcs:
Code changes look good - I think you just need to run go mod tidy again so the SA checks pass =)

Issue #1053, opened 2023-01-06
grype is showing (suppressed) issues even though they aren't requested.


josesa-xx:
**What happened**:

when I run grype with only the option "--only-fixed" the output generated includes "(suppressed)" issues even though I'm not using the option "--show-suppressed".

**What you expected to happen**:
Results with "(suppressed)" should only be shown if explicitly using the option "--show-suppressed"

**How to reproduce it (as minimally and precisely as possible)**:

run `grype --only-fixed httpd:2.4.54-alpine`

Output I get
```
 ✔ Vulnerability DB        [no update available]
 ✔ Loaded image
 ✔ Parsed image
 ✔ Cataloged packages      [36 packages]
 ✔ Scanned image           [5 vulnerabilities]
NAME        INSTALLED  FIXED-IN   TYPE  VULNERABILITY   SEVERITY
libcrypto3  3.0.7-r0   3.0.7-r2   apk   CVE-2022-3996   High
libcurl     7.86.0-r1  7.87.0-r0  apk   CVE-2022-43551  Unknown
libcurl     7.86.0-r1  7.87.0-r0  apk   CVE-2022-43552  Unknown
libldap     2.6.3-r6              apk   CVE-2015-3276   Medium (suppressed)   <<< Problem
libssl3     3.0.7-r0   3.0.7-r2   apk   CVE-2022-3996   High
```

**Anything else we need to know?**:

**Environment**:
```
$ grype version
Application:          grype
Version:              0.55.0
Syft Version:         v0.65.0
BuildDate:            2023-01-04T21:53:44Z
GitCommit:            c559833c7e02aed827773adb4f9677cb398577be
GitDescription:       v0.55.0
Platform:             linux/amd64
GoVersion:            go1.18.9
Compiler:             gc
Supported DB Schema:  5
```

wagoodman:
Thanks for reporting! We've confirmed the behavior and will get a fix in soon for this.

miquella:
I was looking through the v0.54.0..v0.55.0 diff and noticed this change: https://github.com/anchore/grype/compare/v0.54.0..v0.55.0#diff-e12805a26e8031bd6f10e235003454ad4ebd52a6af6c78c740df26193e41d28dL29-L33

The original code was:
```go
	case tableFormat:
		if presenterConfig.showSuppressed {
			return table.NewPresenter(matches, packages, metadataProvider, ignoredMatches)
		}
		return table.NewPresenter(matches, packages, metadataProvider, nil)
[...]

PR #1322, opened 2023-05-29
Hide suppressed vulnerabilities when --show-suppressed is not given

jamestran201:
Closes https://github.com/anchore/grype/issues/1053

This PR fixes a bug where suppressed vulnerabilities are displayed even though the `--show-suppressed` flag is not provided. The fix is made by checking the value of the `showSuppressed` option before including the suppressed vulns in the result. This PR also adds tests to assert the behaviour with/without `--show-suppressed`.

### DIFF
--- a/grype/presenter/models/models_helpers.go
+++ b/grype/presenter/models/models_helpers.go
@@ -29,6 +29,17 @@ func GenerateAnalysis(t *testing.T, scheme syftSource.Scheme) (match.Matches, []
 	return matches, packages, context, NewMetadataMock(), nil, nil
 }
 
+func GenerateAnalysisWithIgnoredMatches(t *testing.T, scheme syftSource.Scheme) (match.Matches, []match.IgnoredMatch, []pkg.Package, pkg.Context, vulnerability.MetadataProvider, interface{}, interface{}) {
+	t.Helper()
+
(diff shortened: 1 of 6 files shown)


grype/presenter/models/models_helpers.go
```diff
@@ -29,6 +29,17 @@ func GenerateAnalysis(t *testing.T, scheme syftSource.Scheme) (match.Matches, []
 	return matches, packages, context, NewMetadataMock(), nil, nil
 }
 
+func GenerateAnalysisWithIgnoredMatches(t *testing.T, scheme syftSource.Scheme) (match.Matches, []match.IgnoredMatch, []pkg.Package, pkg.Context, vulnerability.MetadataProvider, interface{}, interface{}) {
+	t.Helper()
+
+	packages := generatePackages(t)
+	matches := generateMatches(t, packages[0], packages[0])
+	ignoredMatches := generateIgnoredMatches(t, packages[1])
```
jamestran201, line 37: `generatePackages()` returns 2 packages. I wanted to use one package for generating ignored matches, so I passed the first package twice to `generateMatches()`. This feels a bit strange, but I'm not sure how to change it.

wagoodman (approved):

PR #1328, opened 2023-06-01
feat: fix summary output (#1312)

### LINKED ISSUE #1312: The summary by severity is confusing
**What happened**:

Back in August, someone opened an issue asking Grype to show a summary of vulnerability matches by severity: https://github.com/anchore/grype/issues/877. This was a very logical idea, it's a great way to get an overall sense of a scan result.

The implementation ended up leveraging Grype's internal eventing system to "live update" the severity counts, but I think some tradeoffs were made that resulted in a user experience that raises more questions than answers.

It's now not uncommon to see output from Grype like this:

```text
 ✔ Loaded image
 ✔ Parsed image
 ✔ Cataloged packages      [404 packages]
 ✔ Scanning image...       [2 vulnerabilities]
   ├── 0 critical, 1 high, 1 medium, 0 low, 0 negligible
   └── 0 fixed

No vulnerabilities found
```

Since this UX change, I've personally been asked by confused users, "what is this trying to tell me?" or "what is this hiding?" at least **once a week**.

If Grype is deciding conclusively that a vulnerability **does not** match a package, it should never imply to the user that they might need to worry about that match.

This means two things:

1. The sum of severity counts should ALWYAS equal the count of vuln matches surfaces in the final output.
2. A vuln match should never be represented in the severity counts unless it's going to appear in the final output.

I mention the second point because it could be tempting to lean further into the eventing system, and have the severity counts decrement during Grype's filtering stage. I would caution against going this route — because that will raise even more questions about what Grype is doing (and possibly hiding). Users want this summary because they want to get a high level impression at a glance, not to dive into the inner workings of Grype's implementation.

My overall advice is to keep this simple, and that will go a long way for Grype's users.

jneate:
This resolves #1312 

Opening the PR slightly earlier than normal, just to get peoples views on the proposed output:

image

And then with supressed/ignored vulns

image

Happy to hear any thoughts on the above.

- TODO
I removed the log summary so need to add that back in once I find a nice way of handling it e.g.
image

Signed-off-by: James Neate <jamesmneate@gmail.com>

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -391,7 +391,7 @@ func startWorker(userInput string, failOnSeverity *vulnerability.Severity) <-cha
 		}
 
 		bus.Publish(partybus.Event{
-			Type:  event.VulnerabilityScanningFinished,
+			Type:  event.PresentationStarted,
 			Value: presenter.GetPresenter(presenterConfig, pb),
 		})
 	}()
--- a/grype/event/event.go
+++ b/grype/event/event.go
@@ -9,4 +9,5 @@ const (
 	VulnerabilityScanningFinished partybus.EventType = "grype-vulnerability-scanning-finished"
 	NonRootCommandFinished        partybus.EventType = "grype-non-root-command-finished"
 	DatabaseDiffingStarted        partybus.EventType = "grype-database-diffing-started"
+	PresentationStarted           partybus.EventType = "grype-presentation-started"
 )
--- a/grype/event/parsers/parsers.go
+++ b/grype/event/parsers/parsers.go
@@ -6,6 +6,7 @@ import (
 	"github.com/wagoodman/go-partybus"
 	"github.com/wagoodman/go-progress"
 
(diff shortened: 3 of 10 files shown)


kzantow:
Hi @jneate I didn't realize you had already put some work into fixing this issue and picked it up myself. I think there's a much simpler solution to this problem. Would it be a problem to close this PR in favor of a different one?

jneate:
I spoke with @wagoodman about it and this PR needed some refactoring anyway so if you've already opened another PR then go for it

Curious to see the PR if you can link? I think Alex was saying we'd keep the monitors and do the work when the monitor was completed as that's the architectural pattern the team had gone for, as opposed to emitting an "end" event

kzantow:
@jneate here's the alternate PR: https://github.com/anchore/grype/pull/1359

kzantow:
Since I think this is now fixed, I'm going to close this PR but thank you very much @jneate !!

PR #1339, opened 2023-06-06
feat: add package info to search by for all match details

westonsteimel:
This adds the package name and version that were searched by to the match details in grype output for all matchers.  Previously this detail was only surfaced when searching by distro.


kzantow (approved):
👍

PR #1341, opened 2023-06-07
feat: disable Java CPE matching by default

kzantow:
This PR disables CPE matching by default for Java.


wagoodman:
Superseded by https://github.com/anchore/grype/pull/1412

Issue #1316, opened 2023-05-25
A folder with go templates like in Trivy repo


YevheniiPokhvalii:
It would be nice to have a folder with go templates with examples, like Trivy has: https://github.com/aquasecurity/trivy/tree/main/contrib

Their templates are not compatible with Grype.
I need an XML JUnit output format and it's a pain to create it from scratch.

YevheniiPokhvalii:
I made a simple XML Go template that works with Grype. It might be useful to someone:

```
grype docker_image -o template -t grype-junit.tpl
```

```
<?xml version="1.0" ?>
<testsuites name="grype-junit">
{{- $failures := len $.Matches }}
    <testsuite tests="{{ $failures }}" failures="{{ $failures }}" name="{{ $.Distro.Name }}:{{ $.Distro.Version }}" errors="0" skipped="0">
        <properties>
            <property name="type" value="{{ $.Distro.Name }}"></property>
        </properties>
        {{- range .Matches }}
        <testcase classname="{{ .Artifact.Name }}-{{ .Artifact.Version }} ({{ .Artifact.Type }})" name="[{{ .Vulnerability.Severity }}] {{ .Vulnerability.ID }}">
            <failure message="{{ .Artifact.Name }}: {{ .Vulnerability.ID }}" type="description">{{ .Vulnerability.Description }} {{ .Artifact.CPEs }} {{ .Vulnerability.DataSource }} </failure>
        </testcase>
        {{- end }}
    </testsuite>
</testsuites>

```

spiffcs:
This sounds like a good idea! I'll throw it into our backlog for a team member to grab when they have cycles.

We also would love to have community PR that gets this effort started - 

The best kind of PR would be an example CSV or HTML template that we could use to help people get started

If someone contributes this please put your name at the top of the template so users can understand who to query if there are any future issues =)

spiffcs:
@YevheniiPokhvalii #1343 was merged - so we now have a folder where you can contribute your Junit template along with some starter documentation. Let me know if you need help getting it contributed!

anaynayak:
Thanks @YevheniiPokhvalii for sharing that snippet. Are you planning to contribute that template? Otherwise I don't mind helping and would love to see it integrated into grype.

kzantow:
Yeah, that looks like a great one to add to the sample templates here: https://github.com/anchore/grype/tree/main/templates -- I'd love to see you get credit for the contribution @YevheniiPokhvalii !

[...]

PR #1343, opened 2023-06-09
feat: add community template folder and new table template

### LINKED ISSUE #629: Expose Go template file that produces the table report
Go template files that Grype would use internally to generate reports.:

Currently if someone wants to compose a Go template using the guidelines from the README file, there is a lot of logic that needs to be reverse engineered form the Grype code if you want to get the same output as Grype's default table (columnar) summary report.:

I am composing stylized XML (for Jenkins) and HTML (for emails) template and the data I work during templating is far less valuable in its current form than the table summary report.:

jneate:
This is a start on #1316 - Unsure whether this fixes #629 or more work is needed to fix the issues.

I've included a template that mimics the table output of Grype, it's not great (see the README) and it can be optimised further by adding some helper functions into the presenter FuncMap but it works and it's a start.

image

Biggest pain is it's sorted by the CVE (I think) but like I said, it's a start.

Fixed-In is a list so I concatenated the fields in the template but I couldn't find an example with multiple fix versions so can't provide a shiny screenshot sorry

Potentially more documentation could go in the README about extracting the json models so people know what fields are available etc.

### DIFF
--- a/templates/README.md
+++ b/templates/README.md
@@ -1,0 +1,23 @@
+# Grype Templates
+
+This folder contains a set of "helper" go templates you can use for your own reports.
+
+Please feel free to extend and/or update the templates for your needs, be sure to contribute back into this folder any new templates!
+
+Current templates:
+
+<pre>
+.
+├── README.md
+└── table.tmpl
+</pre>
+
+## Table
+
+This template mimics the "default" table output of Grype, there are some drawbacks using the template vs the native output such as:
+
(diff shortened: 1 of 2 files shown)


spiffcs:
Thanks for the PR @jneate! This is a great start and I don't see any reason why we can't use this as the foundation and add more/improve on the template you added going forward. Thanks again for the documentation too!

PR #1345, opened 2023-06-09
feat: add ability to ignore packages on pom properties

### LINKED ISSUE #985: Ignoring Java test dependencies

Some way to ignore Java test dependencies like this vulnerable-legacy log4j:

```xml
<dependency>
    <groupId>log4j</groupId>
    <artifactId>log4j</artifactId>
    <version>1.2.17</version>
    <scope>test</scope>
</dependency>
```

**Why is this needed**:

Test libs are not packaged in final JAR/WAR file. Scope "provided" is not package too.

**Additional context**:

Grype reporting log4j test libs:

```
log4j                 1.2.17                       java-archive  CVE-2022-23307       High
log4j                 1.2.17                       java-archive  GHSA-2qrg-x229-3v8q  Critical
log4j                 1.2.17                       java-archive  CVE-2022-23305       Critical
log4j                 1.2.17                       java-archive  GHSA-w9p3-5cr8-m3jj  High
log4j                 1.2.17                       java-archive  CVE-2020-9493        Critical
log4j                 1.2.17                       java-archive  CVE-2022-23302       High
log4j                 1.2.17                       java-archive  CVE-2019-17571       Critical
log4j                 1.2.17                       java-archive  GHSA-65fg-84f6-3jq3  Critical
log4j                 1.2.17                       java-archive  GHSA-fp5r-v3w9-4333  High
```

jneate:
This resolves #985 

I was 50/50 about adding GroupID and ArtifactID, happy to remove artifact ID as you could use package-name in the package ignore rules but opened the PR anyway just in case it added value.

The artifact ID ignore rules are dependant on https://github.com/anchore/syft/pull/1870 as currently it's blank

Group ID filter
image

Scope filter
image

Pom used for testing:

```xml
<dependency>
        <groupId>log4j</groupId>
        <artifactId>log4j</artifactId>
        <version>1.2.17</version>
</dependency>		
<dependency>
	<groupId>org.springframework.boot</groupId>
	<artifactId>spring-boot-starter-webflux</artifactId>
	<scope>test</scope>
</dependency>
```

### DIFF
--- a/grype/match/ignore.go
+++ b/grype/match/ignore.go
@@ -2,6 +2,8 @@ package match
 
 import (
 	"github.com/bmatcuk/doublestar/v2"
+
+	"github.com/anchore/grype/grype/pkg"
 )
 
 // An IgnoredMatch is a vulnerability Match that has been ignored because one or more IgnoreRules applied to the match.
@@ -21,6 +23,7 @@ type IgnoreRule struct {
 	Namespace     string            `yaml:"namespace" json:"namespace" mapstructure:"namespace"`
 	FixState      string            `yaml:"fix-state" json:"fix-state" mapstructure:"fix-state"`
 	Package       IgnoreRulePackage `yaml:"package" json:"package" mapstructure:"package"`
+	Pom           IgnoreRulePom     `yaml:"pom" json:"pom" mapstructure:"pom"`
 }
 
 // IgnoreRulePackage describes the Package-specific fields that comprise the IgnoreRule.
@@ -30,6 +33,13 @@ type IgnoreRulePackage struct {
 	Language string `yaml:"language" json:"language" mapstructure:"language"`
 	Type     string `yaml:"type" json:"type" mapstructure:"type"`
 	Location string `yaml:"location" json:"location" mapstructure:"location"`
+}
+
+// IgnoreRulePom describes the Pom-specific fields that comprise the IgnoreRule.
+type IgnoreRulePom struct {
+	Scope      string `yaml:"scope" json:"scope" mapstructure:"scope"`
+	GroupID    string `yaml:"group-id" json:"group-id" mapstructure:"group-id"`
+	ArtifactID string `yaml:"artifact-id" json:"artifact-id" mapstructure:"artifact-id"`
 }
 
 // ApplyIgnoreRules iterates through the provided matches and, for each match,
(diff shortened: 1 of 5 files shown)


grype/match/ignore_test.go
```diff
@@ -70,19 +70,22 @@ var (
 		{
 			Vulnerability: vulnerability.Vulnerability{
 				ID:        "CVE-458",
-				Namespace: "ruby-vulns",
+				Namespace: "java-vulns",
```
kzantow, line 73: Has the change to this vulnerability test removed the ruby case? Unclear why it should be removed, should this be added back in as another test case?

grype/match/ignore.go
```diff
@@ -21,6 +23,7 @@ type IgnoreRule struct {
 	Namespace     string            `yaml:"namespace" json:"namespace" mapstructure:"namespace"`
 	FixState      string            `yaml:"fix-state" json:"fix-state" mapstructure:"fix-state"`
 	Package       IgnoreRulePackage `yaml:"package" json:"package" mapstructure:"package"`
+	Pom           IgnoreRulePom     `yaml:"pom" json:"pom" mapstructure:"pom"`
```
kzantow, line 26: I wonder if this should be nested under a Java ignore rule container? E.g. the configuration would be:
```yaml
  # ...or just by a single pom field:
  - java:
      pom:
        scope: test
```

kzantow (commented):
Thanks @jneate -- my main question is if the pom should be a top-level field on the IgnoreRule (I don't know, just posing the question)

grype/match/ignore.go
kzantow: @jneate the team talked about this PR a bit and had a few comments:

The first thing to note is 

The simplest change to get this to the finish line: since this is for a package rather than something else we believe this should be nested under the `IgnoreRulePackage`, probably for now as `java` (since the metadata is `pkg.JavaMetadata` rather than Maven POM specific), so the configuration would look something like:
```yaml
- package:
    java:
      scope: 'test'
      artifact-id: 'log4j'
      group-id: 'org.apache.logging.log4j'
```
... I think this would basically just involve moving the
```
Pom           IgnoreRulePom     `yaml:"pom" json:"pom" mapstructure:"pom"`
```
under the `IgnoreRulePackage`, as something more like:
```
Java           IgnoreRuleJava     `yaml:"java" json:"java" mapstructure:"java"`
```

--

But... I'd also mention there was another thought that we might like to make this more generic, where we could potentially specify any metadata information, something like:
```yaml
- package:
    metadata:
      - field : .PomScope
        value: test
      - field :  .PomArtifactID
        value: log4j
      - field :  .PomGroupID
        value: 'org.apache.logging.log4j'
```

Do you think either of these options would be something you'd like to do? If not, we could probably help at least get the first change done to get this to the finish line.

wagoodman:
This to me feels more like a matcher configuration and not additional fields on an ignore rule. The reason being is that matchers are already ecosystem specific and can handle cases where we're adding new attributes to consider for that entire ecosystem.

spiffcs:
@jneate thanks for the contribution and apologies that this got stale. I've rebased the changes here and will work on getting the commnets from @wagoodman and @kzantow addressed!

spiffcs:
We're back to passing on this PR - I'm now taking it in the direction suggested by @wagoodman where we update the matcher configuration rather than the ignore rules

PR #1346, opened 2023-06-10
feat(outputs): allow to set multiple outputs (#648)

### LINKED ISSUE #648: Write to multiple output files (like syft)
Unlike `syft`, `grype` does not seem to support writing to (multiple) output files:
`grype ubuntu:20.04 -o json=docker.vulnerabilities.grype.json -o table=docker.vulnerabilities.grype.table -o table`

Works with `syft`:
`syft ubuntu:20.04 -o json=docker.vulnerabilities.grype.json -o table=docker.vulnerabilities.grype.table -o table`

**Why is this needed**:
So that `grype` and `syft` are more similar :)

**Additional context**:

olivierboudet:
I tried to resolve #648 with this PR but I am not a Go developer so you may have remarks...

This PR allows to specify multiple outputs like in syft, for example `-o json=report.json -o template=report.html -t template.html`

I need to add more tests but if someone could tell me if it looks ok overall ?


kzantow (commented):
Are you able to add tests for this change?

cmd/writer.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/hashicorp/go-multierror"
+
+	"github.com/anchore/syft/syft/formats"
+	"github.com/anchore/syft/syft/formats/table"
+	"github.com/anchore/syft/syft/formats/template"
+	"github.com/anchore/syft/syft/sbom"
```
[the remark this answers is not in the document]
olivierboudet: you are right, I deleted this file, it was not used...

olivierboudet:
> Are you able to add tests for this change?

I just added some tests

wagoodman:
Right now this doesn't seem to work when providing multiple files:
```
$ go run . alpine:3.2 -o json=/tmp/json -o table -o cyclonedx=/tmp/cyclone

# stdout is cyclonedx, not the table output
```

That being said, I think I can update this PR to include additional pattern changes to how we encode the final reports. I'll think about it this afternoon and try to get something pushed before the end of the day.

test/quality/vulnerability-match-labels
wagoodman, line 1: was updating this submodule intentional?

olivierboudet: No it was not intentional, sorry for this.

grype/presenter/json/presenter.go
```diff
@@ -31,11 +37,25 @@ func NewPresenter(pb models.PresenterConfig) *Presenter {
...
 		appConfig:        pb.AppConfig,
 		dbStatus:         pb.DBStatus,
+		outputFilePath:   outputFilePath,
+		fs:               fs,
 	}
 }
 
 // Present creates a JSON-based reporting
-func (pres *Presenter) Present(output io.Writer) error {
+func (pres *Presenter) Present(defaultOutput io.Writer) error {
```
wagoodman, line 46: unfortunately this is breaking the presenter abstraction by adding filesystem concerns.

Today the presenter acts as a two-way facade... the caller provides doesn't need to know about any format concerns (is this json, table, xml, etc) and the presenter only needs to write bytes to the writer, but does not need to know about the concerns of that writer (is it to a file, the screen, streaming to a socket, etc).

wagoodman: I think ultimately we should probably be getting rid of the presenter abstraction and inverting this entire flow... so instead of providing a writer to an object, we should ask the object to return bytes. If this is the way to go, we should try to do this incrementally. A good approach to this would be to take the GetPresenter() function and adapt that to return the new object and write a short generalized helper function that takes a presenter and returns bytes.   Initially we'd leave all of the presenters alone and fix everything above the GetPresenter() call (maybe call it GetWriter instead since it would no longer return a presenter). Then over time we can swap all of the presenters to encoders/writers.

I haven't thought this entirely though, but these are some initial thoughts.

wagoodman: @olivierboudet I can try out the above suggestion on your branch, as it would ultimately enable the multi -o feature (which would be awesome). Would you mind if I pushed a lot of updates to your branch? 

(Alternatively I could try and get these changes in another PR then rebase this, which isn't a problem, but I feel the rebase would be nasty and the features on this branch nicely motivate the need for the change.)

olivierboudet: Thanks for the review ! Of course you can push on my branch ☺️

wagoodman:
@olivierboudet I updated the PR based on https://github.com/anchore/grype/pull/1346#discussion_r1258887699 by leveraging the multi-writer patterns in syft. In doing so I needed to change some underlying eventing and presentation functionality, which we were overdue for. This doesn't address larger concerns with the presenter pattern and how it works with the execution flow, but instead nudges grype in the right direction (larger refactors can be done in future PRs)

wagoodman (approved):
this one has been a long time coming!

olivierboudet:
Thanks a lot @wagoodman

PR #1347, opened 2023-06-12
Update the contributing guide

joshbressers:
This updates the Grype contributing guide to be very similar to the Syft contributing guide (which was updated a few months ago to be more useful).


kzantow (approved):
👍

PR #1358, opened 2023-06-22
Add more log4j-adjacent package ignore rules

luhring:
The added rules handle false positives against `log4j-over-slf4j`, where the vulnerability is meant for Log4j itself.

I added these here because there was a precedent for this log4j-adjacent situation. A more holistic (and larger scope) approach would be some of the ideas we've talked about recently for more precise CPE locating.

Happy to add test coverage for these, but looking at the existing tests it seemed like it might be overly tautological. I defer to you all!


luhring:
I could use help debugging the quality tests failure 😅 

I see this error in the logs, not sure if it's related:

```console
git submodule update vulnerability-match-labels
. venv/bin/activate && ./gate.py
Error: -22 20:25:19,857 [ERROR] unable to open grype DB [Errno 2] No such file or directory: 'grype'
```

And I see several excerpts like this:

```console
Match differences between tooling (with labels):
   TOOL PARTITION      PACKAGE                  VULNERABILITY  LABEL          COMMENTARY
   grype@v0.63.0 ONLY  log4j-over-slf4j@1.7.30  CVE-2020-9493  FalsePositive  (got rid of a former FP 🙌)

Failed quality gate
```

Is that "commentary" acknowledging that this PR got rid of an FP? (I would think that's a good thing?)

kzantow (approved):

kzantow:
Thanks @luhring!

PR #1366, opened 2023-06-28
Add a simple CSV format template to the templates/ directory and tweak docs

tgerla:
- add the csv.tmpl to templates/ directory
- edit README.md to match

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,20 +267,9 @@ Grype lets you define custom output formats, using [Go templates](https://golang
 
 - Grype's template processing uses the same data models as the `json` output format — so if you're wondering what data is available as you author a template, you can use the output from `grype <image> -o json` as a reference.
 
-**Example:** You could make Grype output data in CSV format by writing a Go template that renders CSV data and then running `grype <image> -o template -t ~/path/to/csv.tmpl`.
-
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -271,7 +271,7 @@ Grype lets you define custom output formats, using [Go templates](https://golang
 
 **Please note:** Templates can access information about the system they are running on, such as environment variables. You should never run untrusted templates.
 
-Here's what the `csv.tmpl` file might look like:
+There are several example templates in the [templates/](https://github.com/anchore/grype/tree/main/templates) directory. Here's what the `(csv.tmpl)[https://github.com/anchore/grype/blob/main/templates/csv.tmpl]` file looks like:
 
 ```gotemplate
```
kzantow, line 276: Should we take the example out of the readme and just point users to the templates directory to see examples? If nothing else, it at least eliminates one location to update if the data model ever changes enough that these need to be updated

kzantow (approved):

README.md
tgerla: Yeah, that's a good idea, I will update the patch.

kzantow (approved):

Issue #1234, opened 2023-04-13
Development Documentation Out of Date?


shanedell:
**What happened**:

`DEVELOPING.md` seems to be out dated. Believe this is because these commands are inside the `Docker Development` section:

```bash
docker-compose build grype
docker-compose run --rm grype bash
```

However, there is no `docker-compose.yml` inside of the repo. 

Also when trying to use docker commands:

```bash
docker build -t grype -f Dockerfile .
docker run --rm grype bash
```

An error like this below is returned:

```bash
docker: Error response from daemon: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "/grype": stat /grype: no such file or directory: unknown.
```

So the `build` works but not the `run`

**What you expected to happen**:

Does the development README need updated? I would have expected `docker run --rm grype bash` to work after building the image so maybe I also missing something there.

This really isn't necessarily a bug more of documentation question to help me get setup locally but I wasn't sure what category to pick.

tgerla:
Hi @Shanedell, thanks for the issue. You're right that the DEVELOPING.md document needs to be updated, since we are no longer using docker-compose. For your second question, this method with Docker isn't really a supported way of developing Grype. For the most part we use `go run ./main.go` for development. Will that get you going for now? Thanks!

PR #1384, opened 2023-07-11
Remove Docker section from DEVELOPING.md

willmurphyscode:
Developing in Docker is no longer explicitly supported. Update developing docs to reflect this.

Fixes https://github.com/anchore/grype/issues/1234


kzantow (approved):
👍

PR #1390, opened 2023-07-17
Refactor integ test to table test

willmurphyscode:
To make it easier to see which tests fail if there's a failure.


spiffcs (approved):
While it's more verbose I really like how it works so far as having individual tables that can be easily identified if failing to investigate +1 🟢

PR #1394, opened 2023-07-18
chore: bump yardstick to address PyYAML cython compatibility issues

westonsteimel:


westonsteimel:
Hmm, this seems to still be picking up the old yardstick version from something

westonsteimel:
Seems to be installing from the labels requirements, so bumping that in https://github.com/anchore/vulnerability-match-labels/pull/82

PR #1400, opened 2023-07-24
feat: update table sort to be name, version, type, severity, vulnerability

spiffcs:
## Update table sort in default presenter

This PR updates the table sort in the default table presenter with the following priority:
- name
- version
- type
- severity
- vulnerability

### Before the table output would have severities interleaved along with an unspecific ordering of CVE by date filed
Screenshot 2023-07-25 at 1 58 33 PM

### After the table output now collates the results by severity (highest to lowest), further sorting those by when the CVE was filed (most recent year/number at the top)
Screenshot 2023-07-25 at 1 58 45 PM

TODO

- [x] Static Code linting
- [x] Update presenter tests to account for latest sort order

### DIFF
--- a/grype/presenter/table/presenter_test.go
+++ b/grype/presenter/table/presenter_test.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 
 	"github.com/go-test/deep"
+	"github.com/google/go-cmp/cmp"
 	"github.com/sergi/go-diff/diffmatchpatch"
 	"github.com/stretchr/testify/assert"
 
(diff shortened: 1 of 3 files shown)


grype/presenter/table/presenter_test.go
```diff
@@ -18,7 +19,7 @@ import (
 	syftPkg "github.com/anchore/syft/syft/pkg"
 )
 
-var update = flag.Bool("update", false, "update the *.golden files for table presenters")
+var update = flag.Bool("update", true, "update the *.golden files for table presenters")
```
wagoodman, line 22: this should never be checked in

Issue #709, opened 2022-04-05
Consistent sort order for grype output


exortech:
**What would you like to be added**:
It would be fantastic if the output produced by grype was consistently ordered. It does appear to be sorted by NAME, but a secondary sort by VULNERABILITY would ensure that the output is rendered consistently across runs.

**Why is this needed**:
We commit the output of our grype runs to version control after updating dependencies. The inconsistent sorting creates unnecessarily complex diffs.

freedom-isnotanarchy:
I agree with the value of this enhancement. The proposed "consistently  rendered output", any consistent sort whatsoever, makes diffs possible.

adriens:
Also, is sorted from `Critical`-> Negligible` it would make `| head`` much more interesting as people usually are looking for the worst one.

So finally it leads to make `grype` more efficient for end user :muscle: :pray:

spiffcs:
Thanks for the issue @exortech! This looks like a pretty good option for a community PR or first issue if anyone on this thread wants to take a stab. If not I'll try and get around to this as soon as I have some time.

maqzee-git:
➕ 1️⃣ for this one

spiffcs:
I've added a PR to this issue for the table sort - if that's what we're looking for I'll go ahead and get one added for machine readable formats

PR #1403, opened 2023-07-26
feat: implement secondary sorting for default json output

spiffcs:
## Follow up to https://github.com/anchore/grype/pull/1400

Another submission for #709 

Implements the new table sorting secondary sort for the default json output.

There is a small side effect which influences the template presenter since it uses the same sorting interface as the json output.

### For reviewer

Do you think we should do the same here for sarif and cyclonedx? I'm unsure here and only have made changes for formats we have agency over at the moment (table, json, template).

### DIFF
--- a/grype/presenter/models/match.go
+++ b/grype/presenter/models/match.go
@@ -60,30 +60,36 @@ func newMatch(m match.Match, p pkg.Package, metadataProvider vulnerability.Metad
 	}, nil
 }
 
-var _ sort.Interface = (*ByName)(nil)
+var _ sort.Interface = (*MatchSort)(nil)
 
-type ByName []Match
+type MatchSort []Match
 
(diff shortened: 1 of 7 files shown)


grype/presenter/models/match.go
```diff
@@ -60,30 +60,36 @@ func newMatch(m match.Match, p pkg.Package, metadataProvider vulnerability.Metad
 	}, nil
 }
 
-var _ sort.Interface = (*ByName)(nil)
+var _ sort.Interface = (*MatchSort)(nil)
```
spiffcs, line 63: Renamed since there is more than just `ByName` being sorted here. This is now the default match sort. We can always enhance this by adding other sort methods that a user could influence via flag or config option

willmurphyscode (approved):

PR #1408, opened 2023-07-29
fix: prevent hang if stderr is tty but stdout is not

willmurphyscode:
Previously, invocations of grype could hang if, for example, stdout was redirect but stderr was a TTY. Fall back to non-fancy GUI if either stdout or stderr is not a TTY, instead of only when both are directed.

## Manual testing done

```
# hangs forever:
❯ cat <(grype alpine:latest | head -n 1) <(grype alpine:latest | rg binary)
^C
# runs correctly:
❯ cat <(go run cmd/grype/main.go alpine:latest | head -n 1) <(go run cmd/grype/main.go alpine:latest | rg binary)
NAME        INSTALLED  FIXED-IN  TYPE  VULNERABILITY  SEVERITY
```

I believe this is same problem that affects syft, see https://github.com/anchore/syft/issues/1439 and https://github.com/anchore/syft/pull/1977

### DIFF
--- a/cmd/grype/internal/ui/select.go
+++ b/cmd/grype/internal/ui/select.go
@@ -19,7 +19,7 @@ import (
 func Select(verbose, quiet bool) (uis []clio.UI) {
 	isStdoutATty := term.IsTerminal(int(os.Stdout.Fd()))
 	isStderrATty := term.IsTerminal(int(os.Stderr.Fd()))
-	notATerminal := !isStderrATty && !isStdoutATty
+	notATerminal := !isStderrATty || !isStdoutATty
 
 	switch {
 	case runtime.GOOS == "windows" || verbose || quiet || notATerminal || !isStderrATty:


willmurphyscode:
I'm reconsidering this fix, because:

```
❯ go run cmd/grype/main.go alpine:latest | rg python

❯ grype alpine:latest | rg python
 ✔ Loaded image                                                                                                                                                     alpine:latest
 ✔ Parsed image                                                                                           sha256:5053b247d78b5e43b5543fec77c856ce70b8dc705d9f38336fa77736f25ff47c
 ✔ Cataloged packages              [16 packages]
 ✔ Scanned for vulnerabilities     [4 vulnerabilities]
   ├── 0 critical, 0 high, 0 medium, 0 low, 0 negligible (4 unknown)
   └── 4 fixed
```

It's nice to get the stderr stuff when doing a simple pipe. Maybe there's a better approach?

Issue #811, opened 2022-06-27
feat: use ghsa to improve matching for cpes


cpendery:
Another thing which might potentially be useful (and should probably be configurable on individual ecosystem matcher level) could be if we find an NVD match and that CVE has a GHSA for another ecosystem (and the eocsystem we're matching for is supported by GitHub) we filter it out.  Hopefully that makes sense but if not I can try and explain it better.

_Originally posted by @westonsteimel in https://github.com/anchore/grype/issues/390#issuecomment-1052132016_

cpendery:
I wanted to keep track of this idea when my pr closes its source issue in order to investigate in the future

westonsteimel:
Yes, so I'm already working on a feature for this internally.  It will be configurable to define some hierarchy of namespaces which should be considered authoritative.  I'll let you know when I'm ready with it and you can try it out.

PR #1412, opened 2023-08-02
feat: disable CPE-based matching for GHSA ecosystems by default

westonsteimel:
Disables CPE-based matching for ecosystems which are covered by GitHub Security Advisories.  Also adds a separate rust matcher and related configuration to allow configuring CPE-based matching off for rust packages while still leaving it enabled for anything falling into the stock matcher.

Fixes: #811


spiffcs:
Added blocked for now so that we investigate the quality gate a bit more to see why this is passing when we expect a failure given the number of false negatives

wagoodman (approved):

PR #1418, opened 2023-08-04
chore: use syft v0.86.1 in the quality gate tests

westonsteimel:
This ensures the CPE dict enhancements are taken into account for future quality gate comparisons.  This should wait until https://github.com/anchore/vulnerability-match-labels/pull/92 makes it in


spiffcs:
https://github.com/anchore/vulnerability-match-labels/pull/92 was merged with the new base runner for extra disk space - we can now kick off these workflows to get them merged for grype (also with a larger disk for the GH actions runner)

spiffcs (approved):

PR #1434, opened 2023-08-15
fix: update semver regular expression constraint to allow for 1.20rc1 cases no '-'

spiffcs:
### Summary
Related to https://github.com/anchore/grype-db/pull/145

Grype was seeing issues where semantic versions like `1.20rc1` would fall into the fuzzy matcher and not use the correct `Satisfied` method. This PR updates the regex so that cases without the dash will be considered by the semantic version matcher while still preserving the openssl fuzzy matcher gate of checking cases like `1.20.0a-rc0`

Note - should these cases NOT use the semantic matcher there is a chance the fuzzy matcher is behaving incorrectly:
https://github.com/anchore/grype/blob/main/grype/version/fuzzy_constraint_test.go needs an extra look in a separate PR to account for cases that are not valid semver.

TOOO:
- [x] Add example python and openssl test cases to the fuzzy constraint harness to protect against regressions


westonsteimel (approved):
Thanks for diving into this one for me @spiffcs .

Issue #1435, opened 2023-08-15
I would like a legacy.Execute() without the os.Exit(1)


harve106:
**What would you like to be added**:
These lines of code in github.com\anchore\grype@v0.65.1\cmd\grype\cli\legacy\cmd.go
```
func ExecuteNoExit() {
	if err := rootCmd.Execute(); err != nil {
		_ = stderrPrintLnf(err.Error())
	}
}
```
**Why is this needed**:
I need it in order to use this function as I would the command line app because it's way easier to screen scrape than to go through APIs for simple usage

harve106:
I have a PR that implements the functionality I would like with your old method of calling Execute(), https://github.com/anchore/grype/pull/1436

kzantow:
Hi @harve106, the aforementioned PR has been merged and you can run grype CLI using this code:
```go
package main

import (
	"fmt"
	"github.com/anchore/clio"
	"github.com/anchore/grype/cmd/grype/cli"
)

func executeGrype() {
	cmd := cli.Command(clio.Identification{
		Name:    "my-app-name",
		Version: "[not provided]",
	})
	_ = cmd.Execute()
}

func main() {
	executeGrype()
	fmt.Println("post-execute")
}
```
It does not call `os.Exit`.

However, I think it would be better to use Grype as a library. I took a stab at this and ended up running into a few issues, but would suggest we look at how to improve the experience of using Grype as a library. Could you expand on your use case at all?

tgerla:
Hey @harve106, we'll go ahead and close this issue out but please feel free to let us know if you need any more help here. Thanks!

PR #1436, opened 2023-08-15
Added ExecuteNoExit that executes without calling exit afterwards

harve106:
#1435 
Signed-off-by: Kyle Harvey <harve106@purdue.edu>

### DIFF
--- a/cmd/grype/cli/legacy/cmd.go
+++ b/cmd/grype/cli/legacy/cmd.go
@@ -45,6 +45,12 @@ func Execute() {
 	}
 }
 
+func ExecuteNoExit() {
+	if err := rootCmd.Execute(); err != nil {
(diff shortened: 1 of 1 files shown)


kzantow:
Thanks for the contribution @harve106 , but I'm going to close this PR as there is now a simple way to accomplish the same thing by using code like this sample:
```go
package main

import (
	"fmt"
	"github.com/anchore/clio"
	"github.com/anchore/grype/cmd/grype/cli"
)

func executeGrype() {
	cmd := cli.Command(clio.Identification{
		Name:    "my-app-name",
		Version: "[not provided]",
	})
	_ = cmd.Execute()
}

func main() {
	executeGrype()
	fmt.Println("post-execute")
}
```

PR #1452, opened 2023-08-22
fix: set correct default to exclude overlapping binaries

kzantow:
This PR sets the same behavior as Syft, which excludes binaries that overlap with package manager packages.


internal/config/search.go
```diff
@@ -39,5 +39,6 @@ func (cfg search) ToConfig() cataloger.Config {
 			IncludeUnindexedArchives: cfg.IncludeUnindexedArchives,
 			Scope:                    cfg.ScopeOpt,
 		},
+		ExcludeBinaryOverlapByOwnership: true,
```
willmurphyscode, line 42: I wanted to make sure I understand the config options we have here.

1. This change tells syft not to emit binary packages if they are in files that are owned by another package
2. There is code in grype today that tells grype not to _match_ on binary packages that are owned by other packages. I think this code is run unconditionally.
3. There is code in grype today (unreleased) that, on certain distros, also removes non-binary packages if they are owned by an OS package manager package. This code will also run unconditionally.

I think the change governed by this config is a subset of the code in item 2, so I just wanted to make sure we understand what configs are needed and what configs are available.

willmurphyscode: Talked with @kzantow about this. It turns out that Syft, when `ExcludeBinaryOverlapsByOwnership` is turned on, removes more packages than Grype will. Specifically, Grype checks whether the child package's version is a prefix of the parent package's version before removal, but Syft does not.

Here's the removal code in grype: https://github.com/anchore/grype/blob/e906e685a109fa524d03c9c55b6b45515cefbccf/grype/pkg/package.go#L129

Here's the removal code in syft:
https://github.com/anchore/syft/blob/cedfa05e93f9ad9649b514e5a8b9127fe1d44f85/syft/pkg/cataloger/package_exclusions.go#L30-L51

willmurphyscode (approved):

PR #1461, opened 2023-08-24
Track vulnerability match label submodule to main branch

wagoodman:
This PR adjusts the vulnerability-match-labels submodule to always track the main branch. This prevents the need to pull in new labels continually. If there is ever an issue upstream we can pin back to a specific commit until the issue is resolved (then retrack to main).


wagoodman:
nvm, this is not working the way I expected it to work.

PR #1463, opened 2023-08-24
Introduce exit code failure option for db update check

devfbe:
Introduce the -f / --fail-on-update-available flag for the "db check" command. If set, the update check command will now return exit code 100 if an db upgrade is available.

This is useful in CICD environments where grype container images, containing the database, should automatically be rebuilt when an database upgrade is available . I have this use case in my container image build pipelines where I have many check jobs for my container images (e.g. my custom grype image) which decide, if a rebuild is necessary.

Use the exit code 100 here (instead of 1), because 1 indicates an error and you want to distinguish errors like a broken internet connection and the fact, that an update is available.

Exit code 100 was a free exit code which is not reserved by any other meaning, so I chose this randomly.


tgerla:
Hey @devfbe, thank you for this patch! We are talking over this idea and wondering if the -f flag is even necessary--what if we skipped the flag and just had "grype db check" return the error code 100 on an available update?

devfbe:
That's a good question @tgerla!

My thoughts were that this could be a breaking change for people that currently use | grep to check If there is an update available (when pipefail is set as shell opt).


On the other hand side, it's definitely a better approach to signal the result directly without the explicit flag - feels more unix like.

So depending on your choice (breaking change or not) I'll update the merge request. At the Moment I personally would prefer removing the -f flag.

devfbe:
Because there were breaking changes in the earlier releases of grype, I guess this is okay to do it again here - It should not affect many users and it's really simple to fix. And returning 100 when db updates are available, really makes sense without a toggle.

I updated the merge request, added only the os.exit(100) lines (removed the old stuff).
Rebased and reworded the commit.

tgerla:
Thanks @devfbe, we had the same thoughts--better to get a change like this in before 1.0, anyway! I agree that it probably won't affect too many folks.

Thanks for the updated patch, we'll take a look ASAP.

spiffcs:
Looks like updating to `1.21.1` had some unseen affects and has made contributing harder - I'll take a look at the Go 1.21 articles to see if we can pin down the best setup here so we're not constantly seeing `go mod tidy` update the toolchain directive across different branches and commits.

ahoz:
Hi all 👋
when can we expect this pr to be merged?
It would be a nice feature for grype automations in pipelines 🚀

kzantow (approved):
Thanks for the contribution @devfbe!

PR #1470, opened 2023-08-28
Pin the vulnerability DB used in quality gate testing

wagoodman:
Ultimately the goal of the quality gate in grype is to test the effectiveness of grype itself to detect vulnerabilities in an artifact. This is done relative to the set of labels we know about these artifacts. The gate today has two moving targets: the grype code itself and the vulnerability DB. This PR changes this such that the only independent variable is the grype code itself by pinning the DB to a single version.

### DIFF
--- a/test/quality/Makefile
+++ b/test/quality/Makefile
@@ -7,6 +7,11 @@ YARDSTICK_LABELS_DIR = .yardstick/labels
 VULNERABILITY_LABELS = ./vulnerability-labels
 RESULT_SET = pr_vs_latest_via_sbom
 
+# update periodically with values from "grype db list"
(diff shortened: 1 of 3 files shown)


kzantow (approved):
🎉

test/quality/Makefile
```diff
@@ -7,6 +7,11 @@ YARDSTICK_LABELS_DIR = .yardstick/labels
 VULNERABILITY_LABELS = ./vulnerability-labels
 RESULT_SET = pr_vs_latest_via_sbom
 
+# update periodically with values from "grype db list"
+TEST_DB_URL = https://toolbox-data.anchore.io/grype/databases/vulnerability-db_v5_2023-08-24T01:23:40Z_fd07204627d474f68f90.tar.gz
```
willmurphyscode, line 11: This will become increasingly stale over time. How often should we bump it?

PR #1471, opened 2023-08-31
Bump version of syft to v0.89.0

wagoodman:


wagoodman:
Closing and using https://github.com/anchore/grype/pull/1472 instead

Issue #1430, opened 2023-08-12
False positive? Trailing zeros seen as difference in version numbers


FlorianPfaff:
Thanks for the nice tool. I am encountering an issue and am not 100 % sure if this is intended behavior or not. If this is intended behavior, poetry (which generated my requirements.txt) should have done this differently.

**What happened**:
I got the report
`❌ Linted [REPOSITORY] files with [grype]: Found 1 error(s) - (11.02s) (expand for details)
  - Using [grype v0.63.1] https://megalinter.io/7.3.0/descriptors/repository_grype
  - MegaLinter key: [REPOSITORY_GRYPE]
  - Rules config: [.grype.yaml]
  --Error detail:
  NAME     INSTALLED  FIXED-IN    TYPE    VULNERABILITY        SEVERITY 
  certifi  2023.7.22  2023.07.22  python  GHSA-xqr8-7jwr-rhp7  High      
  1 error occurred:
  	* discovered vulnerabilities at or above the severity threshold`

**What you expected to happen**:
No error because it is the version in which it is already fixed.

**How to reproduce it (as minimally and precisely as possible)**:
Have
certifi==2023.7.22 ; python_version >= "3.10" and python_version < "3.13"
in the requirements.txt

**Anything else we need to know?**:
This may be intended behavior - I am not sure. In this case, poetry should properly write the requirements.txt.

**Environment**:
- Output of `grype version`:
 v0.63.1
- OS (e.g: `cat /etc/os-release` or similar):
MegaLinter image based on alpine linux.
- Image Creation Date: 2023-08-09T22:49:15Z
 - Image Revision: fda6ac3a38be0e969820709ac16e[44](https://github.com/FlorianPfaff/pyRecEst/actions/runs/5841327748/job/15841424213#step:12:45)2464e5a035
 - Image Version: v7.3.0

kzantow:
Looking at [GHSA-xqr8-7jwr-rhp7](https://github.com/advisories/GHSA-xqr8-7jwr-rhp7), you'll see that it does, in fact have the `07` in the fixed-in version: 2023.07.22. I'm not entirely sure what the correct format is, but I think this is something Grype should be able to handle when doing version comparisons, at least for Python packages.

crimsonknave:
I've submitted and update to the advisory db which has been merged and the advisory how has `2023.7.22` as the patched version.

[...]

Issue #1417, opened 2023-08-04
False Positive: CVE-2023-37920 reported for certifi library in python


chidambaranathan-r:
**What happened**:

I have a Python Image "certifi" installed with "2023.7.22". As per https://github.com/certifi/python-certifi/security/advisories/GHSA-xqr8-7jwr-rhp7, this version is not impacted.

But grype is marking this CVE as applicable for my image:
```
   "matchDetails": [
    {
     "type": "exact-direct-match",
     "matcher": "python-matcher",
     "searchedBy": {
      "language": "python",
      "namespace": "github:language:python",
      "package": {
       "name": "certifi",
       "version": "2023.7.22"
      }
     },
     "found": {
      "versionConstraint": ">=2015.04.28,<2023.07.22 (python)",
      "vulnerabilityID": "GHSA-xqr8-7jwr-rhp7"
     }
    }
   ],
```

`versionConstraint` used by the matcher is incorrect. As per GH advisory, patched versions are `>=2023.07.22`.

**What you expected to happen**:
Ideally, grype should not report the patched version as vulnerable.

**How to reproduce it (as minimally and precisely as possible)**:
Scan any python image containing certifi with version `2023.07.22`.

**Environment**:
- Output of `grype version`: 
```
bash-4.4# grype version
Application:          grype
Version:              0.65.0
Syft Version:         v0.86.1
BuildDate:            2023-08-01T00:36:47Z
GitCommit:            c97048baa1595a481a26f7add8b18d59ec65838a
GitDescription:       v0.65.0
Platform:             linux/amd64
GoVersion:            go1.20.1
Compiler:             gc
Supported DB Schema:  5
```

pcreager23:
I will point out that the lack of leading zeroes in any of these dated versions is standard practice for Python, per canonical PEP 440, so `2023.07.22` is the correct format.
Ref: [Version specifiers](https://packaging.python.org/en/latest/specifications/version-specifiers/)

spiffcs:
Linking https://github.com/anchore/grype/issues/1172 to this

Issue #1172, opened 2023-03-11
Grype is not recognizing python-certifi is patched for GHSA-43fp-rhv2-5gv8


ssullivan:
**What happened**:

Grype is detecting that I have an unpatched version of certifi when it is patched. I've seen Grype
report this error on multiple platforms and operating systems. The following logs were generated on a Windows 10 system.

**What you expected to happen**:

I expected the CVE to not show up in the Grype output.

**How to reproduce it (as minimally and precisely as possible)**:

## Test 1

I created a Dockerfile like the following

```docker
FROM python:3.10.10-alpine3.16
RUN python -m pip install certifi==2022.12.07
```

```shell
docker build . -t example:latest
```

```shell
$ grype example:latest
NAME         INSTALLED  FIXED-IN    TYPE    VULNERABILITY        SEVERITY
certifi      2022.12.7  2022.12.07  python  GHSA-43fp-rhv2-5gv8  Medium
libcom_err   1.46.5-r0  1.46.6-r0   apk     CVE-2022-1304        High
pip          22.3.1                 python  CVE-2018-20225       High
python       3.10.10                binary  CVE-2007-4559        Medium
python       3.10.10                binary  CVE-2023-24329       High
sqlite-libs  3.38.5-r0              apk     CVE-2022-46908       High
```

## Test 2

I noticed that the python certifi library has defined their version as 2022.12.07 in their 
[__init__.py](https://github.com/certifi/python-certifi/blob/master/certifi/__init__.py) file here like this

```python
from .core import contents, where

__all__ = ["contents", "where"]
__version__ = "2022.12.07"
```

I decided to create a new Dockerfile like the following to install from source since 2022.12.07 is not
on PYPI and the source repo has the version that Grype says the CVE is patched.

```docker
FROM python:3.10.10-alpine3.16
RUN apk add git && python -m pip install git+https://github.com/certifi/python-certifi.git
```

and got the following output

```shell
$ docker build --no-cache --progress=plain -t example:latest .
#1 [internal] load build definition from Dockerfile
#1 sha256:a1283ec5588a18bc133b0e663cbc731257a537d894c1aaaf511240b3a22cd841
#1 transferring dockerfile: 31B done
#1 DONE 0.0s

#2 [internal] load .dockerignore
#2 sha256:1b44120402eebbe7b7f8eaf90cde234cd91cac0b16b12f1a32807f069424c7f0
#2 transferring context: 2B done
#2 DONE 0.0s

[...]

Issue #1034, opened 2022-12-12
False positive on certifi 2022.12.07


mrginglymus:
**What happened**:
```
TYPE    NAME                    INSTALLED               VULN                  SEVERITY      FIXED?     FIXED IN              UPSTREAM PACKAGE           DATASOURCE
python  certifi                 2022.12.7               GHSA-43fp-rhv2-5gv8   Medium        fixed      2022.12.07                                       https://github.com/advisories/GHSA-43fp-rhv2-5gv8
```

**What you expected to happen**:
No security advisory

**How to reproduce it (as minimally and precisely as possible)**:
Python docker image with certifi==2022.12.7 installed

**Anything else we need to know?**:

`certifi` publish their python packages with a 'human formatted' zero-padded date format. However, PEP440 canonical versions have their leading zeros removed. Thus, `2022.12.7` is equivalent to `2022.12.07`.

What makes this odd is that even on a string comparison, `2022.12.7` is greater than `2022.12.07`.

**Environment**:
```
Application:          grype
Version:              0.50.2
Syft Version:         v0.57.0
BuildDate:            2022-09-20T19:54:28Z
GitCommit:            f094b860b9b9ebcdf06d9b9c6f7d5c000fad395b
GitDescription:       v0.50.2
Platform:             windows/amd64
GoVersion:            go1.18.6
Compiler:             gc
Supported DB Schema:  4
```

westonsteimel:
Thanks for the report @mrginglymus!  Yes, this is because we don't currently have exact support for python version comparisons according to PEP 440 and it does cause some false positives. https://github.com/anchore/grype/issues/986 is another case of this.

ilanKeshet:
reproduces for me as well 
```
Package                                   Type              Installed Version                  Fixed Version             Vulnerability             Severity
certifi                                 python                      2022.12.7                     2022.12.07       GHSA-43fp-rhv2-5gv8               Medium
```

danekantner:
This CVE is recorded wrong by Github IMO. The actual correct version is 2022.12.7 but the CVE is wrong in that it reports the vulnerable version is < 2022.12.07 and fixed version >= 2022.12.07 -- there is no .07 it is .7 

This problem is happening with this CVE in other security tools too.

ilanKeshet:
@danekantner This sort of stuff is bound to happen again and again. 
I get that your not in the wrong here, but it would be truly awesome if you (grype) could shield users from such inevitable inconveniences 
🙏 

Thank You

[...]

PR #1510, opened 2023-09-20
fix: use PEP440 for Python package version comparison

willmurphyscode:
There have been several issues opened where `certifi` Python package, which uses a version string like `2022.07.22`, which is PEP440 compliant but may or may not look like semver depending on the regex in use, results in incorrect version comparisons. Switch python version comparison to use Python PEP440 directly.

Fixes https://github.com/anchore/grype/issues/1430
Fixes https://github.com/anchore/grype/issues/1417
Fixes https://github.com/anchore/grype/issues/1172
Fixes https://github.com/anchore/grype/issues/1034
Fixes https://github.com/anchore/grype/issues/986

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,7 @@
 /.grype.yaml
 
 CHANGELOG.md
-VERSION
+/VERSION
 /snapshot/
 /dist/
 *.profile
--- a/grype/version/constraint.go
+++ b/grype/version/constraint.go
@@ -25,7 +25,7 @@ func GetConstraint(constStr string, format Format) (Constraint, error) {
 		// Although this will work in most cases, some oddities aren't supported, like:
 		// 1.0b2.post345.dev456 which is allowed by the spec. In that case (a dev release of a post release)
 		// the comparator will fail. See https://www.python.org/dev/peps/pep-0440
-		return newFuzzyConstraint(constStr, "python")
+		return newPep440Constraint(constStr)
 	case KBFormat:
 		return newKBConstraint(constStr)
 	case PortageFormat:
--- a/grype/version/pep440_constraint_test.go
+++ b/grype/version/pep440_constraint_test.go
@@ -1,8 +1,9 @@
 package version
 
 import (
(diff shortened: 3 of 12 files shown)


.gitignore
```diff
@@ -4,7 +4,7 @@
 /.grype.yaml
 
 CHANGELOG.md
-VERSION
+/VERSION
```
willmurphyscode, line 7: One case insensitive file systems (such as macOS), this line previously resulted in ignoring files under `./grype/version/`.

grype/version/pep440_constraint_test.go
```diff
@@ -0,0 +1,305 @@
...
+		{
+			name:       "date based pep440 version string boundary condition",
+			version:    "2022.12.7",
+			constraint: ">=2017.11.05,<2022.12.07",
+		},
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			_, err := newPep440Constraint(tc.constraint)
+			require.NoError(t, err)
```
willmurphyscode, line 302: Big `TODO` here - didn't actually implement these tests yet. Test cases are copied from the old fuzzy matcher.

kzantow (approved):
LGTM

wagoodman (approved):

alex:
As a certifi maintainer: thank you!

grype/version/constraint.go
```diff
@@ -25,7 +25,7 @@ func GetConstraint(constStr string, format Format) (Constraint, error) {
 		// Although this will work in most cases, some oddities aren't supported, like:
 		// 1.0b2.post345.dev456 which is allowed by the spec. In that case (a dev release of a post release)
 		// the comparator will fail. See https://www.python.org/dev/peps/pep-0440
```
danielsamuels, line 27: Presumably this comment is outdated now?

willmurphyscode: Good catch @danielsamuels! I'll send a PR to clean up the comment. Thanks!

PR #1530, opened 2023-09-29
Feat/grype 86 distro from purl

willmurphyscode:


kzantow:
Hey @willmurphyscode -- now that the v6 work has merged, `Distro` is [an optional field on the grype package](https://github.com/anchore/grype/blob/main/grype/pkg/package.go#L38) struct, so we should be able to simply populate it from PURLs wherever it's applicable.  👍

PR #1539, opened 2023-10-04
fix: `if` with identical `then` and `else` blocks in `internal/format/writer_test.go`

### LINKED ISSUE #1534: `if` with identical `then` and `else` blocks in `internal/format/writer_test.go`
https://github.com/anchore/grype/blob/45315280992cc13a0f5c1b85368ab9e472233be3/internal/format/writer_test.go#L175-L179

Either the whole conditional can be replaced by `assert.NotNil(t, w.out) ` or the `else` branch should be fixed.

(found by [revive](https://github.com/mgechev/revive) while testing it on grype code base)

chavacava:
Closes #1534 by replacing the conditional with just the content of one of its branches.


kzantow (approved):
LGTM

Issue #1526, opened 2023-09-27
Incorrect version comparisons for maven packages


westonsteimel:
**What happened**:

Currently grype is just using the fuzzy matcher (which falls back to semver when the version is somewhat semver like) for java packages; however, it should really be using the [Maven version spec](https://maven.apache.org/pom.html#Dependency_Version_Requirement_Specification) as the semver logic leads to wrong comparisons in some specific cases.

A specific case where grype currently fails is below:  

Cloudbees append a `-cb-<patch number>` to the version when issuing patches to an existing version and currently grype does not consider these correctly in match comparisons.

An example is 2.414.2-cb-5 should be considered > 2.414.2, but grype currently treats it as a pre-releaase

**What you expected to happen**:

 2.414.2-cb-5 should be treated > 2.414.2

**How to reproduce it (as minimally and precisely as possible)**:

```
grype docker.io/cloudbees/cloudbees-core-mm:2.414.2.2 | grep GHSA-5j46-5hwq-gwh7
jenkins-core                                2.414.2-cb-5           2.414.2                   java-archive    GHSA-5j46-5hwq-gwh7  High
```

**Anything else we need to know?**:

**Environment**:
- Output of `grype version`:
```
Application:         grype
Version:             0.69.0
BuildDate:           2023-09-20T20:56:04Z
GitCommit:           da3de94842f51059f32409289d863792726f83ba
GitDescription:      v0.69.0
Platform:            darwin/arm64
GoVersion:           go1.21.1
Compiler:            gc
Syft Version:        v0.91.0
Supported DB Schema: 5
```
- OS (e.g: `cat /etc/os-release` or similar):

wagoodman:
We don't want this to fall through to fuzzy version comparator, nor do we want the semver comparator since this is against the semver spec (-cb... looks like a pre-release). It looks like we need a new comparator that would take these cases into consideration, or a way to configure the semver comparator we have to treat prereleases as after the versions the reference in a configurable way and select on this when we see `-cb...`.

westonsteimel:
The existing maven version comparison library that trivy uses: https://github.com/masahiro331/go-mvn-version, already handles this case correctly: https://play.golang.com/p/7b9M5QcUaxu

So maybe we should just be switching to use that?

westonsteimel:
There are also a good number of test cases at https://github.com/google/osv.dev/blob/master/osv/ecosystems/maven_test.py that we should add in

PR #1571, opened 2023-10-24
feat: add custom maven comparator

spiffcs:
## Summary

Fix for #1526.

This PR takes the recommendation from #1526 and adapts the [go-mvn-version](https://github.com/masahiro331/go-mvn-version) to be used as a custom comparator for matching against packages that have the `JavaPkg` type.  Packages of type `JavaPkg` will no longer use the stock matcher.

The specific case mentioned in the issue has been included in both new test harnesses.

Here is a screenshot of the issues test case working as intended with the versions for `jenkins-core` being correctly compared. The FP is eliminated:
Screenshot 2023-10-25 at 1 50 09 PM

#### Related Labeled PR for updating the quality gate:
- https://github.com/anchore/vulnerability-match-labels/pull/116
- https://github.com/anchore/vulnerability-match-labels/pull/117
- https://github.com/anchore/vulnerability-match-labels/pull/118

Local Sample showing FP improvements with this change:
Screenshot 2023-10-25 at 10 00 34 PM

### Todo
- [x] Constraint Tests
- [x] Version Tests
- [x] Unit cases from https://github.com/google/osv.dev/blob/master/osv/ecosystems/maven_test.py


grype/version/format.go
```diff
@@ -11,6 +11,7 @@ const (
 	SemanticFormat
 	ApkFormat
 	DebFormat
+	JavaFormat
```
wagoodman, line 14: I think this should be oriented around maven ecosystem instead of the higher order java language based on where the spec is https://maven.apache.org/pom.html#Version_Order_Specification

spiffcs: I've reoriented the naming of the additions in this PR around `maven` given that they are the authors of the version order specification we're now following for packages with type `JavaPkg` going forward.

I think you also mentioned a change for https://github.com/anchore/grype-db where we should start labeling these constraints correctly?

spiffcs:
Putting this back in draft - I think it needs another look as I'm still getting the faulty match mentioned in the issue. While we're testing correctly against the comparator being used in the unit tests something is not hooked up correctly

spiffcs:
[ad906d4](https://github.com/anchore/grype/pull/1571/commits/ad906d4e11aad71f68b404060c9f29a37df61fe0) updates the `fuzzyComparitor` to trust the version object being passed to it. Trust in this case means that it will attempt to recreate a temporary comparator based on the verObjs type to see if `Satisfied` is true.

This has a minor downside.  If the package being examined is not actually of the assumed type(from the version objects) , then the satisfied logic for the new comparitor might generate an incorrect result. If allowed to progress the fuzzy matcher may have returned a correct result.

The trade off here is that we can allow for new comparators to be tried by trusting the versionObj knows what it's looking for without an update to the `version_format` field in grype-db.

The best case here is that `version_format` is already populated and we're not using the fuzzy comparator to being with, but this change attempts to handle the middle ground a bit better so we don't always need an update to the DB.

spiffcs:
Blocked on https://github.com/anchore/yardstick/pull/171

The following is being returned by the quality gate locally:
```
Failed quality gate
   - current indeterminate matches % is greater than 10%: 
   - current=10.22% image=docker.io/cloudbees/cloudbees-core-oc@sha256:9cd85ee84e401dc27e3a8268aae67b594a651b2f4c7fc056ca14c7b0a0a6b82d
```

Exploring the labeled data for this image shows that a majority of the Labeled data that is categorized as `Unclear` contains `DISPUTED` in the seciton. The above PR removes `Unclear` from the considered set of unlabeled data.

wagoodman (approved):

PR #1582, opened 2023-10-30
Incorporate format API changes from syft

wagoodman:
Incorporate changes from anchore/syft#2228


grype/presenter/cyclonedx/test-fixtures/snapshot/TestCycloneDxPresenterDir.golden
```diff
@@ -26,14 +26,18 @@
           "name": "syft:package:type",
           "value": "rpm"
         },
+        {
```
willmurphyscode, line 29: Should we have a snapshot test that asserts the right value ends up here? I was surprised to see this being set to `""` in both snapshots for this PR.

grype/pkg/syft_sbom_provider.go
```diff
@@ -106,7 +106,7 @@ func parseSBOM(scheme, path string) (io.Reader, *inputInfo, error) {
 	return r, info, nil
 }
 
-func decodeStdin(r io.Reader) (io.Reader, *inputInfo, error) {
+func decodeStdin(r io.Reader) (io.ReadSeeker, *inputInfo, error) {
 	b, err := io.ReadAll(r)
```
willmurphyscode, line 110: Totally unrelated to this review, but it might be nice to be able to check whether the `ReadAll` is strictly necessary. This could already be a `ReadSeeker`, in which case `b` might be a large and unnecessary allocation.

grype/pkg/package_test.go
```diff
@@ -198,375 +227,14 @@ func TestNew(t *testing.T) {
...
 		t.Run(test.name, func(t *testing.T) {
-			if string(test.syftPkg.MetadataType) != "" {
-				observedMetadataTypes.Add(string(test.syftPkg.MetadataType))
-			}
 			assert.Equal(t, test.metadata, New(test.syftPkg).Metadata, "unexpected metadata")
 			assert.Equal(t, test.upstreams, New(test.syftPkg).Upstreams, "unexpected upstream")
 		})
 	}
-
-	// did we see all possible metadata types? if not, then there is an uncovered case and this test should error out
```
willmurphyscode, line 563: Is it asserted elsewhere that we cover all metadata types? Maybe this logic moved into Syft and I don't know about it? Or maybe I'm missing something and our mechanism changed so that it's no longer necessary to assert completeness here?

willmurphyscode (commented):
I had a couple of questions on this.

grype/presenter/cyclonedx/test-fixtures/snapshot/TestCycloneDxPresenterDir.golden
wagoodman: that's a good point, let me update the test packages to have a metadata object (thus they'll have a metadata type specified)

wagoodman: actually, there is a metadata type, let me dig more I think something might be broken

grype/pkg/package_test.go
wagoodman: I should have made a comment in the description about this removal; essentially the fact that we were extracting upstream packages from the metadata was the important part of this test. We were following a completeness testing pattern here to ensure that as new metadata types are added that there is a test that shows what the new behavior should be. Since the metadata type definitions have been moved from a collection of strings to a collection of structs that get a label, most of this has been moved to an `internal` package within syft to keep the API a little more defensible. That also makes completion testing here intractable, since there isn't a single source of truth to get the full collection of expected types to process. For that reason I reduced the scope of this test to simply ensure we are covering known types correctly (only a few types support the upstream packages concept).

grype/pkg/syft_sbom_provider.go
wagoodman: the reader being passed in is always stdin (an `*os.File`), which the types support the notion of seeking, however in practice if you try to seek at all you'll get an error. This seems like a bad idea from the stdlib to express stdin as a file at all (see https://github.com/golang/go/issues/13473).

grype/presenter/cyclonedx/test-fixtures/snapshot/TestCycloneDxPresenterDir.golden
wagoodman: Ok, the test fixture was using grype packages as the basis, not syft packages. I've updated the fixture now and this issue should be resolved.

grype/pkg/syft_sbom_provider.go
willmurphyscode: That makes sense - I hadn't noticed that it was always stdin. It also hadn't occurred to me that we might get a poisoned `ReadSeeker` (something that implements the interface but always errors if you call `Seek`).

willmurphyscode (approved):

PR #1598, opened 2023-11-09
chore: pin action to correct sha

spiffcs:


.github/workflows/validations.yaml
```diff
@@ -60,7 +60,7 @@ jobs:
 
       - name: Upload the provider state archive
         if: ${{ failure() }}
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32
```
willmurphyscode, line 63: Does this need a version comment that dependabot can keep up to date?

willmurphyscode (approved):

Issue #1581, opened 2023-10-30
Vulnerabilities in go packages without go modules are not detected


akpsgit:
**What happened**:
When scanning an image that contains a Go binary with a dependency of a package that doesn't use go modules (appears as `+incompatible` in go.mod), the expected vulnerability wasn't not detected by Grype (`github.com/dgrijalva/jwt-go` in my case).


**What you expected to happen**:
The expected vulnerability is detected. e.g, for `github.com/dgrijalva/jwt-go`:
```
NAME                         INSTALLED  FIXED-IN  TYPE       VULNERABILITY        SEVERITY
github.com/dgrijalva/jwt-go  v3.2.0               go-module  GHSA-w73w-5m7g-f7qc  High
```

**How to reproduce it (as minimally and precisely as possible)**:

1. Create a minimal go program that uses only one dependency that has no go modules:
```
package main

import (
	"github.com/dgrijalva/jwt-go"
)

func main() {
	jwt.EncodeSegment(nil)
}
```
2. Run `go mod init` and `go mod tidy` and the following go.mod should be created:
```
module github.com/test

go 1.21.1

require github.com/dgrijalva/jwt-go v3.2.0+incompatible

```
3. Scan the directory with the Go program with Grype:
```
grype dir:./test
 ✔ Vulnerability DB                [no update available]
 ✔ Indexed file system                                                                                                                                                                                                                                                  test
 ✔ Cataloged packages              [1 packages]
 ✔ Scanned for vulnerabilities     [0 vulnerability matches]
   ├── by severity: 0 critical, 0 high, 0 medium, 0 low, 0 negligible
   └── by status:   0 fixed, 0 not-fixed, 0 ignored
[0000]  WARN no explicit name and version provided for directory source, deriving artifact ID from the given path (which is not ideal)
No vulnerabilities found
``` 


**Anything else we need to know?**:
When I tried to create an SBOM using Syft first and to scan with Grype I got:
```

syft dir:./test -o json > test.sbom

cat test.sbom
{
 "artifacts": [
  {
   "id": "89592a8d9d48cbdf",
   "name": "github.com/dgrijalva/jwt-go",
   "version": "v3.2.0+incompatible",
   "type": "go-module",
   "foundBy": "go-mod-file-cataloger",
   "locations": [
    {
     "path": "/go.mod",
     "annotations": {
      "evidence": "primary"
     }
    }
   ],
   "licenses": [],
   "language": "go",
   "cpes": [
[...]

Issue #1416, opened 2023-08-04
False positive GHSA-jr9c-h74f-2v28/CVE-2022-0905 reported for Non-vulnerable Gitea version


ramanNarasimhan77:
**What happened**:

I am using `gitea/gitea:1.16.7` image in my application.  Version from the binary within the Image is 1.16.7 as shown below

```shell 
❯ docker run -it --entrypoint bash gitea/gitea:1.16.7
bash-5.1# gitea -h
NAME:
   Gitea - A painless self-hosted Git service

USAGE:
   gitea [global options] command [command options] [arguments...]

VERSION:
   1.16.7 built with GNU Make 4.3, go1.18.1 : bindata, timetzdata, sqlite, sqlite_unlock_notify

DESCRIPTION:
   By default, gitea will start serving using the webserver with no
arguments - which can alternatively be run by running the subcommand web.

COMMANDS:
   web              Start Gitea web server
   serv             This command should only be called by SSH shell
   hook             Delegate commands to corresponding Git hooks
   dump             Dump Gitea files and database
   cert             Generate self-signed certificate
   admin            Command line interface to perform common administrative operations
   generate         Command line interface for running generators
   migrate          Migrate the database
   keys             This command queries the Gitea database to get the authorized command for a given ssh key fingerprint
   convert          Convert the database
   doctor           Diagnose problems
   manager          Manage the running gitea process
   embedded         Extract embedded resources
   migrate-storage  Migrate the storage
   docs             Output CLI documentation
   dump-repo        Dump the repository from git/github/gitea/gitlab
   restore-repo     Restore the repository from disk
   help, h          Shows a list of commands or help for one command

GLOBAL OPTIONS:
   --port value, -p value         Temporary port number to prevent conflict (default: "3000")
   --install-port value           Temporary port number to run the install page on to prevent conflict (default: "3000")
   --pid value, -P value          Custom pid file path (default: "/run/gitea.pid")
   --quiet, -q                    Only display Fatal logging errors until logging is set-up
   --verbose                      Set initial logging to TRACE level until logging is properly set-up
   --custom-path value, -C value  Custom path file path (default: "/data/gitea")
   --config value, -c value       Custom configuration file path (default: "/data/gitea/conf/app.ini")
   --version, -v                  print the version
   --work-path value, -w value    Set the gitea working path (default: "/app/gitea")
   --help, -h                     show help

DEFAULT CONFIGURATION:
[...]

PR #1599, opened 2023-11-10
fix: golang version parsing

willmurphyscode:
Resolves https://github.com/anchore/grype/issues/1581

Right now, go versions are handled by generic semantic version comparison (or fall through to fuzzy version comparison sometimes). There are a few ways in which the version found in a `go.mod` for example is not exactly a semver string:

1. Sometimes, the go tooling appends `+incompatible` to a version string. This causes the comparison to fall through from semantic version to fuzzy version, which is the root cause of #1581.
2. Versions of the go standard library itself start with `go` (e.g. `go1.24.1`) instead of `v`.
3. Go dependencies can be declared based on commits, rather than semver tags, which results in a string that looks like `v0.0.0-20180116102854-5a71ef0e047d`. Right now, these are working because the lexicographical sorting of the strings gives the right order, which is great, but a bit of a coincidence since we're letting them fall through to the "just guessing based on strings" part of the version comparison logic.
4. Sometimes syft emits `(devel)` for a go version.

This change introduces a golangVersion and golangConstraint structs in the grype/version package, so that we have a single place to write down rules like "`v3.2.0+incompatible` should compare as equal to `v3.2.0`".

Note that this change does not yet specify how `(devel)` should compare to other go versions. All the answer have potential pitfalls (e.g. saying `(devel)` is greater than `v1.2.3` might not be true; someone could check out `v1.2.2` and build it, and get `(devel)` as the version number, or vice versa if we say `(devel)` is less than `v1.2.3`).

This also changes a CPE test that, as far as I can tell, was relying on go being a language without specific versioning logic. I've changed this test to use erlang which, for the moment, does not have specific versioning logic.

It could also be that there's a nice go library somewhere for parsing and comparing go versions, and we shouldn't maintain this code (in fact, something like this code must exist in Go's own build tools), but I haven't been able to find it, and this is a pretty thin wrapper around semver anyway.

TODO:
- [x] decide how to handle `(devel)` in comparison

### DIFF
--- a/grype/version/golang_version.go
+++ b/grype/version/golang_version.go
@@ -8,12 +8,40 @@ import (
 	hashiVer "github.com/anchore/go-version"
 )
 
+var _ Comparator = (*golangVersion)(nil)
+
 type golangVersion struct {
 	raw              string
 	semVer           *hashiVer.Version
 	timestamp        string
 	commitSHA        string
 	incompatibleFlag bool
+}
+
+func (g golangVersion) Compare(version *Version) (int, error) {
+	if version.Format != GolangFormat {
+		return -1, fmt.Errorf("cannot compare %v to golang version", version.Format)
+	}
+	if version.rich.golangVersion == nil {
+		return -1, fmt.Errorf("cannot compare version with nil golang version to golang version")
+	}
+	if version.rich.golangVersion.raw == g.raw {
+		return 0, nil
+	}
+	if version.rich.golangVersion.semVer != nil && g.semVer != nil {
+		return g.semVer.Compare(version.rich.golangVersion.semVer), nil
+	}
+	if version.rich.golangVersion.semVer != nil && g.semVer == nil {
+		// semvers are greater than tag versions
+		return -1, nil
+	}
+	if g.semVer != nil && version.rich.golangVersion.semVer == nil {
+		return 1, nil
+	}
+	if version.rich.golangVersion.timestamp != "" && g.timestamp != "" {
+		return strings.Compare(g.timestamp, version.rich.golangVersion.timestamp), nil
+	}
+	return 0, nil
 }
 
 var startsWithSemver = regexp.MustCompile(`v\d+\.\d+\.\d+`)
--- a/grype/version/golang_version.go
+++ b/grype/version/golang_version.go
@@ -48,6 +48,9 @@ func (g golangVersion) compare(o golangVersion) int {
 var startsWithSemver = regexp.MustCompile(`(v|go){0,1}\d+\.\d+\.\d+`)
 
 func newGolangVersion(v string) (*golangVersion, error) {
+	if v == "(devel)" {
+		return &golangVersion{raw: v}, nil
+	}
 	if !startsWithSemver.MatchString(v) {
 		return nil, fmt.Errorf("%s is not a go version", v)
 	}
@@ -69,7 +72,13 @@ func newGolangVersion(v string) (*golangVersion, error) {
 
 	if semver, err := hashiVer.NewSemver(strings.TrimPrefix(version, "v")); err == nil {
 		result.semVer = semver
+		return result, nil
 	}
 
+	// go stdlib is reported by syft as a go package with version like "go1.24.1"
+	if semver, err := hashiVer.NewSemver(strings.TrimPrefix(version, "go")); err == nil {
+		result.semVer = semver
+		return result, nil
+	}
 	return result, nil
 }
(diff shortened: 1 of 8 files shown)


spiffcs:
Feedback and thoughts here are based off of the `Versions` section found under the `go mod` technical documentation found [here](https://go.dev/ref/mod).

1) With regards to an external library choice: given that the golang specification informs us that it follows Semantic Versioning 2.0.0, what's to stop us from using: https://pkg.go.dev/github.com/hashicorp/go-version as the basis for our comparison logic? It looks like it handles the build metadata suffixes correctly by ignoring them:  https://go.dev/play/p/48oM8Umjw2d

2) given the special prefix `go` or `v`  it might be wise to strip all prefix in this case before using the above library provided by hashicorp.

3) I think the commit section is only a third of the story here.
    - v0.0.0-20191109021931-daa7c04131f5 is a pseudo-version
    - Each Pseudo verion has three parts segmented by `-`
    - The second part is a timestamp (yyyymmddhhmmss), which is the UTC time the revision was created. In Git, this is the commit time, not the author time. Is it reasonable to construct matching based on this knowledge? I think the above library also has this feature - see: https://go.dev/play/p/0GYH5f-mk65

4) The examples under `go version -m` on the `mod ref` explain why `(devel)` is used :

> mod: the module containing the main package. The columns are the module path, version, and sum, respectively. The [main module](https://go.dev/ref/mod#glos-main-module) has the version (devel) and no sum.

If grype is already at a point where the package has been created and `devel` has been determined to be the version, then it means upstream tooling/package creation failed to determine the version by other context aware means. Given the discovery of a devel version I think we should forfeit matching the case. A true positive is luck at best, and a false negative is an indicator that upstream tooling or go itself must provide a common way to identify the main modules published version. See this comment on the upstream syft tooling: https://github.com/anchore/grype/issues/1416#issuecomment-1665172919

@willmurphyscode let me know if I missed anything in your questions or comments about the change! 

Thanks for the detail and explanation on this change

willmurphyscode:
Thanks for the comments @spiffcs! It seems like the semver library we already use handles most of the comparisons we want. I can make the golangVersion an even thinner wrapper around semver. Do you think it makes sense to adjust the regexes we use to detect semver so that they (1) trim the `go` prefix and (2) don't reject a version just because it's a pseudoversion or just contains `+incompatible`? That would be less code to maintain. I didn't realize that some of what go's versioning is doing is part of a standard semver library.

spiffcs:
Yea I think the thinner approach is is correct here. As long as those regex are specific to the golang matcher than that seems like the best way. Special case to strip go for things like `go1.21.1`. We don't even need to strip `v` as the library works with both:

https://go.dev/play/p/wN80SUZJziJ

 We also should not reject a version with those suffixes. Less code 👍

grype/version/golang_version.go
```diff
@@ -0,0 +1,58 @@
+package version
+
+import (
+	"fmt"
+	"strings"
+
+	hashiVer "github.com/anchore/go-version"
```
spiffcs, line 7: Nice thanks for using the internal library!

grype/version/version.go
```diff
@@ -63,6 +64,10 @@ func (v *Version) populate() error {
 		ver, err := newDebVersion(v.Raw)
 		v.rich.debVer = ver
 		return err
+	case GolangFormat:
+		ver, err := newGolangVersion(v.Raw)
+		v.rich.golangVersion = ver
+		return err
```
spiffcs, line 70: Taking a look here - if we do get the error for when something is like `(devel)` then there are 3 references to `NewVersionFromPkg` where the error could be returned. I think the only instance for the case of code changes here would be under `grype/search/language.go` - This is eventually surfaced as a warning so we don't exit grype based on an invalid package version.

I think the warning is good for now since it's how we've handled this kind of error case in the past.

https://github.com/anchore/grype/blob/83e5176127d0d654d9d671504e30d89fa5fd1070/grype/search/criteria.go#L34C1-L38C5

spiffcs (approved):

grype/version/version.go
willmurphyscode: It looks like the devel version is already handled specially by the Go Matcher:

https://github.com/anchore/grype/blob/9750ef245281262b2d606967fd3ce665117b7c0c/grype/matcher/golang/matcher.go#L45-L52

And manual testing indicates that we don't get this error logged as a warning.

It might be worth consolidating our golang version handling in the future, but I think for now at least this won't spam a warning for every go executable.

PR #1602, opened 2023-11-13
fix(apk): find secdb entries for origin packages

luhring:
This fixes a bug in Grype's APK matching logic (affecting Alpine and Wolfi) where if there was an NVD match to a package, Grype wasn't doing the resolution-to-origin-package to find relevant data in the secdb for the match. This created a case where Grype would show a distro package as vulnerable even though the distro had marked the package as fixed or a false positive already.

This PR adds the missing check to the matcher and adds a unit test to cover this case.


luhring:
@wagoodman Just hit this bug again (which produces false positives) — maybe we could try to land this? I can add a test to `matcher_test.go` and then anything else?

PR #1387, opened 2023-07-13
feat: filter out packages owned by OS packages

willmurphyscode:
For example, if the rpm "python3-rpm" is installed, it brings a python package called "rpm" with it, which is just python bindings to RPM. But this python package is part of "python3-rpm", and should not be matched against directly.

Fixes https://github.com/anchore/grype/issues/1373
https://github.com/anchore/grype/issues/1362 won't be fixed anymore because of the comprehensive distro check.

PR #1603, opened 2023-11-15
fix: distro FP data not applied correctly

### LINKED ISSUE #1236: Vulnerabilities marked as fixed in distro packages should be reported as fixed for all contained packages too
**What happened**:

Today Grype applies "fix" data from distro advisory sources to vulnerabilities matched to the distro package (e.g. an `apk`, `rpm`, etc.), which makes sense. But, it doesn't apply the fix data to the software that comprises the distro package. IMHO, this doesn't make sense.

#### Example

The Wolfi package `py3.10-pip` was matched to CVE-2018-20225, which is disputed. After looking into it, the Wolfi maintainers determined the correct action was to NAK the vulnerability in the Wolfi secdb.

When you run Grype against an image with this package, here's what happens:

```console
$ grype -q cgr.dev/chainguard/python:3.10-dev
NAME         INSTALLED   FIXED-IN  TYPE    VULNERABILITY   SEVERITY
pip          23.0.1                python  CVE-2018-20225  High
python-3.10  3.10.11-r0            apk     CVE-2023-24329  High
```

Of note, all of the files that Grype lists as `locations` for `pip` (2 files in this case) are owned by the `py3.10-pip` Wolfi package, according to `/lib/apk/db/installed`.

In this example, Grype shouldn't report the match of CVE-2018-20225 to the python `pip` package. It doesn't make sense for the distro package to be not affected by the vulnerability, but the python package described by the distro package to still be vulnerable _to that same vulnerability_.

```sh
grype -q cgr.dev/chainguard/python:3.10-dev
```

In case it helps for extra context / ringing any bells, I mentioned this to @spiffcs [in Slack](https://anchorecommunity.slack.com/archives/C027JE5M345/p1680286989357059?thread_ts=1680124457.120959&cid=C027JE5M345) and to @wagoodman IRL recently. 😃 

**Environment**:
- Output of `grype version`:
```text
Application:          grype
Version:              0.61.0
Syft Version:         v0.76.0
BuildDate:            2023-04-04T15:11:17Z
GitCommit:            d8c0c0805b59659c4d6e49d6806a0eba11bdc2ee
GitDescription:       v0.61.0
Platform:             darwin/arm64
GoVersion:            go1.19.7
Compiler:             gc
Supported DB Schema:  5
```
- OS (e.g: `cat /etc/os-release` or similar): Windows Vista

luhring:
fixes #1236 

This demonstrates an implementation that more correctly applies distro false positive data to package matches across all matcher types, for the specific case when the file(s) that comprise the package are claimed by the distro's package DB.

In talking with @wagoodman, we agreed that in the long term, this kind of second-pass filtering won't be needed: once Grype has fully implemented "negative matches", a more elegant approach becomes possible; and at the same time, that it still makes sense to pursue more accurate results for Grype in short term, via leveraging the distro's security data correctly to avoid including (language ecosystem) matches to packages owned by distro packages that have been analyzed by the distro's security team.

This implementation appears to be tailored for just a few distros... It would normally apply across all distros, but the matching for many distros (excluding Alpine, Wolfi, and Chainguard) was recently changed in #1387 such that distro-owned language ecosystem packages are exempt from matching altogether (IIUC). Per the discussion in #1236, this approach isn't suitable for the "non-comprehensive-feed" distros since Grype would have no means of finding unfixed vulnerabilities in distro packages.

So instead, this PR proposes a solution that enables Grype to continue finding unfixed vulnerabilities in these distros, while still correctly applying the security data from the distro's security team.

~This PR still needs tests and what not, but~ wanted to start the discussion on the approach itself. Thanks for your time!


luhring:
One secondary issue I noticed while working on this is that it looks like the GHSA entries from the Wolfi and Chainguard secdbs aren't making it into the Grype DB. This contributes to the distro data not getting applied correctly, since fixes/nacks in GHSAs are not available to Grype at time of matching.

Is this something we can look at adjusting?

Example using the "apko" package:

```jsonc
// ...
    {
      "pkg": {
        "name": "apko",
        "secfixes": {
          "0": [
            "CVE-2023-45283",
            "CVE-2023-45284",
            "GHSA-jq35-85cj-fj4p" //  <-- this vuln doesn't have a CVE ID assigned!
          ],
          "0.10.0-r6": [
            "CVE-2023-39325",
            "CVE-2023-3978"
          ],
          "0.7.3-r1": [
            "CVE-2023-28840",
            "CVE-2023-28841",
            "CVE-2023-28842"
          ],
          "0.8.0-r1": [
            "CVE-2023-30551"
          ]
        }
      }
    },
// ...
```

```console
sqlite> SELECT * FROM vulnerability WHERE namespace LIKE '%wolfi%' AND package_name = 'apko';
pk       id              package_name  namespace                   package_qualifiers  version_constraint  version_format  cpes  related_vulnerabilities                          fixed_in_versions  fix_state  advisories
-------  --------------  ------------  --------------------------  ------------------  ------------------  --------------  ----  -----------------------------------------------  -----------------  ---------  ----------
1822413  CVE-2023-28840  apko          wolfi:distro:wolfi:rolling                      < 0.7.3-r1          apk                   [{"id":"CVE-2023-28840","namespace":"nvd:cpe"}]  ["0.7.3-r1"]       fixed
1822417  CVE-2023-28841  apko          wolfi:distro:wolfi:rolling                      < 0.7.3-r1          apk                   [{"id":"CVE-2023-28841","namespace":"nvd:cpe"}]  ["0.7.3-r1"]       fixed
1822421  CVE-2023-28842  apko          wolfi:distro:wolfi:rolling                      < 0.7.3-r1          apk                   [{"id":"CVE-2023-28842","namespace":"nvd:cpe"}]  ["0.7.3-r1"]       fixed
1822441  CVE-2023-30551  apko          wolfi:distro:wolfi:rolling                      < 0.8.0-r1          apk                   [{"id":"CVE-2023-30551","namespace":"nvd:cpe"}]  ["0.8.0-r1"]       fixed
1822558  CVE-2023-39325  apko          wolfi:distro:wolfi:rolling                      < 0.10.0-r6         apk                   [{"id":"CVE-2023-39325","namespace":"nvd:cpe"}]  ["0.10.0-r6"]      fixed
1822724  CVE-2023-3978   apko          wolfi:distro:wolfi:rolling                      < 0.10.0-r6         apk                   [{"id":"CVE-2023-3978","namespace":"nvd:cpe"}]   ["0.10.0-r6"]      fixed
1823032  CVE-2023-45283  apko          wolfi:distro:wolfi:rolling                      < 0                 apk                   [{"id":"CVE-2023-45283","namespace":"nvd:cpe"}]  ["0"]              fixed
1823380  CVE-2023-45284  apko          wolfi:distro:wolfi:rolling                      < 0                 apk                   [{"id":"CVE-2023-45284","namespace":"nvd:cpe"}]  ["0"]              fixed
```

westonsteimel:
Yeah, looks like just an adjustment to vunnel to either not do this filtering at all or adjust it to allow GHSA prefix https://github.com/anchore/vunnel/blob/main/src/vunnel/providers/wolfi/parser.py#L100-L104

westonsteimel:
https://github.com/anchore/vunnel/pull/399 removes the id filtering from the wolfi provider entirely.  I can make the same change for the chainguard provider if everyone agrees with the approach

luhring:
Wow, thank you @westonsteimel!

grype/pkg/package.go
```diff
@@ -216,6 +216,7 @@ func dataFromPkg(p pkg.Package) (interface{}, []UpstreamPackage) {
 			metadata = *m
 		}
 	case pkg.ApkDBEntry:
+		metadata = p.Metadata
```
luhring, line 219: Use a new struct for apkdb specific to what Grype needs for matching

luhring:
@wagoodman This should be ready for a look now! I _think_ it incorporates all of the changes we talked about offline.

Thanks again for the discussion on this ❤️

luhring:
I've found one outstanding issue, related to how upstream vuln data gets discovered — which we had already started talking about with https://github.com/anchore/grype/pull/1602.

I'm wondering if we should tackle it in this PR or in 1602 — but leaning toward this PR...

Here's what Grype shows in a particular scan (incorporating the work from this PR 1603):

```console
$ grype -q cgr.dev/chainguard/argo-workflowcontroller:latest
NAME                      INSTALLED             FIXED-IN  TYPE       VULNERABILITY        SEVERITY
github.com/docker/docker  v24.0.0+incompatible  24.0.7    go-module  GHSA-jq35-85cj-fj4p  Medium
```

We can see that Syft was able to identify the origin package for the APK package that owns the Go module shown above (for brevity, I'm skipping the step of inspecting Grype's JSON for the above scan, but you can trust me!):

```console
$ syft -q cgr.dev/chainguard/argo-workflowcontroller:latest -o json | jq '.artifacts | map(select(.name == "argo-workflow-controller")) | first'
{
  "id": "64b20153a4589427",
  "name": "argo-workflow-controller",
  "version": "3.5.2-r3",
  "type": "apk",
  "foundBy": "apk-db-cataloger",
  // ...
  "metadata": {
    "package": "argo-workflow-controller",
    "originPackage": "argo-workflows",
    // ...
  }
}
```

And, we can see that the Wolfi secdb shows the vulnerability reported from Grype (`GHSA-jq35-85cj-fj4p`) is a false positive:

```console
$ curl -sS https://packages.wolfi.dev/os/security.json | jq '.packages | map(select(.pkg.name == "argo-workflows")) | first'
{
  "pkg": {
    "name": "argo-workflows",
    "secfixes": {
      "0": [
        "CVE-2023-45283",
        "GHSA-vvjp-q62m-2vph",
        "CVE-2023-45284",
        "GHSA-rq3x-83w4-p28c",
        "GHSA-jq35-85cj-fj4p"
      ],
  // ...
}
```

So that data should be consulted when generating the false positive index at the end of matching.

luhring:
Addressed the scenario from my comment above, and added coverage for that scenario in the unit and integration tests ✅

grype/vulnerability_matcher.go
```diff
@@ -122,7 +125,7 @@ func (m *VulnerabilityMatcher) searchDBForMatches(
 	release *linux.Release,
 	packages []pkg.Package,
 	progressMonitor *monitorWriter,
-) match.Matches {
+) (match.Matches, error) {
```
spiffcs, line 128: Does surfacing this error here create a new case where grype can now fail for the following types when we cannot indexFalsePositivesByLocation?
```
`d.Type == distro.Wolfi || d.Type == distro.Chainguard || d.Type == distro.Alpine)`
```

Should we log the error and continue executing grype as we do on main now, or is this a good spot to add this new `grype exits on fail` condition?

grype/vulnerability_matcher.go
```diff
@@ -93,7 +94,10 @@ func (m *VulnerabilityMatcher) findDBMatches(pkgs []pkg.Package, context pkg.Con
 	var ignoredMatches []match.IgnoredMatch
 
 	log.Trace("finding matches against DB")
-	matches := m.searchDBForMatches(context.Distro, pkgs, progressMonitor)
+	matches, err := m.searchDBForMatches(context.Distro, pkgs, progressMonitor)
```
spiffcs, line 97: this is where the error is surfaced from my previous comment for the new `indexFalsePositivesByLocation` error

grype/vulnerability_matcher.go
```diff
@@ -154,26 +165,114 @@ func (m *VulnerabilityMatcher) searchDBForMatches(
...
+					}
+				}
+			}
+		}
+	}
+
+	return result, nil
+}
+
+func filterMatchesUsingDistroFalsePositives(ms []match.Match, falsePositivesByLocation map[string][]string) []match.Match {
```
spiffcs, line 247: 👍

grype/vulnerability_matcher.go
```diff
@@ -154,26 +165,114 @@ func (m *VulnerabilityMatcher) searchDBForMatches(
...
+			// ignored: matches that are filtered out due to user-provided ignore rules
+			// dropped: matches that are filtered out due to hard-coded rules
+			updateVulnerabilityList(progressMonitor, additionalMatches, nil, dropped, m.Store)
+		}
+	}
+
+	return res, nil
+}
+
+func indexFalsePositivesByLocation(
```
spiffcs, line 193: 👍

spiffcs (commented):
Tests and behavior work as described in the PR with the new false positive information being flagged and removed based on the locations used by `getDistroFalsePositivesByLocation`

The only comment I had for this is the new addition of the grype error case - I agree that we should definitely error in the case that the new functionality fails, but am undecided on if it should be `grype [exit]` vs `grype logs some error and moves on with other information`

grype/pkg/syft_sbom_provider_test.go
```diff
@@ -47,6 +47,107 @@ func TestParseSyftJSON(t *testing.T) {
 							Name: "alpine-baselayout",
 						},
 					},
+					Metadata: ApkMetadata{
+						Files: []ApkFileRecord{
+							{Path: "/dev"},
```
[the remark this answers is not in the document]
luhring: Sounds good to me! Yeah I didn't want to leave this test hanging since it started failing, but it'd be awesome to not have to add these!

grype/vulnerability_matcher.go
luhring: That's a good question. I defer to you and @wagoodman here... (Fun fact: we made this particular change standing over the same laptop at ALX!)

I got the impression from Alex that it felt like time to add the `error` return parameter at this point, but I'm not sure where else we'd want to return non-nil errors now.

Let me know what change you'd like me to make, if any — more than happy to.

wagoodman (approved):
thanks for the fix here 🙌

grype/vulnerability_matcher.go
wagoodman: this is a pretty fundamental function -- if there really is an error condition here it could lead to missed results, which should always return 1 on the CLI (thus we should not log and continue).

grype/pkg/syft_sbom_provider_test.go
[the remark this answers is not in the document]
wagoodman: I think it's a little unfortunate that I picked a test fixture that had so many file entires... but this probably is not a blocker for the fix

PR #1646, opened 2023-12-28
Fix logging configuration in README

### LINKED ISSUE #1645: Parameter `quiet` is ignored in configuration file
**What happened**: Setting `quiet: true` in YAML config file produces no effect, even though this parameter is mentioned in README. Same applies to `GRYPE_QUIET` env variable.

**What you expected to happen**: `quiet` from YAML file and `GRYPE_QUIET` form env should be taken into account.

**How to reproduce it (as minimally and precisely as possible)**: Set `quiet: true` in `.grype.yaml`. There is no effect.

**Anything else we need to know?**: I've investigated the source code and found that there is no parameter for quiet in file `/cmd/grype/cli/options/grype.go`. I've also found PR https://github.com/anchore/grype/pull/1437 that introduced CLIO as the base for CLI, which delegated handling of quiet to CLIO, and, I believe, removed it from the Grype config.

**Environment**:
- Output of `grype version`: 0.73.5
- OS (e.g: `cat /etc/os-release` or similar): Ubuntu 22.04.3 LTS

plavy:
After porting to CLIO, README was not updated to reflect changes in the logging configuration. The file is now updated based on `grype --help` output which is already up to date.

This closes #1645

### DIFF
--- a/README.md
+++ b/README.md
@@ -683,10 +683,6 @@ fail-on-severity: ""
 # same as -o ; GRYPE_OUTPUT env var
 output: "table"
 
-# suppress all output (except for the vulnerability list)
-# same as -q ; GRYPE_QUIET env var
-quiet: false
-
 # write output report to a file (default is to write to stdout)
 # same as --file; GRYPE_FILE env var
 file: ""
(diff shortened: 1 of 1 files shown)


kzantow (approved):

kzantow:
Thanks for the contribution, @plavy !

Issue #1652, opened 2024-01-08
Build Docker Image with Database Packaged


kchason:
**What would you like to be added**:

A Docker image that has the database packaged into the image and disables the automatic update by default.

**Why is this needed**:

The offline/air-gapped instructions still require the database to be hosted or sideloaded. This is an extra step versus downloading and copying a Docker image to an offline environment.

**Additional context**:

There is a pull request with this capability using the `packaged` suffix (similar to the `debug` one) at https://github.com/anchore/grype/pull/1653.

wagoodman:
Adding this image wouldn't be as beneficial as it initially appears. Grype has a default configuration to fail when the DB is > 5 days old to ensure that users do not use old DBs by default. For this reason we would need to publish images nearly daily to ensure that users have the most up to data DB (which is much more frequently than we're releasing today). This means for an air-gapped solution, the DB would either need to be pulled often anyway or the configuration changed to a longer stale period for the DB age.

Alternatively, when preparing the image for transfer to air-gapped systems, you could build your own image easily in an ad-hoc way:

```dockerfile
FROM anchore/grype:latest

# disable auto updating the DB
ENV GRYPE_DB_AUTO_UPDATE false

# set this to what you need in hours
ENV GRYPE_DB_MAX_ALLOWED_BUILT_AGE 120h

RUN ["/grype", "db", "update"]
```

```
$ docker build -t grype-with-db:latest .
$ docker run --rm -it grype-with-db:latest alpine:latest
 ✔ Parsed image                                                                                                                        sha256:1dc785547989b0db1c3cd9949c57574393e69bea98bfe044b0588e24721aa402
 ✔ Scanned for vulnerabilities     [14 vulnerability matches]
   ├── by severity: 0 critical, 0 high, 12 medium, 0 low, 0 negligible (2 unknown)
   └── by status:   2 fixed, 12 not-fixed, 0 ignored
NAME           INSTALLED   FIXED-IN  TYPE  VULNERABILITY   SEVERITY
busybox        1.36.1-r15            apk   CVE-2023-42366  Medium
busybox        1.36.1-r15            apk   CVE-2023-42365  Medium
busybox        1.36.1-r15            apk   CVE-2023-42364  Medium
busybox        1.36.1-r15            apk   CVE-2023-42363  Medium
busybox-binsh  1.36.1-r15            apk   CVE-2023-42366  Medium
busybox-binsh  1.36.1-r15            apk   CVE-2023-42365  Medium
[...]

PR #1653, opened 2024-01-08
Add Packaged Dockerfile (#1652)

kchason:
This addresses #1652 to build a Docker image that has the vulnerability database packaged in the image.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -205,6 +205,60 @@ dockers:
       - "--build-arg=VCS_REF={{.FullCommit}}"
       - "--build-arg=VCS_URL={{.GitURL}}"
 
+  - image_templates:
+      - anchore/grype:packaged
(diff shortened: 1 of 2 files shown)


wagoodman:
As mentioned [on the issue](https://github.com/anchore/grype/issues/1652#issuecomment-1888051744) there is a straightforward workaround for this:
```dockerfile
FROM anchore/grype:latest

# disable auto updating the DB
ENV GRYPE_DB_AUTO_UPDATE false

# set this to what you need in hours
ENV GRYPE_DB_MAX_ALLOWED_BUILT_AGE 120h

RUN ["/grype", "db", "update"]
```

... and there are aspects of the image distribution that don't make sense with the default configuration of grype and the needed release cadence of the images today. If I missed an aspect of this solution, please shout out here or on the issue, but in the meantime I'm going to close this PR.

Issue #1639, opened 2023-12-18
VEX documents not taken into account when `--fail-on` is set


ferozsalam:
**What happened**:

Using the `--fail-on` flag doesn't appear to take into account the contents of a VEX document that has been provided using the `--vex` flag - even if several CVEs have been marked as `not_affected` in the VEX document, the command still fails when the `--fail-on` flag is set.

This is inconsistent with the behaviour shown when a `.grype.yaml` file is used with an `ignore` block defined, despite VEX documents and the `ignore` block having similar purposes.

**What you expected to happen**:

If Grype is invoked with both `--vex` and `--fail-on` set, and all CVEs in the scanned image are defined as `not_affected` in the VEX document, I would expect the command to pass successfully, as it does when all such CVEs are defined in a `.grype.yaml` `ignore` block.

**How to reproduce it (as minimally and precisely as possible)**:

Create the following VEX doc and save it as `test-vex.json`:

```
{
  "@context": "https://openvex.dev/ns/v0.2.0",
  "@id": "https://openvex.dev/docs/public/vex-cb58dca06abcf3da095d3f5e6126128aeb39b145e57d4eac468483eb75b5c014",
  "author": "Unknown Author",
  "timestamp": "2023-12-18T12:00:54.292255Z",
  "last_updated": "2023-12-18T12:01:58.35081Z",
  "version": 4,
  "statements": [
    {
      "vulnerability": {
        "name": "CVE-2023-42366"
      },
      "timestamp": "2023-12-18T12:00:54.292256Z",
      "products": [
        {
          "@id": "pkg:oci/busybox"
        }
      ],
      "status": "not_affected",
      "justification": "vulnerable_code_not_in_execute_path"
    },
    {
      "vulnerability": {
        "name": "CVE-2023-42365"
      },
      "timestamp": "2023-12-18T12:01:17.799359Z",
      "products": [
        {
          "@id": "pkg:oci/busybox"
        }
      ],
      "status": "not_affected",
      "justification": "vulnerable_code_not_in_execute_path"
    },
    {
      "vulnerability": {
        "name": "CVE-2023-42364"
      },
      "timestamp": "2023-12-18T12:01:41.069995Z",
      "products": [
        {
          "@id": "pkg:oci/busybox"
        }
      ],
      "status": "not_affected",
      "justification": "vulnerable_code_not_in_execute_path"
    },
    {
      "vulnerability": {
        "name": "CVE-2023-42363"
      },
[...]

PR #1657, opened 2024-01-14
fix: take VEX docs into account when --fail-on is set

ferozsalam:
Previously, VEX documents provided to Grype when --fail-on was set were not
taken into account. That led to inconsistent behaviour where a vulnerability
would be ignored when only `--vex` was specified, but would be included in
Grype output when both `--vex` and `--fail-on` were specified.

This change fixes that by moving the failure severity check to after the VEX
documents provided are tested.

I have also added a unit test to check that the combination of VEX docs and
failure severity checks works as expected.

I also took the opportunity to fix some typos while I was poking around.

fixes https://github.com/anchore/grype/issues/1639


ferozsalam:
Hi @tgerla – I was wondering what sort of timescales I should expect for a review on this change?

wagoodman (approved):

PR #1675, opened 2024-01-24
chore: break assumption that syft cpe.CPE is wfn.Attributes

willmurphyscode:
Previously, Syft's cpe.CPE type was an alias for wfn.Attributes. Fix a couple places where Grype's compilation depended on that fact, since it will stop being true in the next Syft release.

See anchore/syft#2529

### DIFF
--- a/grype/cpe/cpe.go
+++ b/grype/cpe/cpe.go
@@ -1,9 +1,10 @@
 package cpe
 
 import (
+	"github.com/facebookincubator/nvdtools/wfn"
+
 	"github.com/anchore/grype/internal/log"
 	"github.com/anchore/syft/syft/cpe"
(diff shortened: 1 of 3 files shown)


willmurphyscode:
Note that this code probably can't be merged until the next Syft is out, because of this linter failure:

```
grype/cpe/cpe.go:26:21: unnecessary conversion (unconvert)
	a := wfn.Attributes(c)
	                   ^
grype/cpe/cpe.go:28:28: unnecessary conversion (unconvert)
		canCopy := wfn.Attributes(candidate)
		                         ^
make: *** [lint] Error 1
```

I'm leaving this as a draft PR whose commit should be pulled into the next PR that will bump Syft in Grype, to fix a couple small compile-time errors.

wagoodman (approved):

PR #1720, opened 2024-02-17
fix: improve sarif descriptive text and fingerprint

### LINKED ISSUE #1715: GitHub code scanning alerts missing information
I'm using [scan-action](https://github.com/anchore/scan-action) to scan multiple images and upload the results to GitHub code scanning, but it's not immediately clear which image is affected at first glance.

image

From the screenshot, you can see that the title just says the name of the package and then there's what seems like an URI that seems to be missing something:
```
image//lib/apk/db/installed:1
```

I can see that the distro is alpine if I click "show more", but if I have multiple images based on alpine I don't have a way to know which one is affected.

It would be nice to have the name of the docker image somewhere in the report. 

I previously opened https://github.com/anchore/scan-action/issues/227 but I think the problem originates here.

kzantow:
This PR improves the descriptive text when outputting SARIF to display more information to the user including the image/ and tag or sha that was specified. [An example of what this looks like in GitHub can be seen here](https://github.com/kzantow-anchore/test-sarif/security/code-scanning/38).

Fixes #1715

### DIFF
--- a/grype/presenter/sarif/presenter.go
+++ b/grype/presenter/sarif/presenter.go
@@ -1,6 +1,7 @@
 package sarif
 
 import (
+	"crypto/sha256"
 	"fmt"
 	"io"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


jodyrhodes1 (commented):
Rhodes@ETradeenterprises.net

spiffcs:
@kzantow I saw you added another commit for checks against the image name to make sure it's valid for githubs API. Are there are changes going into this PR?

The new commit with tests LGTM

kzantow:
@spiffcs -- just waiting a bit to see if the original issue reporter, @lucacome, is able to confirm this should work for their needs; no other planned changes unless feedback dictates

PR #1721, opened 2024-02-17
Remove unused file internal/logger/logrus.go

seiyab:
I encountered a problem reported in https://github.com/anchore/go-logger/pull/13. When I tried to fix it, this misleading dead code confused me. This file looks to get unused at https://github.com/anchore/grype/pull/1026 so I remove it to prevent confusion.


kzantow (approved):
Good catch, thanks @seiyab !

PR #1724, opened 2024-02-21
remove unused file internal/file/tar.go and its test

seiyab:
**TL;DR**
internal/file/tar.go looks unused since https://github.com/anchore/grype/pull/657.

**Context**
I scanned bugs similar to https://github.com/anchore/go-logger/pull/13 just with curiosity and found that internal/file/tar.go has one (written file can be broken forgetting truncation). Investigating it, I also found that it isn't used. Running `git bisect run sh -c "git grep -q 'file.UnTarGz'"`, https://github.com/anchore/grype/pull/657 looks to remove the last use of the file.


kzantow (approved):
Nice catch, thanks again @seiyab!

PR #1725, opened 2024-02-23
fix(install): return right exit codes

hacst:
As the main function was not the last command, it did always return a zero status code. Exiting right after main allows us to return the actual return code instead.

Analogous to https://github.com/anchore/syft/pull/2664 .

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -695,5 +695,6 @@ set +u
 if [ -z "${TEST_INSTALL_SH}" ]; then
   set -u
   main "$@"
+  exit $?
 fi
 set -u


kzantow (approved):
Thanks again @hacst

hacst:
No problem. Thanks for the great project!

I see a validation failure complaining about a test db that is too old. Will you take care of that or should I try to figure that out?

kzantow:
@hacst -- I'll get this updated when https://github.com/anchore/grype/pull/1726 merges

PR #1739, opened 2024-03-06
fix: make bootstrap-tools failed

testwill:
GO111MODULE=off GOBIN=/home/xxx/code/grype/.tmp go get -u golang.org/x/perf/cmd/benchstat
go: modules disabled by GO111MODULE=off; see 'go help modules'
make: *** [Makefile:112: bootstrap-tools] Error 1


kzantow:
@testwill could you give more explanation why this change is needed? the `make bootstrap` command works fine for me without the change here on macOS and `ubuntu:latest`.

spiffcs (approved):

Issue #1731, opened 2024-02-28
Grype scan command appears to hang when downloading db or listing file


githala-deepak:
**What happened**:
Grype command gets stuck and I get the error after 3 hours 
failed to load vulnerability db: unable to update vulnerability database: unable to download db: stream error: stream ID 1; INTERNAL_ERROR; received from peer
**What you expected to happen**:
Grype scan should get completed in under a minute
**How to reproduce it (as minimally and precisely as possible)**:
Occurs randomly, can't reproduce
**Anything else we need to know?**:

**Environment**:
- Output of `grype version`: Application:         grype
Version:             0.74.5
BuildDate:           2024-02-07T21:34:47Z
GitCommit:           74780902eda090afa5ca36699da2453f2df2e655
GitDescription:      v0.74.5
Platform:            linux/amd64
GoVersion:           go1.21.6
Compiler:            gc
Syft Version:        v0.104.0
Supported DB Schema: 5

- OS (e.g: `cat /etc/os-release` or similar): PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
NAME="Debian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"

willmurphyscode:
Hi @githala-deepak,

Thanks for the report.

It sounds like grype is having trouble downloading its updated vulnerability DB, which it will try to do about once per day.

If you run `grype db update -vvv`, do you seen any errors?

If you download the db directly, with a command like this:

``` sh
curl -vvv -o /tmp/db.tar.gz 'https://toolbox-data.anchore.io/grype/databases/vulnerability-db_v5_2024-02-28T01:23:28Z_ea5efb77a61bf939917f.tar.gz'
```

Do you see any errors? Does the download succeed? I think you probably need to troubleshoot a network issue, and that curl command will start you in the right direction.

hkadakia:
I am having a similar issue. 
```
Syft: Summary of packages by <count> <type>
00:03:13 See mediaimage.syft.json for full package details
00:03:13     122 "go-module"
00:03:13       3 "python"
00:03:13     159 "rpm"
00:03:13 
00:03:13 Grype: scanning for vulnerabilities 
[...]

PR #1777, opened 2024-04-01
fix: enable http timeout

willmurphyscode:
Otherwise grype db commands can hang if the CDN is having issues.

Related to https://github.com/anchore/grype/issues/1731

## Some notes

We don't want commands like `grype db update` to hang indefinitely, and we certain don't want `grype <some-source>` to hang indefinitely.

Grype is already intended to continue on error if db updates are not available: https://github.com/anchore/grype/blob/fix%2Foverall-http-timeout/grype/db/curator.go#L142, so after this change, a timeout error when checking for update or downloading the database will result in a logged warning.

30 seconds overall timeout seems be reasonable. The compressed grype db as of this writing is ~152MB .tar.gz file when downloaded. I've done a few `time curl -o /dev/null "${DB_URL}"` for grype dbs published on different days, and the highest I've seen is about 8 seconds.

## Questions

1. Do we want to expose the timeout as a config variable?
2. Do we want to use separate timeouts for downloading listing.json (small JSON doc) vs database (much larger tar.gz), maybe making a shorter timeout for the listing.json and a longer timeout for the database?

## Manual testing done:

With this change:
```
( run a server on localhost:5000 that accepts connections but never replies)
$ GRYPE_DB_UPDATE_URL=http://localhost:5000/listing.json go run cmd/grype/main.go alpine:latest
[0000]  INFO grype version: [not provided]
[0060]  WARN unable to check for vulnerability database update << prints warning about failed update, but continues
[0060]  INFO found 12 vulnerability matches across 15 packages
NAME           INSTALLED   FIXED-IN  TYPE  VULNERABILITY   SEVERITY
busybox        1.36.1-r15            apk   CVE-2023-42366  Medium
busybox        1.36.1-r15            apk   CVE-2023-42365  Medium
busybox        1.36.1-r15            apk   CVE-2023-42364  Medium
... snip ...
```

### DIFF
--- a/grype/db/curator.go
+++ b/grype/db/curator.go
@@ -390,6 +390,7 @@ func (c Curator) ListingFromURL() (Listing, error) {
 
 func defaultHTTPClient(fs afero.Fs, caCertPath string) (*http.Client, error) {
 	httpClient := cleanhttp.DefaultClient()
+	httpClient.Timeout = 30 * time.Second
 	if caCertPath != "" {
 		rootCAs := x509.NewCertPool()
 
--- a/grype/db/curator.go
+++ b/grype/db/curator.go
@@ -37,11 +37,14 @@ type Config struct {
 	ValidateByHashOnGet bool
 	ValidateAge         bool
 	MaxAllowedBuiltAge  time.Duration
+	ListingFileTimeout  time.Duration
+	UpdateTimeout       time.Duration
 }
 
 type Curator struct {
 	fs                  afero.Fs
-	downloader          file.Getter
+	listingDownloader   file.Getter
+	updateDownloader    file.Getter
 	targetSchema        int
 	dbDir               string
 	dbPath              string
@@ -55,7 +58,14 @@ func NewCurator(cfg Config) (Curator, error) {
 	dbDir := path.Join(cfg.DBRootDir, strconv.Itoa(vulnerability.SchemaVersion))
 
 	fs := afero.NewOsFs()
-	httpClient, err := defaultHTTPClient(fs, cfg.CACert)
+	listingClient, err := defaultHTTPClient(fs, cfg.CACert)
+	listingClient.Timeout = cfg.ListingFileTimeout
+	if err != nil {
+		return Curator{}, err
+	}
+
+	dbClient, err := defaultHTTPClient(fs, cfg.CACert)
+	dbClient.Timeout = cfg.UpdateTimeout
 	if err != nil {
 		return Curator{}, err
 	}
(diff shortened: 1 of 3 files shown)


kzantow (dismissed):

willmurphyscode:
I think there's some legitimate concern that 30 seconds is not always long enough to download the 150 mb grype-db, and the timeouts should definitely be configurable. I'll push an update to:

1. Put the default timeout for fetching the listing file to be much lower than for fetching the database, since it's a much smaller file, and it gets fetched every time (to see whether there's a newer database available)
2. Expose both as config and environment variables, so that users can make their own decision on how long it's reasonable to wait for the download.

The particular complaint in #1731 is that the download starts promptly but is very slow. This symptom means that we can't use a dial timeout or a response headers timeout, since the delay is in actually sending the bytes of the response _body_. However, it might be worth also trying to tune these finer grained timeouts a bit, with the goal that if the network or the database or listing file downloads are just unreachable, grype would block for as little time as possible.

grype/db/curator.go
```diff
@@ -55,15 +58,23 @@ func NewCurator(cfg Config) (Curator, error) {
 	dbDir := path.Join(cfg.DBRootDir, strconv.Itoa(vulnerability.SchemaVersion))
 
 	fs := afero.NewOsFs()
-	httpClient, err := defaultHTTPClient(fs, cfg.CACert)
+	listingClient, err := defaultHTTPClient(fs, cfg.CACert)
```
willmurphyscode, line 61: It would be nice to have better testing on this logic, which suggests there could be some refactoring that would help here.

grype/db/curator_test.go
```diff
@@ -364,3 +373,77 @@ func TestCurator_validateStaleness(t *testing.T) {
...
+		success <- struct{}{}
+	}
+	go runTheTest(success, errs)
+
+	select {
+	case <-success:
+		return
+	case err := <-errs:
+		t.Error(err)
+	case <-timeout:
```
spiffcs, line 446: Since we're doing a test that combines the calls of `IsUpdateAvailable` and `UpdateTo` is there a chance this can race incorrectly? 

I guess the gap is so large from the configured timeouts and `failAfter` (more than 10x) that it doesn't matter.

LGTM

spiffcs (approved):

grype/db/curator_test.go
willmurphyscode: @spiffcs I wanted to make sure I understand what you mean by "race incorrectly." I think you're saying that since (1) the test fails on a timeout and (2) the assertion expects a different timeout, we could get the wrong timeout. Is that correct?

I think that since we're effectively racing a 0.8 second timer against a 10 second timer, the test should be pretty stable. I put in the 10 second timer because otherwise removing one of the timeouts causes the test suite to hang for 10 minutes while the overall testing timer runs out, which seems like worse behavior.

Please let me know if I've addressed these concerns adequately and then I'll merge.

Issue #1762, opened 2024-03-20
Grype returns indirect vulnerability matches on 'kernel' source rpms for the 'kernel-headers' rpm for container images that don't have the kernel


zhill:
**What happened**:
Build an image with `gcc` installed from RedHat/UBI 8, and scan it with Grype and note vulnerabilities found on the 'kernel-headers' package that are matched via indirect lookups on the 'kernel' source RPM.

**What you expected to happen**:
I don't expect to see CVEs on the kernel-headers package since it does not contain the actual kernel.

**How to reproduce it (as minimally and precisely as possible)**:
Build an image with this Dockerfile:
```
FROM registry.access.redhat.com/ubi8/ubi
RUN yum -y install --setopt=tsflags=nodocs gcc
```

Match Details:
```
 ...
 {
   "vulnerability": {
    "id": "CVE-2022-0886",
    "dataSource": "https://access.redhat.com/security/cve/CVE-2022-0886",
    "namespace": "redhat:distro:redhat:8",
    ...
    
 "matchDetails": [
    {
     "type": "exact-indirect-match",
     "matcher": "rpm-matcher",
     "searchedBy": {
      "distro": {
       "type": "redhat",
       "version": "8.9"
      },
      "namespace": "redhat:distro:redhat:8",
      "package": {
       "name": "kernel",
       "version": "4.18.0-513.18.1.el8_9"
      }
     },
     "found": {
      "versionConstraint": "none (rpm)",
      "vulnerabilityID": "CVE-2022-0886"
     }
    }
   ],
    "artifact": {
    "id": "1468511e21d69381",
    "name": "kernel-headers",
    "version": "4.18.0-513.18.1.el8_9",
    "type": "rpm",
    "locations": [
     {
      "path": "/var/lib/rpm/Packages",
      "layerID": "sha256:d46a662dc8788fc7780786c5d9ff576b615ebcea65b6027e9736aa022f3c84c0"
     }
    ],
    "language": "",
    "licenses": [
     "GPLv2 and Redistributable, no modification permitted"
    ],
    "cpes": [
     "cpe:2.3:a:kernel-headers:kernel-headers:4.18.0-513.18.1.el8_9:*:*:*:*:*:*:*",
     "cpe:2.3:a:kernel-headers:kernel_headers:4.18.0-513.18.1.el8_9:*:*:*:*:*:*:*",
[...]

PR #1787, opened 2024-04-04
fix: adds ignore rules for kernel-headers indirect matches

zhill:
Adds ignoring of kernel-headers indirect matches on kernel vulns since the kernel-headers package does not have the kernel code in it that kernel vulns are actually referring to.

Adds a config value to control this ignore behavior that defaults to enabling the ignore rules.

Fixes: #1762

* Adds ignore rule support for match types and upstream package names.
* Adds default ignore rules for kernel-headers indirect matches on kernel for rpms.

### DIFF
--- a/cmd/grype/cli/options/grype.go
+++ b/cmd/grype/cli/options/grype.go
@@ -11,30 +11,31 @@ import (
 )
 
 type Grype struct {
-	Outputs                []string           `yaml:"output" json:"output" mapstructure:"output"`                                           // -o, <presenter>=<file> the Presenter hint string to use for report formatting and the output file
-	File                   string             `yaml:"file" json:"file" mapstructure:"file"`                                                 // --file, the file to write report output to
-	Distro                 string             `yaml:"distro" json:"distro" mapstructure:"distro"`                                           // --distro, specify a distro to explicitly use
-	GenerateMissingCPEs    bool               `yaml:"add-cpes-if-none" json:"add-cpes-if-none" mapstructure:"add-cpes-if-none"`             // --add-cpes-if-none, automatically generate CPEs if they are not present in import (e.g. from a 3rd party SPDX document)
-	OutputTemplateFile     string             `yaml:"output-template-file" json:"output-template-file" mapstructure:"output-template-file"` // -t, the template file to use for formatting the final report
-	CheckForAppUpdate      bool               `yaml:"check-for-app-update" json:"check-for-app-update" mapstructure:"check-for-app-update"` // whether to check for an application update on start up or not
-	OnlyFixed              bool               `yaml:"only-fixed" json:"only-fixed" mapstructure:"only-fixed"`                               // only fail if detected vulns have a fix
-	OnlyNotFixed           bool               `yaml:"only-notfixed" json:"only-notfixed" mapstructure:"only-notfixed"`                      // only fail if detected vulns don't have a fix
-	IgnoreStates           string             `yaml:"ignore-states" json:"ignore-wontfix" mapstructure:"ignore-wontfix"`                    // ignore detections for vulnerabilities matching these comma-separated fix states
-	Platform               string             `yaml:"platform" json:"platform" mapstructure:"platform"`                                     // --platform, override the target platform for a container image
-	Search                 search             `yaml:"search" json:"search" mapstructure:"search"`
-	Ignore                 []match.IgnoreRule `yaml:"ignore" json:"ignore" mapstructure:"ignore"`
-	Exclusions             []string           `yaml:"exclude" json:"exclude" mapstructure:"exclude"`
-	DB                     Database           `yaml:"db" json:"db" mapstructure:"db"`
-	ExternalSources        externalSources    `yaml:"external-sources" json:"externalSources" mapstructure:"external-sources"`
-	Match                  matchConfig        `yaml:"match" json:"match" mapstructure:"match"`
-	FailOn                 string             `yaml:"fail-on-severity" json:"fail-on-severity" mapstructure:"fail-on-severity"`
-	Registry               registry           `yaml:"registry" json:"registry" mapstructure:"registry"`
-	ShowSuppressed         bool               `yaml:"show-suppressed" json:"show-suppressed" mapstructure:"show-suppressed"`
-	ByCVE                  bool               `yaml:"by-cve" json:"by-cve" mapstructure:"by-cve"` // --by-cve, indicates if the original match vulnerability IDs should be preserved or the CVE should be used instead
-	Name                   string             `yaml:"name" json:"name" mapstructure:"name"`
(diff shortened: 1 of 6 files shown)


zhill:
Opening as a draft for initial feedback. I'll address any concerns in UX, testing, etc and then move it to ready for review

spiffcs (commented):
Thanks for the PR @zhill - I took a look and like this approach of using the built in ignore rules as the hook for this option. 

I'll take a look this afternoon at what unit tests should look like for this feature.

I also noted that the quality gate is now failing with this PR so will examine to see how I can help fix that for you as well.

cmd/grype/cli/options/grype.go
```diff
@@ -34,7 +34,8 @@ type Grype struct {
 	Name                   string             `yaml:"name" json:"name" mapstructure:"name"`
 	DefaultImagePullSource string             `yaml:"default-image-pull-source" json:"default-image-pull-source" mapstructure:"default-image-pull-source"`
 	VexDocuments           []string           `yaml:"vex-documents" json:"vex-documents" mapstructure:"vex-documents"`
-	VexAdd                 []string           `yaml:"vex-add" json:"vex-add" mapstructure:"vex-add"` // GRYPE_VEX_ADD
+	VexAdd                 []string           `yaml:"vex-add" json:"vex-add" mapstructure:"vex-add"`                                     // GRYPE_VEX_ADD
+	ShowKernelHeaders      bool               `yaml:"show-kernel-headers" json:"show-kernel-headers" mapstructure:"show-kernel-headers"` // Show matches on kernel-headers packages where the match is on kernel upstream instead of marking them as ignored, default=false
```
kzantow, line 38: Maybe this should be a little more specific like `include-kernel-headers-with-kernel-upstream` or similar, what do you think?

kzantow (commented):
Overall this looks pretty good to me

zhill:
> Thanks for the PR @zhill - I took a look and like this approach of using the built in ignore rules as the hook for this option.
> 
> I'll take a look this afternoon at what unit tests should look like for this feature.
> 
> I also noted that the quality gate is now failing with this PR so will examine to see how I can help fix that for you as well.

Ack, yes I knew the quality gate would need work once i get 👍 on the approach. Happy to resolve that once we're good on the things like option naming and rule details.

cmd/grype/cli/options/grype.go
```diff
@@ -133,6 +135,11 @@ func (o *Grype) AddFlags(flags clio.FlagSet) {
 		"vex", "",
 		"a list of VEX documents to consider when producing scanning results",
 	)
+
+	flags.BoolVarP(&o.ShowKernelHeaders,
+		"show-kernel-headers", "",
+		"do not ignore matches for kernel-headers packages if the match is on the kernel upstream package",
+	)
```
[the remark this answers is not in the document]
zhill: ack. I can do that.

grype/match/ignore.go
```diff
@@ -24,15 +24,17 @@ type IgnoreRule struct {
...
-	Version  string `yaml:"version" json:"version" mapstructure:"version"`
-	Language string `yaml:"language" json:"language" mapstructure:"language"`
-	Type     string `yaml:"type" json:"type" mapstructure:"type"`
-	Location string `yaml:"location" json:"location" mapstructure:"location"`
+	Name         string `yaml:"name" json:"name" mapstructure:"name"`
+	Version      string `yaml:"version" json:"version" mapstructure:"version"`
+	Language     string `yaml:"language" json:"language" mapstructure:"language"`
+	Type         string `yaml:"type" json:"type" mapstructure:"type"`
+	Location     string `yaml:"location" json:"location" mapstructure:"location"`
+	UpstreamName string `yaml:"upstream-name" json:"upstream-name" mapstructure:"upstream-name"`
```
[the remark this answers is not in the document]
zhill: ah, good catch. I wasn't sure if there would later need to be things like upstream-version constraints as well. My gut right now is to make the json match "upstream-name" instead.

cmd/grype/cli/options/grype.go
zhill: yeah, I agree with specificity, just gets very verbose. I'm thinking "include-upstream-kernel-matches-on-headers"? which is a mouthful, but if not a command-line option, then may be not too bad.

cmd/grype/cli/options/grype.go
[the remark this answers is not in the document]
zhill: should be good now.

cmd/grype/cli/options/grype.go
zhill: I went with: `show-kernel-headers-matched-upstream-kernel`. Good, or do you prefer 'include' to 'show'? I was used 'show' to indicate the matches are present just ignored vs not included at all, but I wasn't sure about the precedent for that. I only saw the `show-suppressed` option that was similar. I'm not strongly opinionated on this, so if 'include' is better, I'll change it.

zhill:
I've added tests and I have a pr for the quality tests: https://github.com/anchore/vulnerability-match-labels/pull/132

What's the process for getting the quality to pass? Do I wait for that to merge and then update the submodule to that commit in main?

willmurphyscode:
@zhill I've approved anchore/vulnerability-match-labels#132.

> What's the process for getting the quality to pass? Do I wait for that to merge and then update the submodule to that commit in main?

I'd recommend updating the submodule in this branch. It's also fine to update the submodule in `main` in a separate PR, and merge/rebase that change into this PR, but in the case where the set of expected matches changes, like in this case, it's possible that a PR to main that changes labels will fail, because it's missing the logic in this PR. Does that make sense?

cmd/grype/cli/options/grype.go
```diff
@@ -11,30 +11,31 @@ import (
...
+	Name                                   string             `yaml:"name" json:"name" mapstructure:"name"`
+	DefaultImagePullSource                 string             `yaml:"default-image-pull-source" json:"default-image-pull-source" mapstructure:"default-image-pull-source"`
+	VexDocuments                           []string           `yaml:"vex-documents" json:"vex-documents" mapstructure:"vex-documents"`
+	VexAdd                                 []string           `yaml:"vex-add" json:"vex-add" mapstructure:"vex-add"`                                                                                                             // GRYPE_VEX_ADD
+	ShowKernelHeadersMatchedUpstreamKernel bool               `yaml:"show-kernel-headers-matched-upstream-kernel" json:"show-kernel-headers-matched-upstream-kernel" mapstructure:"show-kernel-headers-matched-upstream-kernel"` // Show matches on kernel-headers packages where the match is on kernel upstream instead of marking them as ignored, default=false
```
kzantow, line 38: The reason I was looking to avoid the term "show" is that the `show-suppressed` is actually only for the presentation in the table view whereas this is excluding matches, affecting other parts like `--fail-on`.

How about calling this `MatchUpstreamKernelHeaders`?

zhill: I'll make that change asap.

PR #1797, opened 2024-04-10
fix: use Go main module version when possible

luhring:
Previously, if Grype saw a Go binary's main module package version was set to a string prefixed by `v0.0.0-`, it would **avoid any kind of vulnerability matching** for that package.

Meanwhile, Syft makes [several reasonable attempts](https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/golang/parse_go_binary.go#L132-L149) to find a sane version to use as this main module's version. In the case of VCS information being embedded in the Go binary, Syft has enough information to produce a Go pseudoversion.

The "no vulnerability matching at all" approach may have been to avoid some false positive edge cases, or it may have predated some of the enhancements to Syft's version computation. But in any case, we were seeing this behavior result in false negatives, because many Go projects produce Go binaries (and are thus the main module for that binary), and when Syft can compute a pseudoversion from VCS data, and these projects have GHSAs filed for them using pseudoversions (e.g. GHSA-xx8w-mq23-29g4), Grype was missing these matches 100% of the time.

To be upfront, I do expect this change to yield an increase in false positives. But I expect this change to yield a _much bigger_ decrease in false negatives. If I understand the current Grype philosophy, distros with "comprehensive feeds" aren't even subject to this kind of matching, so it's really just the distros that are left (Wolfi, Chainguard, and Alpine), and then non-distro Go binaries where the developer is likely in much more control of the version information embedded in the binary.

For context, this change was discussed in advance in Anchore's community slack with @kzantow. ([thread link](https://anchorecommunity.slack.com/archives/C027JE5M345/p1712697730521309))

Let me know if I can answer any questions!

### DIFF
--- a/grype/matcher/golang/matcher.go
+++ b/grype/matcher/golang/matcher.go
@@ -42,11 +42,14 @@ func (m *Matcher) Match(store vulnerability.Provider, d *distro.Distro, p pkg.Pa
 		mainModule = m.MainModule
 	}
 
-	// Golang currently does not have a standard way of incorporating the vcs version
-	// into the compiled binary: https://github.com/golang/go/issues/50603
-	// current version information for the main module is incomplete leading to multiple FP
-	// TODO: remove this exclusion when vcs information is included in future go version
-	isNotCorrected := strings.HasPrefix(p.Version, "v0.0.0-") || strings.HasPrefix(p.Version, "(devel)")
+	// Golang currently does not have a standard way of incorporating the main
+	// module's version into the compiled binary:
+	// https://github.com/golang/go/issues/50603.
+	//
+	// Syft has some fallback mechanisms to come up with a more sane version value
+	// depending on the scenario. But if none of these apply, the Go-set value of
+	// "(devel)" is used, which is altogether unhelpful for vulnerability matching.
+	isNotCorrected := strings.HasPrefix(p.Version, "(devel)")
 	if p.Name == mainModule && isNotCorrected {
 		return matches, nil
 	}
--- a/grype/matcher/golang/matcher_test.go
+++ b/grype/matcher/golang/matcher_test.go
@@ -15,7 +15,7 @@ import (
 	syftPkg "github.com/anchore/syft/syft/pkg"
 )
 
-func TestMatcher_DropMainPackage(t *testing.T) {
+func TestMatcher_DropMainPackageIfNoVersion(t *testing.T) {
 
 	mainModuleMetadata := pkg.GolangBinMetadata{
 		MainModule: "istio.io/istio",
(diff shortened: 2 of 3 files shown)


spiffcs (approved):

spiffcs:
Good news! I cannot find an increase in FP that this PR introduces from a first glance. @luhring Is there a case you saw where a new FP was introduced by this change? I can confirm that the FN example you provided does work and is now showing up in results.

luhring:
> @luhring Is there a case you saw where a new FP was introduced by this change?

No actually, good question. My reason for expecting them is that I can dream up edge cases, like when the GHSA fixed version is `v3.0.1`, but the computed pseudoversion would be `v0.0.0-<some commit that's later than v3.0.1>`, and so Grype would show the main module as vulnerable when it's not. But I think this is mitigated by multiple factors, like I mentioned above. So this seems like a well worthwhile tradeoff in the meantime.

spiffcs:
I did a bit more digging here to find an FP that would get surfaced by this change. Something like this:
```
syft -o json ollama/ollama:0.1.32 | ~/go/bin/grype



github.com/ollama/ollama    v0.0.0-20240414223325-7027f264fbb3  0.1.29             go-module  GHSA-5jx5-hqx5-2vrj  High
```

Here grype now surfaces `github.com/ollama/ollama` even though `7027f264fbb3` is a fixed version.

I still agree with this change in so much that grype should not be assuming SBOM are incorrect and discounting pseudo versions from matching.

The cataloging of main module versions has to get better on the SBOM side. Current challenges exist for the Golang ecosystem summarized here:

spiffcs:
I think @willmurphyscode had a comment about understanding the new FP introduced here more concretely so I'll let him chime in before merging:

Another query I put together:
```
sqlite> select count(*) from vulnerability WHERE namespace = "github:language:go" AND fixed_in_versions LIKE '%0.0.0%';
57
``` 

So this has the potential to FIX 57 false negatives. I don't have a good impression of the false positive impact beyond my experimentation with the ollama image that showed how a potential new FP could be introduced.

Final Thoughts:

I’m in favor of not tying Syft improvements to Golang main module version discovery to this change.

I think if a user brings an SBOM that was not generated with syft, then grype should be considering all valid cases.

Versions prefixed with v0.0.0 are valid cases:
https://github.com/advisories/GHSA-xx8w-mq23-29g4

Any introduction of FP is regrettable here, but if there are two outcomes for grype, then I’d rather go down the path where Grype is correct for ANY given SBOM input rather than the narrow case of trying to shore up edges of main module version discovery in Syft’s Golang cataloger.

willmurphyscode:
Thanks for the investigation @spiffcs! I think @luhring is right that the `v0.0.0` prefix defensive check pre-dates some of the attempts in syft to get a main module version. Syft's main module detection was improved as recently as https://github.com/anchore/syft/pull/2608 (merged Feb 9 2024).

I think I'm comfortable approving this because, without this change, Grype is effectively saying, "the sbom must be wrong; no one ships v0.0.0- main module versions," but sometimes there are `v0.0.0-` versions that are shipped, and have vulnerabilities.

PR #1809, opened 2024-04-17
fix: add linux and libc-dev headers ignore rules for debian packages

zhill:
Follow up on #1787 for debian packages, not just rpms.

Adds ignore rules for indirect matches on "linux" upstream package if the binary package is the linux-headers-* or linux-libc-dev packages since those are just headers files not the code that has the linux source that is vulnerable.


kzantow (approved):
LGTM 👍

PR #1810, opened 2024-04-17
WIP: narrow golang match comparison for pseudo versions

spiffcs:
## Summary

This PR attempts to narrow the golang_constraint `Satisfied` logic as a follow up to #1797

#1797 allows grype to proceed with matches when it encounters a package with a `pseudo version`. This PR limits those pseudo versions to only be compared against constraints that also contain pseudo versions.

This eliminates a case of FP where an incomplete `pseudo version` (which doesn't have the correct main module information) is compared against a valid semver constraint.

Example of this FP:
```
syft -o json ollama/ollama:0.1.32 | go run cmd/grype/main.go

...

github.com/ollama/ollama    v0.0.0-20240414223325-7027f264fbb3  0.1.29             go-module  GHSA-5jx5-hqx5-2vrj  High
```

In the above case `v0.0.0-20240414223325-7027f264fbb3` is not < `0.1.29`. Syft is unable to determine the main module version for ollama. By comparing the incomplete pseudo version to the semver constraint grype produces a FP.

This PR makes it so that packages with versions like `v0.0.0-20240414223325-7027f264fbb3` should only be compared to constraints that also have a similar format.

### DIFF
--- a/grype/version/golang_constraint.go
+++ b/grype/version/golang_constraint.go
@@ -36,16 +36,26 @@ func (g golangConstraint) Satisfied(version *Version) (bool, error) {
 		return true, nil // the empty constraint is always satisfied
 	}
 
-	// when we get a pseudo version from the package and the constraint is not a pseudo version, we should not consider it as satisfied
-	// ex: constraint = ">=v1.0.0", version = "v0.0.0-0.20210101000000-abcdef123456"
-	if isPseudoVersion(version.String()) && !isPseudoVersion(g.raw) {
+	var constraintContainsPseudoVersion bool
+	for _, units := range g.expression.units {
+		for _, unit := range units {
+			if isPseudoVersion(unit.version) {
+				constraintContainsPseudoVersion = true
+				break
+			}
+		}
+	}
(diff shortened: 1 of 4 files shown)


grype/version/golang_constraint.go
```diff
@@ -31,9 +35,37 @@ func (g golangConstraint) Satisfied(version *Version) (bool, error) {
...
 	return g.expression.satisfied(version)
 }
 
+// PseudoVersionPattern is a regular expression pattern to match pseudo versions
+const PseudoVersionPattern = `^v0\.0\.0[-+].*$`
+
+// Check if a version string is a pseudo version
+func isPseudoVersion(version string) bool {
+	// List of prefixes commonly used for pseudo versions
+	regex := regexp.MustCompile(PseudoVersionPattern)
```
kzantow, line 64: This could be moved to the `PseudoVersionPattern` definition so it only is parsed once (the pattern also does not need to be exported, I think)

spiffcs:
Quality tests are now failing - let me go check the labels

Issue #1624, opened 2023-11-30
Allow configurting timeout for external-sources


abelsromero:
**What would you like to be added**:
A new option to limit the time spent checking external sources. It could be a total time or a number of retries. I assume the latter is simpler but exposes more of the implementation, I leave that to the project leads 🙇 

```
external-sources:
  enable: true
  maven:
    search-upstream-by-sha1: true
    base-url: https://search.maven.org/solrsearch/select
    abort-after: 10m
```

**Why is this needed**:

Scan just stays for an indefinite time when there's an issue in the search services and nothing is shown in the console, only in debug.
This is happening right now due to issues with `search.maven.org/`, and we found what was causing issues when enabled debug and found the logs. However, it takes a long time to even show that.

```
failed to match against upstream data for HdrHistogram: status 504 Gateway Timeout from [https://search](https://search.maven.org/solrsearch/select?q=1%3A%226eb7552156e0d517ae80cc2247be1427c8d90452%22&rows=1&wt=json)
```

**Additional context**:

tgerla:
Hi @abelsromero, thank you for the suggestion. We'll put this in the backlog for consideration. Are you interested in working on this? We'd be happy to point you in the right direction. Thanks again!

abelsromero:
If you can provide specific of what configuration options you feel comfortable I can give it a try. Like what I mentioned, do you think it'd be better to have a number of retries or a total time?

tgerla:
That would be great! I think total time would be a good starting point.

One thing that would be nice is if the setting cascaded--in other words, you could set a global abort-after as well as abort-after values for individual sources:

``` yaml
external-sources:
  enable: true
  abort-after: 10m
  maven:
    search-upstream-by-sha1: true
    base-url: https://search.maven.org/solrsearch/select
    abort-after: 5m  # override the global config
  another-source:
    blah-blah: true
    # aborts after 10m by inheriting the global config
```

PR #1812, opened 2024-04-18
Feat/allow configuring timeout for external sources

pouyan021:
This pull request addresses the issue [1624](https://github.com/anchore/grype/issues/1624). It adds the ability to set a new property `abort-after` to `external` sources`. As discussed in the issue, it supports both a global prop and a `maven` property that overrides the global if it is set.


README.md
```diff
@@ -278,9 +278,11 @@ feature is currently disabled by default. To enable this feature add the followi
 ```yaml
 external-sources:
   enable: true
+  abort-after: 10m
   maven:
-    search-upstream-by-sha1: true
```
spiffcs, line 282: 👋 thanks for adding the new field! Would it be possible to keep the field name here so that we don't release a breaking change? I know we're pre v1.0, but the best way to handle this is to deprecate this field and add the new one (allowing people to migrate) rather than changing it on the fly =)

spiffcs: Oh wow - on second pass it looks like the json/yaml tags here were never correct here so thanks for the catch and docs update. I've made a note to change the json tag to be consistent going forward.

spiffcs (approved):

spiffcs:
@pouyan021 Approved and running checks now - Thank you so much for the contribution!

spiffcs:
@pouyan021 looks like there is a small change needed where `go mod tidy` is run

pouyan021:
@spiffcs Appreciate your support, thanks a lot! I'm sorry I forgot to sign-off my commit on the 2nd push! Could you kindly approve again?

README.md
```diff
@@ -278,9 +278,11 @@ feature is currently disabled by default. To enable this feature add the followi
...
+    search-upstream: true
     base-url: https://search.maven.org/solrsearch/select
+    abort-after: 5m #override the global config
```
spiffcs, line 285: cc @wagoodman - I know he's pretty sensitive to duplicate fields that override each other so I'd like him to chime in on where he sees this going or what his preference would be

spiffcs (commented):
Everything else LGTM - Just waiting for @wagoodman and his final say on the config direction he wants to go here

cmd/grype/cli/options/datasources.go
```diff
@@ -6,20 +6,24 @@ import (
...
+	BaseURL              string  `yaml:"base-url" json:"baseUrl" mapstructure:"base-url"`
+	AbortAfter           *string `yaml:"abort-after" json:"abortAfter" mapstructure:"abort-after"`
 }
 
 func defaultExternalSources() externalSources {
 	return externalSources{
+		AbortAfter: defaultAbortAfter,
 		Maven: maven{
 			SearchUpstreamBySha1: true,
 			BaseURL:              defaultMavenBaseURL,
```
wagoodman, line 29: We want the following quality to multi-teir config items like this:

-  a user specifying top-level fields overrides UNspecified nested fields
-  a user specifying top-level fields does NOT override specified nested fields
-  a user specifying nested fields does NOT top-level fields

Where:
- specified: a user explicitly provided a value
- unspecified: a user did NOT explicitly provide a value. This might mean no value should be set or a default value should be set instead.

The implementation does follow this, however, the rendered configuration unintentionally conveys that there is no maven abort-after value:
```yaml
# running: grype -vv
...
  external-sources:
      abort-after: 10m
      maven:
          abort-after: null
```

Here from a user perspective it appears as if an override has been set at the top level, but has not been inherited. This is because we're conflating inheriting default values vs overriding nested values.

What we really want is to convey that the user did not specify overriding values in top-level fields, and that the individual sources have sane defaults:

```yaml
# running: grype -vv
...
  external-sources:
      abort-after: null
      maven:
          abort-after: 10m0s
```

This has the added benefit that different sources can have different default values (since the default value for nested properties is not inheriting a top-level property).

We've dealt with multi-level properties in syft, see the `pretty` fields for an example: https://github.com/anchore/syft/blob/v1.4.1/cmd/syft/internal/options/format.go . We end up making both fields optional to show what is specified vs unspecified, then reason about the final values in a `PostLoad` function, which is invoked while rendering the configuration.

Specifying an override then appears in both places correctly:
```yaml
# run: GRYPE_EXTERNAL_SOURCES_ABORT_AFTER=40m grype -vv 
  external-sources:
      abort-after: 40m0s
      maven:
          abort-after: 40m0s
```

Instead of no value being set:
```yaml
# run: GRYPE_EXTERNAL_SOURCES_ABORT_AFTER=40m grype -vv 
  external-sources:
      abort-after: 40m
      maven:
          abort-after: null
```

wagoodman (commented):
Overall the functionality looks good, but have some comments on testing and configuration. I'll push up some changes shortly to help out.

I do think that `RequestTimeout` is a better name for this config item -- what do you think @pouyan021 ?

wagoodman:
note: I force pushed to get this branch rebased onto the latest commit on main

cmd/grype/cli/options/datasources.go
wagoodman: FYI, no action needed, I made the adjustment to use the `PostLoad` in a follow up commit.

grype/matcher/java/matcher_mocks_test.go
```diff
@@ -52,13 +56,22 @@ type mockMavenSearcher struct {
...
 }
 
-func (m mockMavenSearcher) GetMavenPackageBySha(string) (*pkg.Package, error) {
-	return &m.pkg, nil
+func (m mockMavenSearcher) GetMavenPackageBySha(ctx context.Context, sha1 string) (*pkg.Package, error) {
+	deadline, ok := ctx.Deadline()
+	fmt.Println("GetMavenPackageBySha called with deadline:", deadline, "deadline set:", ok)
+	// Sleep for a duration longer than the context's deadline
+	select {
+	case <-time.After(10 * time.Second):
```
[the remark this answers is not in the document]
wagoodman: FYI, no action needed, I made the adjustment to change the mock to not require sleeping in problematic cases.

grype/matcher/java/matcher_mocks_test.go
```diff
@@ -52,13 +56,22 @@ type mockMavenSearcher struct {
 	pkg pkg.Package
 }
 
-func (m mockMavenSearcher) GetMavenPackageBySha(string) (*pkg.Package, error) {
-	return &m.pkg, nil
+func (m mockMavenSearcher) GetMavenPackageBySha(ctx context.Context, sha1 string) (*pkg.Package, error) {
+	deadline, ok := ctx.Deadline()
+	fmt.Println("GetMavenPackageBySha called with deadline:", deadline, "deadline set:", ok)
```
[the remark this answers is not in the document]
wagoodman: FYI, no action needed, I made the mock to take `testing.TB` for logging.

cmd/grype/cli/options/datasources.go
```diff
@@ -6,20 +6,24 @@ import (
...
+	AbortAfter string `yaml:"abort-after" json:"abortAfter" mapstructure:"abort-after"`
+	Maven      maven  `yaml:"maven" json:"maven" mapstructure:"maven"`
 }
 
 type maven struct {
-	SearchUpstreamBySha1 bool   `yaml:"search-upstream" json:"searchUpstreamBySha1" mapstructure:"search-maven-upstream"`
-	BaseURL              string `yaml:"base-url" json:"baseUrl" mapstructure:"base-url"`
+	SearchUpstreamBySha1 bool    `yaml:"search-upstream" json:"searchUpstreamBySha1" mapstructure:"search-maven-upstream"`
+	BaseURL              string  `yaml:"base-url" json:"baseUrl" mapstructure:"base-url"`
+	AbortAfter           *string `yaml:"abort-after" json:"abortAfter" mapstructure:"abort-after"`
```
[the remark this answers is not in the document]
wagoodman: FYI, no action needed, I made the adjustment to use the `time.Duration` in the application configuration.

pouyan021:
> Overall the functionality looks good, but have some comments on testing and configuration. I'll push up some changes shortly to help out.
> 
> 
> 
> I do think that `RequestTimeout` is a better name for this config item -- what do you think @pouyan021 ?

Hey @wagoodman thanks a lot for the thorough review and your additional changes. The name was chosen as `abort-after` based on the suggestion of one of the contributors in the issue. I agree with you that `RequestTimeout` is more self-explanatory.

pouyan021:
Hey @wagoodman should I go for the  `requestTimeout` as per your suggestion?

pouyan021:
Hey @wagoodman @spiffcs I resolved the conflict on this branch recently, any plans for this to move forward?

spiffcs:
@pouyan021 I'm so sorry here - I've approved and run the final checks and will get this merged - I did not see the notification for this and apologize for letting it sit

pouyan021:
Hey @spiffcs, I rebased this branch again, could you please trigger the checks so we can hopefully merge this?

pouyan021:
cc @wagoodman

kzantow:
Hi @pouyan021, it looks like this PR has been through a lot! There have been some other enhancements to the Maven search function, including rate limiting, but I think that's not exactly the same thing as you are hoping to get merged here. Another change will currently result in a scan failure if a user does end up getting rate limited. I'd like to make sure this doesn't get lost in the noise, and get it moved forward if we can.

I hate to ask you to rebase this again; let me know if you'd like help getting this to the finish line. It seems like we might need to think about how all these options/behaviors should work together.

PR #1819, opened 2024-04-22
docs: update README with newer data sources

luhring:
- Adds data sources added since this list was last updated: Chainguard and Wolfi SecDBs
- Minor alphabetizing and case corrections


luhring:
I can update the distro list in this section too if it makes sense! https://github.com/anchore/grype?tab=readme-ov-file#features

spiffcs (approved):

spiffcs:
Thanks @luhring! - I think just the read me updates are good here.

PR #1857, opened 2024-05-10
fix: ask catalog for package, rather than type asserting

willmurphyscode:
## TODO

- [x] tests
- [x] research fixing in Syft instead - nothing to be done; grype wasn't using Syft's API very well.

## Explanation

Basically, when we're cataloging we remove some packages by ownership overlap, (e.g. if you're on redhat and do `yum install python3-urllib`, grype will drop the PyPI package `urllib` in favor of the RPM `python3-urllib`, because we have better match information for the distro package than for PyPI, e.g. because RedHad backported a fix to an old version of the python code or something).

We do this:
``` go
   		from, ok := r.From.(pkg.Package)
		if ok && excludePackage(comprehensiveDistroFeed, p, from) {
			continue
		}
```

Which seems fine, _however_, some relationships have a From of type `*pkg.Package`, not type `pkg.Package` (pointer vs struct). This causes the type assertion in grype to _always_ fail, resulting FPs.

The underlying reason is that the syft scanner does this:

``` go
		parent := catalog.Package(parentID)
		child := catalog.Package(childID)
```

when building the relationship. And catalog.Package(id) returns a `*Package`.

However, when we're decoding an SBOM, the relationship builder looks over `catalog.Sorted()`, which is a `[]Package`.

We can compensate for this in grype by trying to type assert to both `pkg.Package` and `*pkg.Package`. I think that's the easier way, and I feel like the "right" way (making Syft either always put a pointer in the To/From field of a relationship or never do so) might end up breaking the type signature of something exported, which we wanted to avoid in 1.0. Grype PR soon, but definitely open to feedback on the approach.

### DIFF
--- a/grype/pkg/package_test.go
+++ b/grype/pkg/package_test.go
@@ -874,7 +874,7 @@ func catalogWithOverlaps(packages []string, overlaps []string) *sbom.SBOM {
 		pkgs = append(pkgs, p)
 	}
 
-	for _, overlap := range overlaps {
+	for i, overlap := range overlaps {
 		parts := strings.Split(overlap, "->")
 		if len(parts) < 2 {
 			panic("invalid overlap, use -> to specify, e.g.: pkg1->pkg2")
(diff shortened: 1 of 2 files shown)


grype/pkg/package_test.go
```diff
@@ -883,8 +883,8 @@ func catalogWithOverlaps(packages []string, overlaps []string) *sbom.SBOM {
 		to := toPkg(parts[1])
 
 		relationships = append(relationships, artifact.Relationship{
-			From: from,
-			To:   to,
+			From: &from,
+			To:   &to,
```
kzantow, line 887: Is there a way we could have _both_ pointers and non-pointers? maybe the first iteration add a pointer and subsequent ones add non-pointers?

kzantow (approved):
LGTM as is, but possible test improvement suggestion

grype/pkg/package_test.go
willmurphyscode: @kzantow I've added your suggestion - how does that look?

kzantow (approved):
👍

Issue #572, opened 2021-12-30
Can't download db because not available space on /tmp/


PawelAntosik-TomTom:
**What happened**:
/usr/local/bin/grype db update
Unable to update vulnerability database
unable to update vulnerability database: unable to update vulnerability database: unable to download db: write /tmp/getter160826247/archive: no space left on device
**What you expected to happen**:
To install a new database
**How to reproduce it (as minimally and precisely as possible)**:
You need to have a system with filesystem /tmp/ of size 504M or less
**Anything else we need to know?**:
It would be good if instead of using /tmp/ be default we could specify our own temp directory.
**Environment**:
- Output of `grype version`: 0.28.0
- OS (e.g: `cat /etc/os-release` or similar): CentOS 6.5

kzantow:
@PawelAntosik-TomTom you should be able to do this already (see: https://github.com/anchore/grype#local-database-cache-directory) does this not work for you?

PawelAntosik-TomTom:
Nope it doesn't work for me

[antosik@ ~]$ export GRYPE_DB_CACHE_DIR=/opt/grype/db
[antosik@ ~]$ grype db update
Unable to update vulnerability database
unable to update vulnerability database: unable to update vulnerability database: unable to download db: write /tmp/getter140947589/archive: no space left on device

PawelAntosik-TomTom:
it doesn't work even if i try to import db

[antosik@ grype]$ pwd
/opt/grype
[antosik@ grype]$ grype db status
Location:
Built:     0001-01-01 00:00:00 +0000 UTC
Schema:    0
Checksum:
Status:    invalid
database metadata not found at "/opt/grype/db/3"
[antosik@ grype]$ grype db import vulnerability-db_v1_2022-01-12T08\:16\:10Z.tar.gz
unable to import vulnerability database: unable to create db temp dir: mkdir /tmp/grype-import3712712505: no space left on device

PawelAntosik-TomTom:
setting TMPDIR works!

luhring:
>setting TMPDIR works!

Awesome! Let us know if you need anything else.

PR #1879, opened 2024-05-23
fix: add note about TMPDIR env var

avtar:
As described in https://github.com/anchore/grype/issues/572 setting just the GRYPE_DB_CACHE_DIR environment variable doesn't successfully make grype use the custom cache location. TMPDIR needs to be set as well.


avtar:
Closed in favour of #1880

PR #1894, opened 2024-05-28
fix: Go main module pseudo version matching: turn off by default

luhring:
About 6 weeks ago, I opened https://github.com/anchore/grype/pull/1797 in an effort to **reduce false negatives** for the main modules of Go binaries. Before that PR, Grype's previous behavior was to avoid vulnerability matching for the main module altogether if the detected module version was prefixed with `v0.0.0-`. Grype's reasoning for this had been that Syft will manufacture pseudo versions that start with `v0.0.0-` in the case where it's not able to find a real version in the Go binary itself.

By making the change in the PR linked above, we indeed saw some reduction in false negatives 🎉 , which was great. We weren't sure about the larger impact of false positives. I only [speculated](https://github.com/anchore/grype/pull/1797#issuecomment-2052672419) about the potential effects:

>My reason for expecting them is that I can dream up edge cases, like when the GHSA fixed version is `v3.0.1`, but the computed pseudoversion would be `v0.0.0-<some commit that's later than v3.0.1>`, and so Grype would show the main module as vulnerable when it's not.

**The bad news:** After seeing this change released and used on a broader set of images, it looks like this net effect is that Grype's F1 score is **worsened** considerably more than we anticipated. We've been able to recover some recall, but at a painful hit to precision.

**The good news:** I saw that after my original PR was merged, @spiffcs opened https://github.com/anchore/grype/pull/1816, which made this new behavior **configurable** and defaulted to on. IMHO, that was brilliant, and in hindsight I probably should've added that in my original PR, too. 😞 

At scale, this now-configurable behavior is working very much like CPE matching: it's great for scenarios where the user is willing to pay a high FP cost for the chance at not missing any matches, but on average it adds more noise than signal.

So with this PR, I suggest we adjust the noisy config option to work just like language package CPE matching: off by default but configurable to on, such that consumers can get more sane behavior by default, but they can crank up the sensitivity knob if they know what they're doing.

Curious for your thoughts!

### DIFF
--- a/README.md
+++ b/README.md
@@ -855,7 +855,8 @@ match:
     using-cpes: false
     # even if CPE matching is disabled, make an exception when scanning for "stdlib".
     always-use-cpe-for-stdlib: true
-    allow-main-module-pseudo-version-comparison: true
+    # allow main module pseudo versions, which may have only been "guessed at" by Syft, to be used in vulnerability matching
+    allow-main-module-pseudo-version-comparison: false
   stock:
     using-cpes: true
 ```
--- a/cmd/grype/cli/options/match.go
+++ b/cmd/grype/cli/options/match.go
@@ -34,7 +34,7 @@ func defaultGolangConfig() golangConfig {
 			UseCPEs: false,
 		},
 		AlwaysUseCPEForStdlib:                  true,
-		AllowMainModulePseudoVersionComparison: true,
+		AllowMainModulePseudoVersionComparison: false,
 	}
 }
 


kzantow:
I could certainly be wrong here, but for what it's worth, I _think_ the right behavior is to consider only comparing pseudo-versions with pseudo-versions, but maybe consider using the timestamp portion. What I mean by this is as follows: I've seen go module versions be one of 3 distinct things (although the [require syntax docs only mention 2](https://go.dev/doc/modules/gomod-ref#require-syntax)):
* a complete pseudo-version: `v0.0.0-<date>-<hash>`
* a speculated pseudo-version: `v3.2.4-<date>-<hash>`
* a semver-like version: `v2.5.6`

Vulnerabilities are reported against the best version someone can find, and people use modules from different sources including, say a `main` branch, which means we will potentially see any combination of these versions on both sides of the matching equation.
If we have a pseudo-version with all zeroes, there's no direct relation to a real semver-like version and we really just can't compare them -- it's the proverbial apples and oranges. If we have a speculated pseudo-version, there's a relation to both and we could potentially compare to either but comparing to a full-zero pseudo-version again isn't going to be especially fruitful. So my proposal would be as follows:
```
if <both versions have semver-like numbering>
  compare semver section
else if <both versions have date-like numbering>
  compare date section
else
  don't compare
```
The `<date>` portion is supposed to be repository revision date, which admittedly is not a substitute for proper version lineage. But I believe it's the best we can do without having more context about the hash info. Or I could be all wrong!

spiffcs:
Throwing in a 👍 to the @luhring comment of:
> I think the change in this PR will help recover Grype's accuracy the fastest. 

I'm pro changes that make our tool by default better for people to use. If we merge this PR I can file an issue summarizing the other problems we want to fix going forward surrounding syft's hallucinated Golang versions and it's ability to communicate to scanners whether the version in the package is a best guess or actually read from the metadata

spiffcs (approved):

luhring:
Thanks @spiffcs!

PR #1944, opened 2024-06-14
fix match sort ordering for different locations

luhring:
A common case in vulnerability matching is when two vulnerability matches are the same except for their location. I noticed this case was causing Grype to produce inconsistent output when using its match sorting, e.g. `sort.Sort(match.ByElements(matches))`.

This PR adds consideration for locations into the match sort ordering logic.


wagoodman (approved):
🙌

PR #1945, opened 2024-06-14
chore: Update syft v1.7.0

spiffcs:
## Summary

This PR bumps grype to use the latest syft version v 1.7.0

The bulk of changes come from updating `schema/cyclonedx/cyclonedx.xsd` so that our cyclonedx documents can be measured and validated against the latest released version of cdx (v1.6)

These document updates were pulled from:
https://github.com/CycloneDX/specification/tree/1.6/schema


go.mod
```diff
@@ -1,9 +1,11 @@
 module github.com/anchore/grype
 
-go 1.21.1
+go 1.22.0
+
+toolchain go1.22.4
```
wagoodman, line 5: do we need the toolchain declaration?

kzantow, line 5: It would be great if we could omit the toolchain directive. The syft go mod only has `go 1.22.0`

wagoodman (approved):

go.mod
spiffcs: Ahh yea - new laptop with up to date toolchain installed so it does this automagically on go mod tidy =(

PR #1946, opened 2024-06-14
feat: pass thru the cpe source if available

zhill:
For CPE-based matches, display the detail about the CPE used to match.

This includes the "source" indicating if the cpe was syft-generated, or a lookup from the cpe-dictionary.

### DIFF
--- a/grype/search/cpe.go
+++ b/grype/search/cpe.go
@@ -24,9 +24,10 @@ type CPEPackageParameter struct {
 }
 
 type CPEParameters struct {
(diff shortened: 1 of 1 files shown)


zhill:
Opened this for comments, will fix tests etc once there is a 👍 on this as something to do. Understood that Grype is reducing reliance on CPEs but some folks still use/need it and want to understand the matches better when they do appear.

zhill:
This branch is quite stale and the merge conflicts make it not worth fixing, so I'll redo it with the same objective on the latest code

PR #1953, opened 2024-06-19
chore: enable dependabot to keep boostrap action updated

westonsteimel:
Allows dependabot to also keep the boostrap action updated


kzantow (approved):
LGTM 👍

PR #1961, opened 2024-06-24
chore: add workflow to update quality test db

spiffcs:
## Summary
Add monthly job to update quality gate db automatically via PR

I can update it to pull the latest version of grype rather than compile it in CI. Since it's a monthly job I don't have a strong opinion either way =)

### Green Auto Run
https://github.com/anchore/grype/actions/runs/9650508688/job/26616274691


.github/scripts/update-quality-gate-db.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+# Run your command and capture the output
+output=$(go run ./cmd/grype/main.go db list)
+
+# Extract the first instance of URL using grep, cut, and sed to trim leading whitespace
+url=$(echo "$output" | grep -m 1 -o 'URL: .*' | cut -d' ' -f2- | sed 's/^[[:space:]]*//')
```
wagoodman, line 7: if we are ok with `jq` then we could use `grype db list -o json` or `-o raw` and make this a little simpler

.github/workflows/update-quality-gate-db.yml
```diff
@@ -0,0 +1,38 @@
...
+        with:
+          app_id: ${{ secrets.TOKEN_APP_ID }}
+          private_key: ${{ secrets.TOKEN_APP_PRIVATE_KEY }}
+
+      - uses: peter-evans/create-pull-request@c5a7806660adbe173f04e3e038b0ccdcd758773c #v6.1.0
+        with:
+          signoff: true
+          delete-branch: true
+          branch: auto/update-quality-test-db
+          labels: dependencies
```
wagoodman, line 33: is `dependencies` the right label?

.github/scripts/update-quality-gate-db.sh
```diff
@@ -0,0 +1,13 @@
...
+
+# Escape special characters in the URL for sed substitution
+escaped_url=$(printf '%s\n' "$url" | sed -e 's/[\/&]/\\&/g')
+
+# Replace TEST_DB_URL in specific Makefile using sed
+sed --e "s|^TEST_DB_URL = .*|TEST_DB_URL = $escaped_url|" test/quality/Makefile
```
wagoodman, line 13: we could probably update the Makefile to read this from a file by default then this wouldn't need to be a sed command on the makefile... just overwrite the contents of a single file that only has the URL in question

.github/workflows/update-quality-gate-db.yml
spiffcs: Yea I think so - it's ignored by the changelog generator by default and is updating a test dependency from out of date to valid. I can add a different label if you want a subsection for these kinds of updates

.github/scripts/update-quality-gate-db.sh
spiffcs: 👍 Sure! I'll make it so we read from a file and update to use jq to reduce the complexity here

.github/scripts/update-quality-gate-db.sh
kzantow: FWIW I thought I posted this but it was just pending 🤦 I think this could work:
```
url=$(go run ./cmd/grype db list -o json | jq '.[0].url')
```
... and `jq` is available on the runners by default

.github/workflows/update-quality-gate-db.yml
```diff
@@ -0,0 +1,38 @@
...
+          private_key: ${{ secrets.TOKEN_APP_PRIVATE_KEY }}
+
+      - uses: peter-evans/create-pull-request@c5a7806660adbe173f04e3e038b0ccdcd758773c #v6.1.0
+        with:
+          signoff: true
+          delete-branch: true
+          branch: auto/update-quality-test-db
+          labels: test, changelog-ignore
+          commit-message: 'chore(deps): update quality db to latest version'
+          title: 'chore(deps): update quality db to latest version'
```
wagoodman, line 35: nit: the conventional commit type isn't right here (should be `test` I think)

kzantow (approved):
🎉

Issue #1975, opened 2024-07-03
`db status` does not validate `vulnerability.db`


joshuai96:
**What happened**:

`grype db status` does not validate `vulnerability.db` hash.

**What you expected to happen**:

`grype db status` fully validates the `vulnerability.db` and fails on an invalid db.

**How to reproduce it (as minimally and precisely as possible)**:

```bash
$ cd $HOME/.cache/grype/db/5/
$ mv vulnerability.db vulnerability.db.bak

# with an invalid db
$ tail -n 50 vulnerability.db.bak > vulnerability.db
$ grype db status
Location:  /home/joshua/.cache/grype/db/5
Built:     2024-07-03 01:30:39 +0000 UTC
Schema:    5
Checksum:  sha256:a7920011e5de7de5e0acafb614f94c38dc4353d34784de07cf16fa765c07904f
Status:    valid

# with missing db
$ rm vulnerability.db
$ grype db status
Location:  /home/joshua/.cache/grype/db/5
Built:     2024-07-03 01:30:39 +0000 UTC
Schema:    5
Checksum:  sha256:a7920011e5de7de5e0acafb614f94c38dc4353d34784de07cf16fa765c07904f
Status:    valid
```

**Anything else we need to know?**:

Scans with `grype`, do not report a invalid DB either:

```bash
# with invalid db
$ grype alpine:latest
 ✔ Vulnerability DB                [no update available]  
 ✔ Parsed image                              sha256:a606584aa9aa875552092ec9e1d62cb98d486f51f389609914039aabd9414687
 ✔ Cataloged contents                               dabf91b69c191a1a0a1628fd6bdd029c0c4018041c7f052870bb13c5a222ae76
   ├── ✔ Packages                        [14 packages]  
   ├── ✔ File digests                    [77 files]  
   ├── ✔ File metadata                   [77 locations]  
   └── ✔ Executables                     [17 executables]  
failed to load vulnerability db: unable to get namespaces from store: file is not a database (26)

# with missing db
$ grype alpine:latest
 ✔ Vulnerability DB                [no update available]  
 ✔ Parsed image                              sha256:a606584aa9aa875552092ec9e1d62cb98d486f51f389609914039aabd9414687
 ✔ Cataloged contents                               dabf91b69c191a1a0a1628fd6bdd029c0c4018041c7f052870bb13c5a222ae76
   ├── ✔ Packages                        [14 packages]  
   ├── ✔ File digests                    [77 files]  
[...]

Issue #1648, opened 2024-01-04
`grype db status` doesn't always check the db's checksum and validity


luhring:
**What happened**:

A coworker was getting this output from Grype:

```console
1 error occurred:
	* failed to load vulnerability db: unable to get namespaces from store: database disk image is malformed (11)
```

This error is originating from the sqlite driver, and it looks like other Grype users encounter it on occasion, e.g. https://github.com/anchore/grype/issues/1150. The underlying cause is that the database has been corrupted (somehow).

But... I asked them to run `grype db status`, and here's what they got:

```console
grype db status
Location:  /Users/<snip>/Library/Caches/grype/db/5
Built:     2024-01-04 01:26:41 +0000 UTC
Schema:    5
Checksum:  sha256:d853702af304b86fb2180b83d8fcef1ddf7f6578df8216cfe427d42abb2b5d26
Status:    valid
```

This could be way more helpful than it currently is, especially when debugging database issues like this one.

It's not actually checking the real database's checksum, or the real database's validity. It's just reporting what's in the metadata JSON file. This makes it harder to find out why Grype isn't working correctly, when it's a database issue. This is frustrating for users, too, because such database issues **usually render Grype inoperable** until you run `grype db delete && grype db update` or something (which the average user doesn't know about).

**What you expected to happen**:

From [talking on Anchore Slack](https://anchorecommunity.slack.com/archives/C027JE5M345/p1704403793917949), it sounds like the config check used during the execution of `grype db status` should become a non-conditional, "always do the check" execution path.

https://github.com/anchore/grype/blob/d5ced7fb81c8155faa010bc38cec0c3a0672513a/grype/db/curator.go#L323

**How to reproduce it (as minimally and precisely as possible)**:

```console
$ echo "booboo" > ~/Library/Caches/grype/db/5/vulnerability.db
$ grype db status
Location:  /Users/dan/Library/Caches/grype/db/5
Built:     2024-01-04 01:26:41 +0000 UTC
Schema:    5
Checksum:  sha256:d853702af304b86fb2180b83d8fcef1ddf7f6578df8216cfe427d42abb2b5d26
Status:    valid
```

**Anything else we need to know?**:

**Environment**:
- Output of `grype version`:
- OS (e.g: `cat /etc/os-release` or similar):

tgerla:
Thanks @luhring, we'll add this to the backlog.

wagoodman:
[...]

PR #1976, opened 2024-07-03
Fix: Fully validate `vulnerability.db` by hash

joshuai96:
Fixes: #1975
Fixes: #1648

Setting `ValidateByHashOnStart` to `true` by default, enables the database curator to execute `ValidateByHash` in `validateIntegrity` and report an invalid database.

```bash
$ cd $HOME/.cache/grype/db/5/
$ mv vulnerability.db vulnerability.db.bak

# with an invalid db
$ tail -n 50 vulnerability.db.bak > vulnerability.db
$ grype db status
Location:  /home/joshua/.cache/grype/db/5
Built:     2024-07-03 01:30:39 +0000 UTC
Schema:    5
Checksum:  sha256:a7920011e5de7de5e0acafb614f94c38dc4353d34784de07cf16fa765c07904f
Status:    invalid
bad db checksum (/home/joshua/.cache/grype/db/5/vulnerability.db): "sha256:a7920011e5de7de5e0acafb614f94c38dc4353d34784de07cf16fa765c07904f" vs "sha256:d3b1df971b1c7753cb658a2c0cb1046b781bc97f0727682bf6eacada5e92e5c8"

# with missing db
$ rm vulnerability.db
$ grype db status
Location:  /home/joshua/.cache/grype/db/5
Built:     2024-07-03 01:30:39 +0000 UTC
Schema:    5
Checksum:  sha256:a7920011e5de7de5e0acafb614f94c38dc4353d34784de07cf16fa765c07904f
Status:    invalid
failed to open file '/home/joshua/.cache/grype/db/5/vulnerability.db': open /home/joshua/.cache/grype/db/5/vulnerability.db: no such file or directory
```

Scans with `grype` now give a better error too:

```bash
# with invalid db
$ grype alpine:latest
 ✔ Parsed image                              sha256:a606584aa9aa875552092ec9e1d62cb98d486f51f389609914039aabd9414687
 ✔ Cataloged contents                               dabf91b69c191a1a0a1628fd6bdd029c0c4018041c7f052870bb13c5a222ae76
   ├── ✔ Packages                        [14 packages]  
   ├── ✔ File digests                    [77 files]  
   ├── ✔ File metadata                   [77 locations]  
   └── ✔ Executables                     [17 executables]  
failed to load vulnerability db: vulnerability database is invalid (run db update to correct): bad db checksum (/home/joshua/.cache/grype/db/5/vulnerability.db): "sha256:a7920011e5de7de5e0acafb614f94c38dc4353d34784de07cf16fa765c07904f" vs "sha256:d3b1df971b1c7753cb658a2c0cb1046b781bc97f0727682bf6eacada5e92e5c8"

# with missing db
$ grype alpine:latest
 ✔ Parsed image                              sha256:a606584aa9aa875552092ec9e1d62cb98d486f51f389609914039aabd9414687
 ✔ Cataloged contents                               dabf91b69c191a1a0a1628fd6bdd029c0c4018041c7f052870bb13c5a222ae76
   ├── ✔ Packages                        [14 packages]  
   ├── ✔ File digests                    [77 files]  
   ├── ✔ File metadata                   [77 locations]  
   └── ✔ Executables                     [17 executables]  
failed to load vulnerability db: vulnerability database is invalid (run db update to correct): failed to open file '/home/joshua/.cache/grype/db/5/vulnerability.db': open /home/joshua/.cache/grype/db/5/vulnerability.db: no such file or directory
```

### DIFF
--- a/cmd/grype/cli/options/database.go
+++ b/cmd/grype/cli/options/database.go
@@ -35,10 +35,11 @@ const (
 
 func DefaultDatabase(id clio.Identification) Database {
 	return Database{
-		Dir:         path.Join(xdg.CacheHome, id.Name, "db"),
-		UpdateURL:   internal.DBUpdateURL,
-		AutoUpdate:  true,
-		ValidateAge: true,
+		Dir:                   path.Join(xdg.CacheHome, id.Name, "db"),
+		UpdateURL:             internal.DBUpdateURL,
+		AutoUpdate:            true,
+		ValidateByHashOnStart: true,
+		ValidateAge:           true,
 		// After this period (5 days) the db data is considered stale
 		MaxAllowedBuiltAge:     defaultMaxDBAge,
 		UpdateAvailableTimeout: defaultUpdateAvailableTimeout,


wagoodman:
It was intentional to only validate the DB on download and not on start (for performance reasons) since validating every time could lead to [startup performance issues](https://github.com/anchore/grype/pull/224/commits/45d393804698782ab5d45e22af2782304f90e5b6). That being said, I can also see the point where getting grype results with a known bad database is also not ideal, since we can no longer trust those results. 

I think ultimately it's the right call to turn on validation by default, however, I feel that it should coincide with attempting to improve performance. Right now we're working on grype DB v6, which is very different than DB v1-5. Part of this work may also switch the distributing sha256 digests to xxh64, which is much faster.

One question I have: I have an anecdote from the past that startup was taking an extra second due to hashing the DB -- what extra time at startup are you seeing with this change @joshuai96 ?

joshuai96:
In the setup I described in https://github.com/anchore/grype/issues/1975#issuecomment-2208251742 I've seen `grype db update` stating `Vulnerability database updated to latest version!` and then having some files missing for example `metadata.json`, what leads to failing scan runs. So I think validation on download seems a little bit buggy, too. I am positive this is some kind of freak accident based on network hiccups, as our GitLab CI runners are known to have some networking issue.

You mean the startup before a scan runs, @wagoodman ? I didn't notice any inconvenient increase in time, but I can run some tests, to have some numbers backing that up. IMHO a second or two, has no impact on a scan that usually takes a few seconds anyway, but I'm not as sensitive to time, as this scans run over night in a CI/CD setup.

kzantow:
FWIW, enabling this option in the latest grype seems to increase the scan time of `alpine:latest` by ~7 seconds on my x86 mac.

joshuai96:
@wagoodman any idea, why I don't see a slowdown of scans?

kzantow:
FYI this has been updated for v6 in the default config, and as of the v6 enablement PR, these defaults are what Grype uses: https://github.com/anchore/grype/pull/2439/files#diff-deeafa2398d21fd34ccc42187013f75d7de4e363745ad76df57ab590e8573bb2R54

kzantow:
Hi @joshuai96 -- thanks for this PR. I'm sorry we weren't able to get it merged, but the good news is [this will now be the default behavior when the next Grype release is published](https://github.com/anchore/grype/blob/main/grype/db/v6/installation/curator.go#L53), with V6 database support.

PR #2005, opened 2024-07-24
chore: only check for db update if local db is old enough

willmurphyscode:
Previously, grype would check for a database update on every invocation. However, this was causing slow downs in the CDN. Since new databases are typically made available every 24 hours, change grype to, by default, only check if a new database is available if the local database is more than 12 hours old.

See #1731

### DIFF
--- a/cmd/grype/cli/commands/db_check.go
+++ b/cmd/grype/cli/commands/db_check.go
@@ -30,6 +30,8 @@ func DBCheck(app clio.Application) *cobra.Command {
 }
 
 func runDBCheck(opts options.Database) error {
+	// `grype db check` should _always_ check for updates, regardless of config
+	opts.MinAgeToCheckForUpdate = 0
 	dbCurator, err := db.NewCurator(opts.ToCuratorConfig())
 	if err != nil {
 		return err
--- a/cmd/grype/cli/commands/db_check.go
+++ b/cmd/grype/cli/commands/db_check.go
@@ -19,10 +19,15 @@ func DBCheck(app clio.Application) *cobra.Command {
 	opts := dbOptionsDefault(app.ID())
 
 	return app.SetupCommand(&cobra.Command{
-		Use:     "check",
-		Short:   "check to see if there is a database update available",
-		PreRunE: disableUI(app),
-		Args:    cobra.ExactArgs(0),
+		Use:   "check",
+		Short: "check to see if there is a database update available",
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			// `grype db check` should _always_ check for updates, regardless of config
+			opts.DB.MinAgeToCheckForUpdate = 0
+
+			return disableUI(app)(cmd, args)
+		},
+		Args: cobra.ExactArgs(0),
 		RunE: func(_ *cobra.Command, _ []string) error {
 			return runDBCheck(opts.DB)
 		},
@@ -30,8 +35,6 @@ func DBCheck(app clio.Application) *cobra.Command {
 }
 
 func runDBCheck(opts options.Database) error {
-	// `grype db check` should _always_ check for updates, regardless of config
-	opts.MinAgeToCheckForUpdate = 0
 	dbCurator, err := db.NewCurator(opts.ToCuratorConfig())
 	if err != nil {
 		return err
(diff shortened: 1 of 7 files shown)


cmd/grype/cli/commands/db_check.go
```diff
@@ -30,6 +30,8 @@ func DBCheck(app clio.Application) *cobra.Command {
 }
 
 func runDBCheck(opts options.Database) error {
+	// `grype db check` should _always_ check for updates, regardless of config
+	opts.MinAgeToCheckForUpdate = 0
```
willmurphyscode, line 34: I thought about adding a separate bool "force check" option, or a force check parameter to IsUpdateAvailable and Update. Would that have been preferable?

wagoodman: I feel that not changing the behavior of db check is the right call for now from an options-perspective. We can always add more config items later to opt into `force`.

But, this param should be changed within `PreRunE` such that the config options are finalized before logging them.

cmd/grype/cli/options/database.go
```diff
@@ -18,6 +18,7 @@ type Database struct {
 	AutoUpdate             bool          `yaml:"auto-update" json:"auto-update" mapstructure:"auto-update"`
 	ValidateByHashOnStart  bool          `yaml:"validate-by-hash-on-start" json:"validate-by-hash-on-start" mapstructure:"validate-by-hash-on-start"`
 	ValidateAge            bool          `yaml:"validate-age" json:"validate-age" mapstructure:"validate-age"`
+	MinAgeToCheckForUpdate time.Duration `yaml:"min-age-to-check-for-update" json:"min-age-to-check-for-update" mapstructure:"min-age-to-check-for-update"`
```
wagoodman, line 21: naming suggestion: what about `stale-age` with an attached description?

willmurphyscode:
``` yaml
db:
   min-age-to-check-for-update: 12h
```

vs

``` yaml
db:
   stale-age: 12h
```

The first one seems clearer to me in terms of what it controls. But I agree the name is pretty wordy. How do we communicate to a user what `stale-age` does?

wagoodman: I'm open to other naming suggestions too. Here's how we attach descriptions to fields https://github.com/anchore/syft/blob/9573f557d12756f8c873b98a6af476835f611055/cmd/syft/internal/options/update_check.go#L15-L17

willmurphyscode: I was trying to be parallel to `MaxAllowedBuiltAge`, itself perhaps not an ideal age, since "built" seems a bit of an implementation concern.

wagoodman: What about `min-age-to-update` ?

kzantow: What about something like `update-check-interval`?

grype/db/curator_test.go
```diff
@@ -447,3 +448,107 @@ func TestCuratorTimeoutBehavior(t *testing.T) {
...
+			files[listingURL] = string(b)
+			dirs := map[string]string{}
+			fs := afero.NewMemMapFs()
+			fs.MkdirAll(path.Join(testDir, "5"), 0755)
+			f, err := fs.Create(path.Join(testDir, "5", "metadata.json"))
+			require.NoError(t, err)
+			defer f.Close()
+			err = json.NewEncoder(f).Encode(tt.existingMetadata)
+			require.NoError(t, err)
+			getter := newTestGetter(fs, files, dirs)
```
kzantow, line 512: I had some issues getting this working in all the scenarios I wanted to test because it seemed as though I couldn't mock every call and the getter code was ultimately calling `os.Open`, instead of using the `afero.FS`. If you happen to run into similar issues, I [wrote a mocked but functional server to emulate the listing file / db download](https://github.com/anchore/grype/pull/1247/files#diff-bc68f6371e1fd9156df23d07f9939a6bf90495b4d41668489f65383700cf710cR383).

willmurphyscode: I agree testing with a local http server is a cleaner solution. I'll wait for that PR to get in and then rebase and change these tests to use that, assuming we still decide we want this behavior.

willmurphyscode:
We will probably close this in favor of #2020, which seems like a much better approach. Thanks @kzantow!

willmurphyscode:
This is probably the wrong approach, because once the local DB is old enough, grype will still check every time, so from the CDN's perspective, there will still be a window from when the N hours expires until when the new database is downloaded by most clients that will have the same requests per second as today, so it won't increase reliability during that time window.

Issue #1939, opened 2024-06-13
grype db is not being downloaded


tomersein:
**What happened**:
trying to scan images like alpine \ ubuntu.
it starts to download grype-db and then hangs.
this issue started today.
**What you expected to happen**:
download the db
**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:
attaching logs:
```
0000] DEBUG found database update candidate: Listing(url=https://toolbox-data.anchore.io/grype/databases/vulnerability-db_v5_2024-06-13T01:31:30Z_1718252496.tar.gz)
[0000] DEBUG existing database (2024-06-10 01:30:58 +0000 UTC) is older than candidate update (2024-06-13 01:31:30 +0000 UTC), using update...
[0000] DEBUG database update available: Listing(url=https://toolbox-data.anchore.io/grype/databases/vulnerability-db_v5_2024-06-13T01:31:30Z_1718252496.tar.gz)
[0000]  INFO downloading new vulnerability DB
[0000]  INFO new version of grype is available: 0.78.0 (currently running: 0.74.7)
```
**Environment**:
- Output of `grype version`: 0.78.0
- OS (e.g: `cat /etc/os-release` or similar): mac

tgerla:
Hi @tomersein, thanks for the report. I just tried to trivially reproduce this and I was able to successfully use "grype db update" to download the database. Can you try this manual download command using curl and let us know the output?

```
curl -vO https://toolbox-data.anchore.io/grype/databases/vulnerability-db_v5_2024-06-13T01:31:30Z_1718252496.tar.gz
```

tomerse-sg:
I guess it is something with my network..
40m and it is still downloading it

tgerla:
Let us know if you continue to see problems, we have had other issue with our CDN in the past.

tomersein:
example from today:
`unable to download listing: context deadline exceeded (Client.Timeout or context cancellation while reading body)`

last log i can see - 
`[0000]  INFO base_db_url and target_db_url not provided; fetching most recent`

willmurphyscode:
Thanks all! We are taking a look at some CDN troubleshooting now.

max-allan:
Hello, the same thing is happening today. It downloads 75M  very quickly and then goes down to a few kb/s or less (or zero).

I included the Cloudflare headers below, I don't know if you (or they) can do much with them to debug!

```
grype db update -vv
[0000]  INFO grype version: 0.79.6
[...]

PR #2020, opened 2024-07-31
feat: improve db update check

kzantow:
There have been [multiple](https://github.com/anchore/grype/issues/1939) [reports](https://github.com/anchore/grype/issues/1731) about issues while attempting to check for database updates. We think it will be good to address these on multiple fronts. One of the things we can do instead of downloading a listing file of ~150Kb every time Grype is run is to use the standard `If-Modified-Since` behavior and track a useable time we obtained the database to include in this header.

This PR adds an `Updated` field to the metadata, which is set at db download time and used when further requests are made for the database to prevent downloading the file when it won't be used anyway.


grype/db/curator.go
```diff
@@ -148,13 +152,14 @@ func (c *Curator) Update() (bool, error) {
 	defer downloadProgress.SetCompleted()
 	defer importProgress.SetCompleted()
 
-	updateAvailable, metadata, updateEntry, err := c.IsUpdateAvailable()
+	metadata := c.GetMetadata()
```
willmurphyscode, line 155: Should we always download the listing on `grype db check` and `grype db update`?

kzantow: I don't think so, if we reliably know they are already up-to-date, why waste CPU cycles/network?

willmurphyscode: I am pessimistically assuming that something will go wrong, and having a command that means, "no really, I mean it, download that database whether you need to or not" might be helpful.

The goal of this change is to alleviate traffic. Making rare commands less useful isn't going to alleviate traffic.

But I do see your point.

kzantow: I'm not sure how this is making commands less useful, they all continue to function exactly as they do: if we downloaded a listing file and got the latest database, we have the latest database, so downloading a listing again and downloading the database again is rather pointless. I think I understand your point, though. I'll see about making this particular function isolated as much as possible.

kzantow: I think as a user, I would prefer `grype db update` to be efficient and use this function to eliminate unnecessary downloads if it's already up-to-date. Maybe we add a `--force` argument or similar to make it download everything anew?

Issue #1885, opened 2024-05-24
Grype failed to load vulnerability database: database metadata not found


Atharex:
**What happened**:
Running Grype on an Azure DevOps Pipeline task occasionally fails with the following error: 

```
[info] checking github for the current release tag 
[info] fetching release script for tag='v0.77.4' 
[info] checking github for the current release tag 
[info] using release tag='v0.77.4' version='0.77.4' os='linux' arch='amd64' 
[info] installed /usr/local/bin/grype 
failed to load vulnerability db: vulnerability database is invalid (run db update to correct): database metadata not found: /home/vsts/.cache/grype/db/5

##[error]Bash exited with code '1'.

```

Rerunning the pipeline several times eventually brings it to completion, yet this transient error prevents us from relying on grype scanning automation.

**What you expected to happen**:
The pipeline to run consistently successfully 

**How to reproduce it (as minimally and precisely as possible)**:
Run an Azure DevOps pipeline task where Grype is used (my current version is v0.77.4)

**Environment**:
- Output of `grype version`: v.0.77.4
- OS (e.g: `cat /etc/os-release` or similar): Linux, amd64

tgerla:
Hi @Atharex, thanks for the report. I think we will need more information to help you troubleshoot this. This doesn't look like a network error reaching the Grype DB's servers. It looks like an error on the filesystem. Can you share some more details about the Azure setup? What Linux distribution, etc? It would probably be worthwhile to add "-vv" to your Grype command line for more verbose logging. You might also want to try adding a "grype db update" command to the top of your pipeline and see if the failures continue. Just a hunch, but do you by chance have multiple Grype processes running simultaneously on the same underlying volume?

westonsteimel:
The issue is because they are downloading a v1 schema db from the listing file and should instead be downloading a v5 schema db

djclarkson:
Same issue with grype installed via brew on mac sonoma. 

```
grype ./
 ✔ Vulnerability DB                [138 MB / 166 MB]  
 ✔ Indexed file system                                                                                                                                                                                         .
 ✔ Cataloged contents                                                                                                                           cdb4ee2aea69cc6a83331bbe96dc2caa9a299d21329efb0336fc02a82e1839a8
   ├── ✔ Packages                        [1,239 packages]  
   └── ✔ Executables                     [0 executables]  
[0000]  WARN no explicit name and version provided for directory source, deriving artifact ID from the given path (which is not ideal)
[...]

PR #2033, opened 2024-08-05
fix: higher default timeout for database download

willmurphyscode:
Depending on region and network conditions, 120s was not enough time for many clients, leading to some complaints. Raise the default timeout to five minutes.

There's an open question whether we should wait for more data from the [latest attempt to resolve the CDN issues](https://anchorecommunity.discourse.group/t/grype-db-network-and-cdn-issues/48/4?u=willmurphy)

Fixes https://github.com/anchore/grype/issues/1885

### DIFF
--- a/cmd/grype/cli/options/database.go
+++ b/cmd/grype/cli/options/database.go
@@ -30,7 +30,7 @@ var _ interface {
 const (
 	defaultMaxDBAge               time.Duration = time.Hour * 24 * 5
 	defaultUpdateAvailableTimeout               = time.Second * 30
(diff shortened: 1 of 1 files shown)


spiffcs:
> Separately, the DataDog synthetics are telling us that the default timeout for downloading the new database, 2 minutes, is definitely too low for some regions and networks, so I’ll raise that, probably to 5 minutes.

Given the above comment from discourseI think this is ok to merge now. 

I've approved this @willmurphyscode but left it to you to merge if you're looking to wait for any more data 👍

spiffcs (approved):

PR #2036, opened 2024-08-06
chore: remove quality gate Makefile db age check

kzantow:
Since we lowered the number of days in the database listing file, it causes an issue in the quality tests here, since the Makefile validates the database exists in the listing file and would require updating the database URL quite frequently. This PR just removes the pseudo-age-check since we're already updating this every month with an automated workflow.

### DIFF
--- a/test/quality/Makefile
+++ b/test/quality/Makefile
@@ -11,7 +11,6 @@ RESULT_SET = pr_vs_latest_via_sbom
 TEST_DB_URL_FILE = ./test-db-url
 TEST_DB_URL = $(shell cat $(TEST_DB_URL_FILE))
 TEST_DB = db.tar.gz
-LISTING_FILE = https://toolbox-data.anchore.io/grype/databases/listing.json
 
 # formatting variables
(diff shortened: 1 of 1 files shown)


test/quality/Makefile
```diff
@@ -34,15 +33,9 @@ validate: venv $(VULNERABILITY_LABELS)/Makefile ## Run all quality checks agains
 capture: sboms vulns ## Collect and store all syft and grype results
 
 .PHONY: capture
-vulns: venv $(TEST_DB) check-db ## Collect and store all grype results
+vulns: venv $(TEST_DB) ## Collect and store all grype results
 	$(YARDSTICK) -v result capture -r $(RESULT_SET)
 
-.PHONY: check-db
-check-db:
```
willmurphyscode, line 41: Do we need to set a DB max age in yardstick config or something to ensure that we don't run grype quality against a super old DB forever?

wagoodman (approved):

wagoodman:
I'll make a placeholder issue for this to capture for v6

PR #2049, opened 2024-08-12
Do not fail when inflating DB records

### LINKED ISSUE #2048: failed to parse constraint of CVE-2024-6345 which fails the scan
**What happened**:
I have an image based on alpine 3.20 and the scan fails. The error is:
`unable to find matches in DB: provider failed to inflate vulnerability record (namespace="alpine:distro:alpine:3.20" id="CVE-2024-6345" distro="alpine 3.20.2"): failed to parse constraint='< 70.3.0-rc0' format='Apk': unable to parse apk constraint phrase: failed to create comparator for '&{< 70.3.0-rc0}': unable to parse constraint version (70.3.0-rc0): invalid version`
Scan should pass
Here is a Dockerfile to reproduce:
```
# Use the Alpine 3.20 image as the base image
FROM alpine:3.20

# Install the py3-setuptools package
RUN apk add --no-cache py3-setuptools

# Set a default command, you can replace this with the command you want the container to run
CMD ["sh"]
```

I checked on alpine 3.19 and it works.
might it be related to the phrase "rc" inside the constraint?
**Environment**:
- Output of `grype version`: 0.79.5
- OS (e.g: `cat /etc/os-release` or similar): mac

wagoodman:
Fixes #2048

Today when we are fetching a entry from the DB we will raise up failures when attempting to craft vulnerability objects all the way back up through the matcher. This is not correct behavior since there may have been matched we could have raised up but did not. Instead we should be logging warnings when this occurs (since it affects the results of the report, even if it's not actionable) and still allow for other matches to be made.

### DIFF
--- a/grype/db/vulnerability_provider.go
+++ b/grype/db/vulnerability_provider.go
@@ -52,7 +52,8 @@ func (pr *VulnerabilityProvider) Get(id, namespace string) ([]vulnerability.Vuln
 	for _, vuln := range vulns {
 		vulnObj, err := vulnerability.NewVulnerability(vuln)
 		if err != nil {
-			return nil, fmt.Errorf("provider failed to inflate vulnerability record (namespace=%q id=%q): %w", vuln.Namespace, vuln.ID, err)
+			log.WithFields("namespace", vuln.Namespace, "id", vuln.ID).Errorf("failed to inflate vulnerability record: %v", err)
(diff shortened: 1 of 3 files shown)


kzantow (approved):

wagoodman:
rebasing to account for DB pin issue

PR #2054, opened 2024-08-14
fix: hash vuln db only once on load

### LINKED ISSUE #1502: LoadVulnerabilityDB could be faster with ValidateByHashOnGet
**What happened**:

Calling `LoadVulnerabilityDB` with `ValidateByHashOnGet` set to true ends up hashing the DB twice.

We would hash the DB once.

```
diff --git a/cmd/grype/cli/options/database.go b/cmd/grype/cli/options/database.go
index 83fd177..3563e6d 100644
--- a/cmd/grype/cli/options/database.go
+++ b/cmd/grype/cli/options/database.go
@@ -29,6 +29,8 @@ func DefaultDatabase(id clio.Identification) Database {
                ValidateAge: true,
                // After this period (5 days) the db data is considered stale
                MaxAllowedBuiltAge: time.Hour * 24 * 5,
+               // Repro:
+               ValidateByHashOnStart: true,
        }
 }
```

Then run grype. It is slower than you would expect because it's hashing the DB twice:

image

This ends up hashing the whole DB: https://github.com/anchore/grype/blob/b81340c7c666eee0366122ed7d220431eae3357b/grype/load_vulnerability_db.go#L23-L26
This also ends up hashing the whole DB: https://github.com/anchore/grype/blob/b81340c7c666eee0366122ed7d220431eae3357b/grype/load_vulnerability_db.go#L28

Also, this could probably be even faster if [`HashFile`](https://github.com/anchore/grype/blob/b81340c7c666eee0366122ed7d220431eae3357b/internal/file/hasher.go#L42) used `io.CopyBuffer` or a `bufio.Reader` with a larger buffer than the default used by `io.Copy`.

Since we are calling `LoadVulnerabilityDB` as a library, we are probably hitting this where nobody else would (via the `grype` cli), and we don't even look at the `Status` that gets returned, so perhaps we should just reimplement what `LoadVulnerabilityDB` is doing, but figured it would be good to file an issue.

**Environment**:
- Output of `grype version`:

```
Application:         grype
Version:             0.68.1
BuildDate:           2023-09-15T18:35:30Z
GitCommit:           brew
GoVersion:           go1.21.1
Compiler:            gc
Syft Version:        v0.90.0
Supported DB Schema: 5
```

```
$ sw_vers
ProductName:		macOS
ProductVersion:		13.5.2
BuildVersion:		22G91
```

lucasrod16:
Fixes #1502

This PR aims to improve the performance of `LoadVulnerabilityDB` when `ValidateByHashOnGet` is set to `true`

As described in the linked issue, `GetStore` and `Status` both hash the db. This PR changes `Status` to not hash the db by removing the call to `c.Validate`

As far as I can tell, `Status` is only used in a few places:

1. `LoadVulnerabilityDB`

    This should not be an issue since `GetStore` already hashes the db
    
    https://github.com/anchore/grype/blob/e7ceffadc856ce70b2854eefa3fc78f8770365e7/grype/load_vulnerability_db.go#L28

1. `runDBStatus`

    I am unsure whether hashing the db is desired behavior for `grype db status`. If so, perhaps a new field could be added to `db.Config` to allow CLI and library users to configure this behavior?
    
    https://github.com/anchore/grype/blob/e7ceffadc856ce70b2854eefa3fc78f8770365e7/cmd/grype/cli/commands/db_status.go#L33

1. `Test_DifferDirectory`

    This test doesn't appear to rely on the hashing behavior in `Status`

    https://github.com/anchore/grype/blob/e7ceffadc856ce70b2854eefa3fc78f8770365e7/grype/differ/differ_test.go#L41
    https://github.com/anchore/grype/blob/e7ceffadc856ce70b2854eefa3fc78f8770365e7/grype/differ/differ_test.go#L47

### DIFF
--- a/grype/db/curator.go
+++ b/grype/db/curator.go
@@ -120,7 +120,7 @@ func (c *Curator) Status() Status {
 		SchemaVersion: metadata.Version,
 		Location:      c.dbDir,
 		Checksum:      metadata.Checksum,
-		Err:           c.Validate(),
+		Err:           nil,
 	}
 }
 
--- a/grype/load_vulnerability_db_bench_test.go
+++ b/grype/load_vulnerability_db_bench_test.go
@@ -1,0 +1,26 @@
+package grype
+
+import (
+	"path/filepath"
+	"testing"
+
+	"github.com/anchore/grype/grype/db"
+	"github.com/anchore/grype/internal"
+)
+
+// this benchmark was added to measure the performance
+// of LoadVulnerabilityDB, specifically in regards to hash validation.
+// https://github.com/anchore/grype/issues/1502
+func BenchmarkLoadVulnerabilityDB(b *testing.B) {
+	cfg := db.Config{
+		DBRootDir:           filepath.Join(".tmp", "grype-db"),
+		ListingURL:          internal.DBUpdateURL,
+		ValidateByHashOnGet: true,
+	}
+	for range b.N {
+		_, _, _, err := LoadVulnerabilityDB(cfg, true)
+		if err != nil {
+			b.Fatal(err)
+		}
+	}
+}


lucasrod16:
Based on my local benchmark results and a sample program that calls `LoadVulnerabilityDB` executed with `time`, removing hash validation from `Status` reduced runtime duration by ~0.8s 

```shell
go test ./grype -bench=BenchmarkLoadVulnerabilityDB -benchtime=20x
```

lucasrod16:
As suggested in the linked issue, I also ran benchmarks using `io.CopyBuffer` and a `bufio.Reader` with buffer sizes ranging from 64kb - 1mb in `HashFile`, and there were no noticeable performance improvements.

lucasrod16:
Hi @wagoodman,

I noticed that this pull request has been pending for a while. Is there anything needed from me? Please let me know if there are any updates required.

Thanks!

wagoodman (approved):
thanks for the work on this @lucasrod16 I've also attached this to a v6 schema issue to help prevent regression here

PR #2064, opened 2024-08-18
fix: do not panic when given empty string arg

### LINKED ISSUE #2063: Grype panics with a nil pointer dereference error when given an empty string argument
**What happened**:
Grype panics with a nil pointer dereference error when given an empty string argument

For Grype to exit gracefully with a descriptive error message

```shell
grype ""
```

`extractReaderAndInfo` calls `stdinReader` and `decodeStdin`

https://github.com/anchore/grype/blob/89c419091483d8f4f5e83a046dd0b45aaaba3eb5/grype/pkg/syft_sbom_provider.go#L83-L86

`stdinReader` returns a `nil` `io.Reader` if no stdin input was provided (as in this case). This should likely return an error instead because calling `Read` on a `nil` `io.Reader` will always cause a panic.

https://github.com/anchore/grype/blob/89c419091483d8f4f5e83a046dd0b45aaaba3eb5/grype/pkg/syft_sbom_provider.go#L150-L152

`decodeStdin` calls `io.ReadAll` which calls `Read` on the `nil` `io.Reader` causing the panic

https://github.com/anchore/grype/blob/89c419091483d8f4f5e83a046dd0b45aaaba3eb5/grype/pkg/syft_sbom_provider.go#L110

**Environment**:
- Output of `grype version`:

```shell
Application:         grype
Version:             0.79.6
BuildDate:           2024-08-12T16:20:46Z
GitCommit:           brew
GoVersion:           go1.22.6
Compiler:            gc
Syft Version:        v1.11.0
Supported DB Schema: 5
```

M1 Mac

lucasrod16:
Fixes #2063

### DIFF
--- a/cmd/grype/cli/commands/root.go
+++ b/cmd/grype/cli/commands/root.go
@@ -347,6 +347,14 @@ func validateRootArgs(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("an image/directory argument is required")
 	}
 
+	// in the case that a single empty string argument ("") is given and there is no piped input we want to show the help text and return with a non-0 return code.
+	if len(args) != 0 && args[0] == "" && !isStdinPipeOrRedirect {
+		if err := cmd.Help(); err != nil {
+			return fmt.Errorf("unable to display help: %w", err)
+		}
+		return fmt.Errorf("an image/directory argument is required")
+	}
+
 	return cobra.MaximumNArgs(1)(cmd, args)
 }
 
--- a/grype/pkg/syft_sbom_provider.go
+++ b/grype/pkg/syft_sbom_provider.go
@@ -2,6 +2,7 @@ package pkg
 
 import (
 	"bytes"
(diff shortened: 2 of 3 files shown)


spiffcs (approved):
Thanks @lucasrod16 - great issue description, easy to reproduce, and awesome PR with the added test for regression protection.

I was able to reproduce this on latest grype, pulled your change, and ran with "" to verify the fix.

👍 🚀  Thank you very much for the bug fix!

lucasrod16:
> Thanks @lucasrod16 - great issue description, easy to reproduce, and awesome PR with the added test for regression protection.
> 
> I was able to reproduce this on latest grype, pulled your change, and ran with "" to verify the fix.
> 
> 👍 🚀 Thank you very much for the bug fix!

@spiffcs Much appreciated! I really enjoy grype as a CLI and library user and appreciate the work that has been put into it. I'm glad to contribute back!

PR #2066, opened 2024-08-19
fix: correctly close the db file in v4/v5 stores

AndreiStefanie:
This PR fixes the closing of the DB file when calling the `Close()` method.

Before, the close method on the SQLite handler was never called. This leads to issues when running Grype as a server and trying to update the database on an NFS (because the files cannot be fully deleted until the process releases them - there will be a temporary file like `.nfs047ff9ac82cfe60d00000002`)


grype/db/v5/store/store.go
```diff
@@ -290,7 +290,7 @@ func (s *store) Close() {
 	s.db.Exec("VACUUM;")
 
 	sqlDB, err := s.db.DB()
-	if err != nil {
+	if err == nil {
```
kzantow, line 293: Wouldn't we just always want to try to close this, regardless of error status?

AndreiStefanie: Indeed, it seems that the DB method might return both an error and a valid instance of the DB. Changed the logic. Thanks!

kzantow:
Thanks for the contribution, @AndreiStefanie

PR #2088, opened 2024-08-29
Escape description in junit.tmpl

felixbuenemann:
Fixes broken XML for description like "ruby < 3.0" by html escaping the text.


kzantow (approved):
Thanks @felixbuenemann !

PR #2139, opened 2024-09-20
Use faster xz lib for decompression

wagoodman:
Moving from github.com/ulikunitz/xz to github.com/xi2/xz saves upwards of half the time when decompressing archives on download. This PR is the spiritual mirror to https://github.com/hashicorp/go-getter/pull/262 , which was never merged and closed for some reason.

Partially addresses https://github.com/anchore/grype-db/issues/367


wagoodman:
added `changelog-ignore` since this does not impact the system yet (there are no DBs yet published that can leverage this).

kzantow (approved):

PR #2143, opened 2024-09-25
docs: start documenting how to add new provider

willmurphyscode:
Many of these docs are over at Vunnel, but Grype is sort of the central repo for how these vulnerabilities are reported, so an outline at least belongs in Grype.


willmurphyscode:
This documentation is published in a much better form at https://oss.anchore.com/docs/contributing/vunnel/

Issue #2132, opened 2024-09-17
Simplify grype DB access abstractions


wagoodman:
Here are the rough steps today with v1-5 to get a match from the DB, starting within a matcher:

1. Matchers use the [search package Criteria](https://github.com/anchore/grype/blob/89c419091483d8f4f5e83a046dd0b45aaaba3eb5/grype/matcher/python/matcher.go#L34-L39) to access the given vulnerability.Provider , where the provider is DB schema agnostic and passed to the matcher
2. [Search package static helper functions ](https://github.com/anchore/grype/blob/89c419091483d8f4f5e83a046dd0b45aaaba3eb5/grype/search/distro.go#L29)query and refine the final set of vulnerability candidates, using the given vulnerability.Provider 
3. The vulnerability.Provider then uses the [DB specific store reader](https://github.com/anchore/grype/blob/89c419091483d8f4f5e83a046dd0b45aaaba3eb5/grype/db/vulnerability_provider.go#L82) to find vulnerabilities, and return [normalized vulnerability objects](https://github.com/anchore/grype/blob/89c419091483d8f4f5e83a046dd0b45aaaba3eb5/grype/vulnerability/vulnerability.go#L17) (agnostic to DB schema)
4. The store reader [accesses the underlying sqlite store directly](https://github.com/anchore/grype/blob/89c419091483d8f4f5e83a046dd0b45aaaba3eb5/grype/db/v5/store/store.go#L117) and raises up [DB specific vulnerability objects](https://github.com/anchore/grype/blob/89c419091483d8f4f5e83a046dd0b45aaaba3eb5/grype/db/v5/vulnerability.go#L11)

Some observations out of this are:
- The search functions are static and require continually passing a provider
    - In v6 we improve on this by requiring the matcher to instantiate a client with specific configuration and access to the store at construction, and the client provides the raw search functionalities
- The provider is DB agnostic, which requires that all data is fully deserialized from the DB (even if it’s not needed)
    - In v6 we improve on this by searching against indexed tables and only deserialize related blobs if they are needed beyond a specific step

## Changes

([feel free to browse the prototype](https://github.com/anchore/grype/tree/db-v6-blob-store))

The search package should be where DB model deserialization occurs to leverage as many optimizations as possible while searching. This would remove some unnecessary abstractions (the vulnerability.Provider).

Matchers search by criteria against a client (where the client is driven by search criteria and is passed into the matcher at matcher construction)

<details>
[...]

Issue #2128, opened 2024-09-17
Add DB v6 reader/writers


wagoodman:
DB v6 is meant to cover [several use cases](https://github.com/anchore/grype-db/issues/108) (you can safely ignore this link). The  high level goals are:
- Make the distributed DB smaller
  - Via removing necessary information: for instance, "namespace" is a hold over from earlier projects and is no longer strictly necessary.
  - Via minimizing redundant information: in order to optimize the matching process, the v5 Vulnerability table is filled with redundant information (rows are duplicated for every unique fix-in found in upstream data).
- Make adding new use cases to the DB easier (e.g. add EPSS or KEV data sources for grype to functionally use them without needing to make structural changes to grype)

The high level design is as follows:
- Add a table-per-search-usecase. The bare minimum information needed to search records are in these tables, which are indexed. These tables allow you to identify a record ID. _We'll refer to these tables as "Search" or "*Hanadle" tables_.
- Have a pseudo content addressable (single) blobs table that maps record ID to JSON blobs. When writing to the DB, these record IDs are 1:1 to JSON content digests, so that we cheaply don't write duplicate information. These digests should be stripped away before distribution, since they are not strictly needed at runtime. These JSON blobs represent the business elements (e.g. Vulnerability, Severity, AffectedPackage, etc). All payloads are self-describing with a named/version "kind" field used for deserialization concerns. _We'll refer to these as "*Blob" models_.
- Have auxiliary tables related to search tables to aid in refinement / augmentation to searches. 

Here are a list of the “Handle” tables to search against:
- AffectedPackageHandles: search by package name within an ecosystem
- AffectedCPEHandles: search by CPE fields
- VulnerabilityHandles: search by vulnerability ID

These are related to two other auxiliary tables:
- Packages: listing of all unique pairings of package name and ecosystem
- OperatingSystems: listing of all unique tuples of distro name, major version, minor version (with optional codename)

Here's how they AffectedPackageHandle table would relate to auxiliary tables:

```sql
CREATE TABLE AffectedPackageHandle (
    id INTEGER PRIMARY KEY,
    vulnerability_id INTEGER NOT NULL,
    operating_system_id INTEGER,
    package_id INTEGER NOT NULL,
    blob_id INTEGER NOT NULL,
    FOREIGN KEY(vulnerability_id) REFERENCES VulnerabilityHandle(id),
    FOREIGN KEY(operating_system_id) REFERENCES OperatingSystem(id),
[...]

PR #2146, opened 2024-09-26
Add v6 DB metadata store

wagoodman:
Partial implementation for #2128 and #2132 in terms of the `DBMetadata` object

Adds:
- the `DBMetadata` table for v6
- the store reader and writer implementation

### DIFF
--- a/grype/db/internal/gormadapter/open.go
+++ b/grype/db/internal/gormadapter/open.go
@@ -23,11 +23,60 @@ var readOptions = []string{
 	"mode=ro",
 }
 
-// Open a new connection to a sqlite3 database file
-func Open(path string, write bool) (*gorm.DB, error) {
-	memory := len(path) == 0
+type config struct {
+	path   string
+	write  bool
+	memory bool
+}
 
-	if write && !memory {
+type Option func(*config)
+
+func WithTruncate(truncate bool) Option {
+	return func(c *config) {
+		c.write = truncate
+	}
+}
+
+func newConfig(path string, opts []Option) config {
+	c := config{}
+	c.apply(path, opts)
+	return c
+}
+
+func (c *config) apply(path string, opts []Option) {
+	for _, o := range opts {
+		o(c)
+	}
+	c.memory = len(path) == 0
+	c.path = path
+}
+
+func (c config) shouldTruncate() bool {
+	return c.write && !c.memory
+}
+
+func (c config) connectionString() string {
+	var conn string
+	if c.path == "" {
+		conn = ":memory:"
+	} else {
+		conn = fmt.Sprintf("file:%s?cache=shared", c.path)
+	}
+
+	if !c.write && !c.memory {
+		// &immutable=1&cache=shared&mode=ro
+		for _, o := range readOptions {
+			conn += fmt.Sprintf("&%s", o)
+		}
+	}
+	return conn
+}
+
+// Open a new connection to a sqlite3 database file
+func Open(path string, options ...Option) (*gorm.DB, error) {
+	cfg := newConfig(path, options)
+
+	if cfg.shouldTruncate() {
 		if _, err := os.Stat(path); err == nil {
 			if err := os.Remove(path); err != nil {
 				return nil, fmt.Errorf("unable to remove existing DB file: %w", err)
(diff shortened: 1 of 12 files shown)


grype/db/internal/gormadapter/open.go
```diff
@@ -23,9 +25,18 @@ var readOptions = []string{
...
 // Open a new connection to a sqlite3 database file
 func Open(path string, write bool) (*gorm.DB, error) {
-	if write {
-		// the file may or may not exist, so we ignore the error explicitly
-		_ = os.Remove(path)
+	memory := len(path) == 0
+
+	if write && !memory {
+		if _, err := os.Stat(path); err == nil {
+			if err := os.Remove(path); err != nil {
```
willmurphyscode, line 32: This really surprised me the first time I read it.

What I think is going on after thinking about it is: If we're opening the database for _write_ access, it's because we're making a whole new Grype DB and we should blow away the old one. But it requires a lot of knowledge of how Grype / Grype DB work to know that.

What if we split into two functions, `Create` and `Open`, and `Open` is always read-only? I think it would be less surprising for clients if calling `Create("/some/path")` when `/some/path` exists destroys the old data than calling `Open`.

I don't love that `""` as the path means, "using sqlite in memory," but I can't think of a good alternative right now.

willmurphyscode: I didn't realize we were fitting into an existing interface. I can probably live with the current implementation in that case.

willmurphyscode (commented):
I think I'd like to do a sync review on this, since it's foundational, to make sure I understand how it.

grype/db/internal/gormadapter/open.go
willmurphyscode: On second thought, aren't we swapping out a bunch of the management of the database in the v6 change? In other words, this meets the interface we used for v5, but is it time to change that interface?

grype/db/internal/gormadapter/open.go
```diff
@@ -13,6 +13,8 @@ var writerStatements = []string{
 	// on my box it reduces the time to write from 10 minutes to 10 seconds (with ~1GB memory utilization spikes)
 	`PRAGMA synchronous = OFF`,
 	`PRAGMA journal_mode = MEMORY`,
+	`PRAGMA cache_size = 100000`,
+	`PRAGMA mmap_size = 268435456`, // 256 MB
```
willmurphyscode, line 17: notes from sync review with @wagoodman:

Cache size:
* These are not required, but improve performance
* The db write is much faster if more of the db is cached in memory
mmap_size:
* when reading the database, use more RAM to go faster

These numbers are both higher than the defaults.

grype/db/internal/gormadapter/open.go
willmurphyscode: Notes from sync review: we'll split this into two functions, `Create` which truncates the existing, and `Open` (or another name) that opens the database for read operations. Grype never updates a database, so Create and OpenForReading are the only logical operations.

grype/db/v6/store.go
```diff
@@ -0,0 +1,61 @@
...
+	write  bool
+}
+
+func newStore(cfg Config, write bool) (*store, error) {
+	db, err := gormadapter.Open(cfg.DBFilePath(), write)
+	if err != nil {
+		return nil, err
+	}
+
+	db.Exec("PRAGMA foreign_keys = ON")
```
willmurphyscode, line 25: notes from sync review:

Why isn't this in the list of PRAGMA statements at `grype/db/internal/gormadapter/open.go` line 15 or so?

Answer: This `PRAGMA` only applies to v6; the others apply to all schemas. It's probably safe to move though - the other schemas lack foreign keys, so the pragma should have no effect.

grype/db/v6/store.go
```diff
@@ -0,0 +1,61 @@
...
+
+	return &store{
+		dbMetadataStore: newDBMetadataStore(db),
+		db:              db,
+		config:          cfg,
+		write:           write,
+	}, nil
+}
+
+func migrate(db *gorm.DB, models ...any) error {
```
willmurphyscode, line 41: take or leave feedback: `migrate` sounds like we're changing an existing database schema, but really we're migrating from an empty database to one with all our models in it. A name with `create` in it might make more sense?

grype/db/internal/gormadapter/open.go
wagoodman: I came to a slightly different conclusion when thinking about this further: this function returns `*gorm.DB` which has read and write capabilities. Having separate readonly and write functions doesn't really help here if the gorm type can do both reads and writes anyway. We really want a helper to make clear what the connection string will be and add clear options (replace `write` with `truncate`).

I made a config object to encapsulate and test this logic which is controlled by functional options (currently only one).

grype/db/v6/store.go
wagoodman: it's too small of a function anyway to be useful, I rolled it into the `Open` function

grype/db/v6/store.go
wagoodman: I missed this during the review: those pragmas are for writers only, but we want FKs for readers and writers

wagoodman: but the spirit of the comment is essentially: move these pragmas within the `Open` call -- I'll make that change

willmurphyscode (approved):

Issue #2125, opened 2024-09-17
DB v6 distribution approach


wagoodman:
Today the grype DB is distributed via a hosted `listing.json` file with URLs to DBs, listing out historical entries to N many days. There are a few points here:
- The listing file serves two purposes: to find the latest DB and access historical DBs. The former is the primary use case of the listing file, the latter is added weight.
- The listing file takes absolute URLs, not relative paths. This makes crafting a listing file not portable between environments, thus, needs to be rebuilt for each environment deployed to.
- Distribution definitions are not tied to a DB schema version, and the listing contains entries for all schema versions. This prevents from being able to make breaking changes to the listing file format itself. This also requires more coordination when updating the listing file nightly (upload all databases in a fan-out, then fan in to update the listing -- ideally there would be no need to fan in [which causes lots of fun failure modes to think through]). (this is a grype-db repo concern, but motivates the changes here)
- We do not check the DB checksum on start, mostly because this is expensive, but also there have been inefficiencies in this load path too. 
  - https://github.com/anchore/grype/issues/1648
  - https://github.com/anchore/grype/issues/2077
  - https://github.com/anchore/grype/pull/2054 

Based on these points here are the suggested changes:
- we support listings for a single DB schema only -- each new schema will be hosted in a new location.
- replace the single listing file with two files: `latest.json` and `history.json`, split based on use case. This means that the most common use case (latest.json) is as small as possible, removing pressure from the CDN.
- split the `db.Curator` by use case: DB distribution vs access to an already installed DB.
- use xxh64 for DB checksum (not sha256), which is rather quick when checking large DB files
- use relative URLs (relative to where the latest.json/history.json files are hosted, not absolute ones). Note: we should still be able to express absolute URLs for operational fallback positions, but this should be the exception, not the norm.

## `latest.json` file

```json
{
  "schemaVersion": 6,
  "status": "active",
  "archive": {
    "database": {
      "built": "2024-08-23T12:34:56Z",
      "checksum": "xxhash64:1a2b3c4d5e6f7g8h", 
      "providers": [
        {
          "name": "nvd",
          "compiled": "2024-08-23T08:00:00Z"
        },
        {
          "name": "github",
          "compiled": "2024-08-23T09:00:00Z"
        },
        ...
      ]
    },
    "path": "databases/v6/grype-db_v6_2024-08-23T11:22:22Z_1724213998.tar.gz",
[...]

PR #2150, opened 2024-09-30
Add v6 distribution client

wagoodman:
Adds a new `distribution.Client` (for v6) responsible for:
- fetching a `latest.json` file
- determining if a given existing DB description is older or newer than a candidate update
- downloads candidate DB archives into a given directory, automatically failing if an expected archive checksum differs

One benefit of this approach is that the client is unaware of any installation concerns.

Partially implements #2125

### DIFF
--- a/grype/db/v6/description.go
+++ b/grype/db/v6/description.go
@@ -1,0 +1,87 @@
+package v6
+
+import (
+	"fmt"
+	"path"
+	"time"
+
+	"github.com/OneOfOne/xxhash"
+	"github.com/spf13/afero"
+
+	"github.com/anchore/grype/grype/db/internal/schemaver"
+	"github.com/anchore/grype/internal/file"
+)
+
+const DescriptionFileName = "metadata.json"
+
+type Description struct {
+	// SchemaVersion is the version of the DB schema
+	SchemaVersion schemaver.SchemaVer `json:"schemaVersion,omitempty"`
+
+	// Built is the timestamp the database was built
+	Built Time `json:"built"`
+
+	// Checksum is the self-describing digest of the database file
+	Checksum string `json:"checksum"`
+}
+
+type Time struct {
+	time.Time
+}
+
+func (t Time) MarshalJSON() ([]byte, error) {
+	return []byte(fmt.Sprintf("%q", t.String())), nil
+}
+
+func (t *Time) UnmarshalJSON(data []byte) error {
+	str := string(data)
+	if len(str) < 2 || str[0] != '"' || str[len(str)-1] != '"' {
+		return fmt.Errorf("invalid time format")
+	}
+	str = str[1 : len(str)-1]
+
+	parsedTime, err := time.Parse(time.RFC3339, str)
+	if err != nil {
+		return err
+	}
+
+	t.Time = parsedTime.In(time.UTC)
+	return nil
+}
+
+func (t Time) String() string {
+	return t.Time.UTC().Round(time.Second).Format(time.RFC3339)
+}
+
+func NewDescriptionFromDir(fs afero.Fs, dir string) (*Description, error) {
+	// checksum the DB file
+	dbFilePath := path.Join(dir, VulnerabilityDBFileName)
+	digest, err := file.HashFile(fs, dbFilePath, xxhash.New64())
+	if err != nil {
+		return nil, fmt.Errorf("failed to calculate checksum for DB file (%s): %w", dbFilePath, err)
+	}
+	namedDigest := fmt.Sprintf("xxh64:%s", digest)
+
+	// access the DB to get the built time and schema version
+	r, err := NewReader(Config{
+		DBDirPath: dir,
+	})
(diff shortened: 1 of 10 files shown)


grype/db/v6/description.go
```diff
@@ -0,0 +1,87 @@
...
+	"time"
+
+	"github.com/OneOfOne/xxhash"
+	"github.com/spf13/afero"
+
+	"github.com/anchore/grype/grype/db/internal/schemaver"
+	"github.com/anchore/grype/internal/file"
+)
+
+const DescriptionFileName = "metadata.json"
```
willmurphyscode, line 15: It always throws me off a bit that this is sometimes called Description and sometimes called metadata. Is moving to schema v6 the right time to either call it `description.json` or call the struct `Metadata`? Not a blocker, but having two names has always confused me.

grype/db/v6/distribution/client.go
```diff
@@ -0,0 +1,234 @@
...
+		log.Debugf("check for vulnerability update failed: %+v", err)
+	}
+	return c.isUpdateAvailable(current, updateEntry)
+}
+
+func (c client) isUpdateAvailable(current *v6.Description, candidate *LatestDocument) (*Archive, error) {
+	if candidate == nil {
+		return nil, nil
+	}
+	log.Debugf("found database update candidate: %s", candidate.Archive.Description)
```
willmurphyscode, line 101: This log message is a little confusing - it sounds like we've already decided to update. Maybe something like `"comparing current db against ..."` or just skip the log here since we'll always say the update is or isn't available.

grype/db/v6/distribution/client.go
```diff
@@ -0,0 +1,234 @@
...
+}
+
+func isSupersededBy(m *v6.Description, other v6.Description) bool {
+	if m == nil {
+		log.Debugf("cannot find existing metadata, using update...")
+		// any valid update beats no database, use it!
+		return true
+	}
+
+	otherModel, otherOk := other.SchemaVersion.Model()
```
willmurphyscode, line 208: Can we rename or comment this? The `model` is the 6 of `6.0.0` but I thought it was going to be the whole `6.0.0` and misunderstood the rest of the function. I know we don't want `major` because that's semver, not schemaver, but `otherModelComponent` might be nice because it calls attention to the fact that this isn't the whole thing. (`Model` is a word that doesn't mean much due to overloading, like `instance` or `container` or something, so I didn't realize the method call on this line subsetted the SchemaVersion).

Maybe I just need to know schemaver better 🤷

grype/db/v6/distribution/client.go
```diff
@@ -0,0 +1,234 @@
...
+		return false
+	}
+
+	if !currentOk {
+		log.Errorf("update has no schema version, doing nothing...")
+		return false
+	}
+
+	if otherModel != currentModel {
+		log.Debugf("update is for a different DB schema, skipping...")
```
willmurphyscode, line 222: This is quite unexpected. Maybe `log.Warn`? (indicates that a grype built for schema v6 tries to download a v5 or v7 database, for example).

grype/db/v6/distribution/client.go
```diff
@@ -0,0 +1,234 @@
...
+		return false
+	}
+
+	if other.Built.After(m.Built.Time) {
```
willmurphyscode, line 226: note: `other.Built` is now the time when the database was _packaged_. This is a change from v5 where the same field means the time when the data was pulled, not when the data was packaged.

This change is important because if there's a flaw in the packaging, we can re-run the package step without re-running the build step and still make a new database that can be deterministically used.

We could _also_ just make grype always take the new update if the files differ at all, which would allow for server side rolling back of the database, and for always updating when switching mirrors, which is something we might discuss.

grype/db/v6/distribution/status.go
```diff
@@ -0,0 +1,16 @@
...
+type Status string
+
+const LifecycleStatus = StatusActive
+
+const (
+	// StatusActive indicates the database is actively being maintained and distributed
+	StatusActive Status = "active"
+
+	// StatusDeprecated indicates the database is still being distributed but is approaching end of life. Upgrade grype to avoid future disruptions.
+	StatusDeprecated Status = "deprecated"
```
willmurphyscode, line 12: This isn't used anywhere, but when downloading updates if the `latest.json` says we're updating a deprecated or eol schema version, we should `log.Warn` and partybus.Notify people that they need to upgrade.

grype/db/v6/distribution/client.go
wagoodman: removed

grype/db/v6/distribution/client.go
wagoodman: I renamed those helper functions to `*Part` (e.g. `ModelPart`)

grype/db/v6/description.go
wagoodman: I went with description.json

PR #2155, opened 2024-10-02
Add a space following the "Name:" label in html.tmpl

deftdawg:
Should be a space between "Name:" and "container:tag" currently that doesn't render, so adding a space here fixes that.  Tested on Firefox


spiffcs (approved):

spiffcs:
Thanks @deftdawg!

PR #2162, opened 2024-10-04
fix: use fix info from secDB in APK matcher even if NVD fix info present

willmurphyscode:
This is to address false fixed ins in Alpine, Chainguard, or Wolfi that might be introduced by anchore/grype-db#112.

# TODO

- [x] refactor to pass lints once approach is approved
- [x] add a test asserting that sec db fixes do stay in

### DIFF
--- a/grype/matcher/apk/matcher.go
+++ b/grype/matcher/apk/matcher.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 
+	grypeDB "github.com/anchore/grype/grype/db/v5"
 	"github.com/anchore/grype/grype/distro"
 	"github.com/anchore/grype/grype/match"
 	"github.com/anchore/grype/grype/pkg"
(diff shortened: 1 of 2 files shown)


grype/matcher/apk/matcher.go
```diff
@@ -88,7 +88,12 @@ cveLoop:
 		secDBVulnerabilitiesForID, exists := secDBVulnerabilitiesByID[id]
 		if !exists {
 			// does not exist in secdb, so the CPE record(s) should be added to the final results
-			finalCpeMatches = append(finalCpeMatches, cpeMatchesForID...)
+
+			// remove fixed-in versions, since NVD doesn't know when Alpine will fix things
+			for _, nvdOnlyMatch := range cpeMatchesForID {
+				nvdOnlyMatch.Vulnerability.Fix = vulnerability.Fix{}
```
willmurphyscode, line 94: TODO: state unknown

willmurphyscode:
I hesitated over the `bug` label for a minute, since it's only a bug once https://github.com/anchore/grype-db/pull/112 is merged, but I decided that was still the right label.

spiffcs (approved):

spiffcs:
This is the correct fix info to surface for secdb.

@willmurphyscode I noticed that your PR also mentioned chainguard and wolfi. I have not looked at those vuln providers in a while, but it seems these changes are only for the APK matcher. 

Are there more changes you're looking to make?

This does look 🟢 from the correctness of secdb fixed in is being respected over nvd for the APK matcher and the tests prove it.

willmurphyscode:
@spiffcs thanks for the question. Chainguard and Wolfi images also use APK packages, so this code change will affect them as well. They also use secDB fixes, but with different URLs (the vunnel providers are very similar). As far as I can tell, this change covers Chainguard and Wolfi images as well.

PR #2174, opened 2024-10-09
Add `grype db providers` command

### LINKED ISSUE #2131: Add DB providers command
It should be trivial to see what data providers (from vunnel) populated the DB, along with timestamp information:
```
grype db providers

alpine       2024-09-16T01:34:04.087474Z      ...
amazon       2024-09-16T01:34:04.087474Z      ...
chainguard   2024-10-16T01:34:04.087474Z      ...
debian       2024-09-16T01:34:04.087474Z      ...
github       2024-09-28T01:34:04.087474Z      ...
mariner      2024-09-16T01:34:04.087474Z
nvd          2023-11-16T01:34:04.087474Z
oracle       2024-09-16T01:34:04.087474Z
rhel         2024-09-16T01:34:04.087474Z
sles         2024-09-16T01:34:04.087474Z
ubuntu       2024-09-16T01:34:04.087474Z
wolfi        2024-09-16T01:34:04.087474Z
```

This should be available as:
- table output
- json output

Example JSON output:
```json
{
  "providers": [
    {
      "name": "nvd",
      "dateCaptured": "2024-09-16T01:34:04.087474Z",
      ...
    },
    ...
  ]
}
```

Ideally we'd capture vunnel version information either in bulk or with each provider. For instance, see the prototype information capture here: https://github.com/anchore/grype/blob/db-v6-blob-store/grype/db/v6/models.go#L71-L95

ADorigi:
# Issue

resolves #2131 

# Changes

- Provider data is read from `provider-metadata.json` file.
- Added flag `-o`/`--output` flags which accept `json` and `table`.
- Display logic is separate from the provider data collection. 
- Can be easily updated for db schema `v6`.
- Fails gracefully, with relevant error message, for absence of `provider-metadata.json` file.
- Updated `Readme.md` for the new command.

### DIFF
--- a/cmd/grype/cli/commands/db_diff.go
+++ b/cmd/grype/cli/commands/db_diff.go
@@ -10,7 +10,6 @@ import (
 	"github.com/anchore/grype/cmd/grype/cli/options"
 	"github.com/anchore/grype/grype/db/legacy/distribution"
 	"github.com/anchore/grype/grype/differ"
-	"github.com/anchore/grype/internal"
 	"github.com/anchore/grype/internal/bus"
 	"github.com/anchore/grype/internal/log"
 )
@@ -30,7 +29,7 @@ func (d *dbDiffOptions) AddFlags(flags clio.FlagSet) {
 
 func DBDiff(app clio.Application) *cobra.Command {
 	opts := &dbDiffOptions{
-		Output:    internal.TableOutputFormat,
+		Output:    "table",
 		DBOptions: *dbOptionsDefault(app.ID()),
 	}
 
--- a/cmd/grype/cli/commands/db_list.go
+++ b/cmd/grype/cli/commands/db_list.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/anchore/clio"
 	"github.com/anchore/grype/grype/db/legacy/distribution"
-	"github.com/anchore/grype/internal"
 )
 
 type dbListOptions struct {
@@ -25,7 +24,7 @@ func (d *dbListOptions) AddFlags(flags clio.FlagSet) {
 
 func DBList(app clio.Application) *cobra.Command {
 	opts := &dbListOptions{
-		Output:    internal.TextOutputFormat,
+		Output:    "text",
 		DBOptions: *dbOptionsDefault(app.ID()),
 	}
 
@@ -59,7 +58,7 @@ func runDBList(opts *dbListOptions) error {
 	}
 
 	switch opts.Output {
-	case internal.TextOutputFormat:
+	case "text":
 		// summarize each listing entry for the current DB schema
 		for _, l := range available {
 			fmt.Printf("Built:    %s\n", l.Built)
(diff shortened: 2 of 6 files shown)


willmurphyscode:
Hi @ADorigi thanks for the quick work here!

We need a couple things before we can merge it.

1. Can you clean up static analysis so that it passes? I think running `make lint-fix` and committing the resulting diff is all that's needed. You can check locally whether `make static-analysis` passes - that's the same check as runs in CI.
2. Can you add a test for this command? You could either add a unit test as a sibling of your new file, or a CLI test in https://github.com/anchore/grype/tree/main/test/cli that proves the new command exists and prints the right sort of thing.

Let me know if you have any questions! Thanks!

ADorigi:
@willmurphyscode
I have addressed the following:
1. Cleaned the branch with `make lint-fix` and `make static-analysis`, ensuring it passes.
2. Added a test for the `grype db providers` command in the `test/cli` directory.

willmurphyscode:
Hi @ADorigi! Thanks for getting up a rev that passes the validations. I have being trying out the command locally, and I have some additional feedback. There are some things that need to change in order to merge this, and some things that are merely nice to have, listed separately here: 

Things that need to change

* `grype db providers` defaults to _raw_ output, which is surprising; all our other commands without `--output / -o` default to a table output, so we need this to also default to table output.
* `grybe db providers` uses cell delimiting in the table output, which puts in extra lines and delimiters like ` | ` show up in grepped output. Please copy the table writers settings from https://github.com/anchore/syft/blob/326754509791f55d051bb9808009de1e2fc0add5/syft/format/table/encoder.go#L71-L82 - it should look like the table from grype or syft's default command.
* Remove `-o raw`: We're going to change the internal representation of this data very soon, so providing access to the raw data is setting people up to count on we know will change. Let's remove `-o raw` and just have table and JSON output.
* testing: the CLI test for the JSON output doesn't assert that the output is actually JSON. Copying the trait assertion from https://github.com/anchore/syft/blob/326754509791f55d051bb9808009de1e2fc0add5/test/cli/trait_assertions_test.go#L33 and invoking in in the `-o json` test would make these tests a lot more valuable.

Nice to have changes:

* nitpick: `internal.TableOutput` etc. constants are too broadly scoped - I realize that you did this to make the linter stop complaining about duplicate string literals, but `db diff` and `db providers` are totally different, so they should really each have their own constants.
* For extra credit: create some exemplary unit tests in `cmd/grype/cli/commands/db_providers.go`. I realize there aren't great examples to work from, but if you could get a small table test going that checks on different outputs, you could move cases besides `-o json` into unit tests and keep the CLI tests small.

ADorigi:
Hello @willmurphyscode 
Thank you for explaining the required changes in detail. Greatly appreciate the same. 

I have made the following changes:
- Updated the default output format to be `table`, to align it with the other commands.
- Updated the formatting of the tablewriter to match the one in [syft](https://github.com/anchore/syft/blob/326754509791f55d051bb9808009de1e2fc0add5/syft/format/table/encoder.go#L71-L82).
- The `db providers` command supports only `table` and `json` as the output formats, with `table` now as default.
- Added `json` and `table` trait assertions for the cli tests. 
- Limited the use of constants to only the `db providers` command
- Updated test files
  - added a test file alongside `cmd/grype/cli/commands/db_providers.go`, to test individual components of the implementation
  - tests in `test/cli/db_providers_test.go` deal with how the command will actually be used.
  - as mentioned, I have not used any concrete example for these tests and will be willing to update them if required.

willmurphyscode (approved):

PR #2188, opened 2024-10-14
Account for implicit 0s in rpm release versions

### LINKED ISSUE #398: RPM version comparison oddity with release field
When taking a look at another RPM version-related investigation @luhring  noticed the following oddity:
```
# this resolves to true
8.3.1-5.el8 < 0:8.3.1-5.el8.0.0

# this resolves to true
8.3.1-5.el8.0.0 > 0:8.3.1-5.el8
```
It appears that this could be falling back to string length (i.e. if the string is a subset of the other, then the longest string is "greater" than the other) --but I haven't verified this.

Both of these statements (I think) should resolve as "false". I've added these failing tests here: https://github.com/anchore/grype/commit/138b1a22f7226cc64df718993c331293cd4ace3c

This issue should:
1. at least determine if this behavior is unexpected or not
2. if there is a problem, fix it
3. identify other cases that may exhibit similar behavior (and add them as tests)

wagoodman:
Closes #398

### DIFF
--- a/grype/version/rpm_version.go
+++ b/grype/version/rpm_version.go
@@ -151,10 +151,11 @@ func compareRpmVersions(a, b string) int {
 	// get alpha/numeric segments
 	segsa := alphanumPattern.FindAllString(a, -1)
 	segsb := alphanumPattern.FindAllString(b, -1)
-	segs := int(math.Min(float64(len(segsa)), float64(len(segsb))))
+	maxSegs := int(math.Max(float64(len(segsa)), float64(len(segsb))))
+	minSegs := int(math.Min(float64(len(segsa)), float64(len(segsb))))
 
 	// compare each segment
-	for i := 0; i < segs; i++ {
+	for i := 0; i < minSegs; i++ {
 		a := segsa[i]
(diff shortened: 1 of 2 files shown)


kzantow (approved):

grype/version/rpm_version.go
```diff
@@ -151,10 +151,11 @@ func compareRpmVersions(a, b string) int {
 	// get alpha/numeric segments
 	segsa := alphanumPattern.FindAllString(a, -1)
 	segsb := alphanumPattern.FindAllString(b, -1)
-	segs := int(math.Min(float64(len(segsa)), float64(len(segsb))))
+	maxSegs := int(math.Max(float64(len(segsa)), float64(len(segsb))))
```
willmurphyscode, line 154: Why are we casting to a float here? Also, I think this code may pre-date built-in `min` and `max` functions, and should just use them? https://go.dev/ref/spec#Min_and_max

wagoodman: TIL! I'll update it 👍

PR #2197, opened 2024-10-17
fix: azurelinux considered as comprehensive distro

westonsteimel:
azurelinux should be considered as providing a comprehensive vulnerability data source


westonsteimel:
An example image is `mcr.microsoft.com/azurelinux/base/python:3.12.3-1-azl3.0.20240727` which should no longer display GHSA matches on `setuptools`

spiffcs (approved):

PR #2226, opened 2024-10-30
Search for distro namespaces with incomplete information

### LINKED ISSUE #2222: Missing alpine patch version yields inaccurate results
**What happened**:
I'm seeing that when providing your own package details and alpine distro version details independent of an image, the distro patch version being left off may result in incorrect results:
```
❯ echo "pkg:apk/openssh@9.6_p1-r1" |  grype purl:/dev/stdin --distro 'alpine:3.19.4' -q
No vulnerabilities found

❯ echo "pkg:apk/openssh@9.6_p1-r1" |  grype purl:/dev/stdin --distro 'alpine:3.19' -q
NAME     INSTALLED  FIXED-IN   TYPE  VULNERABILITY  SEVERITY
openssh  9.6_p1-r1  9.8_p1-r0  apk   CVE-2024-6387  High
```

Please take note the vuln info for this record:
```
❯ sqlite3 --header --column /Users/wagoodman/Library/Caches/grype/db/5/vulnerability.db 'select id,package_name,namespace,version_format,version_constraint,fixed_in_versions from vulnerability where namespace like "%alpine%" and id == "CVE-2024-6387"'

id             package_name  namespace                  version_format  version_constraint  fixed_in_versions
-------------  ------------  -------------------------  --------------  ------------------  -----------------
CVE-2024-6387  openssh       alpine:distro:alpine:3.17  apk             < 9.1_p1-r6         ["9.1_p1-r6"]
CVE-2024-6387  openssh       alpine:distro:alpine:3.18  apk             < 9.3_p2-r2         ["9.3_p2-r2"]
CVE-2024-6387  openssh       alpine:distro:alpine:3.19  apk             < 9.6_p1-r1         ["9.6_p1-r1"]
CVE-2024-6387  openssh       alpine:distro:alpine:3.20  apk             < 9.7_p1-r4         ["9.7_p1-r4"]
CVE-2024-6387  openssh       alpine:distro:alpine:edge  apk             < 9.8_p1-r0         ["9.8_p1-r0"]
```

The results with `--distro 'alpine:3.19.4' ` is correct/expected: this is a valid secdb entry that should prune the NVD match from results entirely.

What appears to be happening is that when the patch version is not specified this is using the `alpine:distro:alpine:edge` namespace to search against instead of `alpine:distro:alpine:3.19` which has a different package fix version entirely (thus won't be pruned).

Using grype with this purl with both `--distro 'alpine:3.19.4' ` and `--distro 'alpine:3.19' ` should have the same result: `CVE-2024-6387` should be absent from the results.

wagoodman:
Fixes #2222 

Today when specifying `--distro` you need to provide exactly the correct version information. This is not great since some distros track based on `major` version only and others with `major.minor`. This PR changes this behavior by searching for the closest matching namespace based on the provided distro information. So if the user provides `--distro alpine:3.20` then we'll correctly line this up with the `alpine:3.20` namespace (where as today this will match against `edge`).

This same loosening of namespace searching based on arbitrary version info has been applied to all distros/namespaces within the v5 schema only.

### DIFF
--- a/grype/db/v5/namespace/index.go
+++ b/grype/db/v5/namespace/index.go
@@ -3,17 +3,20 @@ package namespace
 import (
 	"fmt"
 	"regexp"
+	"sort"
 	"strings"
 
+	hashiVer "github.com/anchore/go-version"
 	"github.com/anchore/grype/grype/db/v5/namespace/cpe"
 	"github.com/anchore/grype/grype/db/v5/namespace/distro"
 	"github.com/anchore/grype/grype/db/v5/namespace/language"
 	grypeDistro "github.com/anchore/grype/grype/distro"
+	"github.com/anchore/grype/internal"
 	"github.com/anchore/grype/internal/log"
 	syftPkg "github.com/anchore/syft/syft/pkg"
 )
 
-var alpineVersionRegularExpression = regexp.MustCompile(`^(\d+)\.(\d+)\.(\d+)$`)
+var simpleSemVer = regexp.MustCompile(`^(?P<major>\d+)(\.(?P<minor>\d+)(\.(?P<patch>\d+(?P<remaining>[^-_]+)*))?)?$`)
 
 type Index struct {
 	all         []Namespace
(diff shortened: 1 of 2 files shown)


kzantow (approved):
LGTM

Issue #744, opened 2022-05-11
add Gentoo Linux support


williamh:
Hi all,

I am a packager for Gentoo Linux, and I am interested in adding both this package and syft to our distribution.
I see that you list several distributions you support, but Gentoo is not on that list, so I thought I would open up an issue and ask before I attempt to package them. Is there anything specific that I can share with you about Gentoo so that it can be supported?

If you need a separate issue for syft, let me know and I'll open one there.

Thanks for your time.

William

williamh:
It looks like @zmedico is going to be able to assist with this.

wagoodman:
It looks like this work has not been completed: though matching capabilities have been implemented there the data is not being imported into the grype DB. The original PR references an DB import helper that was used for testing:

> Here is a sample program that I have used to import Gentoo GLSA metadata into a grype database:
> https://github.com/zmedico/grype-glsa-import/blob/main/main.go

Which is great! But we should not have closed this until there was a way for grype users to use this via the nightly DB publisher path. I'm going to reopen this with the added action needed is [adding a vunnel provider](https://github.com/anchore/vunnel/blob/main/DEVELOPING.md#adding-a-new-provider).

PR #2227, opened 2024-10-30
Remove gentoo integration test

wagoodman:
Based on https://github.com/anchore/grype/issues/744#issuecomment-2448163737 we still have more work to do before we can claim completion on gentoo support. This was revealed by a fix made in #2226 .


kzantow (approved):
👍

PR #2243, opened 2024-11-06
Add v6 vulnerability & blob stores

wagoodman:
Adds the `VulnerabilityHandle`, `Blob`, and `BlobDigest` models as well as the first blob, `VulnerabilityBlob` with supporting structs and store accessors (only in the Db v6 schema). 

When adding a new vulnerability to the DB the store objects handle serializing the object to JSON and storing to the DB within the `blobs` table (and the same is true for fetching in reverse). When the store is open for writing the `blob_digests` table is created and populated with each write to the `blobs` table; since this takes up a good amount of space in the DB and isn't needed for distribution, upon closing the store this table is dropped entirely.

Partially addresses #2128 #2132

### DIFF
--- a/grype/db/v6/blob_store.go
+++ b/grype/db/v6/blob_store.go
@@ -13,7 +13,7 @@ import (
 
 type blobable interface {
 	getBlobValue() any
-	setBlobID(int64)
+	setBlobID(ID)
 }
 
 type blobStore struct {
@@ -75,7 +75,7 @@ func (s *blobStore) addBlobs(blobs ...*Blob) error {
 	return nil
 }
 
-func (s *blobStore) getBlobValue(id int64) (string, error) {
+func (s *blobStore) getBlobValue(id ID) (string, error) {
 	var blob Blob
 	if err := s.db.First(&blob, id).Error; err != nil {
 		return "", err
--- a/grype/db/v6/blob_store.go
+++ b/grype/db/v6/blob_store.go
@@ -91,6 +91,9 @@ func (s *blobStore) Close() error {
 
 	log.WithFields("records", count).Trace("finalizing blobs")
 
+	// we use the blob_digests table when writing entries to ensure we have unique blobs, but for distribution this
+	// is no longer needed and saves on space considerably. For this reason, we drop the table after we are
+	// done writing blobs so that the DB is always in a distributable state.
 	if err := s.db.Exec("DROP TABLE blob_digests").Error; err != nil {
 		return fmt.Errorf("failed to drop blob digests: %w", err)
 	}
--- a/grype/db/v6/blob_store.go
+++ b/grype/db/v6/blob_store.go
@@ -1,0 +1,112 @@
+package v6
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strings"
+
+	"gorm.io/gorm"
+
+	"github.com/anchore/grype/internal/log"
+)
+
+type blobable interface {
+	getBlobValue() any
+	setBlobID(int64)
+}
+
+type blobStore struct {
+	db *gorm.DB
+}
+
+func newBlobStore(db *gorm.DB) *blobStore {
+	return &blobStore{
+		db: db,
+	}
+}
+
+func (s *blobStore) addBlobable(bs ...blobable) error {
+	for i := range bs {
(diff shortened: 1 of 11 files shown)


grype/db/v6/vulnerability_store.go
```diff
@@ -0,0 +1,126 @@
...
+
+	"github.com/anchore/grype/internal/log"
+)
+
+type VulnerabilityStoreWriter interface {
+	AddVulnerabilities(vulns ...*VulnerabilityHandle) error
+}
+
+type VulnerabilityStoreReader interface {
+	GetVulnerability(id int64, config *GetVulnerabilityOptions) (*VulnerabilityHandle, error)
```
kzantow, line 17: Should `id` be typedef'd to something, even if it's just `type ID int64`, in case this ever changes or we wanted to pass more context beyond just a db id reference?

kzantow (approved):

grype/db/v6/blobs.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// Severity represents a single string severity record for a vulnerability record
+type Severity struct {
+	// Scheme describes the quantitative method used to determine the Score, such as "CVSS_V3". Alternatively this makes
+	// claim that Value is qualitative, for example "HML" (High, Medium, Low), CHMLN (critical-high-medium-low-negligible)
+	Scheme SeverityScheme `json:"scheme"`
+
+	// Value is the severity score (e.g. "7.5", "CVSS:4.0/AV:N/AC:L/AT:N/PR:H/UI:N/VC:L/VI:L/VA:N/SC:N/SI:N/SA:N",  or "high" )
+	Value string `json:"value"`
```
westonsteimel, line 98: I think I one point we had this as `Score` and then the shape of the object could be determined by the `Scheme` because some scoring mechanisms might not be represented as a single string?  But maybe it was determined that was too complex later, I can't remember back that far now

westonsteimel: Maybe it was that if we want to represent something more complex than a string value here it would be a blob lookup to another table?

westonsteimel: Ah, yeah I think you had moved something like future EPSS over to its own table, so I guess it would be a similar approach for any new measure that has something more complex than a string as a representation

wagoodman: well, what I could do is make value an `any` type so that we could put in various shapes here. I can see this for score/vector for CVSS (even though we could calculate the score out from the vector... but why do that work again if we already have the value).

grype/db/v6/blob_store.go
```diff
@@ -0,0 +1,112 @@
...
+	db *gorm.DB
+}
+
+func newBlobStore(db *gorm.DB) *blobStore {
+	return &blobStore{
+		db: db,
+	}
+}
+
+func (s *blobStore) addBlobable(bs ...blobable) error {
```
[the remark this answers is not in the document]
wagoodman: no -- we really only want folks to add elements to the DB via the store, and blobs themselves are an internal concern for other stores and shouldn't be used externally to add other kinds of blobs

grype/db/v6/vulnerability_store.go
wagoodman: I'll type these to ID, which I think would make their use case a little more clear, but I don't think we should pass anything other than DB ids for simple getters like this (for instance there is a separate getter by name).

grype/db/v6/blob_store.go
```diff
@@ -0,0 +1,112 @@
...
+	var blob Blob
+	if err := s.db.First(&blob, id).Error; err != nil {
+		return "", err
+	}
+	return blob.Value, nil
+}
+
+func (s *blobStore) Close() error {
+	var count int64
+	if err := s.db.Model(&Blob{}).Count(&count).Error; err != nil {
```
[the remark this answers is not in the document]
wagoodman: I don't really want to overrotate on performance concerns so don't want to refactor this to avoid writing temp data (dropping tables is cheap, and writing the digests is also relatively cheap, and the unique constraint is useful on writes). This Close() method is called by the parent store Close() only when we have enabled writes to the DB, so this is a nop when used as a reader. I've added a comment to explain some of this.

PR #2265, opened 2024-11-18
Add data-driven approach to v6 distro search

wagoodman:
Today we have a few hard coded elements that affect which namespace is selected for given distro information:
- debian unstable: look for indications of [`sid` in the pretty-name](https://github.com/anchore/grype/blob/52e21b88c9ece087f25c5ec44d7f0fdb9fb173c8/grype/distro/distro.go#L60) then explicitly set the version to `unstable` which is [acted upon downstream](https://github.com/anchore/grype/blob/52e21b88c9ece087f25c5ec44d7f0fdb9fb173c8/grype/db/v5/namespace/index.go#L120-L124) to map to `sid` vulnerability data
- rolling versions: [manually list](https://github.com/anchore/grype/blob/52e21b88c9ece087f25c5ec44d7f0fdb9fb173c8/grype/distro/distro.go#L98) which distros via code that should match against a [single vulnerability stream](https://github.com/anchore/grype/blob/52e21b88c9ece087f25c5ec44d7f0fdb9fb173c8/grype/db/v5/namespace/index.go#L92-L95), regardless of the version.
- special versions: in alpine processing, [we map alpha versions to a specially named `edge`](https://github.com/anchore/grype/blob/52e21b88c9ece087f25c5ec44d7f0fdb9fb173c8/grype/db/v5/namespace/index.go#L109-L114) vulnerability stream manually
- `major` vs `major.minor` versions: we search [multiple combinations of version segment combinations](https://github.com/anchore/grype/blob/52e21b88c9ece087f25c5ec44d7f0fdb9fb173c8/grype/db/v5/namespace/index.go#L135-L236) to find the most accurate vulnerability stream

The upside of this logic is that it makes matching more accurate, the downside is that this kind of code requires a `grype` release to update (thus users need to update `grype` itself to get these updates). This is not ideal since a distro-related version change or vulnerability data change can happen at any time. 

Ideally this should be fundamentally data driven and flow in from the grype database, which is updated nightly -- this PR nudges us heavily in that direction by migrating this kind of logic to a DB lookup table, which is populated with grype-db, and paired with more agnostic processing of distro data (for example, we still have version specificity searching, but it is unaware of specific ).

Open questions:
- `debian@sid`: ~today we look for "sid" in the PRETTY_NAME for os-release, which is something that is fixed to the artifact, but as I understand it sid is a moving target... so wouldn't it be invalid to match non-trixie sid vuln data (foxy is the next codename) with trixie artifacts? Doesn't that mean we should exclusively look at the codename and ignore sid status altogether?~ We're going to need to appropriate pkg source lists into syft / grype / grype-db / vunnel. That is outside of the scope of this PR
- `alpine@edge`: ~it looks like this would be a single version with multiple pre-releases, but the vunnel data is [specifically hard coding this to "edge"](https://github.com/anchore/vunnel/blob/99acb6323863aa5aa33f2dbdcad022a3049eb8a7/src/vunnel/providers/alpine/parser.py#L165-L166), is this fictitious? (should we just use versions as we find them in the vuln data)~ this is correct, we should be using the alpha versions to correlate with edge, no matter the major/minor version
- ~should this logic be lifted from the DB into memory for faster continual access of OS specific questions? That is, much of this is querying the OS table for each distro search, but that table is relatively small and could be replaced with a single datastructure that is read into memory upon the first query.~ deferred; I don't want to optimize too much in advance though, so elected to not do this first since it could always be done at a later date since that could be implemented in a non-breaking way.

### DIFF
--- a/grype/db/v6/affected_package_store.go
+++ b/grype/db/v6/affected_package_store.go
@@ -2,7 +2,9 @@ package v6
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
+	"regexp"
 
 	"gorm.io/gorm"
 
@@ -11,6 +13,9 @@ import (
 
 var NoDistroSpecified = &DistroSpecifier{}
 var AnyDistroSpecified *DistroSpecifier
+var ErrMissingDistroIdentification = errors.New("missing distro name or codename")
+var ErrDistroNotPresent = errors.New("distro not present")
+var ErrMultipleOSMatches = errors.New("multiple OS matches found but not allowed")
 
 type GetAffectedPackageOptions struct {
 	PreloadOS      bool
@@ -20,11 +25,64 @@ type GetAffectedPackageOptions struct {
 	Distro         *DistroSpecifier
 }
 
+// DistroSpecifier is a struct that represents a distro in a way that can be used to query the affected package store.
 type DistroSpecifier struct {
-	Name         string
+	// Name of the distro as identified by the ID field in /etc/os-release
+	Name string
+
+	// MajorVersion is the first field in the VERSION_ID field in /etc/os-release (e.g. 7 in "7.0.1406")
 	MajorVersion string
+
+	// MinorVersion is the second field in the VERSION_ID field in /etc/os-release (e.g. 0 in "7.0.1406")
 	MinorVersion string
-	Codename     string
+
+	// LabelVersion is mutually exclusive to MajorVersion and MinorVersion and tends to represent the
+	// VERSION_ID when it is not a version number (e.g. "edge" or "unstable")
+	LabelVersion string
+
+	// Codename is the CODENAME field in /etc/os-release (e.g. "wheezy" for debian 7)
+	Codename string
+
+	// AllowMultiple specifies whether we intend to allow for multiple distro identities to be matched.
+	AllowMultiple bool
+}
+
+func (d DistroSpecifier) version() string {
+	if d.MajorVersion != "" && d.MinorVersion != "" {
+		return d.MajorVersion + "." + d.MinorVersion
+	}
+
+	if d.MajorVersion != "" {
+		return d.MajorVersion
+	}
+
+	if d.LabelVersion != "" {
+		return d.LabelVersion
+	}
+
+	if d.Codename != "" {
+		return d.Codename
+	}
+
+	return ""
+}
+
+func (d DistroSpecifier) matchesVersionPattern(pattern string) bool {
+	// check if version or version label matches the given regex
+	r, err := regexp.Compile(pattern)
+	if err != nil {
+		log.Tracef("failed to compile distro specifier regex pattern %q: %v", pattern, err)
+		return false
+	}
+
+	if r.MatchString(d.version()) {
+		return true
+	}
+
+	if d.LabelVersion != "" {
+		return r.MatchString(d.LabelVersion)
+	}
+	return false
 }
 
 type AffectedPackageStoreWriter interface {
@@ -90,7 +148,7 @@ func (s *affectedPackageStore) getNonDistroPackageByName(packageName string, con
 		query = query.Where("operating_system_id IS NULL")
 	}
 
-	query = s.handlePacakge(query, packageName, config)
+	query = s.handlePackage(query, packageName, config)
 	query = s.handlePreload(query, config)
 
 	err := query.Find(&pkgs).Error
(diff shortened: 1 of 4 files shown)


kzantow:
To offer opinions:

> debian@sid: today we look for "sid" in the PRETTY_NAME for os-release, which is something that is fixed to the artifact, but as I understand it sid is a moving target... so wouldn't it be invalid to match non-trixie sid vuln data (foxy is the next codename) with trixie artifacts? Doesn't that mean we should exclusively look at the codename and ignore sid status altogether?

I was under the impression that there is no vuln data for old unstable versions -- that the unstable feed is only for the current unstable branch, is this correct? Do we know which codename is currently in unstable at any given point? If so, I think considering something with an old codename + sid status as a prerelease for that codename would be the right thing to do, if we have enough info to do so.

> alpine@edge: it looks like this would be a single version with multiple pre-releases, but the vunnel data is [specifically hard coding this to "edge"](https://github.com/anchore/vunnel/blob/99acb6323863aa5aa33f2dbdcad022a3049eb8a7/src/vunnel/providers/alpine/parser.py#L165-L166), is this fictitious? (should we just use versions as we find them in the vuln data)

IIRC there was a difficulty mapping the unstable alpine feed without this behavior. Maybe just that the data for the distro isn't published in the normal feed until the release becomes stable? But if the versions align without doing some mapping to "edge", I would agree it sounds better to just use the version name directly.

> should this logic be lifted from the DB into memory for faster continual access of OS specific questions? That is, much of this is querying the OS table for each distro search, but that table is relatively small and could be replaced with a single datastructure that is read into memory upon the first query. I don't want to optimize too much in advance though, so elected to not do this first since it could always be done at a later date since that could be implemented in a non-breaking way.

Probably, yes. Agree it doesn't have to be done here -- only optimize things that need it. I worked on a configuration-heavy app that did something like this: always keep configuration loaded in memory and it worked great. If nothing else, it kept a lot of the programming model easier.

wagoodman:
It looks like there are a couple of answers here:
- debian sid will be dealt with later as we'll need to add another enhancement: searching by source list. This is out of scope for this PR and will be added at a later time
- I've confirmed that alpine edge processing is simply looking for `alpha` in the version

kzantow (approved):

grype/db/v6/affected_package_store.go
```diff
@@ -137,34 +214,171 @@ func (s *affectedPackageStore) getPackageByNameAndDistro(packageName string, con
...
+	}
+
+	if alias.ReplacementLabelVersion != nil {
+		d.LabelVersion = *alias.ReplacementLabelVersion
+	}
+
+	return nil
+}
+
+func (s *affectedPackageStore) handlePacakge(query *gorm.DB, packageName string, config GetAffectedPackageOptions) *gorm.DB {
```
westonsteimel, line 362:
```suggestion
func (s *affectedPackageStore) handlePackage(query *gorm.DB, packageName string, config GetAffectedPackageOptions) *gorm.DB {
```

grype/db/v6/affected_package_store.go
```diff
@@ -112,15 +170,34 @@ func (s *affectedPackageStore) getNonDistroPackageByName(packageName string, con
...
+	var err error
+	if config.Distro != NoDistroSpecified || config.Distro != AnyDistroSpecified {
+		resolvedDistros, err = s.resolveDistro(*config.Distro)
+		if err != nil {
+			return nil, fmt.Errorf("unable to resolve distro: %w", err)
+		}
+
+		if len(resolvedDistros) == 0 {
+			return nil, ErrDistroNotPresent
+		}
```
westonsteimel, line 183: Isn't this unnecessary now as it is covered by the switch statement that follows?

grype/db/v6/affected_package_store.go
```diff
@@ -137,34 +211,170 @@ func (s *affectedPackageStore) getPackageByNameAndDistro(packageName string, con
...
 	}
-	return query
-}
 
-func (s *affectedPackageStore) handleDistro(query *gorm.DB, d *DistroSpecifier) *gorm.DB {
-	if d == AnyDistroSpecified {
-		return query
+	// search for aliases for the given distro; we intentionally map some OSs to other OSs in terms of
+	// vulnerability (e.g. `centos` is an alias for `rhel`). If an alias is found always use that alias in
+	// searches (there will never be anything in the DB for aliased distros).
```
westonsteimel, line 221: So what happens here when we do add affected entries for a distro that was previous an alias (like when we bring in data for `almalinux`)?  I think what would happen is that the alias would no longer exist in the grype db that is published that now contains the almalinux data, but just want to validate that assumption.

westonsteimel: And all prior clients would adjust correctly to it becoming a comprehensive vulnerability dataset rather than an alias to `rhel` data

PR #2277, opened 2024-11-21
Simplify v6 distribution material

wagoodman:
This PR makes adjustments to the v6 schema + supporting code from previous PRs, specifically:
- adjusts DB import to function with both today's and v6 schema automatically
- removes the metadata.json and db.Description constructs -- now there is only the DB file in the distributed tarball
- adjusts the vulnerability blob such that "assigner" is plural (matches the type)
- add application config support to create v6 distribution client and installation curator config objects (not used yet)


grype/db/v6/distribution/latest.go
```diff
@@ -67,6 +73,43 @@ func NewLatestFromReader(reader io.Reader) (*LatestDocument, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to calculate description: %w", err)
+	}
+
+	if desc == nil {
+		return nil, fmt.Errorf("unable to describe the database")
+	}
+
+	// calculate the sh256sum of the archive
+	checksum, err := db.CalculateArchiveDigest(cfg.DBFilePath())
```
willmurphyscode, line 100: Should the parameter here be `path` not `cfg.DBFilePath()`? It seems weird t calculate the archive digest by doing anything besides hashing the tar.

grype/db/v6/distribution/latest.go
```diff
@@ -67,6 +73,43 @@ func NewLatestFromReader(reader io.Reader) (*LatestDocument, error) {
 	return &l, nil
 }
 
+func NewArchive(path string) (*Archive, error) {
```
willmurphyscode, line 76: Is this constructor still necessary if the suggestion at https://github.com/anchore/grype-db/pull/437/files#r1857293906 is implemented?

grype/db/v6/enumerations.go
```diff
@@ -57,6 +60,11 @@ const (
 	NotAffectedFixStatus FixStatus = "not-affected"
 )
 
+const (
+	// AdvisoryReferenceTag is a reference to a vulnerability advisory
+	AdvisoryReferenceTag string = "advisory"
```
willmurphyscode, line 65: Should there be more of these?

willmurphyscode: A reference tag is a string we expect to be used often to tag a reference in https://github.com/anchore/grype/blob/main/grype/db/v6/blobs.go#L51.

What's an Advisory exactly? For example, a Red Hat Security Advisory (RHSA) is a list of patches, but a GitHub Security Advisory is a vulnerability description. Is `advisory` the right tag name? What kind of links or going to be tagged with this tag name?

grype/db/v6/installation/curator.go
```diff
@@ -101,12 +102,23 @@ func (c curator) Status() db.Status {
 		}
 	}
 
+	var persistentErr error
+	digest, persistentErr := db.CalculateDBDigest(c.config.DBFilePath())
+
+	if validateErr := c.validate(); validateErr != nil {
```
willmurphyscode, line 108: How did the `digest` from line 106 get into `c.validate()`? Are we calculating the digest twice?

grype/db/v6/internal/db.go
```diff
@@ -14,7 +14,7 @@ func NewDB(dbFilePath string, models []any, truncate bool) (*gorm.DB, error) {
 		return nil, err
 	}
 
-	if len(models) > 0 {
+	if len(models) > 0 && truncate {
```
willmurphyscode, line 17: This line, and probably this method, could use a comment that we have only 2 modes of interacting with the DB which shouldn't mix:

1. Read mode for Grype at runtime should not migrate, vacuum, etc to avoid changing checksum or other corruption
2. Writing via grype-db should truncate the path to an empty db, apply all migration, migrate.

grype/db/v6/provider_store.go
```diff
@@ -64,3 +65,15 @@ func (s *providerStore) GetProvider(name string) (*Provider, error) {
 
 	return &provider, nil
 }
+
+func (s *providerStore) AllProviders() ([]Provider, error) {
+	log.Trace("fetching all provider records")
+
+	var providers []Provider
+	result := s.db.Find(&providers)
```
willmurphyscode, line 73: Does this need any sort of ordering?

cmd/grype/cli/commands/db_import.go
```diff
@@ -27,14 +31,43 @@ func DBImport(app clio.Application) *cobra.Command {
...
 	if err := dbCurator.ImportFrom(dbArchivePath); err != nil {
-		return fmt.Errorf("unable to import vulnerability database: %+v", err)
+		return fmt.Errorf("unable to import vulnerability database: %w", err)
 	}
 
 	return stderrPrintLnf("Vulnerability database imported")
 }
+
+func isLegacy(path string) bool {
+	return !strings.Contains(filepath.Base(path), "vulnerability-db_v6")
```
willmurphyscode, line 72: Since a rename might make this return true when it shouldn't, it might be worth inverting to `isNew` for now. It might be nice to have a config or a less brittle check or something at some point.

grype/db/v6/enumerations.go
wagoodman: I think I'll add these in another PR, but agree with the overall question

willmurphyscode:
FYI, I re-ran the acceptance tests because they failed after a small commit with `502: bad gateway`, so I assumed a transient network fault. We'll see if they pass on retry.

grype/db/v6/installation/curator.go
```diff
@@ -102,23 +103,15 @@ func (c curator) Status() db.Status {
...
-
-	if validateErr := c.validate(); validateErr != nil {
-		if persistentErr != nil {
-			persistentErr = multierror.Append(persistentErr, validateErr)
-		} else {
-			persistentErr = validateErr
-		}
-	}
+	// override the checksum validation setting to ensure the checksum is always validated
+	digest, validateErr := c.validate(true)
```
willmurphyscode, line 107: I think this makes it so `grype db status` always validates the checksum, so that it can report if the db has been edited since it was fetched. Is that right?

If so, minor nit, the comment should say something like "`grype db status` should always validate"; right now the comment doesn't tell me why we're not passing the config setting.

willmurphyscode (approved):

grype/db/v6/installation/curator.go
wagoodman: righto -- I'll incorporate into another followup PR (this was automerged)

PR #2285, opened 2024-11-25
Add ability to map CPEs directly to packages (v6 schema)

wagoodman:
This PR allows for direct associations of `Packages` to `CPEs` so that raising up more accurate `AffectedPackageHandles` over `AffectedCPEHandles` is possible. The data enrichment process we apply to the NVD data today will be leveraging this functionality (which lands when a new vunnel provider is written). This also makes the nomenclature correction of renaming `cpe.Type` to `cpe.Part`.

### DIFF
--- a/grype/db/v6/affected_package_store.go
+++ b/grype/db/v6/affected_package_store.go
@@ -9,6 +9,7 @@ import (
 	"gorm.io/gorm"
 
 	"github.com/anchore/grype/internal/log"
+	"github.com/anchore/syft/syft/cpe"
 )
 
 var NoDistroSpecified = &DistroSpecifier{}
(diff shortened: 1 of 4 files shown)


grype/db/v6/affected_package_store.go
```diff
@@ -169,10 +178,74 @@ func (s *affectedPackageStore) getNonDistroPackageByName(packageName string, con
...
+		query = query.Where("cpes.product = ?", c.Product)
+	}
+
+	if c.Version != cpe.Any {
+		query = query.Where("cpes.version = ?", c.Version)
+	}
+
+	if c.Update != cpe.Any {
+		query = query.Where("cpes.update = ?", c.Update)
+	}
```
westonsteimel, line 206: We don't keep version and update in the CPE table since those are always just part of the affected version constraints

wagoodman: good catch

wagoodman: updated!

wagoodman:
forced push to rebase on main

PR #2293, opened 2024-12-02
Fix DB v6 curator directory creation

wagoodman:
There are a few assumptions made by the v6 curator surrounding DB directory existence. This PR fixes activation-related issues with the DB not existing.


spiffcs (approved):
👍 - I saw that the `curator_test` was updated to cover some of the replace behavior. Are there any changes to `client_test`? I remember that file already had a lot of coverage so if changing the error string and adding the new MKDIR check are covered by the unhappy path tests already then 🟢

wagoodman:
yeah, I can update the client tests for this -- thanks @spiffcs

PR #2297, opened 2024-12-03
Parameterize v6 DB store

wagoodman:
Today there are multiple DB store functions for each table based off of possible input parameters (one function for by-cpe and another for by-name). This changes this to have a single getter for a noun. So for example, the AffectedPackageStoreReader has been reduced from `GetAffectedPackagesByName` and `GetAffectedPackagesByCPE` functions down to a single `GetAffectedPackages` where you can pass either name or CPE in with the new `PackageSpecifier`.

PR Stack:
1. https://github.com/anchore/grype/pull/2285


kzantow (approved):
👍

wagoodman:
forced pushed to rebase on main

Issue #2130, opened 2024-09-17
Stabilize DB `search` output


wagoodman:
With DB v6 there will be substantial changes made to the DB structure. We'd like to have a way for users to have a common interface with stable output (not directly tied to the DB schema/shapes) across grype versions.

The search command should be expanded to:
- Query a single CVE to get the NVD record (always one)
- Query a single vulnerability ID to get the vulnerability record(s)

_Are there more use cases we should capture?_

This output should be available in:
- table output
- json output

All json output should have a JSON schema describing valid/expected shapes/values.

TimBrown1611:
please notice to this kind of issue in grype db search - https://github.com/anchore/grype/issues/2085
(can be solved in the same task)

wagoodman:
https://github.com/anchore/grype/pull/2288/files#diff-dcf61b30ec74dd958726ebbb9a3ff0b0b271839989f03d8f0d233de89381a727R84-R85 We have the search command partially updated to use the new curator, but does not do anything with the store results until there is  something that inflates the `vulnerability.Vulnerability` struct from the DB models.

Issue #2129, opened 2024-09-17
Replace `grype db diff` with `grype db search --since DATE`


wagoodman:
As currently written the DB diff command will be cumbersome to keep as a feature while working on v6. We could leave DB diff in tact for v1-5 schemas, but this would be confusing for users attempting to use this diff feature for the latest (v6) DBs. 

V6 currently is going to have a blobs table where digests for content could be recomputed and used as a foundation for a diff, however that is out of scope for v6 unless there is a convincing argument / need to keep this feature and add to the v6 scope further.

TimBrown1611:
hi! @wagoodman diff is important since we want to track changes of the DB.
how can it be done on v6 schema? will v6 contain update date for each record?
thanks!

wagoodman:
As eluded to in https://github.com/anchore/grype/issues/2128#issuecomment-2375114119 there is a way to compute the records that differ. 

Is this an important use case to you? If so, can you tell us more about why so?

We dropped figuring this out in v6 to cut scope, but if we have enough users that really depend on this then that would be good reason to keep it in scope (in which case I'd close this issue and make a new one on how to implement the diff command).

TimBrown1611:
I use this command to handle states of last-modified CVEs.
I have to do it separately since the DB is stateless, and some sources don't provide modified date or create date.

willmurphyscode:
@TimBrown1611 when you say some sources don't provide modified or created date, do you mean in the upstream data? Like, are there specific data sources that grype db consumes that are missing this data? If so, which ones?

If grype had a date added and date updated date for most issues in each database, would you still need `grype db diff`?

willmurphyscode:
From discussion on our livestream on 10 October 2024, we had the idea:

Can we add `grype db search --since DATE` to print information about CVEs that were added or changed since `DATE`. Schema v6 would support that, and it seems like this is a simpler, more flexible way of comparing two databases. @TimBrown1611 would that meet your needs?

Things to consider:

1. What if data is deleted between db versions?
2. How to handle differences between what dates are available from different providers?

TimBrown1611:
Hi,
1. If a CVE was deleted its ok since it will not appear in the results.
2. Namespace field will be removed in v6?

wagoodman:
> 1. If a CVE was deleted its ok since it will not appear in the results.

That's correct -- this is one difference between how `db diff` works today and what `db search` can surface.

> 2. Namespace field will be removed in v6?

[...]

PR #2303, opened 2024-12-04
Enhance v6 search command

wagoodman:
Plumbs up the search command for the v6 schema with the store + presenter logic. Functionally this adds the ability to refine affected package searches by vuln published date, vuln modified date, vuln aliases, and vuln status. Note that date searching will be partially functional [until there is more data](https://github.com/anchore/vunnel/issues/742) in the DB for search conditions to key off of.

- Partially implements https://github.com/anchore/grype/issues/2130
- Partially implements https://github.com/anchore/grype/issues/2129

TODO:
- should not access store directly in command package; this should be a pattern finalized in #2132 and we refactor here
- capture a json schema for expected output moving forward
- add search by aliases
- add search by status


cmd/grype/cli/commands/db_search.go
```diff
@@ -164,3 +286,194 @@ func presentLegacy(outputFormat string, vulnerabilities []vulnerability.Vulnerab
...
+			case "1.0", "2.0":
+				family = "mariner"
+			default:
+				family = "azurelinux"
+			}
+		case "oracle":
+			family = "oraclelinux"
+		}
+
+		return fmt.Sprintf("%s:distro:%s:%s", row.Vulnerability.Provider, family, row.OS.Version)
```
willmurphyscode, line 476: I think this will return the wrong string for redhat, since the namespace has `redhat` but the provider is called `rhel`?

wagoodman: resolved with the latest commit

grype/db/v6/models.go
```diff
@@ -275,15 +282,53 @@ func (p *Package) BeforeCreate(tx *gorm.DB) (err error) {
...
+}
+
+func (os *OperatingSystem) VersionNumber() string {
+	if os.MinorVersion != "" {
+		return fmt.Sprintf("%s.%s", os.MajorVersion, os.MinorVersion)
+	}
+	return os.MajorVersion
+}
+
+func (os *OperatingSystem) Version() string {
```
willmurphyscode, line 315: Not a blocking comment, but I'm curious: What motivates this particular fallback logic?

willmurphyscode (approved):

grype/db/v6/models.go
wagoodman: Here's the reasoning that went into field precedence:
- label version tends to override numeric versions ("unstable" is the label for "debian:sid" for instance). There is a good argument that codename should be here, but the only reason why it's not is because vunnel and grype-db reason about a value here over raw values from the vulnerability provider.
- if not found, then we craft the most specific version possible and return that
- we don't expect there to be many cases where there is no version information but if so, codename is the only recourse left (which tends to be empty though).

Issue #1426, opened 2023-08-10
RFC: Explicit reporting of negative matches


willmurphyscode:
## **What would you like to be added?**

Grype should be able to report "negative matches," that is, package/vulnerability pairs where grype found evidence that the package is **not** affected by the vulnerability.

Right now, grype output consists of **positive matches**, that is, package/vulnerability pairs where the package is believed to be affected by the vulnerability. However, when matching is performed, that is, when grype asks whether a given package is vulnerable to a given vulnerability, there are 3 possible outcomes, and grype currently loses the distinction between the second and third:

1. The package is vulnerable -> this yields a match 
2. There is a search miss, for example, the grype database has no record for that package or its upstream -> this does not yield a match
3. There is a search hit, but grype concludes the package is _not_ vulnerable. For example grype's database has a version constraint that excludes the version of the package being considered -> **proposed:** this yields a "negative match"

In other words, this proposal should enable users to distinguish between two situations that are currently lumped together: a search miss, where grype didn't match because there wasn't data in the database indicating a match, and a negative match, where there is evidence in the database that the package is _not_ vulnerable.

In this proposal, grype can *optionally* display negative matches.

## **Why is this needed?**

Why would users of grype want to distinguish between these two cases? Here are a few that come to mind:

### More confidence in the results

Grype being able to say, "yes, package Foo is affected by CVE-123, but you have version 1.2.4 which is patched," may give users more confidence that they are not affected by CVE-123 than simply the absence of CVE-123 in the output.

### Ability to investigate true negatives

Right now, in the case of matches, grype can output match details, that is, information about what artifact was matched, how it was found, why it's believed to be vulnerable. In the case of a negative match, where for example grype matches the package, but then rejects the match based on a version constraint, all of this information is available, but it is dropped in the middle of functions, so that users and even library consumers of grype can't access it.

Additionally, [`grype explain`](https://github.com/anchore/grype/issues/1342) could be enhanced to explain negative matches. 

### Users can report on patching improvements directly

When users undertake the effort to patch dependencies in their project and upgrade their base images, rather than simply seeing grype's output shrink, they could see the set of CVEs reported in negative matches grow.

### Possibility of implementing namespace precedence

In many cases, a distro has more specific vulnerability information than, for example, GHSA or NVD. Being able to encode this precedence provides a good way to implement https://github.com/anchore/grype/issues/1373, and there are likely other business logic enhancements that could be made from surfacing negative matches.

## User experience

[...]

PR #2311, opened 2024-12-10
feat: v6 database support, updated matcher interfaces

### LINKED ISSUE #2132: Simplify grype DB access abstractions

kzantow:
This PR refactors the matching process in an effort to simplify the codebase and to support both V5 and V6 databases.

This change is based on a few philosophical ideals: the Grype `pkg.Package` should contain all the data used to search by, and the `vulnerability.Vulnerability` should contain all of the data used to match against. This stance allows the `vulnerability.Criteria` interface to be very simple -- users can [easily write go code](https://github.com/anchore/grype/pull/2311/files#diff-bc09d233ab846dde8997eab2ae4c29519cba805316287c37971a6a0831989720R26-R31) to determine if the particular vulnerability object matches whatever criteria set forth which is applied during the vulnerability matching process. By association, this means complex logic in the matching process can be easily [tested by providing a list of vulnerabilities](https://github.com/anchore/grype/pull/2311/files#diff-9972e46dc1fc4f4f0c785455fac6686ceca760674cf11de3755978a224c976e5R34) representing "everything in the db", and just running the existing matcher code!

At it's core, this PR introduces a simple vulnerability Criteria interface and refactors the vulnerability Provider to a simple interface accepting these criteria objects. A number of tests had v5-based vulnerability.Provider/Store/etc. implementations, which I've replaced by a simple `MockProvider`, which allows specifying vulnerabilities directly. There is a test that remains with a v5 store exercising the v5 implementation of the interface behaving as expected. And an analogous v6 test has been added. The most important interfaces are updated/introduced:

```
// match.Matcher
type Matcher interface {
	PackageType() syftPkg.Type

	Type() MatcherType

	// Match is called for every package found, returning any matches and an optional Ignorer which will be applied
	// after all matches are found
	Match(vp vulnerability.Provider, p pkg.Package) ([]Match, []IgnoredMatch, error)
}

// vulnerability.Criteria
type Criteria interface {
	// MatchesVulnerability returns true if the provided value meets the criteria
	MatchesVulnerability(value Vulnerability) (bool, error)
}

// vulnerability.Provider
type Provider interface {
	// FindVulnerabilities returns vulnerabilities matching all the provided criteria
	FindVulnerabilities(criteria ...Criteria) ([]Vulnerability, error)
}
```

Fixes #2132

Additionally, I think it fixes: #1426 despite not doing exactly the same thing that was discussed as needed; we should revisit this as a team to understand if the ask has been addressed the right way.


grype/pkg/package.go
```diff
@@ -33,16 +34,17 @@ type Package struct {
 	Name      string           // the package name
 	Version   string           // the version of the package
 	Locations file.LocationSet // the locations that lead to the discovery of this package (note: this is not necessarily the locations that make up this package)
-	Language  pkg.Language     // the language ecosystem this package belongs to (e.g. JavaScript, Python, etc)
+	Language  syftPkg.Language // the language ecosystem this package belongs to (e.g. JavaScript, Python, etc)
+	Distro    *distro.Distro   // a specific distro this package originated from
```
wagoodman, line 38: Matcher was originally

```
Match(v5.VulnerabilityProvider, *distro.Distro, pkg.Package) ([]match.Match, error)
```

which means that it does not necessarily need to be added to the pacakge. When we start matching with per-package distro information keeping these separate means that we can easily separate what was dected from the environment vs what was parsed from the package info (like the pURL). Here they are mixed together.

I don't see a reason why these two changes needed to be made in this PR -- that is I think v6 can be added without needing to break the matcher interface nor add a new type to the core package type.

wagoodman (commented):
just to clarify: is the goal to keep the top-level `search` package? Not a problem if so, just trying to square up the final state of things based on previous conversation.

grype/pkg/package.go
kzantow: The matcher interface needs to change: the existing interface references the V5-specific VulnerabilityProvider, which at a minimum needs to be changed to the `vulnerability.Provider` (or whatever the V6 compatible thing is).

Conversely to this, if I have a PURL with something like: `pkg:deb/debian/adduser@3.134?arch=all&distro=debian-12`, I feel as though _this_ distro information should be associated with the package. Should there be 2 separate places to find distro? I'm not opposed to having some environmental information passed through separately but I couldn't understand why the distro was the only thing. What if we wanted to pass the image name or architecture? By moving this to the package, it seemed to me to be the right delineation that something somewhere determined we should be matching this package with this distro, and every matcher doesn't have to make that determination.

If this is a blocker, we should probably powwow to figure out the path forward.

kzantow: One of the things I was especially interested in doing was making one (or more) top-level structs here that could be updated post-v1. In this case, maybe the `pkg.Context` seems like the right thing to pass in `MatchVulnerabilities(<maybe-other-stuff>, pkg.Context, pkg.Package)`. But as you say doesn't need to be changed as part of this PR.

kzantow:
> is the goal to keep the top-level search package? Not a problem if so, just trying to square up the final state of things based on previous conversation.

Yes, but this package does not have the same purpose as the prior implementation. This is for `search.By<something>` to read nicely and have a package that can be referenced by both db implementations and vulnerability finding logic. I don't really have strong opinions about the naming of this package aside from it reading well with `<name>.By<thing>`.

grype/pkg/package.go
wagoodman: yeah I agree -- lets leave this as is in this PR and in grype 1.0 we can reevaluate (as you pointed out pkg.Context might be a better thing to pass)

grype/load_vulnerability_db.go
```diff
@@ -1,13 +1,55 @@
...
+	v6 "github.com/anchore/grype/grype/db/v6"
+	v6dist "github.com/anchore/grype/grype/db/v6/distribution"
+	v6inst "github.com/anchore/grype/grype/db/v6/installation"
+	"github.com/anchore/grype/grype/vulnerability"
 	"github.com/anchore/grype/internal/log"
 )
 
-func LoadVulnerabilityDB(cfg distribution.Config, update bool) (*v5.ProviderStore, *distribution.Status, error) {
-	dbCurator, err := distribution.NewCurator(cfg)
+func LoadVulnerabilityDBv6(distCfg v6dist.Config, installCfg v6inst.Config, update bool) (vulnerability.Provider, *v5dist.Status, error) {
```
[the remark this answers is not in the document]
kzantow: After a look, it turns out there would end up being a lot of circular references to fix, so from an offline conversation, we've opted not to make a change to the db loading as part of this PR.

grype/db/v6/vulnerability_provider.go
```diff
@@ -0,0 +1,452 @@
...
+		Cvss:        toCvss(vuln.BlobValue.Severities...),
+	}, nil
+}
+
+func (s vulnerabilityProvider) Close() error {
+	return s.reader.(io.Closer).Close()
+}
+
+//nolint:funlen,gocognit,gocyclo
+func (s vulnerabilityProvider) FindVulnerabilities(criteria ...vulnerability.Criteria) ([]vulnerability.Vulnerability, error) {
```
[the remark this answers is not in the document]
kzantow: From discussion: I'm going to add an error condition that prohibits duplicate criteria of known types.

grype/db/v6/db.go
```diff
@@ -65,7 +66,8 @@ func (c Config) DBFilePath() string {
 }
 
 func NewReader(cfg Config) (Reader, error) {
-	return newStore(cfg, false, false)
+	buildIndexes := os.Getenv("GRYPE_DB_BUILD_INDEXES") == "true"
```
wagoodman, line 69: is this for development?

grype/db/v6/store.go
```diff
@@ -18,7 +18,12 @@ type store struct {
 	blobStore *blobStore
 	db        *gorm.DB
 	config    Config
-	readOnly  bool
+	empty     bool
+	writable  bool
+}
+
+func (s *store) GetDB() *gorm.DB {
```
wagoodman, line 25: we really shouldn't leak the raw DB past the store outside of this package

grype/db/v6/fillers.go
```diff
@@ -0,0 +1,160 @@
...
+
+		var t BlobType
+		err = json.Unmarshal([]byte(blob.Value), &t)
+		if err != nil {
+			return err
+		}
+		*ref.ref = &t
+	}
+	return nil
+}
```
wagoodman, line 88: The blobstore already has `attachBlobValue` which does this functionality

grype/db/v6/store.go
kzantow: I need to get the raw DB for `fill` calls -- I can't really use the preload.

grype/db/v6/fillers.go
kzantow: hmm. I've probably missed something, but this is not exported -- how am I supposed to do this while narrowing the initial results?

grype/db/v6/db.go
kzantow: Yep, removed this

grype/db/v6/store.go
kzantow: The problem is: without having a way to "fill-in" specific fields in a set of objects I'm left doing one of two things:
1) preload everything
2) re-fetch every object with preload everything after constraint checks

E.g. the v6 provider's `FindVulnerabilties` call looks like this:
```
			affectedPackages, err = s.reader.GetAffectedPackages(pkgSpec, &GetAffectedPackageOptions{
				OSs:                  osSpecs,
				Vulnerabilities:      vulnSpecs,
				PreloadVulnerability: true,
				PreloadBlob:          true,
				PreloadOS:            true,
				PreloadPackage:       true,
				PreloadPackageCPEs:   true,
			})
```

Neither of these are efficient: I already have the base set of objects, but I don't want to fetch _everything_ initially. This is "typical" ORM stuff -- defer loading relations until one gets asked for but I don't think gorm supports this, or at least it isn't enabled, so we're stuck with option 1, 2 or a third, which is provide these "fill" functions somehow.

wagoodman: I agree with the functionality and filling -- that wasn't what I was commenting on. I really mean that this doesn't need to be exported was all.

grype/db/v6/refs.go
```diff
@@ -0,0 +1,98 @@
...
+// func mapResults[T any](db *gorm.DB, ids []v6.ID, refID idProvider[T]) (map[v6.ID]T, error) {
+//	var results []T
+//	// FIXME probably need to build IN clause
+//	err := db.Where("ID IN (?)", ids).Find(&results).Error
+//	out := map[v6.ID]T{}
+//	for _, result := range results {
+//		out[refID(result)] = result
+//	}
+//	return out, err
+//}
```
wagoodman, line 98: dead code?

grype/db/v6/store.go
wagoodman: I pushed a small change to address this

grype/db/v5/vulnerability_provider.go
```diff
@@ -37,6 +38,87 @@ func NewVulnerabilityProvider(reader VulnerabilityStoreReader) (VulnerabilityPro
...
+		vulns, err = pr.getByCPE(*searchCrit.cpe)
+	default:
+		return nil, fmt.Errorf("unable to find vulnerabilities without distro, language, or CPE: %+v", criteria)
+	}
+
+	// apply criteria
+	for i := 0; i < len(vulns); i++ {
+		v := vulns[i]
+
+		log.Trace(v)
```
wagoodman, line 93: debug code

grype/db/v6/cvss.go
```diff
@@ -0,0 +1,37 @@
...
+			// not a CVSS score
+			continue
+		}
+		out = append(out, vulnerability.Cvss{
+			Source:  sev.Source,
+			Type:    string(sev.Scheme),
+			Version: score.Version,
+			Vector:  score.Vector,
+			Metrics: vulnerability.CvssMetrics{
+				// FIXME: where do these metrics come from?
```
wagoodman, line 28: these have not been used in the past and we should clean this up around grype 1.0

grype/db/v6/vulnerability.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+	if len(vulns) == 0 {
+		return nil, fmt.Errorf("no vulnerabilities found for package %v", id)
+	}
+	return &vulns[0], nil
+}
+
+func newVulnerabilityFromParts(packageName string, vuln *VulnerabilityHandle, affected *AffectedPackageBlob, affectedRange AffectedRange, affectedPackageHandle *AffectedPackageHandle, affectedCpeHandle *AffectedCPEHandle) (*vulnerability.Vulnerability, error) {
+	if vuln.BlobValue == nil {
+		return nil, fmt.Errorf("vuln has no blob value: %+v", vuln) // FIXME we could just query the db
```
wagoodman, line 61:
```suggestion
		return nil, fmt.Errorf("vuln has no blob value: %+v", vuln)
```
within the constructor there really shouldn't be any DB access

grype/db/v6/vulnerability_provider.go
```diff
@@ -0,0 +1,415 @@
...
+	if score > .4 {
+		return vulnerability.MediumSeverity
+	}
+	if score > .2 {
+		return vulnerability.LowSeverity
+	}
+	if score >= .0 {
+		return vulnerability.NegligibleSeverity
+	}
+	return vulnerability.UnknownSeverity
```
wagoodman, line 406: I don't recall this score threshold system before, is this new?

grype/matcher/internal/matcher.go
```diff
@@ -0,0 +1,14 @@
...
+	"github.com/anchore/grype/grype/pkg"
+	"github.com/anchore/grype/grype/vulnerability"
+	syftPkg "github.com/anchore/syft/syft/pkg"
+)
+
+type TypedMatcher interface {
+	PackageTypes() []syftPkg.Type
+	Type() match.MatcherType
+	Match(vulnerability.Provider, pkg.Package) ([]match.Match, []match.IgnoredMatch, error)
+}
```
wagoodman, line 14: dead code?

grype/search/criteria.go
```diff
@@ -0,0 +1,114 @@
...
+					continue
+				}
+				if iType == reflect.TypeOf(row[j]) {
+					return fmt.Errorf("multiple conflicting criteria specified: %+v %+v", row[i], row[j])
+				}
+			}
+		}
+	}
+	return nil
+}
```
wagoodman, line 76: could improve this some with fewer nested loops
```
func ValidateCriteria(criteria []vulnerability.Criteria) error {
	for _, row := range CriteriaIterator(criteria) {
		seenTypes := make(map[reflect.Type]interface{})
		
		for _, criterion := range row {
			criterionType := reflect.TypeOf(criterion)
			
			if slices.Contains(allowedMultipleCriteria, criterionType) {
				continue
			}
			
			if previous, exists := seenTypes[criterionType]; exists {
				return fmt.Errorf("multiple conflicting criteria specified: %+v %+v", previous, criterion)
			}
			
			seenTypes[criterionType] = criterion
		}
	}
	return nil
}
```
(haven't checked this)

grype/vulnerability/log.go
```diff
@@ -0,0 +1,13 @@
+package vulnerability
+
+import "github.com/anchore/grype/internal/log"
+
+// LogDropped is a hook called when vulnerabilities are dropped from consideration in a vulnerability Provider,
+// this offers a convenient location to set a breakpoint
+func LogDropped(vulnerabilityID string, reason any, context ...any) {
```
wagoodman, line 7: can this be an internal concern to v6 instead of a public API helper function in the core vulnerability package?

grype/vulnerability_matcher.go
```diff
@@ -381,6 +303,43 @@ func (m *VulnerabilityMatcher) normalizeByCVE(match match.Match) match.Match {
...
+	return nil
+}
+
+// ignoreMatchFilter creates an ignore filter based on the provided IgnoredMatches to filter out "the same"
+// vulnerabilities reported by other matchers based on overlapping file locations
+func ignoredMatchFilter(ignores []match.IgnoredMatch) match.IgnoreFilter {
+	out := ignoreRulesByLocation{locationToIgnoreRules: map[string][]match.IgnoreRule{}}
+	for _, ignore := range ignores {
+		// TODO should this be syftPkg.FileOwner interface or similar?
+		if m, ok := ignore.Package.Metadata.(pkg.ApkMetadata); ok {
```
wagoodman, line 334: this seems to be part of the migrated apk matcher code, should this have also been moved? or is this intended to be more generic?

wagoodman: from an offline explanation this is left over from the migration but is still needed since the ignore rules need to be applied to the ignored matches

grype/search/version_constraint.go
```diff
@@ -0,0 +1,87 @@
...
+
+// ByConstraintFunc returns criteria which will use the provided function as inclusion criteria
+func ByConstraintFunc(constraintFunc func(constraint version.Constraint) (bool, error)) vulnerability.Criteria {
+	return &constraintFuncCriteria{fn: constraintFunc}
+}
+
+func ByExactConstraint(constraint string) vulnerability.Criteria {
+	return ByConstraintFunc(func(c version.Constraint) (bool, error) {
+		// TODO improve this; constraint could/should have a Constraint() string / Raw() string method
+		return strings.TrimSpace(strings.Split(c.String(), "(")[0]) == constraint, nil
```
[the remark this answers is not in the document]
wagoodman: nvm -- it would be good to migrate this to the apk package and unexport it

grype/db/v6/cvss.go
wagoodman: actually, there might be a change, I'll get back to you

grype/db/v6/vulnerability_provider.go
wagoodman: I'll provide some code in a follow up PR to handle this

wagoodman (approved):
😎 really nice work

PR #2313, opened 2024-12-10
chore: replace archiver with anchore fork

spiffcs:
## Summary
See related [Syft PR](https://github.com/anchore/syft/pull/3516)

This PR Addresses a security fix for syft/grype.

Syft and grype were never vulnerable to https://github.com/advisories/GHSA-rhh4-rh7c-7r5v given the steps taken to fork, fix, and use a replace directive for the upstream repository. This PR moves syft to officially use the fork without a replace directive. This will help reduce false positives in upstream security tools (including grype).

Given the the upstream repository has been archived without a fix we've moved to using our fork officially without the replace directive until another solution can be found.


kzantow (approved):
👍

PR #2295, opened 2024-12-03
ignore linux-aws-headers-.* as well like linux-headers-

barnuri:

PR #2320, opened 2024-12-12
fix upstream match for linux-.*-headers-.*

barnuri:
a fix for this PR, the upstream of linux-.*-headers-.* is linux-aws, linux-azure etc
https://github.com/anchore/grype/pull/2295


kzantow (commented):
We should figure out how to add some tests for this, maybe something in the [test/cli package](https://github.com/anchore/grype/tree/main/test/cli)?

barnuri:
> We should figure out how to add some tests for this, maybe something in the [test/cli package](https://github.com/anchore/grype/tree/main/test/cli)?

i think its very small change and no so familiar with your tests
if you really think its necessary please give me details how to do so 😄

barnuri:
@kzantow ?

tomersein:
hi! any updates? I've tried to look and see if it is easy to add tests here but I didn't find any simple way (and its looks like a small change which can reduce lots of FP)

kzantow:
I still think we should add a test -- a small change like this could easily regress without one. How about updating [these "ignore" tests](https://github.com/anchore/grype/blob/67ac96f986e25cd4cf7a524f9a9209dba4eb1a97/grype/match/ignore_test.go#L174)?

tomersein:
I've noticed the UT is not protecting as expected the code, since the ignore rules are inside the root:
```
var ignoreLinuxKernelHeaders = []match.IgnoreRule{
	{Package: match.IgnoreRulePackage{Name: "kernel-headers", UpstreamName: "kernel", Type: string(syftPkg.RpmPkg)}, MatchType: match.ExactIndirectMatch},
	{Package: match.IgnoreRulePackage{Name: "linux(-.*)?-headers-.*", UpstreamName: "linux.*", Type: string(syftPkg.DebPkg)}, MatchType: match.ExactIndirectMatch},
	{Package: match.IgnoreRulePackage{Name: "linux-libc-dev", UpstreamName: "linux", Type: string(syftPkg.DebPkg)}, MatchType: match.ExactIndirectMatch},
}
```

but they are manually built in the ignore_test, so in case of a change it doesn't detect it.

I suggest moving it or refactor the tests.. not sure how.
@kzantow

barnuri:
@kzantow hi there is any update of this review ?

grype/match/ignore.go
```diff
@@ -215,7 +215,11 @@ func ifPackageLocationApplies(location string) ignoreCondition {
 func ifUpstreamPackageNameApplies(name string) ignoreCondition {
 	return func(match Match) bool {
 		for _, upstream := range match.Package.Upstreams {
-			if name == upstream.Name {
+			pattern, err := packageNameRegex(name)
+			if err != nil {
+				continue
+			}
```
kzantow, line 221: I think this should be moved outside of the returned func, the same way `ifPackageNameApplies` checks the regex once and returns an "always false" func e.g.:

```
func ifUpstreamPackageNameApplies(name string) ignoreCondition {
	pattern, err := packageNameRegex(name)
	if err != nil {
		return func(Match) bool { return false }
	}
	return func(match Match) bool {
		for _, upstream := range match.Package.Upstreams {
			if pattern.MatchString(upstream.Name) {
				return true
			}
		}
		return false
	}
}

kzantow (commented):
This looks great -- one small tweak requested

tomersein:
hi @kzantow , pushed a fix :)

kzantow (approved):
Thanks for the contribution, @barnuri and @tomersein !

PR #2335, opened 2024-12-16
Drop DB v6 indexes on close

wagoodman:
This PR makes the following adjustments to the v6 store:

When opening a DB for reading or writing:
- use 16KB page size (tended to reduce the overall size of the DB for distribution)

When opening an EMPTY db for writing:
- cache size increased to 1GB
- mmap limit increased to 1GB

When closing a newly-populated DB:
- drop all indexes. This helps to keep the distribution size way down. Testing shows that re-building these indexes is quite cheap (~2 seconds, milage may vary)

When importing a DB:
- automigrate is ran, re-creating any indexes that were dropped before distribution (`v6.Hydrater`)

Other changes made in this PR:
- moves `internal.NewDB` to `v6.NewLowLevelDB` so that it has direct access to initial data and models to be migrated (so getting a `gorm.DB` instance is consistent)
- allowed TUI in import DB


cmd/grype/cli/commands/db_import.go
```diff
@@ -11,17 +11,17 @@ import (
...
 
 func DBImport(app clio.Application) *cobra.Command {
 	opts := dbOptionsDefault(app.ID())
 
 	return app.SetupCommand(&cobra.Command{
-		Use:     "import FILE",
-		Short:   "import a vulnerability database archive",
-		Long:    fmt.Sprintf("import a vulnerability database archive from a local FILE.\nDB archives can be obtained from %q.", internal.DBUpdateURL),
-		Args:    cobra.ExactArgs(1),
-		PreRunE: disableUI(app),
```
wagoodman, line 24: import does respond to TUI elements, so this was dropped

grype/db/internal/gormadapter/open.go
```diff
@@ -4,37 +4,81 @@ import (
...
+	}
+}
+
+func WithStatements(statements ...string) Option {
+	return func(c *config) {
+		c.statements = append(c.statements, statements...)
+	}
+}
+
+func WithMigrate(models []any) Option {
```
kzantow, line 67: nit: I think "migrate" can be misconstrued that these models contain migration functionality, maybe this could be named `WithModels`?

grype/db/internal/gormadapter/open.go
```diff
@@ -4,37 +4,81 @@ import (
...
+	// performance improvements (note: will result in lost data on write interruptions)
+	`PRAGMA synchronous = OFF`,     // minimize the amount of syncing to disk, prioritizing write performance over durability
+	`PRAGMA journal_mode = MEMORY`, // do not write the journal to disk (maximizing write performance); OFF is faster but less safe in terms of DB consistency
 }
 
-var readOptions = []string{
-	"immutable=1",
-	"cache=shared",
-	"mode=ro",
+var heavyWriteStatements = []string{
```
kzantow, line 25: nit: this doesn't seem to have options directly related to write performance but rather memory, with an option named `allowLargeMemoryFootprint` should this be `largeMemoryFootprintStatements`?

grype/db/internal/gormadapter/open.go
```diff
@@ -88,22 +134,64 @@ func Open(path string, options ...Option) (*gorm.DB, error) {
...
-				return nil, fmt.Errorf("unable to execute (%s): %w", sqlStmt, dbObj.Error)
-			}
+	if cfg.writable {
+		log.Trace("applying writable DB statements")
+		if err := applyStatements(dbObj, writerStatements); err != nil {
+			return nil, fmt.Errorf("unable to apply DB writer statements: %w", err)
+		}
+	}
+
+	if cfg.truncate && cfg.allowLargeMemoryFootprint {
```
kzantow, line 144: Is the memory footprint related to truncating?

grype/db/v4/store/store.go
```diff
@@ -19,31 +19,22 @@ type store struct {
 	db *gorm.DB
 }
 
+func models() []any {
+	return []any{
+		model.IDModel{},
```
kzantow, line 24: should these be pointers? (the prior AutoMigrate calls appear to be passed pointers) (And same comment for other similar changes.)

cmd/grype/cli/commands/db_import.go
```diff
@@ -48,10 +48,12 @@ func newDBImport(opts options.Database, dbArchivePath string) error {
 		return fmt.Errorf("unable to create curator: %w", err)
 	}
 
+	log.WithFields("path", dbArchivePath).Infof("Importing vulnerability database archive")
 	if err := c.Import(dbArchivePath); err != nil {
 		return fmt.Errorf("unable to import vulnerability database: %w", err)
 	}
-	return stderrPrintLnf("Vulnerability database imported")
+	log.Info("vulnerability database imported")
```
kzantow, line 55: might be nice to include a publish date or something on this line

grype/db/v6/installation/curator_test.go
```diff
@@ -220,6 +226,10 @@ func TestCurator_Update(t *testing.T) {
 	t.Run("error during activation: cannot move dir", func(t *testing.T) {
 		c := setupCuratorForUpdate(t, withWorkingUpdateIntegrations())
 		mc := c.client.(*mockClient)
+		// nop hydrator
+		c.hydrator = func(string) error {
```
kzantow, line 230: hydrator could be optional, checked for nil where it's used instead of requiring empty functions like this

grype/db/v4/store/store.go
wagoodman: These are intended to be passed to automigrate, which only needs instances, not pointers. If being passed to `db.Create` where the instance represents a row in the DB, then a pointer is required (which is not intended with `Models()`.

`InitialData()` does return pointers since they represent rows to be added.

grype/db/v6/installation/curator_test.go
wagoodman: I can change it to nil check

kzantow (approved):
(None of the prior review was blocking; didn't approve then since it was auto-merge)

grype/db/internal/gormadapter/open.go
wagoodman: it is -- when we start with an empty DB we will be doing a lot of writing (and allowing for a larger memory footprint in those cases allows for writing to the DB much faster).

grype/db/internal/gormadapter/open.go
wagoodman: https://github.com/anchore/grype/pull/2335#discussion_r1887662856 so when there are heavy writes expected we want to allow for a larger memory footprint (to improve write performance).

PR #2347, opened 2024-12-20
Clean up config help text

wagoodman:
Today we use the options structs from each command registered with clio to build the output of the `grype config` command. However, not all commands are meant to support options being configured via a config file -- but these are being shown in the `grype config` output anyway:
```
# og-config is `grype config` with the latest grype release
# new-config is `grype config` with grype built from this branch
diff /tmp/og-config /tmp/new-config
```
```diff
228,236d227
< 
< # delete downloaded databases after diff occurs (env: GRYPE_DELETE)
< delete: false
< 
< # format to display results (available=[table, json]) (env: GRYPE_OUTPUT)
< Output: 'table'
< 
< # CVE IDs to explain (env: GRYPE_CVE_IDS)
< cve-ids: []

```
Note that `Output` is for `grype db *` commands and there is still an `output` value documented in the `grype config` output.

This PR removes all subcommands from consideration when generating `grype config` output.

### DIFF
--- a/cmd/grype/cli/commands/db_check.go
+++ b/cmd/grype/cli/commands/db_check.go
@@ -52,10 +52,11 @@ func DBCheck(app clio.Application) *cobra.Command {
 
 	// prevent from being shown in the grype config
 	type configWrapper struct {
-		Opts *dbCheckOptions `json:"-" yaml:"-" mapstructure:"-"`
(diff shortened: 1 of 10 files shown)


kzantow (approved):

wagoodman:
looks like there is something I didn't take into consideration -- I'll take a closer look later today

kzantow (approved):
👍

PR #2349, opened 2024-12-23
Finalize label version and add release id to OS model

wagoodman:
This PR changes the following aspects with regard to the v6 operating system table:

- Searching for an OS combines the codename/label-version (otherwise it would not be clear that these separate fields are OR operations, not AND like everything else)
- Searching for an OS can be done by name or release ID (from /etc/os-release), again as an OR operation
- Operating system table now has the original release id (`ol` for `oraclelinux`)


grype/db/v6/affected_package_store.go
```diff
@@ -100,13 +100,9 @@ type OSSpecifier struct {
 	// MinorVersion is the second field in the VERSION_ID field in /etc/os-release (e.g. 0 in "7.0.1406")
 	MinorVersion string
 
-	// LabelVersion is mutually exclusive to MajorVersion and MinorVersion and tends to represent the
-	// VERSION_ID when it is not a version number (e.g. "edge" or "unstable")
+	// LabelVersion is a string that represents a floating version (e.g. "edge" or "unstable") or is the CODENAME field in /etc/os-release (e.g. "wheezy" for debian 7)
```
spiffcs, line 103: Oh yea this comment is way clearer. This also makes a lot more sense after our Thursday stream where we talked about this table.

grype/db/v6/affected_package_store.go
```diff
@@ -411,15 +403,11 @@ func (s *affectedPackageStore) resolveDistro(d OSSpecifier) ([]OperatingSystem,
 	query := s.db.Model(&OperatingSystem{})
 
 	if d.Name != "" {
-		query = query.Where("name = ?", d.Name)
-	}
-
-	if d.Codename != "" {
-		query = query.Where("codename = ?", d.Codename)
+		query = query.Where("name = ? OR release_id = ?", d.Name, d.Name)
```
spiffcs, line 406: Is `release_id` always going to be queried by `d.Name`? Is there every a chance `release_id` could be the value that's in `d.LabelVersion`?

I guess codename and names never really overlap in the distro space - I just wasn't sure if this case existed at all.

spiffcs (approved):

grype/db/v6/affected_package_store.go
wagoodman: label version represents either a codename or something like "unstable" (a non numeric version). The release ID doesn't stand in for the version, this states more of a distribution family and shouldn't change from release to release like the version will.

PR #2351, opened 2024-12-23
Have aliases for non standard names in DB v6 schema

wagoodman:
It is common to users to input "oracle" or "oraclelinux" instead of "ol", which is the release ID as found in /etc/os-release. This makes it easier to find the correct OS given a fuzzy name.


kzantow (approved):
👍

PR #2376, opened 2025-01-15
Add package spec alias + case insensitivity for v6 DBs

wagoodman:
This adds the capability to search for package specifier ecosystem aliases, such as translating purl types to known syft package types, to ensure we're searching for the best ecosystem values relative to what's in the DB.

This PR also:
- renames the package.type field to package.ecosystem
- adds case insensitivity to commonly searched fields.

### DIFF
--- a/grype/db/v6/affected_package_store.go
+++ b/grype/db/v6/affected_package_store.go
@@ -297,13 +297,17 @@ func (s *affectedPackageStore) handlePackage(query *gorm.DB, config *PackageSpec
 		return query
 	}
 
+	if err := s.applyPackageAlias(config); err != nil {
+		log.Errorf("failed to apply package alias: %v", err)
+	}
+
 	query = query.Joins("JOIN packages ON affected_package_handles.package_id = packages.id")
 
 	if config.Name != "" {
-		query = query.Where("packages.name = ?", config.Name)
+		query = query.Where("packages.name = ? collate nocase", config.Name)
 	}
 	if config.Ecosystem != "" {
-		query = query.Where("packages.type = ?", config.Ecosystem)
+		query = query.Where("packages.ecosystem = ? collate nocase", config.Ecosystem)
 	}
 
 	if config.CPE != nil {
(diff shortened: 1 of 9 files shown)


grype/db/v6/affected_package_store.go
```diff
@@ -529,6 +516,60 @@ func (s *affectedPackageStore) applyAlias(d *OSSpecifier) error {
 	return nil
 }
 
+func (s *affectedPackageStore) searchForDistroVersionVariants(query *gorm.DB, d OSSpecifier) ([]OperatingSystem, error) {
```
wagoodman, line 519: note: this function only migrated to make it more readable (in order of calls from parent functions).

grype/db/v6/affected_package_store.go
```diff
@@ -314,6 +318,43 @@ func (s *affectedPackageStore) handlePackage(query *gorm.DB, config *PackageSpec
...
+		alias = &a
+		break
+	}
+
+	if alias == nil {
+		return nil
+	}
+
+	if alias.ReplacementEcosystem != nil {
+		d.Ecosystem = *alias.ReplacementEcosystem
```
kzantow, line 352: What if there were multiple replacements?

Since we only talked about this offline, I'll just suggest a couple alternate things that, at least to me, could help with transparency when a user is examining what logic was used during a vulnerability match. These are all in the context of [the v6 refactoring](https://github.com/anchore/grype/pull/2311).

* Matcher code could include a call to resolve aliases and maybe make it conditional on something; e.g.:
```
if ecosystemAliaser, ok := vulnerabilityProvider.(EcosystemAliaser); ok {
  for _, ecosystem := ecosystemAliaser.EcosystemAliases(pkg.Type) {
    ...
  }
}
```

* Vulnerability responses could be modified to include data expected to match, e.g.:
```
type Vulnerability struct {
   ... 
  Ecosystems []string
}
```
That said, I'm also not fundamentally opposed to something happening in the provider to provide improved identification like this. I think we'd just want to add some sort criteria instead of `search.ByLanguage()`, it could be `search.ByEcosystemAlias()` or something, and it does not check the actual vulnerability response, only gets interpreted and used by the V6 code. Or... the V6 matcher looks for the language and just removes it from the go-side checks after doing the alias search. Either way, that's not immediately applicable to this PR, anyway. Just a concern when these merge.

kzantow (approved):
LGTM 🎉

grype/db/v6/affected_package_store.go
wagoodman: I'll get this in for now -- but before v6 is finalized let's confer and see if we want to refactor.

PR #2380, opened 2025-01-17
docs(config): add GRYPE_CONFIG docs

kema-dev:
## Why

- I started using [scan-action](https://github.com/anchore/scan-action) and got disappointed when I tried to add a config file, this made me create [this PR](https://github.com/anchore/scan-action/pull/427). However, I discovered [here](https://github.com/anchore/grype/issues/1731#issuecomment-2294148185) that `GRYPE_CONFIG` can be used to do so! I would like to clarify the docs about this possibility.

## How

- Just add a few words in the readme

(Also added some whitespace tidying)


README.md
```diff
@@ -673,10 +673,13 @@ Default configuration search paths (see all with `grype config locations`):
 Use `grype config` to print a sample config file to stdout.
 Use `grype config --load` to print the current config after loading all values to stdout.
 
-You can specify files directly using the `--config` / `-c` flags to provide your own configuration files/paths:
+You can specify files directly using the environment variable `GRYPE_CONFIG` or the `--config` / `-c` flags to provide your own configuration files/paths:
```
kzantow, line 676: I don't think we need to include an example of this, since someone specifying this on the command line probably would just use `-c` and that would be the preferred way rather than using an environment variable that could be overridden. I think it would make sense to revert the dependency here, but add a blurb secondary to the `-c`, so something like: 

> You can specify files directly using the `--config` / `-c` flags (or environment variable `GRYPE_CONFIG`) to provide your own configuration files/paths:

... would this have been sufficient for you to discover this detail?

kema-dev: Completely agree! Fixed

kzantow (approved):
👍

Issue #2383, opened 2025-01-21
external-sources: `GetMavenPackageBySha` can be rate limited by maven central, grype will silently fail which results in inconsistent scan results


rawlingsj:
**What happened**:

We are scanning using Grype with [external-sources](https://github.com/anchore/grype?tab=readme-ov-file#external-sources) enabled which reaches out to maven central.  For some applications that have a large number of java dependencies I noticed inconsistent scan results between runs.  In one case a critical vuln was reported in derby for one run, the next run it wasn't listed by Grype which lead to confusion.

After digging into the code and hacking a little reproducer detailed below I found that maven central (or the ELB infront) will rate limit requests based on IP and return a 403, blocking requests for ~5 mins.  

I also found that when these 403 errors occur there was [no error logged](https://github.com/anchore/grype/blob/66197338303a17cf6ba80d71a2c1d952e06ec68d/grype/db/v5/matcher/java/matcher.go#L58), vulns that would have been discovered are just omitted.

**What you expected to happen**:

- An error logged or returned to indicate something failed in the calls to maven central to help indicate something went wrong with that scan and results may not be complete.
- Consistent scan results - improvement to throttle requests?  It will make the scan a little slower but through my tests it is reliable.

**How to reproduce it (as minimally and precisely as possible)**:

I've put together a little reproducer test that is not intended to be part of the code base, just shows the error we get today after around 2-3 iterations of requesting a large number of sha requests.  The test can take ~ 4-7 mins to fail.

<details>
<summary>Reproducer Code</summary>

Adding to `./grype/db/v5/matcher/java/`

```go
package java

import (
	"net/http"
	"strings"
	"testing"
	"time"
)

// TestMavenSearch_GetMavenPackageBySha tests the GetMavenPackageBySha method of the mavenSearch struct.
// This is a reproducer and not intended to be a real test or part of the code base, in this for at least.
func TestMavenSearch_GetMavenPackageBySha(t *testing.T) {
	ms := mavenSearch{
		client:  http.DefaultClient,
		baseURL: "https://search.maven.org/solrsearch/select",
	}

	// Known SHA1s to test with, using a large number of known good SHA1s to validate rate limiting
	// This is not typical but for Images with a large number of Java packages.
	shas := []string{
		"bb7b7ec0379982b97c62cd17465cb6d9155f68e8",
		"b45b49c1ec5c5fc48580412d0ca635e1833110ea",
		"245ceca7bdf3190fbb977045c852d5f3c8efece1",
[...]

Issue #2216, opened 2024-10-28
Should only check maven central if pom info is missing


wagoodman:
Today when you enable searching for artifact and group ID from maven central with a sha1 then this is done for all java artifacts, not just ones missing artifact and group ID https://github.com/anchore/grype/blob/9b7d752a7901bdd54f5bbda1b19028d571fa3a4f/grype/matcher/java/matcher.go#L79-L83 . This condition should be tightened -- instead of always searching, we should only search if there is missing pom data.

kzantow:
Should we move this functionality to Syft, as part of the family of `--enrich`ments? Looking up package info via archive hash seems like something that would benefit both tools.

wagoodman:
I think we should do this in both places, since there are multiple avenues. That is, agreed, this should be in syft as an opt in enrichment option. But in case you're using an non-syft SBOM or otherwise different input source with grype then we should still allow for looking up this information.

kzantow:
This seems like one of the things that could be done strictly during an "enhancement" phase in Syft, which Grype also runs. In other words:

First, Syft scans, finds JAR files, attempts to identify using bytes available, surfaces packages with all info including SHA-1 hashes for JARs
-- OR --
Syft imports an SBOM into the internal Syft data model

_Then_, Syft passes found packages to enhancement

This would have to happen _before_ compliance rules are applied.

I think this would allow Grype to essentially operate the same as it does today and allow Syft to get the benefit of updating the certain incomplete Java records.

PR #2384, opened 2025-01-21
external-sources: throttle requests to maven central to avoid being rate limited for large sets of java dependencies

rawlingsj:
When [external-sources](https://github.com/anchore/grype?tab=readme-ov-file#external-sources) are enabled if an Image contains a large number of Java dependencies Grype can get rate limited by maven central.

This change will:
- add a rate limiter to throttle requests at 300ms per second to produce reliable results
- if a normal `artifact not found` error is returned by maven central, the existing debug logging happens
- any other error from maven central will result in an error being logged, including the 403 when Grype is rate limited
- adds an integration test that can be used to verify the rate limiter so we can verify against the real external api settings
- adds a unit test to verify the rate limiter behaviour

Related to issue https://github.com/anchore/grype/issues/2383


rawlingsj:
There's more details in https://github.com/anchore/grype/issues/2383 - this is one potential solution, figured it might be helpful, the issue above also includes a reproducer for the current behaviour.

There may be better alternatives, which if so great.  this approach has proved reliable after a fair amount of testing with the integration test included in the PR.

grype/db/v5/matcher/java/matcher.go
```diff
@@ -55,7 +54,10 @@ func (m *Matcher) Match(store v5.VulnerabilityProvider, d *distro.Distro, p pkg.
 	if m.cfg.SearchMavenUpstream {
 		upstreamMatches, err := m.matchUpstreamMavenPackages(store, d, p)
 		if err != nil {
-			log.Debugf("failed to match against upstream data for %s: %v", p.Name, err)
+			if strings.Contains(err.Error(), "no artifact found") {
+				log.Debugf("no upstream maven artifact found for %s", p.Name)
+			}
+			log.Errorf("failed to match against upstream data for %s: %v", p.Name, err)
```
wagoodman, line 60:
```suggestion
			log.WithFields("package", p.Name, "error", err).Warn("failed to resolve package details with maven")
```

grype/db/v5/matcher/java/matcher_integration_test.go
```diff
@@ -0,0 +1,689 @@
...
+// TestMavenSearch_GetMavenPackageBySha tests the GetMavenPackageBySha method of the MavenSearch struct.
+// This is an integration test and requires network access to search.maven.org.
+// It is not intended to be run as part of the normal test suite.
+// Use this to validate rate limiting in [maven_search.go] and the ability to fetch package data from maven.org.
+func TestMavenSearch_GetMavenPackageBySha(t *testing.T) {
+	ctx := context.Background()
+
+	ms := newMavenSearch(http.DefaultClient, "https://search.maven.org/solrsearch/select")
+
+	// Known SHA1s to test with, using a large number of known good SHA1s to validate rate limiting
```
wagoodman, line 23: I don't think we should be hitting maven in this way during a test that could be run by multiple devs at a time and by automation. I'm not against leaving the test in if we add a build tag that isn't on the common path (run by default or in CI). In that way if we run into issues again we can enable the test for debugging purposes.

grype/db/v5/matcher/java/maven_search.go
```diff
@@ -14,13 +17,24 @@ import (
...
+	rateLimiter *rate.Limiter
+}
+
+// newMavenSearch creates a new mavenSearch instance with rate limiting
+// rate is specified as 1 request per 300ms
+func newMavenSearch(client *http.Client, baseURL string) *mavenSearch {
+	return &mavenSearch{
+		client:      client,
+		baseURL:     baseURL,
+		rateLimiter: rate.NewLimiter(rate.Every(300*time.Millisecond), 1), // 1 request per 300ms, using the integration test to determine the best rate.
```
wagoodman, line 36: just out of curiosity, why 1 request per 300 ms? from experimentation? or did you find any useful maven docs on this?

grype/db/v5/matcher/java/matcher_integration_test.go
wagoodman: I think adding a small unit test that ensures that the rate limiter is wired up by default when using `mavenSearch` would be a good enough sanity check.

grype/db/v5/matcher/java/maven_search.go
rawlingsj: yeah through experimenting, I found that it needed a few iterations sometimes to get the 403 but using 300ms I've not seen one failure.  There could be other tuning, I tried a bunch of http client settings too but this was the lowest I found that reliably worked.

grype/db/v5/matcher/java/matcher_integration_test.go
rawlingsj: Doh - I'd originally had
```
//go:build integration
// +build integration
```
At the top but the IDE skips compilation errors with that so been removing it, will add back if that's good enough, totally agree this shouldn't be a unit test :) 

On that, would adding the tags above skip the tests in your CI?

wagoodman: ah, the tag is there and we're not using tags to select integration tests, so it wouldn't be selected today... maybe let's change the tag to something that doesn't look like it overlaps with the average integration test, something like `stress` or `api-limits` as a build tag.

rawlingsj: sounds good!

grype/db/v5/matcher/java/matcher.go
```diff
@@ -55,7 +54,10 @@ func (m *Matcher) Match(store v5.VulnerabilityProvider, d *distro.Distro, p pkg.
 	if m.cfg.SearchMavenUpstream {
 		upstreamMatches, err := m.matchUpstreamMavenPackages(store, d, p)
 		if err != nil {
-			log.Debugf("failed to match against upstream data for %s: %v", p.Name, err)
+			if strings.Contains(err.Error(), "no artifact found") {
+				log.Debugf("no upstream maven artifact found for %s", p.Name)
+			}
+			log.WithFields("package", p.Name, "error", err).Warn("failed to resolve package details with maven")
```
rawlingsj, line 60: @wagoodman was chatting to @luhring and he mentioned that this could still mean that errors could occur and mean results get missed.  During my testing I saw 0 failures but am wondering about the semantics of how Grype handled failures like this.  Is logging the right way to go as opposed to returning the `err`?

rawlingsj: Anyways, not to distract from this PR, maybe it's a followup thing.  Thanks for all the feedback so far!

luhring: Yeah, since we're just logging the error and not returning it, I don't think this completely prevents the "silent failure" case, where vuln matches are missing from scan to scan.

My two cents is that we shouldn't let that stop us from merging this. If we start returning an error instead, we'll have a different kind of problem, even though we'd have solved silent failures.

From what I can tell, this PR is forward progress on the issue, and other work will be needed to make other kinds of improvements to the Maven lookups, like https://github.com/anchore/grype/issues/2216. And then the ideal scenario would be a mechanism where Grype could do these lookups offline.

wagoodman: @kzantow and I were chatting this through -- we agree that there should be some conditions that halt and force grype to return 1 (e.g. http 429 / 503). We also think there are common cases that should be notable but not change execution (keep on matching... e.g. http 404). All of these cases should probably end up as evidence in a new `errors` section in the JSON document, and in the table output be reported as a footer noting any issues found.

I'm ok with merging this PR as is since it alleviates a direct problem and iterating towards these other more impactful suggestions as follow ons.

wagoodman: fyi -- new issue created https://github.com/anchore/grype/issues/2393

wagoodman (approved):

rawlingsj:
Many thanks @wagoodman!

PR #2387, opened 2025-01-21
Remove db debug statements for v6

wagoodman:
This removes the sql debug statements when trace logging when using the v6 exp feature flag. This has no impact on debugging/logging for v1-5 DBs.


spiffcs (approved):

kzantow:
Is there a different way to get the SQL printed when trace logging is enabled?

wagoodman:
@kzantow we can always wire up a config level option for this under `exp` but I didn't feel it should be on by default since this is really only a developer option, not for users in general.

PR #2394, opened 2025-01-22
Performance enhancements for DB v6 writes

wagoodman:
All changes on this PR are purely to decrease the time it takes to build a v6 DB, which when incorporated the build time in CI drops from ~2.5 hours to ~25 minutes 🎉

Changes:
- add in-memory ID lookup for common or expensive models (e.g. CPEs originally had a large where clause to fetch similar rows, now this is an O(1) map lookup based on a string). This has been applied to CPEs, OperatingSystems, VulnerabilityHandles, Packages, and Providers. These are relatively flat records and based on debug logs appear to have a high cache hit rate. This also required ensuring we are correctly using this cache across transactions as well as within a transaction (deduplicating in memory).
- Minimizing use of any queries to the DB within `BeforeCreate` hooks
- keep blob digests in memory, not to a separate digests table (theme: maximize the items not written to the DB).
- removed the pragma for 16KB page sizes (this had little effect on DB size and a small performance boost for write times).

While working on this, I did make some unrelated changes:
- I noticed that pragma statements are not strongly verified by sqlite, so explicit checks were added (when adding a new pragma).

### DIFF
--- a/cmd/grype/cli/commands/internal/dbsearch/affected_packages.go
+++ b/cmd/grype/cli/commands/internal/dbsearch/affected_packages.go
@@ -55,6 +55,7 @@ func (c *CPE) String() string {
 	if c == nil {
 		return ""
 	}
+
 	return v6.Cpe(*c).String()
 }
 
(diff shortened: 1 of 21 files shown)


kzantow (approved):
👍

PR #2397, opened 2025-01-23
external-sources: make maven rate limit configurable

rawlingsj:
Allow overriding the default 300ms rate limit for maven API requests via config. This helps users tune request rates based on their specific environment and maven server capacity/restrictions.

Example config:

```yaml
external-sources:
 maven:
   rate-limit: 400ms # or lower than the default if users prefer
```

This PR also includes:
- addition of unit test for the new configurable rate limits
- removal of single unit test case, testing default config as unnecessary given the `defaultExternalSources`
- doc updates for new config, also updated the base maven URL in the docs to match what is being used
- small fix to warn logging, if `no artifact is found`, we don't want to log a warning

### DIFF
--- a/cmd/grype/cli/options/datasources.go
+++ b/cmd/grype/cli/options/datasources.go
@@ -1,6 +1,8 @@
 package options
 
 import (
+	"time"
+
 	"github.com/anchore/clio"
 	"github.com/anchore/grype/grype/db/v5/matcher/java"
 )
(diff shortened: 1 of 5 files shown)


rawlingsj:
Hi - just wondered if there was any thoughts on this PR, making the rate limit configureable has helped after seeing a few 403's returned in the warn log after working more with extremely large java based images.  `apache/hadoop:3.4.1` for example.  In this case I did have to set the rate limit to 400 but was reluctant to change the default hard coded value I added before and instead make it configurable inline with other external sources maven config.  Hope that helps.

cmd/grype/cli/options/datasources.go
```diff
@@ -19,15 +21,17 @@ var _ interface {
 } = (*externalSources)(nil)
 
 type maven struct {
-	SearchUpstreamBySha1 bool   `yaml:"search-upstream" json:"searchUpstreamBySha1" mapstructure:"search-maven-upstream"`
-	BaseURL              string `yaml:"base-url" json:"baseUrl" mapstructure:"base-url"`
+	SearchUpstreamBySha1 bool          `yaml:"search-upstream" json:"searchUpstreamBySha1" mapstructure:"search-maven-upstream"`
+	BaseURL              string        `yaml:"base-url" json:"baseUrl" mapstructure:"base-url"`
+	RateLimit            time.Duration `yaml:"rate-limit" json:"rateLimit" mapstructure:"rate-limit"`
```
kzantow, line 26: TIL `time.Duration` is supported by Viper, calling [time.ParseDuration](https://pkg.go.dev/time#ParseDuration)

kzantow (approved):
LGTM 👍 Thanks @rawlingsj!

PR #2406, opened 2025-01-27
add pretty format

tomersein:
according to this discussion - https://anchorecommunity.discourse.group/t/grype-scans-in-batches-large-sboms/293/11
we have found out that the pretty mode makes the final json very large, we want to reduce the size by default to avoid OOM.

### DIFF
--- a/grype/presenter/json/presenter.go
+++ b/grype/presenter/json/presenter.go
@@ -25,7 +25,7 @@ type Presenter struct {
 }
 
 // NewPresenter creates a new JSON presenter
-func NewPresenter(pb models.PresenterConfig, pretty bool) *Presenter {
+func NewPresenter(pb models.PresenterConfig) *Presenter {
 	return &Presenter{
 		id:               pb.ID,
 		matches:          pb.Matches,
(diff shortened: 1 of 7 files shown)


grype/presenter/json/presenter.go
```diff
@@ -21,10 +21,11 @@ type Presenter struct {
 	metadataProvider vulnerability.MetadataProvider
 	appConfig        interface{}
 	dbStatus         interface{}
+	pretty           bool
 }
 
 // NewPresenter creates a new JSON presenter
-func NewPresenter(pb models.PresenterConfig) *Presenter {
+func NewPresenter(pb models.PresenterConfig, pretty bool) *Presenter {
```
kzantow, line 28: I think it would make sense to add `Pretty` to the `PresenterConfig`

tomersein:
fixed @kzantow 
i see Validations / CLI tests (Linux) (pull_request)  fails, don't have any further information why

kzantow:
The logs aren't always the easiest to read. [This is the failed test](https://github.com/anchore/grype/actions/runs/13086253419/job/36517887240?pr=2406#step:6:105):
```
--- FAIL: TestCmd (16.22s)
    --- PASS: TestCmd/no-args-shows-help (0.02s)
    --- PASS: TestCmd/empty-string-arg-shows-help (0.02s)
    --- FAIL: TestCmd/ensure_valid_descriptor (10.42s)
```

kzantow (approved):
LGTM, thanks @tomersein! (I fixed the test and added config descriptive text)

PR #2431, opened 2025-02-10
More complete severity parsing for v6 DBs

wagoodman:
This builds on #2311 in the following ways:
- fills in CVSS metrics where possible derived from the CVSS vector string
- removes the Score field from the `CVSSSeverity` blob, changes all downstream processing to derive the base score from the vector
- derives string severity (grype severity) from CVSS v2, v3, v3.1, and v4 vector strings

# Stacked PRs
- https://github.com/anchore/grype/pull/2311


grype/db/v6/severity.go
```diff
@@ -0,0 +1,187 @@
...
+		im := roundScore(cvss.Impact())
+		return &vulnerability.CvssMetrics{
+			BaseScore:           roundScore(cvss.BaseScore()),
+			ExploitabilityScore: &ex,
+			ImpactScore:         &im,
+		}, nil
+	}
+}
+
+func roundScore(score float64) float64 {
```
kzantow, line 101: Is there a reason to round the score internally and not just at display time? For example: if `4.0` was where "medium" started, but the score was `3.96`, it would be "low", but get rounded up to "medium"; this doesn't seem accurate to me... but then if the score was rounded up at display time, I may be confused why it says 4.0 but it's classified as low. I feel like a truncation to 1 decimal point during display time may be the "most correct" thing to do

grype/db/v6/severity.go
```diff
@@ -0,0 +1,187 @@
...
+		return vulnerability.HighSeverity
+	}
+	return vulnerability.UnknownSeverity
+}
+
+func interpretCVSSv3Plus(score float64) vulnerability.Severity {
+	if score == 0 {
+		return vulnerability.NegligibleSeverity
+	}
+	if score > 0 && score <= 3.9 {
```
kzantow, line 137: why have the lower bounds on these? I suspect it all works fine after rounding, but it seems easier/safer to just have checks: `if <= 0, if < 4, if < 7, if < 9, if > 10, else`... I was always told floating point comparisons are very finicky due to internal representations that may not be "exact", in the C++ world at least

grype/db/v6/severity.go
```diff
@@ -0,0 +1,187 @@
...
+)
+
+func extractSeverities(vuln *VulnerabilityHandle) (vulnerability.Severity, []vulnerability.Cvss, error) {
+	if vuln.BlobValue == nil {
+		return vulnerability.UnknownSeverity, nil, nil
+	}
+	sev := vulnerability.UnknownSeverity
+	if len(vuln.BlobValue.Severities) > 0 {
+		var err error
+		sev, err = extractSeverity(vuln.BlobValue.Severities[0].Value)
```
kzantow, line 24: I believe you said this was sorted based on "highest severity first" when building the db? it would probably be good to note that here when someone comes back to this section

kzantow (commented):
Overall this looks 👍 just a couple of comments

kzantow:
Running this on `alpine:3.15`, severities are all lowercased:
```
NAME          INSTALLED  FIXED-IN  TYPE  VULNERABILITY   SEVERITY 
busybox       1.34.1-r7            apk   CVE-2022-48174  critical  
libcrypto1.1  1.1.1w-r1            apk   CVE-2024-9143   medium    
libcrypto1.1  1.1.1w-r1            apk   CVE-2024-4741   high      
...
```

grype/db/v6/severity.go
wagoodman: agreed, I'll update. I chose the values based on the official table https://nvd.nist.gov/vuln-metrics/cvss but feels odd.

wagoodman:
good catch on the casing -- I'll update

kzantow (approved):
LGTM 👍

PR #2432, opened 2025-02-10
Add advisory tag constant

wagoodman:
This will be used in grype-db when transforming OS related records with advisories (so they are explicitly tagged)


kzantow (approved):

kzantow (approved):
Still LGTM 🎉

Issue #2286, opened 2024-11-26
Grype DB schemas v3 & v4 deprecation notice


wagoodman:
This issue tracks the work to turn down the infra related to Grype DB Schemas v3 & v4. On January 25th 2025 there will no longer be vulnerability data published with this schema. All code in the Grype and Grype-DB projects dealing with processing v3 & v4 data shapes will be deleted. If you are using a version of grype before v0.51.0 (Oct 2022) please migrate to a later version of Grype.

PR #2435, opened 2025-02-11
Remove DB schema v3 and v4 code

wagoodman:
Based on #2286 (EOL of v3 and v4 DB schemas) all code relating to processing is being deleted. Any shared code between v3-v5 that was under the `legacy` package is now owned by the `v5` package.


cmd/grype/cli/commands/db_check.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/anchore/clio"
-	legacyDistribution "github.com/anchore/grype/grype/db/legacy/distribution"
+	legacyDistribution "github.com/anchore/grype/grype/db/v5/distribution"
```
kzantow, line 12: the smallest of nits: it would be nice if this was `v5distribution` or something with `v5` rather than legacy (it was really helpful for me when going through the code to see "v5" where it was used and not just the import)

kzantow (approved):
🎉

cmd/grype/cli/commands/db_check.go
wagoodman: I didn't catch this before automerge kicked in -- I'll incorporate on a follow up PR

PR #2439, opened 2025-02-11
feat: enable v6 database

### LINKED ISSUE #1648: `grype db status` doesn't always check the db's checksum and validity

### LINKED ISSUE #2077: `grype db status` reports "valid" when the DB is missing
**What happened**:

```console
$ grype db update
 ✔ Vulnerability DB                [updated]
$ grype db status
Location:  /Users/dan/Library/Caches/grype/db/5
Built:     2024-08-22 01:31:37 +0000 UTC
Schema:    5
Checksum:  sha256:a11915b8368897dd446ad5cbb855414870297841f8bccc3b466c5c5e9ba5539b
Status:    valid
```

So far so good — and then...

```console
$ rm -v /Users/dan/Library/Caches/grype/db/5/vulnerability.db
/Users/dan/Library/Caches/grype/db/5/vulnerability.db
$ grype db status
Location:  /Users/dan/Library/Caches/grype/db/5
Built:     2024-08-22 01:31:37 +0000 UTC
Schema:    5
Checksum:  sha256:a11915b8368897dd446ad5cbb855414870297841f8bccc3b466c5c5e9ba5539b
Status:    valid
```

Some indication that the Grype DB is not valid — that it could not be verified because it could not be found.

(See above)

I don't expect this is a common scenario, FWIW. I got into this state while debugging another DB issue. But still, that this is possible seems not ideal.

**Environment**:
- Output of `grype version`:

### LINKED ISSUE #2126: Configure and use DB distribution URLs
Based on #2125  : 
- distribution configuration should point to the base URL, excluding the DB schema version and filenames. We want default configuration as well as user configuration to not need to change when bumping the schema versions.
- we should point by default to the new latest.json path (in grype.anchore.io, not toolbox-data.anchore.io)
- we should enable `grype db list` to reference both the `latest.json` and `history.json` files hosted without the user needing to know/configure them (these filenames should not be overridable for now out of simplicity)
 
Today we have this configuration (with defaults):
```yaml
db:
  update-url: 'https://toolbox-data.anchore.io/grype/databases/listing.json'
```

I propose this changes to:
```yaml
db:
  update-url: 'https://grype.anchore.io/databases'
```
 
Where the DB curator (distribution side) object is responsible for:
- using the correct schema version
- looking for the schema appropriate files (latest.json and history.json) when needed

### LINKED ISSUE #2129: Replace `grype db diff` with `grype db search --since DATE`

### LINKED ISSUE #2233: Removal of temporary files not working on Windows
**What happened**:

Hi guys!

When I use grype with Git Bash on Windows grype has trouble to remove files and directories:

```
$ grype .
<some stuff about DB (no update needed) and scanned project>
[0000]  WARN no explicit name and version provided for directory source, deriving artifact ID from the given path (which is not ideal)
[0000] ERROR failed to remove file (C:\Users\joerki\AppData\Local\Temp\grype-db-listing974399563): %!w(string=remove C:\Users\joerki\AppData\Local\Temp\grype-db-list
```
(Yes, the filename seems to be incomplete)
Files and (empty) directories stay in temp directory. I see grype-db-listing* and  grype-scratch* relics there.  (* are several digits each)
I don't see this error when I use cmd.

Grype erases temporary date even if (Git) Bash is used on Windows as shell.

Use Git Bash instead of cmd for grype operations.

-
**Environment**:
- Output of `grype version`:
- $ grype version
Application:         grype
Version:             0.83.0
BuildDate:           2024-10-28T21:54:11Z
GitCommit:           0602464ebc9f3c417b1175b3e104b19a006604b7
GitDescription:      v0.83.0
Platform:            windows/amd64
GoVersion:           go1.23.2
Compiler:            gc
Syft Version:        v1.15.0
Supported DB Schema: 5

Windows 11 23H2
Git 2.47.0.2

### LINKED ISSUE #2459: DB list (json) should be a list
**What happened**:
`grype db list -o json ` shows a single object

This should be an array of the current objects (for if/when we implement history.json)

kzantow:
This PR removes v5 database support from Grype, moving support to v6 databases. The `db/v5` package needs to remain, however, as we will continue _building_ v5 databases with functionality from that package in `grype-db`.

Closes #2129 
Closes #2126
Fixes #2459
Fixes #1648 
Fixes #2077 
Fixes #2233

Enables previously implemented functionality:
- #2132
- #2131 
- #2130 
- #2128 
- #2125 
- #2124


grype/db/v6/installation/curator.go
```diff
@@ -79,49 +80,46 @@ func NewCurator(cfg Config, downloader distribution.Client) (db.Curator, error)
...
-	if err != nil {
-		return nil, err
-	}
-
-	m, err := s.GetDBMetadata()
-	if err != nil {
-		return nil, fmt.Errorf("unable to get vulnerability store metadata: %w", err)
-	}
-
-	_, err = c.validate(db.DescriptionFromMetadata(m), c.config.ValidateChecksum)
```
kzantow, line 97: This was causing the db to be hashed twice, since the `.Status()` call was also hashing the db and validating age and is required, I've removed the validation from this call altogether.

grype/db/schema.go
```diff
@@ -1,5 +1,3 @@
 package db
 
-import v5 "github.com/anchore/grype/grype/db/v5"
-
-var SchemaVersion = v5.SchemaVersion
+const SchemaVersion = 6
```
kzantow, line 3: Should this just be entirely removed? We should be reading the version from the DB (we are, at least mostly)

test/cli/db_validations_test.go
```diff
@@ -1,74 +1,465 @@
...
 			assertions: []traitAssertion{
-				assertInOutput("first-run of DB update"),
-				assertInOutput("checking for available database updates"),
-				assertInOutput("max-update-check-frequency: 10h"),
+				assertInOutput("database does not exist"),
+				assertFailingReturnCode,
+			},
+		},
+		{
+			name:           "db status: valid db fails with hash mismatch",
```
kzantow, line 233: This case handles: https://github.com/anchore/grype/issues/1648

test/cli/db_validations_test.go
```diff
@@ -1,74 +1,465 @@
...
+			name:        "db check: missing db does not affect with update",
+			setup:       setup(setupYesterdayDb, deleteDb),
+			dbUpdateURL: todayDbURL,
+			assertions: []traitAssertion{
+				assertDbNotFound,
+				assertFailingReturnCode,
 			},
+		},
+		{
+			name:        "db status: missing db returns error",
```
kzantow, line 224: This test case handles: https://github.com/anchore/grype/issues/2077

grype/db/schema.go
kzantow: To avoid confusion and multiple constants in v6.ModelVersion and db.SchemaVersion, i've removed this

cmd/grype/cli/options/experimental.go
```diff
@@ -1,32 +1,10 @@
...
-	"github.com/anchore/grype/grype/db"
-)
-
 // Experimental options are opt-in features that are...
 // ...not stable
 // ...not yet fully supported
 // ...not necessarily tested
 // ...not ready for production use
 // these may go away at any moment, do not depend on them
 type Experimental struct {
```
kzantow, line 9: should we remove this struct altogether?

wagoodman: my vote would be to keep the struct and leave it empty, that way when we have another feature flag like this it's easier to add than last time

wagoodman:
Open question from community meeting -- does this fix #2233 ?

kzantow:
It seems as of now it does not, and there are other issues on Windows. I'm prioritizing getting these fixed up.

wagoodman (approved):
🎉

PR #2444, opened 2025-02-12
Add more logging and fix search by CPE

wagoodman:
This PR makes the following adjustments:
- adds counts to v6 store searches
- adds stringer functions to models and blobs that are being written to the logger
- when searching for packages with CPEs, if no distro is provided assume this means we should be restricting to records that are not associated with any distros (not search for records that relate to any distro)
- change parameters to `logDroppedVulnerability` to note more detailed reasons and use `logger.Field` for arbitrary key-values (better log presentation)
- adds a new `UnsupportedFormatError` that version constraints can raise up, allowing for the version constraint criteria to not log errors for this case
- exposes a new `dev.db.debug` option to show sql with trace logs

### DIFF
--- a/cmd/grype/cli/commands/db.go
+++ b/cmd/grype/cli/commands/db.go
@@ -4,7 +4,6 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/anchore/clio"
-	"github.com/anchore/grype/cmd/grype/cli/options"
 )
 
 const (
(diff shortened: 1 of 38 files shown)


kzantow (approved):
I see what you mean about the vulnerability name: it hasn't been hydrated at some points. If we could fetch _just that_, maybe it could help too. Nothing to do for this PR, of course. 👍

PR #2449, opened 2025-02-14
Add support for the Mageia distro

dfandrich:
This is basically cargo-culted from existing entries so I don't know if this is
done properly or not. Unfortunately, Mageia vulnerabilities don't seem to be in
the vulnerability database (according to "grype db providers") so I'm not able
to perform a full test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ For commercial support options with Syft or Grype, please [contact Anchore](http
   - CBL-Mariner
   - Debian
   - Distroless
+  - Mageia
   - Oracle Linux
   - Red Hat (RHEL)
   - Ubuntu
--- a/grype/db/v5/namespace/index_test.go
+++ b/grype/db/v5/namespace/index_test.go
@@ -150,6 +150,7 @@ func TestIndex_NamespacesForDistro(t *testing.T) {
 		"wolfi:distro:wolfi:rolling",
 		"chainguard:distro:chainguard:rolling",
 		"archlinux:distro:archlinux:rolling",
+		"mageia:distro:mageia:9",
 	})
 
 	assert.NoError(t, err)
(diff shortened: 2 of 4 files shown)


wagoodman:
Hi @dfandrich ! The first step to adding support for a new distro is to add a new [vunnel](https://github.com/anchore/vunnel) provider that reads [vulnerability advisories from the intended source](https://advisories.mageia.org/). We have some [documentation](https://github.com/anchore/vunnel/blob/main/DEVELOPING.md#adding-a-new-provider) on the best way to get started adding a provider. I don't know much about Mageia, but since it's RPM based there shouldn't be a need to add support in syft to detect packages (🎉 ). Once the vunnel provider is written then it's easier to test the grype changes needed to use that data (we can update this PR later).

Might you be interested in contributing the vunnel provider for this distro?

dfandrich:
I can probably do that. Is there a reason vunnel doesn't pull from osv.dev? That would eliminate the need for all the Alpine, Debian, GSA, Red Hat, SLES, Ubuntu and Wolfi parsers since they're already in osv.dev.

dfandrich:
The data displayed for Mageia vulnerabilities at osv.dev contains all the data of the canonical source, so pulling it from there should be viable. Could you point me to the OSV transformer? Is anyone already working on pulling data from osv.dev?

willmurphyscode:
Hi @dfandrich,

> The data displayed for Mageia vulnerabilities at osv.dev contains all the data of the canonical source, so pulling it from there should be viable. Could you point me to the OSV transformer? Is anyone already working on pulling data from osv.dev?

Grype (or really [vunnel](https://github.com/anchore/vunnel), the client that downloads data to be built into Grype's database), needs to pull down _all_ the vulnerability data for a given provider, and, AFAIK, the osv.dev API doesn't provide and endpoint that can be used to enumerate the entire dataset. Do you know whether Mageia has an upstream source we can access directly?

We do have one provider in progress that uses OSV data, but we get the data by downloading it from a git repo that's upstream of osv.dev.

The transformer is here:
* https://github.com/anchore/grype-db/blob/main/pkg/process/v6/transformers/osv/transform.go
The client is here:
* https://github.com/anchore/vunnel/tree/main/src/vunnel/providers/bitnami

dfandrich:
Mageia serves an index to security advisories at https://advisories.mageia.org/vulns.json
and data on each advisory at https://advisories.mageia.org/{ID}.json
If an OSV parser is available it should be easy to get these data. I'll take a
look at the existing code it won't be for at least a few weeks.

PR #2450, opened 2025-02-15
Additional ecosystem related v6 fixes

wagoodman:
This PR makes several adjustments related to ecosystem (and related) searches based on initial v6 testing:

- separates out name normalization rules for each DB schema. This is done by ensuring that package name resolution flows through the `vx.VulnerabilityProvider` with a new `PackageSearchNames` method on the interface.
- adds name normalization for v6 schema (only python and java, no stock resolver)
- reorients `search.ByLanguage` to be `search.ByEcosystem`, taking both language and package type. This allows v5 to continue using language as the primary way to search while allowing v6 to move towards searching by package type (which is what the DB v6 schema ecosystem column orients around)
- removes duplicate ecosystem alias mappings (jenkins plugins and github actions workflows, both which already have existing mappings)
- adds package types explicitly to the v6 helper function `MimicV5Namespace`
- no longer requires `VulnerabilityHandle` to already exist on vuln reference objects when using fetching `VulnerabilityMetadata` via the `VulnerabilityProvider`. Not all uses of this function have already populated internal instances (such as with related vulnerabilities within the nvd namespace). This now optionally uses `GetVulnerabilities` on the `VulnerabilityStore` in such cases.
- Adds a vulnerability id + provider composite index on the `VulnerabilityHandles` table, speeding up `VulnerabilityMetadata` calls a lot.
- Adds missing OS specifier override for chainguard (making it a "rolling" type)
- only non-rejected vulnerabilities are now considered from any vulnerability provider during matching


grype/db/v6/name/java.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"fmt"
+
+	grypePkg "github.com/anchore/grype/grype/pkg"
+	"github.com/anchore/grype/internal/log"
+	"github.com/anchore/grype/internal/stringutil"
+	"github.com/anchore/packageurl-go"
+)
+
+type JavaResolver struct {
```
kzantow, line 12: I would prefer this to somehow be in the Java matcher, by adding logic in different places like this it becomes harder to figure out what's actually going on during the matching process. It's my feeling that the matchers should be very much representative of the entire process of taking data from a package and mapping it to criteria used to search.

kzantow (approved):

grype/db/v6/name/java.go
wagoodman: yeah, this is probably going to be a grype v1 refactor, but this at least gets matching functional again for v6

PR #2460, opened 2025-02-19
fix(java): error out on maven search rate limiting

luhring:
Following on to #2384 — this makes the failure mode return an error, in addition to logging the issue.

The motivation for this is that otherwise, the unreliability of scan results caused by Maven's rate limiting can't be noticed in programmatic contexts — only by studying log messages.

This adjusts the logic such that if the user wants to use Maven searching to affect vuln matching results, and this is unable to happen successfully, an error is returned indicating as much, which can then prompt the user explicitly to change their approach, such as by turning off the search or adjusting the self rate limiting time.

cc: @rawlingsj @wagoodman

### DIFF
--- a/grype/matcher/java/matcher.go
+++ b/grype/matcher/java/matcher.go
@@ -59,7 +59,8 @@ func (m *Matcher) Match(store vulnerability.Provider, p pkg.Package) ([]match.Ma
 			if strings.Contains(err.Error(), "no artifact found") {
 				log.Debugf("no upstream maven artifact found for %s", p.Name)
 			} else {
-				log.WithFields("package", p.Name, "error", err).Warn("failed to resolve package details with maven")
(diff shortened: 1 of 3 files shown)


luhring:
Friendly ping on this 🙇 ❤️

kzantow (approved):
LGTM, since the Maven sha1 search is an opt-in feature

luhring:
Idk how I missed this, but I just saw https://github.com/anchore/grype/pull/2442/files — and that the error is swallowed. Is there a way to have an error serious enough to bubble up to the caller?

kzantow:
@luhring d'oh! 🤦 Open to suggestions here -- maybe we can add some "stop processing" error?

luhring:
Cool! Tried that out here: https://github.com/anchore/grype/pull/2518

PR #2463, opened 2025-02-19
Add CPE provider

wagoodman:
This PR adjusts input processing for grype to allow for CPE literals from the CLI (in the same way PURLs are supported today):
```
grype 'cpe:2.3:a:zohocorp:manageengine_desktop_central:11'
   ├── by severity: 4 critical, 0 high, 0 medium, 0 low, 0 negligible
   └── by status:   3 fixed, 1 not-fixed, 0 ignored 
NAME                          INSTALLED  FIXED-IN    TYPE  VULNERABILITY   SEVERITY 
manageengine_desktop_central  11         2020-03-07        CVE-2020-8540   Critical  
manageengine_desktop_central  11         100251            CVE-2018-11717  Critical  
manageengine_desktop_central  11         100230            CVE-2018-11716  Critical  
manageengine_desktop_central  11                           CVE-2017-7213   Critical
```

This allows for spot testing CPEs directly from CVE documentation much more easily instead of requiring a full syft package or container image artifact.

It was also found that PURL source formatting when using `-o json` was not functional -- this is also fixed in this PR since the same area was being addressed.

### DIFF
--- a/grype/pkg/cpe_provider.go
+++ b/grype/pkg/cpe_provider.go
@@ -1,0 +1,100 @@
+package pkg
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/anchore/syft/syft/cpe"
+	"github.com/anchore/syft/syft/pkg"
+	"github.com/anchore/syft/syft/sbom"
+	"github.com/anchore/syft/syft/source"
+)
+
+const cpeInputPrefix = "cpe:"
+
+type CPELiteralMetadata struct {
+	CPE string
+}
+
+func cpeProvider(userInput string) ([]Package, Context, *sbom.SBOM, error) {
+	reader, ctx, err := getCPEReader(userInput)
+	if err != nil {
+		return nil, Context{}, nil, err
+	}
+
+	return decodeCPEFile(reader, ctx)
+}
+
+func getCPEReader(userInput string) (r io.Reader, ctx Context, err error) {
+	if strings.HasPrefix(userInput, cpeInputPrefix) {
+		ctx.Source = &source.Description{
+			Metadata: CPELiteralMetadata{
+				CPE: userInput,
+			},
+		}
+		return strings.NewReader(userInput), ctx, nil
+	}
+	return nil, ctx, errDoesNotProvide
+}
+
+func decodeCPEFile(reader io.Reader, ctx Context) ([]Package, Context, *sbom.SBOM, error) {
+	scanner := bufio.NewScanner(reader)
+	var packages []Package
+	var syftPkgs []pkg.Package
+
+	for scanner.Scan() {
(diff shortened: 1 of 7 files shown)


grype/pkg/cpe_provider.go
```diff
@@ -0,0 +1,100 @@
...
+		return nil, nil, fmt.Errorf("unable to decode cpe %q: %w", rawLine, err)
+	}
+
+	syftPkg := pkg.Package{
+		Name:    c.Attributes.Product,
+		Version: c.Attributes.Version,
+		CPEs:    []cpe.CPE{c},
+		// TODO infer from cpe target sw, this is not as important since the only matcher for CPEs is the stock matcher
+		//Type:     pkgType,
+		// Language:...
```
wagoodman, line 87: this is a note for a future enhancement (I am not planning to implement target SW parsing in this PR).

grype/pkg/cpe_provider.go
```diff
@@ -0,0 +1,100 @@
...
+
+		if p != nil {
+			packages = append(packages, *p)
+		}
+		if syftPkg != nil {
+			syftPkgs = append(syftPkgs, *syftPkg)
+		}
+	}
+
+	if err := scanner.Err(); err != nil {
```
willmurphyscode, line 62: TIL `scanner.Err()` returns nil if the only err is `io.EOF`: https://github.com/golang/go/blob/64d82cd72c222caa93b2f71c5970a00ec4e7929a/src/bufio/scan.go#L99

grype/presenter/models/source.go
```diff
@@ -14,6 +15,21 @@ type source struct {
...
+		return source{
+			Type:   "purl-file",
+			Target: m.Path,
+		}, nil
+	case pkg.PURLLiteralMetadata:
+		return source{
+			Type:   "purl",
+			Target: m.PURL,
+		}, nil
+	case pkg.CPELiteralMetadata:
```
willmurphyscode, line 28: Is this missing `CPEFileMetadata`?

willmurphyscode: What about https://github.com/anchore/grype/pull/2463/files#diff-dabafbd923424ab88771650a65b50e02901d044e1466539f0e9d5fd61ed96dcbR42 `decodeCPEFile`?

wagoodman: See the purl processing for this -- it treats cli input and file input the same way (so there is common parsing logic shared). I started with the same approach here, intending to add file support, but at the last second decided against it for simplicity. We can always add file support later, but when we do the shared logic is already in place (this function).

Point taken though on the name being confusing. I'll rename this function and the purl function to just hint they are parsing entries from a reader instead of a file.

willmurphyscode (approved):

Issue #1511, opened 2023-09-21
Show indication of known exploited vulnerabilities (from CISA)


tomerse-sg:
**What would you like to be added**:
CISA provided information regarding if vulnerabilities were exploit.
it will be helpful to present or use this kind of information in grype. I saw we already sync this kind of information from NVD.
link - https://www.cisa.gov/known-exploited-vulnerabilities-catalog
**Why is this needed**:
- prioritize vulnerabilities which were already exploit
- more data about known vulnerabilities
**Additional context**:

tomerse-sg:
is it possible to add it to the grype-db schema change that is planned in the future?
it can be a cool enrichment

alonmaor:
this would be great a enrichment for the existing vulnerabilities
any update on this?

tomerse-sg:
can be relevant for here? https://github.com/anchore/grype-db/issues/108

tomerse-sg:
can be relevant for here? https://github.com/anchore/grype-db/issues/108

wagoodman:
Indeed it will be part of the DB v6 work, which is currently being designed (and this feature is already be incorporated from a schema perspective) 🎉 What will need to happen after the schema lands is to update vunnel/grype-db to slurp in the data and populate it into the DB.

hieunguyentr92:
Hi @wagoodman, we are curious about when this feature is scheduled for implementation. Is there a potential timeline?

wagoodman:
Right now we're working on releasing the v6 DB schema as the default schema in grype --this is a blocker to getting the cisa KEV data into the DB. The good news is that the v6 work will most likely be released this week if all goes well 🎉 ! I don't want to make any date promises but I think this could land on the order of weeks not months.

wagoodman:
I've been working on sorting the grype results using EPSS and KEV info now in the DB, here's what' I'm proposing to do:
- Add a `--sort-by` flag which can take values `package, vulnerability, epss, risk` (today we sort by package)
- Add the EPSS percentile to the table output, which would allow folks to more easily remediate based on the c[overage/efficiency/effort method proposed by the EPSS folks](https://www.first.org/epss/model)
- Define a new metric called "risk" and add it to the table output
- Annotate individual matches in the table output with an indication that there is a known exploit for this vulnerability

The metric "risk" is based off of the `risk ~= threat * severity * impact` from the [epss user guide](https://www.first.org/epss/user-guide) . Where:
[...]

PR #2464, opened 2025-02-19
Add KEV information to v6 DB

wagoodman:
This PR adds the CISA KEV data shape to the DB as a new table and plumbs through `VulnerabilityMetadata` calls to attach matching KEV entries for a given vulnerability.

This adds a new `vulnerability.knownExploited[]` array to the JSON output:
```
$ grype 'cpe:2.3:a:zohocorp:manageengine_desktop_central:1' -o json | jq '.matches[] | select(.vulnerability.id == "CVE-2020-10189").vulnerability.knownExploited'
   ├── by severity: 10 critical, 10 high, 5 medium, 0 low, 0 negligible
   └── by status:   19 fixed, 6 not-fixed, 0 ignored 
```
```json
[
  {
    "cve": "CVE-2020-10189",
    "vendorProject": "Zoho",
    "product": "ManageEngine",
    "dateAdded": "2021-11-03",
    "requiredAction": "Apply updates per vendor instructions.",
    "dueDate": "2022-05-03",
    "knownRansomwareCampaignUse": "unknown",
    "urls": [
      "https://nvd.nist.gov/vuln/detail/CVE-2020-10189"
    ],
    "cwes": [
      "CWE-502"
    ]
  }
]

```

I've removed the KEV `vulnerabilityName` and `shortDescription` from the DB and grype output as it is redundant to the NVD data; here's the removed values for the above record as an example:
```json
{
  "vulnerabilityName": "Zoho ManageEngine Desktop Central File Upload Vulnerability",
  "shortDescription": "Zoho ManageEngine Desktop Central contains a file upload vulnerability that allows for unauthenticated remote code execution."
}
```

Partially addresses https://github.com/anchore/grype/issues/1511


grype/db/v6/store.go
```diff
@@ -55,18 +56,38 @@ func newStore(cfg Config, empty, writable bool) (*store, error) {
 		return nil, fmt.Errorf("failed to open db: %w", err)
 	}
 
+	metadataStore := newDBMetadataStore(db)
```
willmurphyscode, line 59: I don't understand this change, probably because I don't have a good enough mental model of the grype db schema v6 database lifecycle. Can you help me understand?

wagoodman: the goal of this change is to get the DB metadata from the DB since it is used downstream in the vulnerability decorator store (new) to determine client-vs-db capabilities are supported.

Lifecycle wise, there are two times this store is used: when creating a new DB from scratch (in grype-db) or when we are matching packages with vulnerabilities (read only in grype). The first case we start with an empty DB and start adding entries, in which case we need to first write the DB metadata in order to read it. That's why there is a "check if empty is set (which means truncate the DB in preparation for writes) and then set the metadata before getting it"... the read only case will not hit this condition and will simply read the metadata from the DB.

willmurphyscode: That makes sense. The indentation change in the return statement was hiding `vulnerabilityDecoratorStore: newVulnerabilityDecoratorStore(db, bs, dbVersion),` from me, so I didn't understand that previously, we didn't need to get the metadata during this function, and now we do.

grype/presenter/models/vulnerability_metadata.go
```diff
@@ -26,14 +44,41 @@ func NewVulnerabilityMetadata(id, namespace string, metadata *vulnerability.Meta
...
+		}
+	}
+	return result
+}
+
+func formatToDate(t *time.Time) string {
+	if t == nil {
+		return ""
 	}
+	return t.Format("2006-01-02")
```
willmurphyscode, line 81: Does it make sense to use `time.DateOnly` here? https://pkg.go.dev/time#pkg-constants

I had to look up to figure out why this particular string literal is involved :)

internal/schemaver/schema_ver_test.go
```diff
@@ -164,3 +164,92 @@ func TestSchemaVerComparisons(t *testing.T) {
...
+	}{
+		{
+			name:    "valid version",
+			input:   "1.2.3",
+			want:    New(1, 2, 3),
+			wantErr: false,
+		},
+		{
+			name:    "valid version with zeros",
+			input:   "0.0.0",
```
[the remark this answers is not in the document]
wagoodman: good point, this test is not correct (schema ver alwasys starts at 1, not 0)

internal/schemaver/string.go
```diff
@@ -0,0 +1,95 @@
...
+
+func (s String) LessThan(other String) bool {
+	return s.compare(other) < 0
+}
+
+func (s String) GreaterOrEqualTo(other String) bool {
+	return s.compare(other) >= 0
+}
+
+func (s String) compare(other String) int {
```
willmurphyscode, line 60: I'm sort of surprised that we have this logic on both the String representation and Struct representation of the SchemaVersion. Can you help me understand why?

wagoodman: Originally this was used within structs representing JSON that were marshalled to strings. I think at this point I could transition that to only the struct representation (the string approach might be an unnecessary holdover). Let me get back to you.

wagoodman: I've replaced all usages of the string variant with the struct version

grype/db/v6/distribution/client.go
```diff
@@ -230,15 +230,15 @@ func isSupersededBy(current *v6.Description, candidate v6.Description) bool {
 		return true
 	}
 
-	otherModelPart, otherOk := candidate.SchemaVersion.ModelPart()
-	currentModelPart, currentOk := current.SchemaVersion.ModelPart()
+	otherModelPart := candidate.SchemaVersion.Model
```
willmurphyscode, line 233: nit: would this be clearer without temporary variables? e.g. `if candidate.SchemaVersion.Model == 0`?

willmurphyscode (approved):

grype/db/v6/distribution/client.go
wagoodman: done -- also added a Valid() helper function instead of just a 0 check on the single model value

PR #2470, opened 2025-02-20
Detect when DB rehydration is necessary (v6 only)

wagoodman:
This detects the condition when a v6 DB is imported by an older grype client not aware of additional capabilities of the imported DB, and later when a new client which can handle such capabilities attempts to use the DB then the DB is rehydrated (schema migrations are applied and indexes are rebuilt), thus allowing for performant behavior.


grype/db/v6/installation/time.go
```diff
@@ -0,0 +1 @@
+package installation
```
kzantow, line 1: Is this file an accidental commit?

kzantow (approved):

grype/db/v6/installation/time.go
wagoodman: yup

PR #2481, opened 2025-02-24
Add KEV & EPSS to db search schema

wagoodman:
This is a follow up to #2464 and #2472 so that KEV and EPSS information is output when running `db search -o json` and `db search vuln -o json`.


cmd/grype/cli/commands/internal/dbsearch/vulnerabilities_test.go
```diff
@@ -116,8 +211,14 @@ func (m *mockVulnReader) GetAffectedPackages(pkg *v6.PackageSpecifier, config *v
 	return args.Get(0).([]v6.AffectedPackageHandle), args.Error(1)
 }
 
-func ptrTime(t time.Time) *time.Time {
-	return &t
+func (m *mockVulnReader) GetKnownExploitedVulnerabilities(cve string) ([]v6.KnownExploitedVulnerabilityHandle, error) {
```
kzantow, line 214: Where do these values get read from the DB? Could these "decorations" just be regular fields referenced in the vulnerability data model like CPE or Blobs, and we could use a `fill<something>` in `FindVulnerabilities` or similar to efficiently load them?

kzantow (approved):
I think I'm missing something about the decorations. Regardless, this PR is 👍

cmd/grype/cli/commands/internal/dbsearch/vulnerabilities_test.go
wagoodman: I agree with your suggestion. Now that #2492 is going to be merging vulnerability and vulnerability-metadata in most cases I think I can follow up in a separate PR to remove the decorator store -- let's get in the functionality first, then I'll check out that path.

Issue #2482, opened 2025-02-26
grype fails to scan a go binary built using go1.24


xnox:
**What happened**: 

Given this pulumi binary:

```
# syft /usr/bin/pulumi  | grep pulumi/pkg/v3
 ✔ Indexed file system                                                                                                                     /usr/bin/pulumi
 ✔ Cataloged contents                                                                     34581a5a7a042126717559e59c7817f65026ee2fdc977f2a5909bc55536f409b
   ├── ✔ Packages                        [212 packages]  
   ├── ✔ File digests                    [1 files]  
   ├── ✔ File metadata                   [1 locations]  
   └── ✔ Executables                     [1 executables]  
github.com/pulumi/pulumi/pkg/v3                                              v3.152.0+incompatible+dirty            go-module    
```

grype on it fails with:

```
# grype /usr/bin/pulumi 
 ✔ Indexed file system                                                                                                                     /usr/bin/pulumi
 ✔ Cataloged contents                                                                     34581a5a7a042126717559e59c7817f65026ee2fdc977f2a5909bc55536f409b
   ├── ✔ Packages                        [212 packages]  
   ├── ✔ File digests                    [1 files]  
   ├── ✔ File metadata                   [1 locations]  
   └── ✔ Executables                     [1 executables]  
 ✔ Scanned for vulnerabilities     [0 vulnerability matches]  
   ├── by severity: 0 critical, 0 high, 0 medium, 0 low, 0 negligible
   └── by status:   0 fixed, 0 not-fixed, 0 ignored 
[0002]  WARN could not match by package language (package=Pkg(type=go-module, name=github.com/pulumi/pulumi/pkg/v3, version=v3.152.0+incompatible+dirty, ups
No vulnerabilities found
```

**What you expected to happen**:

I expect grype to manage to scan that binary. Note that +incompatible+dirty are largely side-effects of go build processes.

**How to reproduce it (as minimally and precisely as possible)**:

```
$ wget -O pulumi-3.152.0-r0.apk.tar.gz https://packages.wolfi.dev/os/x86_64/pulumi-3.152.0-r0.apk
$ grype pulumi-3.152.0-r0.apk.tar.gz
 ✔ Indexed file system                                                                                               /tmp/syft-archive-contents-2698617573
 ✔ Cataloged contents                                                                     e89677734f8f3392b4f38191417e75fef5c46be991b49e06b32360dc7f645419
   ├── ✔ Packages                        [212 packages]  
[...]

PR #2483, opened 2025-02-26
golang: workaround +incompatible+dirty go1.24 versions

xnox:
Try to workaround the +incompatible+dirty versions that go1.24 generates.

See:
- https://github.com/anchore/grype/issues/2482
- https://github.com/golang/go/issues/71969

### DIFF
--- a/grype/matcher/golang/matcher.go
+++ b/grype/matcher/golang/matcher.go
@@ -7,6 +7,7 @@ import (
 	"github.com/anchore/grype/grype/matcher/internal"
 	"github.com/anchore/grype/grype/pkg"
 	"github.com/anchore/grype/grype/vulnerability"
+	"github.com/anchore/grype/internal/log"
 	syftPkg "github.com/anchore/syft/syft/pkg"
 )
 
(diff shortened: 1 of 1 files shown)


grype/matcher/golang/matcher.go
```diff
@@ -60,6 +61,13 @@ func (m *Matcher) Match(store vulnerability.Provider, p pkg.Package) ([]match.Ma
 		return matches, nil, nil
 	}
 
+	// go1.24 started to generate versions in the form +incompatible+dirty, which
+	// fail to be matched: https://github.com/anchore/grype/issues/2482.
+	if p.Name == mainModule && strings.HasSuffix(p.Version, "+incompatible+dirty") {
```
kzantow, line 66: Instead of adding this change here to update the package's version, I think it would be better to update the creation function to properly account for this: https://github.com/anchore/grype/blob/main/grype/version/golang_version.go#L47

xnox: https://semver.org/#spec-item-10 as per the semver spec all versions should be treated identically irrespective of the build metadata after the '+', thus imho it all should be stripped.

> Build metadata MUST be ignored when determining version precedence.

All of `+.*` should be trimmed from all semver, at syft cataloger level.

To also support any future additions.

Also golang violates the semver spec, as it should be "+incompatible.dirty" as per spec.

kzantow: Right -- so in the `newGolangVersion` function, we're taking action to correct go version strings by trimming `go` prefix, for example. It seems like the right place to look for and trim `+dirty`. Why would it be better to happen in this function?

xnox: @kzantow i wonder how to instrument for such combo to be used for lots of things. Because yeah, it is unlikely that go had main module specific logic at all.

So it would need to be local checkout of a project, use replaces, point to a directory, which is checked out to a tag, with a dirty tree. But I can't remember if this will generate dependencies versions to be +incompatible+dirty or if it will have no module version information at all.

Also it seems golang is just generating versions in violation to the semver spec. Since it should be dot-separate fields as in "+incompatible.dirty". I opened https://github.com/golang/go/issues/71971 about that.

xnox: @kzantow @luhring please see above implemented in https://github.com/anchore/grype/pull/2486

PR #2492, opened 2025-02-27
Refactor presenters to use static model over dynamic lookups

wagoodman:
Today all format presenters have access to the `vulnerability.MetadataProvider` which incurs DB calls to add additional data (such as related vulnerabilities, severity, descriptions, etc) within the presenters. More ideally all presenters would take a static model of finalized data that would simply be formatted without the need to access the DB (thus mock out these calls in tests).

This also makes it easy to add configuration around sorting results -- without a change like this, sorting would be either a function of `match.Matches` or a function of each individual presenter implementation (which required duplicate implementations for similar functionality).

This makes a large change to the relationship between `vulnerability.Vulnerability` and `vulnerability.Metadata`; in the past metadata was fetched just-in-time within the presenters and only the vulnerability object was available during matching (thus in the `Matches` collection object too). This changes the v6 `VulnerabilityProvider` to fetch and attach the metadata objects to vulnerability objects just before returning final results.

This implementation gets us further down the right path here in one dimension: presenters should use static data. However, the core vulnerability objects will probably be refactored in the future to be more data-centric and maybe let us promote some of this work to the core objects. Since none of that work has been planned yet, this at least gets us moving in the right direction (even if this isn't the final implementation, which would come as part of the grype 1.0 work).

# Note
- This depends on v6 behavior entirely and deletes the v5 provider to avoid requiring any changes needed (since it will no longer be used for matching). For that reason this must be merged after #2439 is merged.


grype/db/v6/vulnerability_provider.go
```diff
@@ -35,6 +35,7 @@ var _ interface {
 	vulnerability.Provider
 } = (*vulnerabilityProvider)(nil)
 
+// Deprecated: vulnerability.Vulnerability objects now have metadata included
 func (s vulnerabilityProvider) VulnerabilityMetadata(ref vulnerability.Reference) (*vulnerability.Metadata, error) {
```
kzantow, line 39: We haven't had a Grype release with the `VulnerabilityProvider` yet -- my vote is just take the `VulnerabilityMetadata` function out now, is there a reason not to?

I am 💯 on board with vulnerability metadata being set on the vulnerability objects when returned from `FindVulnerabilities`, I hadn't made the change to limit the size of my PRs 😄

grype/presenter/json/presenter_test.go
```diff
@@ -140,17 +122,14 @@ func TestEmptyJsonPresenter(t *testing.T) {
...
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	if !sort.IsSorted(models.MatchSort(doc.Matches)) {
-		t.Errorf("expected matches to be sorted")
-	}
-}
+// TODO: fix me
+//func TestPresenter_Present_NewDocumentSorted(t *testing.T) {
```
kzantow, line 126: Should this get deleted or uncommented (unclear to me what the issue is that needs fixing)?

grype/presenter/table/__snapshots__/presenter_test.snap
```diff
@@ -1,18 +1,15 @@
 
 [TestTablePresenter/no_color - 1]
 NAME       INSTALLED  FIXED-IN          TYPE  VULNERABILITY  SEVERITY 
-package-1  1.1.1      the-next-version  rpm   CVE-1999-0001  [0;32mLow[0m       
-package-2  2.2.2                        deb   CVE-1999-0002  [1;31mCritical[0m
```
kzantow, line 5: 🎉

grype/vulnerability/provider.go
```diff
@@ -13,8 +13,10 @@ type Criteria interface {
 }
 
 // MetadataProvider implementations provide ways to look up vulnerability metadata
+// Deprecated: vulnerability.Vulnerability objects now have metadata included
```
kzantow, line 16: I think we should at least remove `MetadataProvider` from the provider, below, if possible

kzantow (approved):
LGTM -- only real thing is I would love to get rid of MetadataProvider or at least remove it from the VulnerabilityProvider. I only added it there as a way to minimize code changes during the large refactorings.

grype/db/v6/vulnerability_provider.go
wagoodman: I didn't attempt to solve the case where we're looking back up Metadata objects when there is missing Reference.Internal fields or for RelatedVulnerabilities (which are References). I think it would be another medium lift to finish that off -- but I entirely agree with removing this function/interface, just didn't want to boil the ocean with this PR.

grype/vulnerability/provider.go
wagoodman: I think that can be done -- but in a follow up PR (or part of grype 1.0 work)

grype/presenter/json/presenter_test.go
wagoodman: I'll remove this test, I ended up checking that sorting is occurring in the `models` package instead. We really don't need tests in each presenter package for checking sorting anymore.

PR #2494, opened 2025-02-28
Fix CPE target software filtering + improve logging

wagoodman:


test/quality/.grype.yaml
```diff
@@ -1 +1,22 @@
 by-cve: true
+
+match:
+  java:
+    using-cpes: true
```
kzantow, line 5: Why are all these enabled?

kzantow (approved):
We discussed the reason for returning, well, a "reason" from the criteria. Something about it doesn't sit well for me, though I can't really put my finger on exactly why. I understand there are some strings formatted with some appropriate context this way, but I wonder if the Criteria interface required `fmt.Stringer` to provide "enough" information for logs, or there's a different thing that would sit better. Regardless, this isn't something I want to hold up this PR for. I'll add a 1.0 issue to revisit this.

grype/matcher/internal/only_qualified_packages.go
```diff
@@ -1,23 +1,26 @@
...
 
 // onlyQualifiedPackages returns a criteria object that tests vulnerability qualifiers against the provided package
 func onlyQualifiedPackages(p pkg.Package) vulnerability.Criteria {
-	return search.ByFunc(func(vuln vulnerability.Vulnerability) (bool, error) {
+	return search.ByFunc(func(vuln vulnerability.Vulnerability) (bool, string, error) {
 		for _, qualifier := range vuln.PackageQualifiers {
 			satisfied, err := qualifier.Satisfied(p)
 			if err != nil {
-				return satisfied, err
+				return satisfied, fmt.Sprintf("unable to evaluate qualifier: %s", err), err
```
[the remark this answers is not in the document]
wagoodman: ah, it is not -- I'll adjust

test/quality/.grype.yaml
wagoodman: I had a conversation with @westonsteimel  about this, we want to be able to test this kind of change with as broad of a lens in terms of CPE matching as possible. I'll dig through the QG findings to see if there are additional changes needed. 

I'll also add a comment in this config as to why this is enabled.

wagoodman:
> but I wonder if the Criteria interface required fmt.Stringer to provide "enough" information for logs

I tried that initially but wasn't able to cover all cases. Always up for reevaluating that approach for grype 1.0

wagoodman:
Added label updates to support this PR: https://github.com/anchore/vulnerability-match-labels/pull/141

wagoodman:
I've verified that the QG matching is doing the right thing even though in some cases this is performing a little worse in the examples we have. I'll hold off on merging this though until just before release, otherwise subsequent PRs will also start failing CI for the same reason.

PR #2518, opened 2025-03-08
fix(java): Ensure fatal error from maven search bubbles up

luhring:
This is a follow-up to https://github.com/anchore/grype/pull/2460#issuecomment-2706893918, sticking with the idea that a Maven search error is a big enough deal to consumers that the scan ought to be stopped (since the alternative is a nondeterministic and untrustworthy scan result).

But since matcher errors stopped bubbling up in https://github.com/anchore/grype/pull/2442 by default, this introduces a wrapped error type for the `match` package that allows matchers to declare that certain errors are fatal and must be bubbled up regardless.

This also fixes a minor issue where `errors.Join` was being used incorrectly. (See https://github.com/jonjohnsonjr/til/blob/main/post/errors.md.)

### DIFF
--- a/grype/match/matcher.go
+++ b/grype/match/matcher.go
@@ -27,17 +27,17 @@ type fatalError struct {
 	inner   error
 }
 
-// NewFatalError creates a new FatalError wrapping the given error
+// NewFatalError creates a new fatalError wrapping the given error
 func NewFatalError(matcher MatcherType, e error) error {
 	return fatalError{matcher: matcher, inner: e}
 }
 
-// Error implements the error interface for FatalError.
+// Error implements the error interface for fatalError.
 func (f fatalError) Error() string {
 	return fmt.Sprintf("%s encountered a fatal error: %v", f.matcher, f.inner)
 }
 
-// IsFatalError returns true if err includes a FatalError
+// IsFatalError returns true if err includes a fatalError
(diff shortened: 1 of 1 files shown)


luhring:
Example:

```console
$ go run ./cmd/grype ~/imgs/apache-nifi.syft.json             
 ⠏ Scanning for vulnerabilities    ━━━━━━━━━━━━━━━━━━━━  [0 vulnerability matches]  
   ├── by severity: 0 critical, 0 high, 0 medium, 0 low, 0 negligible
   └── by status:   0 fixed, 0 not-fixed, 0 ignored 
[0232] ERROR unable to find matches against vulnerability database: vulnerability matching cannot continue: resolving details for package "jetty-websocket-core-client" with maven: status 403 Forbidden from https://search.maven.org/solrsearch/select?q=1%3A%22ab0040f5331831417a8ef920242c40404167b9f8%22&rows=1&wt=json
exit status 1
```

kzantow (commented):
LGTM but we should probably have a test validating the behavior for fatal/nonfatal errors

luhring:
> LGTM but we should probably have a test validating the behavior for fatal/nonfatal errors

Done!

kzantow (approved):
I hope you don't mind, I pushed a couple tweaks to the comments and such. Thanks for implementing this! 🎉

PR #2526, opened 2025-03-13
chore: attach snapshot build artifacts to each workflow

kzantow:


.github/workflows/validations.yaml
```diff
@@ -224,3 +265,22 @@ jobs:
...
+  Cleanup-Cache:
+    name: "Cleanup snapshot cache"
+    if: github.event.pull_request.head.repo.full_name == github.repository
+    runs-on: ubuntu-24.04
+    permissions:
+      actions: write
+    needs:
+      - Acceptance-Linux
+      - Acceptance-Mac
+      - Cli-Linux
```
kzantow, line 278: I think this is wrong, should be "CLI tests (Linux)"

PR #2532, opened 2025-03-17
Allow import DB from URL

### LINKED ISSUE #2134: Allow DB import from a URL

wagoodman:
Enables being able to import DBs from a URL (whereas today this must be a local file), so the user does not need to download them first:
```
$ grype db import https://grype.anchore.io/databases/v6/vulnerability-db_v6.0.2_2025-03-14T01:31:06Z_1741925227.tar.zst
 ⠴ Vulnerability DB                ━━━━━━━━━━━━━━━━━━━━  [62 MB / 66 MB]  
```

This means if you have a previous grype scan with a DB URL in it (added in #2529) you can quickly use that reference to import the DB:
```
$grype db import $(cat mygrype.json | jq '.descriptor.db.status.from')
 ✔ Vulnerability DB                [imported]  
```

The value in grype JSON documents has the checksum of the argument as a query parameter (`https://grype.anchore.io/databases/v6/vulnerability-db_v6.0.2_2025-03-14T01:31:06Z_1741925227.tar.zst?checksum=sha256%3Ad4654e3b212f1d8a1aaab979599691099af541568d687c4a7c4e7c1da079b9b8`), which allows for client side validation:
```
$ grype db import https://grype.anchore.io/databases/v6/vulnerability-db_v6.0.2_2025-03-14T01:31:06Z_1741925227.tar.zst?checksum=sha256%3Ad4654e3b212f1d8a1aaab979599691099af541568d687c4a7c4e7c1da079b9b7
[0004] ERROR unable to import vulnerability database: unable to update vulnerability database: unable to download db: Checksums did not match for /tmp/getter3555198916/archive.
Expected: d4654e3b212f1d8a1aaab979599691099af541568d687c4a7c4e7c1da079b9b7
Got:      8a179b9568141aad1092b899fe7c1da09af5a69d4654e3b21b97957c4a7c4d68
```

Closes #2134

## PR Stack
- #2529

### DIFF
--- a/grype/db/v6/installation/curator.go
+++ b/grype/db/v6/installation/curator.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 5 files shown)


grype/db/v6/installation/curator.go
```diff
@@ -443,6 +467,12 @@ func (c curator) Import(path string) error {
 	return nil
 }
 
+var urlPrefixPattern = regexp.MustCompile("^[a-zA-Z]+://")
```
kzantow, line 470: It looks like this will match `file://`; I think this will end up working fine, since it's supported by go-getter, is that right?

kzantow (approved):
LGTM

grype/db/v6/installation/curator.go
wagoodman: yup yup -- it supports handling `file://` 👍

Issue #1609, opened 2023-11-17
Use the upstream Bitmani vulndb data for matching


wagoodman:
Bitnami is providing vulnerability matching data for their contianers, which have embedded SPDX documents outlining the contained components: https://github.com/bitnami/vulndb . This could be leveraged in order to improve matching in grype for those components.

This involves at least the following tasks:
- [ ] Write a new [vunnel provider](https://github.com/anchore/vunnel/blob/main/DEVELOPING.md#adding-a-new-provider) so that we can pull and prepare the data for grype-db. Part of this work is understanding if this data fits into an existing schema, or if we need to create a new one (I think a new one is needed at first glance).
- [ ] Update [grype-db](https://github.com/anchore/grype-db/blob/main/DEVELOPING.md#architecture) to be able to transform and write entries to the DB. This depends on the schema written out by vunnel.
- [ ] Ensure that [syft](https://github.com/anchore/syft/blob/1c787f436fca20cbec4541f978ed2a634b880c97/syft/pkg/cataloger/sbom/cataloger.go#L22-L37) to be able to pick up on SBOMs that are contained within the `/opt/bitnami/*` locations (I think this should already work, but have not verified).
- [ ] Update [grype] to be able to match with these new db records. This will take a little bit of thinking. I don't think we need to add a new `Matcher` object, but most likely enhance the [generic `search.*` functions](https://github.com/anchore/grype/blob/main/grype/search/language.go) to look for bitnami specific material and additionally search those namespaces. This might mean that we need to update the [namespace](https://github.com/anchore/grype/blob/main/grype/db/v5/namespace/language/namespace.go) logic to determine whether to include the additional bitnami namespaces (haven't thought through this entirely yet).

willmurphyscode:
For the last item, updating grype to be able to search by these new records, we're starting to think about that pretty early, because to write a correct namespace we need to know how grype should search.

@wagoodman do you think this is a good time to add a `ByPURL` search function, and emit the namespace `bitnami:purl`? If it's not time to add a new search type, which search type should these be added to?

PR #1826, opened 2024-04-25
feat(vex): CSAF support

juan131:
Following https://github.com/anchore/grype/pull/1397, this PR extends [VEX support for ignoring/adding matches based on Vulnerability Exploitability Exchange data](https://github.com/anchore/grype?tab=readme-ov-file#vex-support) so it also accepts [CSAF format](https://oasis-open.github.io/csaf-documentation/).

The PR also refactors the `grype/vex` package since now there's more than one supported VEX format (CSAF & OpenVEX) and it requires the vex processor to be able to distinguish between VEX formats and apply a different implementation.

Credits to @joancafom as the designer of this implementation.

## Use case

Given a CSAF VEX document such as the one below:

```json
{
  "document": {
    "category": "csaf_vex",
    "csaf_version": "2.0",
    "notes": [
      {
        "category": "summary",
        "text": "Example Company VEX document. Unofficial content for demonstration purposes only.",
        "title": "Author comment"
      }
    ],
    "publisher": {
      "category": "vendor",
      "name": "Example Company ProductCERT",
      "namespace": "https://psirt.example.com"
    },
    "title": "AquaSecurity example VEX document",
    "tracking": {
      "current_release_date": "2022-03-03T11:00:00.000Z",
      "generator": {
        "date": "2022-03-03T11:00:00.000Z",
        "engine": {
          "name": "Secvisogram",
          "version": "1.11.0"
        }
      },
      "id": "2022-EVD-UC-01-A-001",
      "initial_release_date": "2022-03-03T11:00:00.000Z",
      "revision_history": [
        {
          "date": "2022-03-03T11:00:00.000Z",
          "number": "1",
          "summary": "Initial version."
        }
      ],
      "status": "final",
      "version": "1"
    }
  },
  "product_tree": {
    "branches": [
      {
        "branches": [
          {
            "branches": [
              {
                "category": "product_version",
                "name": "2.6.0",
                "product": {
                  "name": "Spring Boot 2.6.0",
                  "product_id": "SPB-00260",
                  "product_identification_helper": {
                    "purl": "pkg:apk/alpine/libcrypto3@3.0.8-r3?arch=aarch64&distro=alpine-3.17.3&upstream=openssl"
                  }
                }
              }
            ],
            "category": "product_name",
            "name": "Spring Boot"
          }
        ],
        "category": "vendor",
        "name": "Spring"
      }
    ]
  },
  "vulnerabilities": [
    {
      "cve": "CVE-2023-1255",
      "notes": [
        {
[...]

PR #2538, opened 2025-03-18
feat: add Bitnami matcher

juan131:
# Summary

After https://github.com/anchore/vunnel/pull/447, https://github.com/anchore/grype-db/pull/217 and https://github.com/anchore/syft/issues/3065 this PR attempts to use Bitmani vulndb data for matching package vulnerabilities.

- fixes https://github.com/anchore/grype/issues/1609

### DIFF
--- a/grype/db/v6/data.go
+++ b/grype/db/v6/data.go
@@ -72,6 +72,8 @@ func KnownPackageSpecifierOverrides() []PackageSpecifierOverride {
 
 		// jenkins plugins are a special case since they are always considered to be within the java ecosystem
 		{Ecosystem: string(pkg.JenkinsPluginPkg), ReplacementEcosystem: ptr(string(pkg.JavaPkg))},
+		// Bitnami is a special case since it's not a language ecosystem but is a package type
+		{Ecosystem: "bitnami", ReplacementEcosystem: ptr(string(pkg.BitnamiPkg))},
 	}
 
 	// remap package URL types to syft package types
--- a/grype/matcher/bitnami/matcher.go
+++ b/grype/matcher/bitnami/matcher.go
@@ -1,0 +1,27 @@
+package bitnami
+
+import (
+	"github.com/anchore/grype/grype/match"
+	"github.com/anchore/grype/grype/matcher/internal"
+	"github.com/anchore/grype/grype/pkg"
+	"github.com/anchore/grype/grype/vulnerability"
+
+	syftPkg "github.com/anchore/syft/syft/pkg"
+)
+
+type Matcher struct{}
+
+func (m *Matcher) PackageTypes() []syftPkg.Type {
+	return []syftPkg.Type{syftPkg.BitnamiPkg}
+}
+
+func (m *Matcher) Type() match.MatcherType {
+	return match.BitnamiMatcher
+}
+
+func (m *Matcher) Match(store vulnerability.Provider, p pkg.Package) ([]match.Match, []match.IgnoredMatch, error) {
+	// Bitnami packages' metadata are built from the package URL (instead of CPE, hence CPE match is not supported)
+	// ref: https://github.com/anchore/syft/blob/main/syft/pkg/bitnami.go#L3-L13
+	// ref: https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/bitnami/package.go#L18-L45
+	return internal.MatchPackageByEcosystemAndCPEs(store, p, m.Type(), false)
+}
--- a/grype/match/matcher_type.go
+++ b/grype/match/matcher_type.go
@@ -16,6 +16,7 @@ const (
 	GoModuleMatcher    MatcherType = "go-module-matcher"
 	OpenVexMatcher     MatcherType = "openvex-matcher"
 	RustMatcher        MatcherType = "rust-matcher"
+	BitnamiMatcher     MatcherType = "bitnami-matcher"
 )
 
 var AllMatcherTypes = []MatcherType{
@@ -32,6 +33,7 @@ var AllMatcherTypes = []MatcherType{
 	GoModuleMatcher,
 	OpenVexMatcher,
 	RustMatcher,
+	BitnamiMatcher,
 }
 
 type MatcherType string
--- a/grype/matcher/matchers.go
+++ b/grype/matcher/matchers.go
@@ -3,6 +3,7 @@ package matcher
 import (
 	"github.com/anchore/grype/grype/match"
 	"github.com/anchore/grype/grype/matcher/apk"
+	"github.com/anchore/grype/grype/matcher/bitnami"
 	"github.com/anchore/grype/grype/matcher/dotnet"
 	"github.com/anchore/grype/grype/matcher/dpkg"
 	"github.com/anchore/grype/grype/matcher/golang"
@@ -44,5 +45,6 @@ func NewDefaultMatchers(mc Config) []match.Matcher {
 		&portage.Matcher{},
 		rust.NewRustMatcher(mc.Rust),
 		stock.NewStockMatcher(mc.Stock),
+		&bitnami.Matcher{},
 	}
 }


grype/matcher/bitnami/matcher.go
```diff
@@ -0,0 +1,27 @@
...
+
+func (m *Matcher) Type() match.MatcherType {
+	return match.BitnamiMatcher
+}
+
+func (m *Matcher) Match(store vulnerability.Provider, p pkg.Package) ([]match.Match, []match.IgnoredMatch, error) {
+	// Bitnami packages' metadata are built from the package URL (instead of CPE, hence CPE match is not supported)
+	// ref: https://github.com/anchore/syft/blob/main/syft/pkg/bitnami.go#L3-L13
+	// ref: https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/bitnami/package.go#L18-L45
+	return internal.MatchPackageByEcosystemAndCPEs(store, p, m.Type(), false)
```
juan131, line 26: @willmurphyscode should we include some mechanisms to match based on the pURL?

juan131: Thanks for the context @willmurphyscode!
FYI: I was assigned a top priority task so you should expect a couple of days delay on my work here. I'll probably resume it next Monday.

juan131: I did a 1st approach at https://github.com/anchore/grype/pull/2538/commits/b6e185830f3016800514356def6486ad20b7f3d9

grype/db/v6/data.go
```diff
@@ -72,6 +72,8 @@ func KnownPackageSpecifierOverrides() []PackageSpecifierOverride {
 
 		// jenkins plugins are a special case since they are always considered to be within the java ecosystem
 		{Ecosystem: string(pkg.JenkinsPluginPkg), ReplacementEcosystem: ptr(string(pkg.JavaPkg))},
+		// Bitnami is a special case since it's not a language ecosystem but is a package type
+		{Ecosystem: "bitnami", ReplacementEcosystem: ptr(string(pkg.BitnamiPkg))},
```
willmurphyscode, line 76: This isn't needed. Since `pkg.Bitnami` is the string `"bitnami"` anyway ( se https://github.com/anchore/syft/blob/main/syft/pkg/type.go#L16 ) this line is a no-op.

grype/match/type.go
```diff
@@ -8,12 +8,14 @@ const (
...
 	ExactIndirectMatch Type = "exact-indirect-match"
 	CPEMatch           Type = "cpe-match"
+	PURLMatch          Type = "purl-match"
 )
 
 var typeOrder = map[Type]int{
 	ExactDirectMatch:   1,
 	ExactIndirectMatch: 2,
 	CPEMatch:           3,
+	PURLMatch:          4,
```
willmurphyscode, line 18: I don't think we want a new match type. This is `exact-direct` since we are searching a vulnerability source we know should govern this package for this exact package. (Indirect is its own type because it signals that we searched via an upstream/source package name, because CPE is its own way of searching. But a PURL search is effectively an exact direct search in this case).

grype/matcher/internal/purl.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/anchore/grype/grype/pkg"
+	"github.com/anchore/grype/grype/search"
+	"github.com/anchore/grype/grype/version"
+	"github.com/anchore/grype/grype/vulnerability"
+	"github.com/anchore/packageurl-go"
+)
+
+var ErrEmptyPURLMatch = errors.New("attempted PURL match against package with no PURL")
+
+func MatchPackageByPURL(provider vulnerability.Provider, p pkg.Package, upstreamMatcher match.MatcherType) ([]match.Match, error) {
```
willmurphyscode, line 17: I wonder if this would be simpler as a facade over https://github.com/anchore/grype/blob/4b35276c4b064f74df0c637bf60c93fafc6f9934/grype/matcher/internal/language.go#L31 ? 

It looks like its copying some things from the CPE matcher that we don't need, like the confidence of `0.9` and the special JVM version handling.

grype/version/format.go
```diff
@@ -85,6 +85,8 @@ func FormatFromPkg(p pkg.Package) Format {
 	switch p.Type {
 	case syftPkg.ApkPkg:
 		return ApkFormat
+	case syftPkg.BitnamiPkg:
```
willmurphyscode, line 88: This surprised me. I'll write a longer comment later, but I thought Bitnami had custom versions, like `7.4.2-12`, and would need a custom version object. Many ecosystems have a custom version object.

grype/match/results.go
```diff
@@ -7,15 +7,15 @@ import (
 	"github.com/scylladb/go-set/strset"
 )
 
-type CPEPackageParameter struct {
+type PackageParameter struct {
```
willmurphyscode, line 10: I think a lot of the changes in this file go away if we're not making PURL its own match type or really its own search type, but using an ecosystem search. (Ecosystem search is in `language.go` because most ecosystems are languages. I thought that Bitnami would be more of an outlier in this regard, but it seems not.)

willmurphyscode (commented):
Thanks for the PR!

I think the big question I have is: Bitnami has its own `go-version` lib, and uses strings like `7.4.2-3` to indicate the third bitnami release of upstream version 7.4.2 (please correct me if I'm wrong). On the other hand, your OSV data has type `SEMVER` on its affected version ranges. But it seems to me that since you have `-3` or whatever on the end of version ranges, you're using an ecosystem-specific matcher.

I was sort of expecting a bitnami-specific entry in the version comparison files at https://github.com/anchore/grype/tree/main/grype/version, and I realize I was sort of expecting the OSV data to have ranges of type `ECOSYSTEM`. Did I miss something here? Do we not need to pull in https://github.com/bitnami/go-version and use it?

Happy to sync up on this as well.

juan131:
Hi @willmurphyscode 

Sorry for the lack of updates on this PR! Easter came and a big part of the team was on vacations so I had to focus on other stuff. 

> I think we should not add whole infrastructure around the PURL search, but should rather do basically a package name and ecosystem like the language matchers do. This will make the diff much smaller.

Great! Also versioning, right?

> Do you happen to know whether regular semver will correctly order versions with the bitnami-added number after a hyphen thing?

As you mentioned, Bitnami does uses its own versioning. It's almost identical to semver with a small modification: pre-releases don't exist and instead they're considered revisions. It's explained in the link below:

- https://github.com/bitnami/go-version?tab=readme-ov-file#usage

> Like `14.4.2-2` is less than `14.4.2-4`?

Exactly! Although we use an `r` prefix on revisions, e.g: `14.4.2-r2`. That's why we created that Go module so it's easier for other projects (such as grype) to compare versions.

willmurphyscode:
@juan131 Thanks for getting back to me! 

I have two big questions regarding the versioning:

1. In your OSV data, you use version type "SEMVER" instead of "ECOSYSTEM", but it sounds like custom version comparison is required beyond strict semver, so I'm surprised here. Is the OSV data encoding the right type of version range events?
2. In the Grype matching, I was expecting to see your Go module for comparing these versions used. You can see examples of other customer comparators being implemented, for example here's one for PEP440-compliant Python package versions: https://github.com/anchore/grype/blob/main/grype/version/pep440_version.go . I was expecting to see this diff add `bitnami_version.go` in that same directory, and I wasn't sure why it didn't.

juan131:
Hi @willmurphyscode 

As you pointed out, we do use `"type": "SEMVER"` on our vulndb OSV records and it's because **those records follow Semantic Versioning strictly**. However, we use our "particular" versioning (using revisions instead of pre-releases) in our SBOM (SPDX files included in Bitnami containers) (that's why we're using it on Syft, see https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/bitnami/package.go#L10)

This is sth we need to take into account so here at Grype. Does it make sense?

juan131:
There's an important caveat with having two versioning: 

What happens when a CVE lists a `last_affected` version? For instance, take [this CVE](https://github.com/bitnami/vulndb/blob/main/data/argo-cd/BIT-argo-cd-2020-11576.json#L29) as reference that says last affected version is `1.5.0`.

If Syft inspects an ArgoCD container and the SPDX shipped on it includes sth like this:

```json
        {
            "SPDXID": "SPDXRef-argo-cd",
            "name": "argo-cd",
            "versionInfo": "1.5.0-1",
            "downloadLocation": "git+https://github.com/argoproj/argo-cd/#refs/tags/v1.5.0",
            "licenseConcluded": "Apache-2.0",
            "licenseDeclared": "Apache-2.0",
            "filesAnalyzed": false,
            "externalRefs": [
                {
                    "referenceCategory": "SECURITY",
                    "referenceType": "cpe23Type",
                    "referenceLocator": "cpe:2.3:*:linuxfoundation:argo_continuous_delivery:1.5.0:*:*:*:*:kubernetes:*:*"
                },
                {
                    "referenceCategory": "PACKAGE-MANAGER",
                    "referenceType": "purl",
                    "referenceLocator": "pkg:bitnami/argo-cd@1.5.0-1?arch=arm64&distro=debian-12"
                }
            ],
            "copyrightText": "NOASSERTION"
        },
```

The version would be `1.5.0-1` which, following our "particular" versioning that considers the `-1` as a revision means that  `1.5.0-1 > 1.5.0`. Hence, in theory, Grype could consider the image is not affected by that CVE since last affected version is older **but I'm afraid that's not correct**.

Why? That revision implies we created a new "Bitnami package revision" for ArgoCD and that new revision may change the bash logic we use during the container initialization or the flags we used to compile the ArgoCD binaries. However, that new package revision **uses exactly the same upstream source code** we used to create the previous one. As a consequence, given Bitnami Vulndb data is created based on the upstream project vulnerabilities, `1.5.0-1` is actually affected by the CVE.

juan131:
> Are you saying we should only use the semver component for version comparisons and drop everything after the hyphen? That is, nothing is every fixed by a hyphen, but is only ever fixed by pulling in more upstream code? I'm never going to see a fixed event with `15.1.1-3` when upstream's `15.1.1` is vulnerable?

Yes, I know it's weird but I think it's the way to go. In the unlikely case that a Bitnami revisions (`-X`) fix an upstream vulnerability (e.g. code is patched at compilation or configuration make it unexploitable), then it's Bitnami responsibliy to emit the corresponding VEX data explaining the reasons why the CVE should be ignored. However, without that VEX data that says the opposite, Grype should assume that, if Bitnami vulndb says a version is affected, there's no unaffected revision.

willmurphyscode:
> However, without that VEX data that says the opposite, Grype should assume that, if Bitnami vulndb says a version is affected, there's no unaffected revision.

Is that VEX data something we'll have access too? Should we be adding that to the Vunnel provider? Has that every been used? How often?

I know you all opened https://github.com/anchore/grype/pull/1826 which is about supporting VEX documents. Is that why?

juan131:
> Is that VEX data something we'll have access too? Has that ever been used? How often?

We are currently producing VEX data but don't expose it publicly. It's something we exclusively offer to our [TAC](https://app-catalog.vmware.com/catalog) customers.

> I know you all opened https://github.com/anchore/grype/pull/1826 which is about supporting VEX documents. Is that why?

Yes, it is important for us that scanners have support for VEX so our customers can consume the VEX data we produce via the scanners. (e.g. we can say "run Grype passing this VEX data via the `xxx` flag").

> Should we be adding that to the Vunnel provider?

I don't think VEX data should be used by default to discard vulnerabilities (regardless we make it public in the future or not). It should be only applied if the user activates some flags acknowledging awareness about its use.

willmurphyscode (commented):
looks like there was one more place to add `Bitnami` to a list of version types. I've pushed a fix - let's see if CI turns green. This PR is correct or very close.

willmurphyscode:
The next steps here are:

1. Merge https://github.com/anchore/grype-db/pull/581 so that there's bitnami data in the default grype db.
2. Add a quality gate image to this PR and vulnerability match labels for that image.
3. Fix the namespace, once the team agrees what the namespace should be.

juan131:
Thanks so much @willmurphyscode !! 

> Add a quality gate image to this PR and vulnerability match labels for that image.

How can I help on this?

juan131:
Thanks so much for making this happen @willmurphyscode !! What are the next steps?

willmurphyscode:
@juan131 all set! Running on the released grype 0.93.0 (just released a few minutes ago):

``` sh
❯ grype bitnami/redis:7.4.0
NAME                    INSTALLED              FIXED-IN                      TYPE       VULNERABILITY         SEVERITY    EPSS%  RISK
redis                   7.4.0-0                7.4.2, 7.2.8, 6.2.17          bitnami    BIT-redis-2024-46981  Unknown     98.72   36.8
... snip
```

juan131:
Hi @willmurphyscode 

I noticed a couple of things by running the command below:

```console
$ grype bitnami/redis:7.4.0 -q | grep BIT
redis                   7.4.0-0                *7.4.2, 7.2.8, 6.2.17          bitnami    BIT-redis-2024-46981  Unknown     98.72   36.8
redis                   7.4.0-0                *7.4.1, 7.2.8, 6.2.16          bitnami    BIT-redis-2024-31449  Unknown     78.34    0.6
redis                   7.4.0-0                *7.4.3, 7.2.8, 6.2.18          bitnami    BIT-redis-2025-21605  Unknown     19.83  < 0.1
redis                   7.4.0-0                *7.4.1, 7.2.8                  bitnami    BIT-redis-2024-31227  Unknown     19.09  < 0.1
redis                   7.4.0-0                *7.4.2, 7.2.8                  bitnami    BIT-redis-2024-51741  Unknown     16.19  < 0.1
redis                   7.4.0-0                8.0.2                          bitnami    BIT-redis-2025-27151  Unknown      5.71  < 0.1
redis                   7.4.0-0                *7.4.1, 7.2.8, 6.2.16          bitnami    BIT-redis-2024-31228  Unknown      0.30  < 0.1
```

1. It reports `BIT-redis...` instead of `CVE-...` for vulnerabilities (when there's an alias, see https://github.com/bitnami/vulndb/blob/main/data/redis/BIT-redis-2024-46981.json#L7)
2. Severity is set tu `Unkown` for every Bitnami vulnerability (but it's something we can obtain from OSV data, see https://github.com/bitnami/vulndb/blob/main/data/redis/BIT-redis-2024-46981.json#L50)

I can contribute to address this if needed.

willmurphyscode:
Hi @juan131 - in grype right now, reporting BIT-... is usual behavior (for example, we report GHSAs by their GHSA number, not their CVE number. This is configurable by passing `--by-cve`, which will cause grype to use the first CVE alias that a given match has.) In Grype 1.0, we will likely change this and make `--by-cve` the default behavior, since using the vuln source's ID instead of the CVE ID often surprises people. Is there a reason to avoid the BIT-... IDs now?

I noticed the `Unknown` severity too - would you mind opening an issue for that? I can take a look as well (I think there are some other data sources where we're incorrectly missing severity.) I think basically the process their is to try to figure out where, between when Vunnel downloads the OSV data and when Grype displays the vuln, the severity was lost. I would work backwards from the display in Grype towards the data. We can talk more on the particular issue. Can you open an issue and let me know there if you get stuck or want me to take a look?

juan131:
Thanks for the context about the CVE, using the `--by-cve` flag works perfectly. 

Regarding the 2nd topic, here you have it: https://github.com/anchore/grype/issues/2725

PR #2540, opened 2025-03-18
wip: working distro flavors

willmurphyscode:


grype/db/v6/models.go
```diff
@@ -470,6 +470,10 @@ type OperatingSystem struct {
 
 	// Codename is the codename of a specific release (e.g. "buster" for debian 10)
 	Codename string `gorm:"column:codename;index,collate:NOCASE"`
+
+	// Designator is a string used to distinguish between different releases of the same version of the operating system
+	// such as RHEL-9.4-EUS vs RHEL-9
+	Designator string `gorm:"column:designator;index:os_idx,unique;index,collate:NOCASE"`
```
wagoodman: the data shape here is a non-breaking change, however based on the data that will be written, this is subtly a breaking change. Since we will now be writing in duplicates of the same fields except for the designator (e.g. rhel 9 & rhel 9+eus) which would break the index as far as older clients are concerned:
```
+----+--------+------------+---------------+---------------+---------------+----------+------------+
| id | name   | release_id | major_version | minor_version | label_version | codename | designator |
+----+--------+------------+---------------+---------------+---------------+----------+------------+
| 23 | redhat | rhel       | 9             |               |               |          | eus        |     <--- this row...
| 24 | redhat | rhel       | 9             | 2             |               |          | eus        |
| 25 | redhat | rhel       | 9             | 4             |               |          | eus        |
| 26 | redhat | rhel       | 9             |               |               |          |            |     <--- ... and this row
+----+--------+------------+---------------+---------------+---------------+----------+------------+
```
So if you use an old client to hydrate a new DB you'll get a constraint failure.

I think the RHEL 9 EUS should really be 9.0 EUS, I'll double check this (in which case this is not a breaking change)

wagoodman: Based on the underlying data what should have been written was 9.0 , not 9; I've fixed the bug and now there are no duplicate rows. However, we're going to want to keep this case in mind when adding tests.

grype/matcher/rpm/matcher.go
```diff
@@ -142,16 +142,132 @@ func (m *Matcher) matchPackage(store vulnerability.Provider, p pkg.Package) ([]m
...
+			FoundByGenerator: func(v vulnerability.Vulnerability) any {
+				return match.DistroResult{
+					VulnerabilityID:   v.ID,
+					VersionConstraint: v.Constraint.String(),
+				}
+			},
+		},
+		disclosures...)
+
+	resolutions, err := provider.FindVulnerabilities(
```
kzantow: What if the results from this call were directly translated into `IgnoreRules`?

grype/matcher/apk/matcher.go
```diff
@@ -26,240 +24,137 @@ func (m *Matcher) Type() match.MatcherType {
...
-				if len(nvdOnlyMatch.Vulnerability.Fix.Versions) > 0 {
-					nvdOnlyMatch.Vulnerability.Fix = vulnerability.Fix{
-						State: vulnerability.FixStateUnknown,
-					}
-				}
-				finalCpeMatches = append(finalCpeMatches, nvdOnlyMatch)
-			}
+	// we still need to raise up explicit ignore rules for every package that has a NAK vulnerability. Note that
+	// this is separate from the combination of disclosures and resolutions that we have already created. This is
+	// because NAKs should apply to results from other matchers, not just the APK matcher.
```
kzantow: This is an important assumption to talk through -- I'm not sure it's universally true

grype/search/version_constraint.go
```diff
@@ -20,27 +20,35 @@ func ByConstraintFunc(constraintFunc func(constraint version.Constraint) (bool,
 	return &constraintFuncCriteria{fn: constraintFunc}
 }
 
+type VersionCriteria struct {
+	Version version.Version
+	vulnerability.Criteria
+}
```
wagoodman, line 31: this might not be correct, since `constraintFuncCriteria` implements additionally the `MatchesConstraint` method (for the `VersionConstraintMatcher` interface)... I don't think type assertion for other callers would work the same way here (needs confirmation).

wagoodman: Fixed https://github.com/anchore/grype/pull/2540/commits/30220d3c02bad469ab01aefe79ae1e56ee61c5fa

wagoodman:
This PR is being replaced by #2787

Issue #1922, opened 2024-06-10
Exit with a different return code for a failed scan


Oh-Py-God:
**What would you like to be added**:
Set the return code to a different number other than `0` or `1` when a scan fails.
There may be 1 or more than 1 error codes for failed scans. For example, the image was not found or the image could not be fetched due to a network issue.

**Why is this needed**:
As of grype version 0.78.0, `-f` / `--fail-on` would set the return code to `1` in case a vulnerability is found greater than the threshold.
However, this does not handle the scenarios where a scan fails due to any particular reason.

While using grype in a CI/CD pipeline, this would be helpful to know if the scan failed or the threshold was crossed.

**Additional context**:
As shown in the attached image, grype returns `1` at both the scenarios:
* Scenario 1: Grype discovered vulnerabilities at or above the severity threshold.
* Scenario 2: Grype scan failed as image pull failed.

![image](https://github.com/anchore/grype/assets/45123449/781ac2ce-d3c4-44fe-b569-8ca855f271ba)

tgerla:
Hi @Oh-Py-God, thank you for the request. We will put this in the backlog. Is this something you're interested in working on? We are happy to help get you started if so. Thanks!

Oh-Py-God:
@tgerla I'd like to work on the fix for this. Let me know the steps please.

devodev:
I'd like to tackle this so we can then update scan-action and close https://github.com/anchore/scan-action/issues/390.

First we need a prerequisite PR in clio to be able to return a different exit code based on the error returned from the root cmd.
I opened a PR to add that capability to clio here: https://github.com/anchore/clio/pull/110

devodev:
@kzantow I think this was closed prematurely. I opened a draft PR to address this issue here: https://github.com/anchore/grype/pull/2575.

devodev:
Also the 0.91.0 release mistakenly mentions this issue as done: https://github.com/anchore/grype/releases/tag/v0.91.0

PR #2575, opened 2025-04-01
feat: return exit code 2 for --fail-on errors

devodev:
Return exit code 2 from the CLI when `--fail-on <severity>` flag is used to differentiate from grype-specific errors.

Use new `MapExitCode` feature of clio added in: https://github.com/anchore/clio/pull/110.

Fixes: #1922


cmd/grype/cli/cli.go
```diff
@@ -70,6 +72,14 @@ func create(id clio.Identification) (clio.Application, *cobra.Command) {
 		).
 		WithPostRuns(func(_ *clio.State, _ error) {
 			stereoscope.Cleanup()
+		}).
+		WithMapExitCode(func(err error) int {
+			// return exit code 2 to indicate when a vulnerability severity is discovered
+			// that is equal or above the given --fail-on severity value.
+			if errors.Is(err, grypeerr.ErrAboveSeverityThreshold) {
+				return 2
```
kzantow, line 80: I think this makes the most sense: error code 1 is "something bad happened, we don't know what it is" and an error code > 1 is a report of a specific condition. We also have [code 100 for db update checks](https://github.com/anchore/grype/blob/0ca1080d18f23fb21c89073651af316e532d39c5/cmd/grype/cli/commands/db_check.go#L88). I think we could handle that condition here, too, if we add a specific error type.

I wonder if this would end up breaking many users or there is a high probability that users are just checking for the nonzero return. I do think the way you've updated here makes the most sense to me. I'll ask the team's opinion on this detail.

devodev: I wholeheartedly hope nobody checks exit=1, best way to shoot yourself in the foot 😄

Good idea to move the check for `exitCodeOnDBUpgradeAvailable` here. When I was going through the code base for alternative exit code, I noticed it but forgot about it. Also the fact that os.Exit is called here means no defers in the callpath will run, which may or may not matter for this particular command, but should probably still be addressed.

I'll add a commit to do just that.

kzantow: Yeah, it would be great to get rid of that `os.Exit` call 👍

devodev:
Thanks @kzantow !
Another ping just to make sure this is brought up, the issue was prematurely closed and the release notes of 0.91.0 mention this feature where I think it will only be available in next release (probably 0.92.0).
https://github.com/anchore/grype/issues/1922#issuecomment-2770054806

PR #2579, opened 2025-04-04
Replace os.ReadDir with afero.ReadDir for consistency

joe-ton:
### Motivation
Replace `os.ReadDir` with `afero.ReadDir` to consistently leverage afero's filesystem abstraction.

### Changes
- **Only** replaced `os.ReadDir(src)` → `afero.ReadDir(fs, src)`.

### Impact
- Minimal, no behavioral changes.

### Compatibility
- Fully backward compatible.

### DIFF
--- a/internal/file/copy.go
+++ b/internal/file/copy.go
@@ -11,7 +11,7 @@ import (
 
 func CopyDir(fs afero.Fs, src string, dst string) error {
 	var err error
-	var fds []os.DirEntry
+	var fds []os.FileInfo // <-- afero.ReadDir returns []os.FileInfo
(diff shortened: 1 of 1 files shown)


popey:
👋  - hello @joe-ton and thanks for the PR. 

I note that we use os.ReadDir in a [few places](https://github.com/search?q=org%3Aanchore%20os.ReadDir&type=code), not just grype. Would you envisage making this kind of change everywhere? Otherwise grype would be different from all our other repos.

Also, please note we require sign-off as per DCO. Thanks again. 🙏

wagoodman (approved):

PR #2587, opened 2025-04-09
Add risk score to results and sort by risk by default

### LINKED ISSUE #1511: Show indication of known exploited vulnerabilities (from CISA)

### LINKED ISSUE #1973: Add EPSS metrics to grype results
for each CVE provide also the epss score based on this - https://www.first.org/epss/
**Why is this needed**:
calculate better the risk for each CVE
**Additional context**:

wagoodman:
Screenshot 2025-04-30 at 9 17 57 AM

This adds a new `risk` metric to the grype results, allowing us to show prioritized vulnerability mathces in a way that helps users remediate the most "important" vulnerabilities first. 

The new `risk` metric is defined as `risk ~= threat * severity`, where:
- `threat` is the EPSS score for the CVE matched with (between 0-1), or if present on the KEV  list, then boosted by 5% (10% if the vulnerability is being used in ransomeware)
- `severity` is the average CVSS base score + string severity on the vulnerability record (scaled to between 0-1)
- the entire result is scaled to be a result from 0-100

**Grype results, regardless of the format, will be sorted by highest-risk-first by default**. Also the raw risk value will be added to the table and json output formats.

A new `--sort-by <value>` flag is being added with the following available options:

- `severity`: sort by severity
- `epss`: sort by EPSS percentile (aka, "threat")
- `risk`: sort by risk score (default)
- `kev`: just like risk, except that KEV entries are always above non-KEV entries
- `package`: sort by package name, version, type
- `vulnerability`: sort by vulnerability ID

Further adjustments have been made to the table output to help users with remediation:

- **add a new EPSS percentile column**: teams typically need to balance risk tolerance and budget when choosing how many vulnerabilities to remediate from the full set of results. The [EPSS documentation](https://www.first.org/epss/model) illustrates that selecting EPSS percentile cutoff could be a good way to maximize coverage and minimize effort relative to your teams needs. Exposing this raw metric makes following this advise much easier.

- **add KEV "pill" annotations next to table rows**: though presence on the KEV list typically means that the vulnerability has a high risk score (thus will float to the top of the results) it does **not** guarantee that a) there are non-KEV entries above it, or b) it is within your teams remediation cutoff (however you choose this). From this perspective it makes sense to make this "exploited in the wild" dimension more visible and not depend on purely sorting methods. If you want a high coupling of KEV evidence and the sort method, I'd recommend using `--sort-by kev`. 

- **emphasize the minimum version upgrade path**: this is really de-emphasizing other upgrade paths and truncating the `fixes` column in the table output when there are several updates (the JSON output will show all values). This better focuses users toward useful information by removing flashier formatting.

Fixes #1511
Fixes #1973

### DIFF
--- a/grype/presenter/table/presenter.go
+++ b/grype/presenter/table/presenter.go
@@ -26,6 +26,14 @@ type Presenter struct {
 	withColor      bool
 
 	recommendedFixStyle lipgloss.Style
+	kevStyle            lipgloss.Style
+	criticalStyle       lipgloss.Style
+	highStyle           lipgloss.Style
+	mediumStyle         lipgloss.Style
+	lowStyle            lipgloss.Style
+	negligibleStyle     lipgloss.Style
+	auxiliaryStyle      lipgloss.Style
+	unknownStyle        lipgloss.Style
 }
 
 type rows []row
@@ -37,7 +45,25 @@ type row struct {
 	PackageType     string
 	VulnerabilityID string
 	Severity        string
+	EPSS            epss
+	Risk            string
 	Annotation      string
+}
+
+type epss struct {
+	Score      float64
+	Percentile float64
+}
+
+func (e epss) String() string {
+	percentile := e.Percentile * 100
+	switch {
+	case percentile == 0:
+		return "  N/A"
+	case percentile < 0.1:
+		return "< 0.1%"
+	}
+	return fmt.Sprintf("%5.2f", percentile)
 }
 
 // NewPresenter is a *Presenter constructor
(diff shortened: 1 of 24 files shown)


grype/presenter/table/presenter.go
```diff
@@ -81,15 +116,16 @@ func (p *Presenter) Present(output io.Writer) error {
...
-			severityColor := getSeverityColor(row.Severity)
 			table.Rich(row.Columns(), []tablewriter.Colors{
-				{},              // name
-				{},              // version
-				{},              // fix
-				{},              // package type
-				{},              // vulnerability ID
-				severityColor,   // severity
-				annotationColor, // annotations
+				{}, // name
```
spiffcs, line 120: Do we need these placeholder empty structs now since we don't use the colors? I don't see where else in the PR we write/read from this list. What happens if we just make this `table.Rich(row.Columns)`

Can we simplify this to just `table.Append`?

Issue #2599, opened 2025-04-16
Grype does not handle cache-dir containing ~ correctly


e4711s:
**What happened**:
When .grype.yaml is present and it contains a cache-dir setting like this

```
db:
  # location to write the vulnerability database cache (env: GRYPE_DB_CACHE_DIR)
  cache-dir: '~/.cache/grype/db'
```

then it actually creates a folder named '~' in the current working directory and below it the cache.
So you have the db in `/home/me/~/.cache/grype/db` .

**What you expected to happen**:

No folder named '~' is created, cache db in `/home/me/.cache/grype/db` .

**How to reproduce it (as minimally and precisely as possible)**:

1. Install grype on Ubuntu 24.04 (I used WSL2, not sure if it makes a difference)
2. run `grype config > ~/.grype.yaml` to create default config containing `cache-dir: '~/.cache/grype/db'`
3. run `grype ubuntu:latest`
4. check the directory contents `ls -la` (it will list the '~' folder)
**Environment**:
- Output of `grype version`: 0.91.0
- OS: Ubuntu 24.04

popey:
Aha! 

Duplicate of #2024 

(which looks like a clio issue under the covers.https://github.com/anchore/clio/issues/77)

kzantow:
FWIW there is nothing in CLIO that handles this today, so the simplest fix would be in Grype directly using something like: `homedir.Expand(cacheDir)` [like we use elsewhere](https://github.com/anchore/grype/blob/86b9742fd36dbb97ee87e1aa68a7332f0cf00f8b/grype/pkg/syft_sbom_provider.go#L146). There was a good suggestion to add a specialized type in CLIO that automatically has this behavior via an implemented `PostLoad` hook, but this hasn't been implemented yet.

PR #2600, opened 2025-04-16
fix: expand db cache dir homedir expressions

### LINKED ISSUE #2024: Grype should expand `~` in paths in config file
**What happened**:

Running `grype config` makes a sample config file. This config file (on macOS) includes 
``` yaml
db:
   cache-dir: '~/Library/Caches/grype/db'
```

This makes sense. There's no need to put my username in the config file. **However**, this config file is interpreted as `$PWD/~/...`. In other words, Grype fails to interpret `~` as the home directory, and instead interprets it as the start of a relative path.

If file paths are specified anywhere in the config file, `~` should be expanded to the user's home directory.

``` sh
cd $(mktemp -d)
grype config >> .grype.yaml
grype alpine:latest
ls -al
```

**Environment**:
- Output of `grype version`:
```
Application:         grype
Version:             0.79.4
BuildDate:           2024-07-31T15:05:32Z
GitCommit:           brew
GoVersion:           go1.22.5
Compiler:            gc
Syft Version:        v1.10.0
Supported DB Schema: 5
```
- OS (e.g: `cat /etc/os-release` or similar): M1 macbook

kzantow:
This PR just calls `homedir.Expand` on the configured DB cache dir. This fixes a longstanding issue that the user's homedir is replaced with `~` in `grype config`, but not expanded back properly as the user's homedir, but rather a subdirectory (`./~`).

Example:
```
$ GRYPE_DB_CACHE_DIR=~/grypedb go run ./cmd/grype alpine:3.15
...
$ ls -l ~/grypedb/6/vulnerability.db
-rw-r--r--@ 1 kzantow  staff  1118699520 Apr 16 08:28 /Users/kzantow/grypedb/6/vulnerability.db
```

Fixes #2024, #2599

### DIFF
--- a/cmd/grype/cli/options/database.go
+++ b/cmd/grype/cli/options/database.go
@@ -4,6 +4,7 @@ import (
 	"time"
 
 	"github.com/anchore/clio"
+	"github.com/anchore/go-homedir"
 	"github.com/anchore/grype/grype/db/v6/distribution"
 	"github.com/anchore/grype/grype/db/v6/installation"
 )
@@ -25,6 +26,7 @@ type Database struct {
 
 var _ interface {
 	clio.FieldDescriber
+	clio.PostLoader
 } = (*Database)(nil)
 
 func DefaultDatabase(id clio.Identification) Database {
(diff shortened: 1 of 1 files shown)


popey:
Works for me!

```
$ snapshot/darwin-build_darwin_arm64_v8.0/grype config > .grype.yaml
$ grep "~" .grype.yaml
  cache-dir: '~/Library/Caches/grype/db'
$ snapshot/darwin-build_darwin_arm64_v8.0/grype db update
 ✔ Vulnerability DB                [updated]
[0000]  WARN overriding db update check
$ ls -ltrha ~/Library/Caches/grype/db/6/vulnerability.db
-rw-r--r--@ 1 alan  staff   1.0G 16 Apr 14:10 vulnerability.db
$ ls -l "~"
ls: ~: No such file or directory
```

popey (approved):
WFM & LGTM FWIW THX

PR #2641, opened 2025-05-05
docs(template): add template reference

OnceUponALoop:
Added a new template to help clarify templated access to fields in the json output.

It was useful for me, thought I'd pass it along :+1:


kzantow:
Thanks for the contribution @OnceUponALoop! I think it's a good idea to have a full reference of the Grype data structure. I should note we are actively working on solidifying Grype APIs and data structures for [Grype 1.0](https://github.com/anchore/grype/milestone/2), so it might be preferred to wait until that's done before adding something like this to avoid having to rework it shortly after it's added.

Issue #2628, opened 2025-04-30
Grype false negatives in versions v0.88.0 and later leading to missed critical vulnerabilities


sschakraborty:
### What happened
I installed two different versions of Grype - v0.87.0 and v0.91.2. I generated a minimalist test SBOM containing vulnerable artifacts. Grype 0.87.0 reports 3 critical and 2 high vulnerabilities whereas any later version (including the latest) does not report anything. This is true for all Grype versions using v6 DB schema.

### What you expected to happen
All versions of Grype should produce consistent results and report all vulnerabilities.

### How to reproduce it (as minimally and precisely as possible)

I used the following ***sbom.json*** file for testing both versions:
```json
{
  "artifacts": [
    {
      "id": "pkg:maven/log4j/log4j@1.2.17",
      "name": "log4j",
      "version": "1.2.17",
      "language": "java",
      "purl": "pkg:maven/log4j/log4j@1.2.17",
      "locations": [
        {
          "path": "META-INF/maven/commons-logging/commons-logging/pom.xml"
        }
      ]
    }
  ],
  "source": {
    "type": "directory"
  },
  "schema": {
    "version": "16.0.9",
    "url": "https://raw.githubusercontent.com/anchore/syft/main/schema/json/schema-16.0.9.json"
  }
}
```

Using **Grype v0.87.0**:
```bash
$> curl -sSfL https://raw.githubusercontent.com/anchore/grype/main/install.sh | sh -s -- -b /usr/local/bin 'v0.87.0'
[info] checking github for release tag='v0.87.0' 
[info] fetching release script for tag='v0.87.0' 
[info] checking github for release tag='v0.87.0' 
[info] using release tag='v0.87.0' version='0.87.0' os='linux' arch='amd64' 
[info] installed /usr/local/bin/grype 
$> grype db update
 ✔ Vulnerability DB                [updated]  
Vulnerability database updated to latest version!

$> grype sbom:sbom.json --add-cpes-if-none
 ✔ Scanned for vulnerabilities     [0 vulnerability matches]  
   ├── by severity: 3 critical, 2 high, 0 medium, 0 low, 0 negligible
   └── by status:   0 fixed, 5 not-fixed, 0 ignored 
NAME   INSTALLED  FIXED-IN  TYPE  VULNERABILITY        SEVERITY 
log4j  1.2.17                     GHSA-f7vh-qwp3-x37m  Critical  
log4j  1.2.17                     GHSA-65fg-84f6-3jq3  Critical  
[...]

Issue #2554, opened 2025-03-21
SBOM many false positives


ionimit:
**What happened**:
The SBOM scanner reported many false positive vulnerabilities
**What you expected to happen**:
Not to report false positives vulnerabilities
**How to reproduce it (as minimally and precisely as possible)**:

Use the following SBOM, all findings are false positive:

[poc-sbom.json](https://github.com/user-attachments/files/19388254/poc-sbom.json)

grype command:
`grype.exe sbom:"C:\path\to\bom.json"`

**Anything else we need to know?**:

The following are identified false positives:

- gen-mapping - https://github.com/advisories/GHSA-8rmg-jf7p-4p22 (actual package is @jridgewell/gen-mapping not gen-mapping)
- middleware-user-agent - https://github.com/advisories/GHSA-7jfr-mfm3-p4mh (actual package is @aws-sdk/middleware-user-agent not middleware-user-agent)
- node-config-provider - https://github.com/advisories/GHSA-p2f3-jr96-8rhf (actual package is @smithy/node-config-provider not node-config-provider)
- protocol-http - https://github.com/advisories/GHSA-p57r-cpw5-9h67 (actual package is @smithy/protocol-http not protocol-http)
- smithy-client - https://github.com/advisories/GHSA-xh9p-w3hh-pqp5 (actual package is @smithy/smithy-client not smithy-client)

Additional misidentified packages that leads to false positive results:

- @types/jose not jose
- @types/mime not mime
- @types/ms not ms
- @types/request not request
- @types/tough-cookie not tough-cookie
- @colors/colors not colors
- @types/debug not debug
- @types/ejs not ejs
- @types/cookiejar not cookiejar
- @types/jsonwebtoken not jsonwebtoken

**Tool Output**:

![Image](https://github.com/user-attachments/assets/05d5f8e7-519a-42a9-99c6-bd18173746e9)

**Environment**:
Application:         grype
Version:             0.89.1
BuildDate:           2025-03-13T20:22:27Z
GitCommit:           718ea3060267edcae7b10a9bf16c0acdad10820a
GitDescription:      v0.89.1
Platform:            windows/amd64
GoVersion:           go1.24.1
Compiler:            gc
Syft Version:        v1.20.0
Supported DB Schema: 6

Issue #1886, opened 2024-05-24
@jridgewell/gen-mapping incorrectly attributed GHSA-8rmg-jf7p-4p22


ataraxus:
**What happened**:

Our pipeline recently broke due to critical finding of GHSA-8rmg-jf7p-4p22
```
[root@d992b56c077e /]# grype --version
grype 0.77.4
[root@d992b56c077e /]# grype bom.json
 ✔ Vulnerability DB                [no update available]
 ✔ Scanned for vulnerabilities     [1 vulnerability matches]
   ├── by severity: 1 critical, 0 high, 0 medium, 0 low, 0 negligible
   └── by status:   0 fixed, 1 not-fixed, 0 ignored
[0000] ERROR failed to fetch latest version: Get "https://toolbox-data.anchore.io/grype/releases/latest/VERSION": tls: failed to verify certificate: x509: certificate has exp
NAME         INSTALLED  FIXED-IN  TYPE  VULNERABILITY        SEVERITY
gen-mapping  0.3.5                npm   GHSA-8rmg-jf7p-4p22  Critical
```

**What you expected to happen**:

The malicious package is: https://www.npmjs.com/package/gen-mapping which is a typesquatting attempt which got cought and squashed. The package in actual usage is: @jridgewell/gen-mapping. Which IMHO is also correctly stated in the provided BOM: 

```
 {
      "type": "library",
      "name": "gen-mapping",
      "group": "@jridgewell",
      "version": "0.3.5",
      "bom-ref": "@jridgewell/gen-mapping@0.3.5",
      "purl": "pkg:npm/%40jridgewell/gen-mapping@0.3.5",
...
}
```

**How to reproduce it (as minimally and precisely as possible)**:

package.json 
````json
{
  "name": "gen-mapping-test",
  "version": "1.0.0",
  "dependencies": {
    "@babel/core": "7.22.10"
  },
  "scripts": {},
  "author": "",
  "license": "ISC"
}
````

````
npm install 
npx -y @cyclonedx/cyclonedx-npm --omit dev --output-reproducible --output-file bom.json --package-lock-only
npx -y @cyclonedx/cyclonedx-npm --omit dev --short-PURLs --output-reproducible --output-file bom-short.json --package-lock-only

grype bom.json
grype bom-short.json
````

**Anything else we need to know?**:

I attached example boms, which trigger the issue.

**Environment**:
- Output of `grype version`:
- OS (e.g: `cat /etc/os-release` or similar):

[...]

Issue #1701, opened 2024-02-06
Vulnerability reported on @group/name dependency when actual vulnerability exists on name dependency


lmco-seth:
**What happened**:

Grype reports a Vulnerability for `colors` when the package in the SBOM is `@colors/colors`. The SBOM is a demo sbom from the [cyclonedx-node-npm](https://github.com/CycloneDX/cyclonedx-node-npm) project. [juice-shop example](https://github.com/CycloneDX/cyclonedx-node-npm/blob/main/demo/juice-shop/example-results/bare/bom.1.5.json#L306-356).

This appears to be caused by Grype ignoring the `group` entry in the SBOM as the NPM `scope` is stored in `group` parameter.

Running `grype --output table <demo-bom.json>` produces --

```
 ✔ Vulnerability DB                [no update available]  
 ✔ Scanned for vulnerabilities     [52 vulnerability matches]  
   ├── by severity: 12 critical, 13 high, 26 medium, 1 low, 0 negligible
   └── by status:   44 fixed, 8 not-fixed, 0 ignored 
[0001]  WARN merging packages have with different pURLs: "bb5ff84367124522"="pkg:npm/%40types/cookie@0.4.1#types/cookie" vs "bb5ff84367124522"="pkg:npm/cookie@0.4.1"
[0001]  WARN merging packages have with different pURLs: "235cdbe593f0b054"="pkg:npm/%40types/strip-bom@3.0.0" vs "235cdbe593f0b054"="pkg:npm/strip-bom@3.0.0"
[0001]  WARN some package(s) are missing CPEs. This may result in missing vulnerabilities. You may autogenerate these using: --add-cpes-if-none
NAME                  INSTALLED  FIXED-IN  TYPE  VULNERABILITY        SEVERITY 
base64url             0.0.6      3.0.0     npm   GHSA-rvg8-pwq2-xj7q  Medium    
colors                1.5.0                npm   GHSA-gh88-3pxp-6fm8  High           
...(excluded for readability)
```

[GHSA-gh88-3pxp-6fm8](https://github.com/advisories/GHSA-gh88-3pxp-6fm8) is for [colors](https://www.npmjs.com/package/colors) not [@colors/colors](https://www.npmjs.com/package/@colors/colors)

**What you expected to happen**:

[...]

PR #2645, opened 2025-05-07
fix: include group/namespace across formats

kzantow:
This PR is a follow-on from [adding a PURL list format to Syft](https://github.com/anchore/syft/pull/3853).

This PR removes the specialized PURL conversion to packages, instead allowing Syft to manage this as well as enhance provided SBOMs using data found in the PURL. The results of this is it fixes a number of issues that many package types were incomplete from SBOMs generated using tools other than Syft, which properly included the CycloneDX `group` field and/or namespace in the PURL. Additionally, per-package distro handling has been centralized to populate package distros regardless of import format, so SBOMs created with tools other than Syft are able to specify upstream packages and linux distributions on a per-package basis.

- Fixes https://github.com/anchore/grype/issues/2628
- Fixes https://github.com/anchore/grype/issues/2554
- Fixes https://github.com/anchore/grype/issues/1886
- Fixes https://github.com/anchore/grype/issues/1701


grype/pkg/package.go
```diff
@@ -44,7 +45,7 @@ type Package struct {
 	Metadata  interface{} // This is NOT 1-for-1 the syft metadata! Only the select data needed for vulnerability matching
 }
 
-func New(p syftPkg.Package) Package {
+func New(p syftPkg.Package, enhancers ...enhancer) Package {
```
wagoodman, line 48: why make an unexposed type that's in a signature of a function in the public api? Though it's still usable it's a little awkward to discover. Any reason not to use the raw function definition here?

wagoodman (approved):

PR #2658, opened 2025-05-13
Refactor match ignore result

kzantow:
While working on v6 I had a bit of an intended goal of making this refactoring, but it got lost along the way. This PR adjusts the return type of the Matcher interface instead of returning `IgnoredMatch`, to return an `IgnoreFilter`. This PR gets rid of all the ecosystem specific code in the `VulnerabilityMatcher`, allowing this to reside in appropriate matchers.


grype/match/ignore.go
```diff
@@ -28,14 +28,15 @@ type IgnoredMatch struct {
...
-	Vulnerability    string            `yaml:"vulnerability" json:"vulnerability" mapstructure:"vulnerability"`
-	Reason           string            `yaml:"reason" json:"reason" mapstructure:"reason"`
-	Namespace        string            `yaml:"namespace" json:"namespace" mapstructure:"namespace"`
-	FixState         string            `yaml:"fix-state" json:"fix-state" mapstructure:"fix-state"`
-	Package          IgnoreRulePackage `yaml:"package" json:"package" mapstructure:"package"`
-	VexStatus        string            `yaml:"vex-status" json:"vex-status" mapstructure:"vex-status"`
-	VexJustification string            `yaml:"vex-justification" json:"vex-justification" mapstructure:"vex-justification"`
-	MatchType        Type              `yaml:"match-type" json:"match-type" mapstructure:"match-type"`
+	Vulnerability        string            `yaml:"vulnerability" json:"vulnerability" mapstructure:"vulnerability"`
+	RelatedVulnerability string            `yaml:"related-vulnerability,omitempty" json:"related-vulnerability,omitempty" mapstructure:"related-vulnerability"`
```
kzantow, line 32: Only change here is adding `RelatedVulnerability`

wagoodman: Should we fill `Vulnerability` here and change `RelatedVulnerability` into a boolean (and probably rename to `IncludeAliases` or similar)? 

This way there aren't two places to put the same information with slightly different semantics, it would be a new field with no overlapping use cases.

grype/vulnerability_matcher.go
```diff
@@ -314,37 +314,42 @@ func (m *VulnerabilityMatcher) normalizeByCVE(match match.Match) match.Match {
...
+		if !ok {
+			continue
+		}
+		if rule.Package.Location != "" && !strings.ContainsRune(rule.Package.Location, '*') {
+			ignores = append(ignores[:i], ignores[i+1:]...) // this rule is handled with location lookups
+			i--
+			out.locationToIgnoreRules[rule.Package.Location] = append(out.locationToIgnoreRules[rule.Package.Location], rule)
 		}
 	}
+	out.remainingFilters = ignores // the remaining rules are not location-based rules
```
wagoodman: nit, consider using slices.DeleteFunc or slices.Retain (new in go1.21) which is easier to read & harder to mess up with subtle index / append issues.

PR #2672, opened 2025-05-19
fix: ensure temp dir on same volume as final db dir

### LINKED ISSUE #2517: Grype can't update DB in docker volume (regression)
**What happened**:
grype db update doesn't work if database in volume.
[0011] ERROR unable to update vulnerability database: unable to activate new vulnerability database: rename /tmp/grype-db-download2458041976 /tmp/grype/6: invalid cross-device link

0.87.0 - works 
0.88.0 - fails
0.89.0 - fails 

`docker run --rm -it -e GRYPE_DB_CACHE_DIR=/tmp/grype --volume cache-grype:/tmp/grype anchore/grype:v0.87.0 db update` - OK

`docker run --rm -it -e GRYPE_DB_CACHE_DIR=/tmp/grype --volume cache-grype:/tmp/grype anchore/grype:v0.88.0 db update` - NOT OK

`docker run --rm -it -e GRYPE_DB_CACHE_DIR=/tmp/grype --volume cache-grype:/tmp/grype anchore/grype:v0.89.0 db update` - NOT OK

`docker run --rm -it anchore/grype:v0.89.0 db update` - OK
`docker run --rm -it anchore/grype:v0.88.0 db update` - OK
`docker run --rm -it anchore/grype:v0.87.0 db update` - OK

**Environment**:
- Image from dockerhub (and self-built too):
- Ubuntu

willmurphyscode:
Previously, when GRYPE_DB_CACHE_DIR was set to a docker volume, the grype temp dir could be created under the /tmp volume in the container but the final destination be under the named volume, which caused os.Rename to fail with an invalid cross-volume link error. Now, ensure that the temporary and final DB location are on the same volume by creating the temp dir under GRYPE_DB_CACHE_DIR.

Fixes #2517

### DIFF
--- a/grype/db/v6/installation/curator.go
+++ b/grype/db/v6/installation/curator.go
@@ -298,7 +298,14 @@ func (c curator) update(current *db.Description) (*distribution.Archive, error) 
 	if err != nil {
 		return nil, fmt.Errorf("unable to resolve vulnerability DB URL: %w", err)
 	}
-	dest, err := c.client.Download(url, filepath.Dir(c.config.DBRootDir), mon.downloadProgress.Manual)
+
+	// Ensure parent of DBRootDir exists for the download client to create a temp dir within DBRootDir
+	// This might be redundant if DBRootDir must already exist, but good for safety.
+	if err := os.MkdirAll(c.config.DBRootDir, 0o700); err != nil {
+		return nil, fmt.Errorf("unable to create db root dir %s for download: %w", c.config.DBRootDir, err)
+	}
+
+	dest, err := c.client.Download(url, c.config.DBRootDir, mon.downloadProgress.Manual)
 	if err != nil {
 		return nil, fmt.Errorf("unable to update vulnerability database: %w", err)
 	}
@@ -307,6 +314,8 @@ func (c curator) update(current *db.Description) (*distribution.Archive, error) 
 
 	mon.downloadProgress.SetCompleted()
 	if err = c.activate(dest, url, mon); err != nil {
+		log.Warnf("Failed to activate downloaded database from %s, attempting cleanup of temporary download directory.", dest)
+		removeAllOrLog(c.fs, dest)
 		return nil, fmt.Errorf("unable to activate new vulnerability database: %w", err)
 	}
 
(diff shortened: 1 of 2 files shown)


grype/db/v6/installation/curator.go
```diff
@@ -298,7 +298,14 @@ func (c curator) update(current *db.Description) (*distribution.Archive, error)
 	if err != nil {
 		return nil, fmt.Errorf("unable to resolve vulnerability DB URL: %w", err)
 	}
-	dest, err := c.client.Download(url, filepath.Dir(c.config.DBRootDir), mon.downloadProgress.Manual)
+
+	// Ensure parent of DBRootDir exists for the download client to create a temp dir within DBRootDir
+	// This might be redundant if DBRootDir must already exist, but good for safety.
+	if err := os.MkdirAll(c.config.DBRootDir, 0o700); err != nil {
```
kzantow, line 304: I had to look up the `0o` prefix -- should we just use `0`, since most examples tend to use this form and the remainder of the app tends to use this form?

grype/db/v6/installation/curator.go
```diff
@@ -307,6 +314,8 @@ func (c curator) update(current *db.Description) (*distribution.Archive, error)
 
 	mon.downloadProgress.SetCompleted()
 	if err = c.activate(dest, url, mon); err != nil {
+		log.Warnf("Failed to activate downloaded database from %s, attempting cleanup of temporary download directory.", dest)
+		removeAllOrLog(c.fs, dest)
```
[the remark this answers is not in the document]
willmurphyscode: I think that we run more risk of filling up disk than of losing debugging clues here. The error messages in this part of the code were pretty clear when debugging the recent issues. What do you think? OK to leave this?

Put another way: we have had users complain about failure to remove temp files. We have not had users complain that they weren't able to investigate a partially activated database.

grype/db/v6/installation/curator.go
willmurphyscode: Hmm, it looks like `0o` was introduced in Go 1.13: https://go.dev/doc/go1.13#language

`gopls` at version `v0.18.1` will re-write `0600` as `0o600` for example, though. It seems like there is momentum (admittedly with mixed opinion) towards the `0o` one. What do you think?

PR #2683, opened 2025-05-23
Added punctuation for one sentence.

patrickjburke245:
I believe you need a period after this sentence. There does not seem to be a reason why this particular sentence isn't punctuated.


kzantow:
I wouldn't be opposed to sneaking this change in with: https://github.com/anchore/grype/pull/2684

patrickjburke245:
Excellent. Let me know if I can assist.

PR #2700, opened 2025-06-02
test: add missing units for old JVM update versions

willmurphyscode:


willmurphyscode:
this doesn't seem worth updating. that part of the code has changed a lot, and this was always a "just in case" sort of PR.

PR #2715, opened 2025-06-06
Update README.md for Accuracy

patrickjburke245:
One cannot sort by the TYPE (package type) or INSTALLED (package version #) fields.

"sort by package name, version, type" should be changed by one of the following:
1. removing ", version, type"
2. changing this to "sort by package name (if there are multiple packages that have the same name, sorting preference is then done based on package version, then package type)"

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,7 +251,7 @@ Results default to sorting by Risk Score and can be overridden with `--sort-by <
 - `epss`: sort by EPSS percentile (aka, "threat")
 - `risk`: sort by risk score
 - `kev`: just like risk, except that KEV entries are always above non-KEV entries
-- `package`: sort by package name, version, type
+- `package`: sort by package name
 - `vulnerability`: sort by vulnerability ID
 
 ### Supported versions


kzantow:
Hey @patrickjburke245 -- the documentation is correct, the sorting of "packages" takes all those fields into account.

patrickjburke245:
Thank you for letting me know, Keith. Is it accurate to say that using `--sort-by package` will sort by package name, then version # if there are duplicates, then package type if there are still duplicates? I'm wondering how all 3 can be sorted upon at the same time; there has to be a priority to them, right?

patrickjburke245:
@kzantow this isn't urgent (can wait a few weeks), just tagging you in case you didn't see my comment.

kzantow:
@patrickjburke245 they should be sorted by the order listed

kzantow:
Hey @patrickjburke245, I'm going to close this PR as I do not believe it is helping to clarify the behavior. Please let me know if I've misunderstood. Thank you for the contribution, though!

Issue #2720, opened 2025-06-10
"Harden Container Runtime with Non-Root User" breaks --output usage


netanel-innovid:
The latest (v0.93.0) release which removed root user from the container, brakes our ability to output the scan results to a file.
We used to mount the current directory (owned by root) to the container, and have it output the result file to the workdir.
```docker run --rm -v /var/run/docker.sock:/var/run/docker.sock:ro -v ${PWD}:/workdir --output json --file my_image_scan_results.json my_image:local```

With the removal of root user from the container image, this obviously brakes:
```[0000] ERROR unable to create report file: my_image_scan_results.json: permission denied```

Would be nice to have the option to pull a root enabled image.

netanel-innovid:
In the meantime I resolved it by adding `--user root` to the run command.

Issue #2721, opened 2025-06-10
Nonroot can't load local docker image with docker socket bind


jtougas:
**What happened**:
Grype fails to find local container image.

**What you expected to happen**:
Grype finds local container image

**How to reproduce it (as minimally and precisely as possible)**:
Run grype from docker, passing it the docker socket (as in https://github.com/anchore/grype?tab=readme-ov-file#getting-started). Get it to scan an image that is available only locally.

**Anything else we need to know?**:
This just started happening after the 0.93 release a couple hours ago. I can see that some changes were made to use a non root user (https://github.com/anchore/grype/pull/2716). I suspect this is the source of the problem, grype no longer has access to the docker socket.

Pinned to 0.92 for now as a workaround.

**Environment**:
- grype v0.93.0
- github actions "runs-on: ubuntu-22.04"

flori:
I can verify that this still works in grype **v0.92.2**.

kzantow:
Hi folks 👋  -- we've reverted the nonroot change and the latest image should be "back to normal" and introduced a "nonroot" tag for users who are able to run the nonroot variant. Please let us know if you continue to have any issues!

flori:
I just tried the newest release **v0.94.0** with a local container image and it works again. 🎉

mundry:
Hi,

thanks for fixing the issue so quickly.

Brief follow up question:
Was this change supposed to also apply to the debug image? The debug image is still using the nonroot distroless base image and therefore defaults to the nonroot user.

For background: I need to copy a custom CA certificate into the image to be able to pull images from a private registry in a CI environment.

kzantow:
@mundry I don't think the debug image needs to be nonroot, it was an oversight 🤦 I was just trying to get the default updated to help avoid breaking people's CI workflows from the latest change. You probably shouldn't be using debug for anything other than actually debugging, though. Why do you need debug image to use a custom CA cert?

schweppestronic:
I'm trying with 0.103.0 and it does not work. Error message is the same:

[0001] ERROR failed to catalog: errors occurred attempting to resolve '[MASKED].azurecr.io/****:****':
  - snap: snap file "[MASKED].azurecr.io/****:****" does not exist
  - docker: docker not available: failed to connect to Docker daemon. Ensure Docker is running and accessible
  - podman: podman not available: no host address
  - containerd: containerd not available: no grpc connection or services is available: unavailable
[...]

PR #2723, opened 2025-06-11
fix: provide separate nonroot image

kzantow:
This PR reverts a change to use `nonroot` images by default, and provides an alternate `nonroot` set of tagged images. Using `nonroot` causes a number of issues causing users friction such as accessing and modifying the `docker.sock` and mount volumes.

NOTE: this will _remove_ the `:latest-<arch>` tags, which I think shouldn't be used since we are providing multi-platform manifests, where an architecture can be selected using `--platform`, e.g. `docker run --rm -it anchore/syft:latest --platform linux/amd64`

Fixes: #2720, #2721

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -93,70 +93,9 @@ brews:
     license: "Apache License 2.0"
 
 dockers:
+  # production images...
   - image_templates:
-      - anchore/grype:debug-amd64
-      - anchore/grype:{{.Tag}}-debug-amd64
-      - ghcr.io/anchore/grype:debug-amd64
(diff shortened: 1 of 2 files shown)


.goreleaser.yaml
```diff
@@ -215,12 +260,12 @@ docker_manifests:
       - anchore/grype:{{.Tag}}-ppc64le
       - anchore/grype:{{.Tag}}-s390x
 
-  - name_template: anchore/grype:debug
+  - name_template: ghcr.io/anchore/grype:latest
```
kzantow, line 263: I reorganized this section to be similar to the order of the section above / and the analogous syft pr: https://github.com/anchore/syft/pull/3998

PR #2724, opened 2025-06-11
Configurable timestamp

### LINKED ISSUE #522: Make timestamp in output configurable (so that results are more reproducible)

gabetrau:
Allows for configuring whether or not to include a timestamp in the final output via config option or env variable. Default is true. Resolves #522

### DIFF
--- a/grype/presenter/models/document.go
+++ b/grype/presenter/models/document.go
@@ -2,7 +2,6 @@ package models
 
 import (
 	"fmt"
-	"os"
 	"time"
 
 	"github.com/anchore/clio"
@@ -25,7 +24,7 @@ type Document struct {
 func NewDocument(id clio.Identification, packages []pkg.Package, context pkg.Context, matches match.Matches, ignoredMatches []match.IgnoredMatch, metadataProvider vulnerability.MetadataProvider, appConfig any, dbInfo any, strategy SortStrategy, outputTimestamp bool) (Document, error) {
 	var timestamp []byte
 
-	if !outputTimestamp || os.Getenv("GRYPE_TIMESTAMP") == "false" {
+	if !outputTimestamp {
 		// can't be nil in string() call
 		timestamp = []byte{}
 	} else {
--- a/grype/presenter/models/document.go
+++ b/grype/presenter/models/document.go
@@ -3,6 +3,7 @@ package models
 import (
 	"fmt"
 	"time"
+	"os"
 
 	"github.com/anchore/clio"
 	"github.com/anchore/grype/grype/distro"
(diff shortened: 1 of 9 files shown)


grype/presenter/models/document.go
```diff
@@ -20,10 +21,18 @@ type Document struct {
...
 
 // NewDocument creates and populates a new Document struct, representing the populated JSON document.
-func NewDocument(id clio.Identification, packages []pkg.Package, context pkg.Context, matches match.Matches, ignoredMatches []match.IgnoredMatch, metadataProvider vulnerability.MetadataProvider, appConfig any, dbInfo any, strategy SortStrategy) (Document, error) {
-	timestamp, timestampErr := time.Now().Local().MarshalText()
-	if timestampErr != nil {
-		return Document{}, timestampErr
+func NewDocument(id clio.Identification, packages []pkg.Package, context pkg.Context, matches match.Matches, ignoredMatches []match.IgnoredMatch, metadataProvider vulnerability.MetadataProvider, appConfig any, dbInfo any, strategy SortStrategy, outputTimestamp bool) (Document, error) {
+	var timestamp []byte
+
+	if !outputTimestamp || os.Getenv("GRYPE_TIMESTAMP") == "false" {
```
wagoodman: All environment vars should be resolved up in the `cmd` package and not down in the library (it seems like this is already accounted for).
```suggestion
	if !outputTimestamp {
```

PR #2731, opened 2025-06-16
fix: properly escape CPE segments

kzantow:
This PR fixes an issue with JSON output where CPE segments are not properly escaped in the same way Syft outputs escaped CPE segments.

### DIFF
--- a/grype/presenter/internal/test_helpers.go
+++ b/grype/presenter/internal/test_helpers.go
@@ -390,8 +390,8 @@ func generatePackages(t *testing.T) []syftPkg.Package {
 				{
 					Attributes: cpe.Attributes{
 						Part:     "a",
-						Vendor:   "ancho:re",
-						Product:  "en/gine",
+						Vendor:   "anchore:oss",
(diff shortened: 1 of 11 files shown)


grype/presenter/internal/test_helpers.go
```diff
@@ -390,8 +390,8 @@ func generatePackages(t *testing.T) []syftPkg.Package {
 				{
 					Attributes: cpe.Attributes{
 						Part:     "a",
-						Vendor:   "anchore",
-						Product:  "engine",
+						Vendor:   "ancho:re",
```
willmurphyscode: I wonder if this test case could be a little clearer? having a vendor called `ancho:re` just looks weird. Maybe the vendor could be like `anchore:oss` or something so it looks plausible that there's meant to be a `:` in there? Not blocking feedback but it would make some of the test cases look a lot less weird/broken.

kzantow: Updated to make prettier test values.

willmurphyscode (approved):

PR #2735, opened 2025-06-16
Refactor the version package

wagoodman:
While I was working on adding RHEL EUS support, I ran across some inconsistencies when attempting to use the `grype/version.semanticVersion`, and when digging deeper, it was apparent that most version structs in the package had similar issues:

- the operand ordering semantics were inconsistent; some comparators returned results in terms of `v1 vs v2` and others did `v2 vs v1`. Eventually, based on the values of `versionObj.Format`, the underlying object structs, and the implementation within each `Compare()` function, the results were correct, but hard to reason which ordering was correct.
- each version object needs to invoke `finalizeComparisonVersion` within their `Compare` method, which can easily be overlooked. Ideally this should be a single choke point in this package where this is done since all of the underlying types are not exported.
- decisions are being driven off of the `versionObj.Format` field deep within each type during comparison even though its the underlying state of the `rich` data structure that really drives this. Since there was no single source of truth on this, different implementations would choose which truth to lean on when making decisions, which made checking which behaviors were right more difficult.

This lead to refactoring much of the `grype/version` package; here are the summary of changes:

- Removes the `rich` struct in favor of type asserting specific `version.comparator` implementations.
- Inverts the value of the comparator results within constraint expression processing; this allows comparators to clearly do a `v1 vs v2` comparison (instead of needing to flip this).
- Migrates all `finalizeComparisonVersion` logic to be done only in `version.Version.Compare()` (not within each version object implementation).
- Removes most of the constraint implementations and uses the `genericConstraint` object (less code).
- All testing now uses `versions.NewVersion(value, format)` so that we are getting a sense of the objects behavior as seen by callers of this package.


internal/ordered_set.go
```diff
@@ -0,0 +1,105 @@
+package internal
+
+import "fmt"
+
+// OrderedSet maintains insertion order while ensuring uniqueness
+type OrderedSet[T comparable] struct {
```
kzantow: After the changes, I think this one is no longer needed? It seems like a good utility for `anchore/go-collections`, though 👍

wagoodman: correct! I'll remove

wagoodman: https://github.com/anchore/go-collections/pull/2

kzantow (approved):
LGTM 👍

PR #2742, opened 2025-06-18
Add package specifier overrides for `kb`, `dpkg`, and `apkg`

westonsteimel:


.github/workflows/validations.yaml
```diff
@@ -280,6 +280,6 @@ jobs:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 #v4.2.2
 
       - name: Delete snapshot cache
-        run: gh cache delete "snapshot-build-${{ github.run_id }}"
+        run: gh cache delete "snapshot-build-${{ github.run_id }}" || echo "Cache deletion failed or cache not found - continuing"
```
wagoodman: Screenshot 2025-06-18 at 12 50 47 PM

ran into issues in CI, however, this should not have failed the build, and it's ok if these caches are not cleaned up occasionally as it's an LRU cache

wagoodman (approved):

Issue #2745, opened 2025-06-18
Use Alma Linux specific advisories for Alma Linux scans


willmurphyscode:
**What would you like to be added**:

Grype should consider Alma Linux fix information specifically when scanning Alma Linux images. (Today Grype treats Alma images identically to RHEL images).

**Why is this needed**:

Even though these distros are very similar by design, the actual fix versions of some packages, especially modules, are different between Alma and RHEL, resulting in some FPs when matching Alma.

This is particularly problematic when modules are involved, because the version strings for modules are different between the distros. Consider for example 

* [ALSA-2025:8506](https://errata.almalinux.org/8/ALSA-2025-8506.html) has `npm-10.9.2-1.22.16.0.1.module_el8.10.0+4006+3c416519.aarch64.rpm`, with `+4006` in the version number.
* [RHSA-2025:8506](https://access.redhat.com/errata/RHSA-2025:8506) has `npm-10.9.2-1.22.16.0.1.module+el8.10.0+23140+4056b950.aarch64.rpm` with `+23140` in the version number.

This causes the fixed version of the Alma Linux package to appear vulnerable, resulting in false positives. Grype should fix this by using Alma Linux specific fix information when scanning Alma Linux images.

willmurphyscode:
Implemented by:

- https://github.com/anchore/grype/pull/2939
- https://github.com/anchore/vunnel/pull/886
- https://github.com/anchore/grype-db/pull/686
- https://github.com/anchore/grype-db/pull/726 will turn this on

PR #2750, opened 2025-06-23
remove alma-rhel alias

willmurphyscode:
In preparation for having Alma Linux specific data, drop the aliases that cause Grype to treat Alma identically to RHEL. Part of https://github.com/anchore/grype/issues/2745, related to https://github.com/anchore/vunnel/pull/826 and https://github.com/anchore/grype-db/pull/592


willmurphyscode:
moving this back to draft because we want to something more like https://github.com/anchore/grype-db/pull/592/files

willmurphyscode:
Closing in favor of https://github.com/anchore/grype/pull/2939/files which allows searching the OS table both with and without aliases.

PR #2754, opened 2025-06-25
feat: enable CSV output

willmurphyscode:
For now, CSV output is the same as table output, but with commas.


grype/presenter/csv/presenter.go
wagoodman, line 1: this is mostly a copy-paste of the table presenter. Are the display semantics similar enough to make the table presenter optionally output CSV?

grype/presenter/csv/presenter.go
```diff
@@ -0,0 +1,311 @@
...
+	if characterCount > maxVersionFieldLength && len(allVersions) > 1 {
+		finalVersions += ", ..."
+	}
+
+	return finalVersions
+}
+
+func (r row) Columns() []string {
+	columns := []string{
+		stripANSI(r.Name),
```
wagoodman: why call the format functions then strip ansi characters? We could not call the format functions at all. Or have an alternative style set that is structural only without any ANSI control characters.

grype/presenter/csv/presenter.go
willmurphyscode: I don't know that they are. For example, we want to pretty print EPSS percents and percentiles, but in the CSV we probably want them to be separate columns and printed as bare floats?

PR #2759, opened 2025-06-30
fix: assume that a vulnerability with no ranges is always vulnerable

wagoodman:
Today it is possible to have NVD entries that have no range information for a CVE, such as with [CVE-2010-4756](https://nvd.nist.gov/vuln/detail/CVE-2010-4756), however, the correct interpretation of this is that all versions are vulnerable. This looks like it was introduced in the v6 development effort (based on the data shape changes I'm seeing, this class of bugs makes sense) and it led to a reduction in FPs with the exception for one finding (which is why it was merged).


kzantow (approved):
Makes sense to me -- I can't see why we'd have entries that we assume the opposite 👍

westonsteimel (approved):

wagoodman:
bypassing quality gate -- it is not expected to pass since fixing this bug necessarily means adding FPs to known results. @westonsteimel and I have verified the results are what we'd expect.

PR #2763, opened 2025-07-01
chore: add version set and combined constraint

wagoodman:
In order to make #2540 smaller, this PR is adding only the following features:
- A `version.Set` type as a format or formatless collection of version objects
- The ability to combine multiple constraints into a single constraint object
- Exposes the underlying string value and format from a constraint


grype/version/generic_constraint.go
```diff
@@ -8,9 +8,9 @@ import (
 var _ Constraint = (*genericConstraint)(nil)
 
 type genericConstraint struct {
-	raw        string
-	expression simpleRangeExpression
-	format     Format
+	Raw        string
+	Expression simpleRangeExpression
+	Fmt        Format
```
kzantow, line 13: nit: do these values need to be exported since a user can just use the interface methods? and then `fmt` could remain `format`?

grype/version/range.go
```diff
@@ -105,7 +105,7 @@ func validateVersion(version string) error {
 			quoteChar = 0
 		case !inQuotes && strings.ContainsRune("><=", r):
 			// invalid character outside of quotes
-			return fmt.Errorf("version %q potentially is a version constraint expression (should not contain '><=' outside of quotes)", version)
+			return fmt.Errorf("version %q potentially is a version constraint Expression (should not contain '><=' outside of quotes)", version)
```
kzantow: nit: a number of these refactors affected strings incorrectly (Expression & Operator, at least), such as here

grype/version/fuzzy_constraint.go
```diff
@@ -14,18 +14,18 @@ import (
 var pseudoSemverPattern = regexp.MustCompile(`^v?(0|[1-9]\d*)(\.(0|[1-9]\d*))?(\.(0|[1-9]\d*))?(?:(-|alpha|beta|rc)((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$`)
 
 type fuzzyConstraint struct {
-	rawPhrase          string
-	phraseHint         string
-	semanticConstraint *hashiVer.Constraints
-	constraints        simpleRangeExpression
+	RawPhrase          string
```
kzantow, line 17: nit: do these need to be exported?

kzantow (approved):
Only small thing I was a little confused by why both export fields and have interface methods to retrieve the same data?

grype/version/generic_constraint.go
wagoodman: I'm with you on the naming (`format` is better than `fmt`). The reason for the change is to help out with cmp.Diff in testing without needing compare options. Since the structs themselves are not exported this seemed ok to do.

grype/version/fuzzy_constraint.go
wagoodman: same reason as https://github.com/anchore/grype/pull/2763#discussion_r2180738214

Issue #2446, opened 2025-02-13
CVE-2024-53104 (Redhat 9.4)


pkeecom:
**What happened**: Redhat has issued a new kernel 5.14.0-427.50.2.el9_4 which fixes the CVE

**What you expected to happen**: Grype should not report the issue as High on the fixed kernel

**How to reproduce it (as minimally and precisely as possible)**: Install Kernel [kernel-headers-5.14.0-427.50.2.el9_4.x86_64.rpm]

**Anything else we need to know?**:
kernel                      5.14.0-427.50.2.el9_4  0:5.14.0-503.23.2.el9_5  rpm     CVE-2024-53104       High      
kernel-core                 5.14.0-427.50.2.el9_4  0:5.14.0-503.23.2.el9_5  rpm     CVE-2024-53104       High      
kernel-modules              5.14.0-427.50.2.el9_4  0:5.14.0-503.23.2.el9_5  rpm     CVE-2024-53104       High      
kernel-modules-core         5.14.0-427.50.2.el9_4  0:5.14.0-503.23.2.el9_5  rpm     CVE-2024-53104       High      
kernel-tools                5.14.0-427.50.2.el9_4  0:5.14.0-503.23.2.el9_5  rpm     CVE-2024-53104       High      
kernel-tools-libs           5.14.0-427.50.2.el9_4  0:5.14.0-503.23.2.el9_5  rpm     CVE-2024-53104       High      
kernel-uki-virt             5.14.0-427.50.2.el9_4  0:5.14.0-503.23.2.el9_5  rpm     CVE-2024-53104       High      
python3-perf                5.14.0-427.50.2.el9_4  0:5.14.0-503.23.2.el9_5  rpm     CVE-2024-53104       High      
High or Critical vulnerabilities exist in grype report!

**Environment**:
- Output of `grype version`: Latest Available
- OS (e.g: `cat /etc/os-release` or similar): RHEL 9.4

pkeecom:
The package was delivered vis Redhat extended update support EUS.

![Image](https://github.com/user-attachments/assets/73176238-1919-4949-b0bb-d6be68aeb866)
![Image](https://github.com/user-attachments/assets/d02f646a-0372-44bf-83ca-c52a1f0e977b)

willmurphyscode:
Hi @pkeecom thanks for the issue! I'll do some digging and get back to you. 

[...]

PR #2782, opened 2025-07-05
chore: add channel to os / distro

wagoodman:
This PR makes the following changes:
- Adds `distro.FixChannel` to be able to specify a channel name that can be applied to one or more distro release IDs within a given version constraint. This will later be used to be able to specify when specific RedHat distros should be using `EUS` (extended support) data when matching packages with vulnerabilities.
- Migrates processing distro override via the `--distro` flag from the `cmd` package to the `pkg` package, within the package provider (as a `pkg.Enhancer` like we do with other distro concerns within `pkg`). This further centralizes distro concerns to be that of the package processing in general (thus in the long term truly belong to `pkg.Package`, and de-leverages using the distro from `pkg.Context`).
- Migrates the `version.NewVersionFromPkg` helper function to `pkg.VersionFormat` (isolating only getting the version format) --this is to prevent a package cycle when using version object within the `distro` package. This also makes a little more sense: a distro has a version, and a package has a version, but a version does not have a distro or package, so the version is a more fundamental package (thus should not depend on either of the other two).
- Changes `version.NewVersion(...)` to `version.New(...)`
- Splits up the distro type IDs from aliases, so that the existing `IDMapping` is purely for IDs (and a new mapping is only for aliases). This makes adding `ID()` to the distro object easier to implement now (leaning toward release IDs over bespoke types as is done now).

This is work broken off from #2540 and partially addresses #2446.

## Work deferred for another PR / for the future
- Introducing an application configuration for distribution fix channels; this will be added in the last PR to complete #2446
- Fully removing the distro object from `pkg.Context` (probably will be part of the grype 1.0 work)

## TODO
- [x] CI will fail until anchore/syft#4023 is merged and incorporated here


grype/distro/fix_channel.go
```diff
@@ -0,0 +1,100 @@
...
+
+// FixChannel represents a subscription or repository where package fixes and updates are provided for a Linux distribution
+type FixChannel struct {
+	// Name is the name of the channel, e.g. "eus" for RHEL
+	Name string
+
+	// IDs is a list of distro release IDs that this channel applies to, e.g. "rhel" for RHEL (this is relative to the /etc/os-release ID field)
+	IDs []string
+
+	// Apply indicates how the channel should be applied to the distrp
```
kzantow: nit: `distrp` -> `distro`

cmd/grype/cli/commands/root.go
```diff
@@ -228,6 +229,23 @@ func runGrype(app clio.Application, opts *options.Grype, userInput string) (errs
 	return errs
 }
 
+func warnWhenDistroHintNeeded(pkgs []pkg.Package, context *pkg.Context) {
+	hasOSPackageWithoutDistro := false
+	for _, p := range pkgs {
+		switch p.Type {
+		case syftPkg.AlpmPkg, syftPkg.DebPkg, syftPkg.RpmPkg, syftPkg.KbPkg:
+			if p.Distro == nil {
+				hasOSPackageWithoutDistro = true
```
kzantow, line 258: nit: `break` loop (I see now that this looks to be a copy/paste, but may as well add a break while refactoring)

grype/presenter/models/document.go
```diff
@@ -76,7 +77,7 @@ func NewDocument(id clio.Identification, packages []pkg.Package, context pkg.Con
 		Matches:        findings,
 		IgnoredMatches: ignoredMatchModels,
 		Source:         src,
-		Distro:         newDistribution(context.Distro),
+		Distro:         newDistribution(context, selectDistro(packages)),
```
wagoodman: need to include channel here https://github.com/anchore/grype/blob/6f8284b6a8b4b01766b29f2bcc32a0f94ff4cdfa/grype/presenter/models/distribution.go#L9-L13

grype/distro/distro.go
```diff
@@ -52,16 +37,50 @@ func New(t Type, version, label string, idLikes ...string) *Distro {
...
+			if len(parts) > 1 {
+				minor = parts[1]
+			}
+			if len(parts) > 2 {
+				remaining = strings.Join(parts[2:], ".")
+			}
+		}
+	}
+
+	return major, minor, remaining, versionWithoutSuffix, stringutil.SplitOnAny(strings.TrimSpace(channelStr), ",", "+")
```
kzantow, line 81: Since there's a separate configuration for applying channels, should that still have special handling in the version string here? I am still under the impression that channels are not the same thing as "a distro", but rather a set of fixes available for a distro. In other words, a user may want to enable multiple channels for a distro, which they can do by setting the auto setting or by explicitly configuring multiple channels to be enabled / auto, but setting a singular channel here would conflict with that, potentially, and probably be more confusing than just leaving the default 'enabled' or 'auto' setting for everything.

I think this actually is due to the search command and from that perspective makes plenty of sense. But what about multiple channels? What happens if I specify `--distro rhel:8+EEUS` for a _scan_, does this also have the affect of disabling the other channels the same way as the configuration does?

I think there's a little bit of an impedance mismatch between channels here and channel behavior configuration.

grype/distro/fix_channel.go
```diff
@@ -0,0 +1,123 @@
...
+	Apply FixChannelEnabled
+
+	// Versions is a version constraint that indicates which versions of the distro this channel applies to (e.g. ">= 8.0" for RHEL 8 and above)
+	Versions version.Constraint
+}
+
+type FixChannels []FixChannel
```
kzantow, line 39: Should this be a
```
type FixChannels struct {
  Apply FixChannelEnabled
  Channels map[string]FixChannel
}
```
... I don't remember exactly where our discussion ended, but I'll point out that if we were to make this data-driven, by requiring all channels get passed in to the configuration, we would have to execute a query to get the results for `DefaultFixChannels` (needs reference to the db, etc.). For now, the `.Apply` function serves to give us the "Apply everything auto by default", but I still think it would be useful to do this at the API configuration level. I don't want this to block anything, however, as it can always be changed later, before 1.0

grype/distro/type.go
```diff
@@ -90,10 +88,30 @@ var IDMapping = map[string]Type{
 	"minimos":       MinimOS,
 }
 
+// aliasTypes maps common aliases to their corresponding Type.
+var aliasTypes = map[string]Type{
+	"Alpine Linux": Alpine, // needed for CPE matching (see #2039)
+	"windows":      Windows,
```
kzantow, line 94: I understand why `Alpine Linux` is here, but why `windows`, I don't see a primary windows mapping?

grype/presenter/models/document.go
```diff
@@ -86,3 +87,30 @@ func NewDocument(id clio.Identification, packages []pkg.Package, context pkg.Con
 		},
 	}, nil
 }
+
+// selectDistro selects the most common distro from the provided packages.
+func selectDistro(pkgs []pkg.Package) *distro.Distro {
```
kzantow: nit: something like `selectMostCommonDistro` could be slightly more clear (thought it might be something else until I read the description)

grype/presenter/json/presenter_test.go
```diff
@@ -20,7 +20,7 @@ import (
 	"github.com/anchore/syft/syft/source"
 )
 
-var update = flag.Bool("update", false, "update the *.golden files for json presenters")
+var update = flag.Bool("update", true, "update the *.golden files for json presenters")
```
kzantow: This should be not be updated here

grype/distro/distro.go
```diff
@@ -4,17 +4,18 @@ import (
...
 )
 
 // Distro represents a Linux Distribution.
 type Distro struct {
 	Type     Type
-	Version  string
-	Codename string
+	Version  string   // major.minor.patch
+	Codename string   // in lieu of a version e.g. "fossa" instead of "20.04"
+	Channels []string // distinguish between different feeds for fix and vulnerability data, e.g. "eus" for RHEL
```
kzantow, line 18: I've flip-flopped on this field, but I want to point out that if I see this set in some JSON output, I would assume this distro _is_ the determined distro which includes the added channel. So, for example, if I have RHEL 8.0 EUS, and I see `eus` as a channel, this makes a lot of sense to me. However, if I don't have `EEUS`, but I start to see both `EUS` and `EEUS` because this is the way that we are passing  _configuration_ to the matcher, I think people will start to get confused.

grype/pkg/purl_provider.go
```diff
@@ -44,3 +52,57 @@ func getPurlReader(userInput string) (r io.Reader, ctx Context, err error) {
...
+	return upstreams
+}
+
+func setDistroFromPURL(applyChannel func(*distro.Distro) bool) func(out *Package, purl packageurl.PackageURL, _ syftPkg.Package) {
+	return func(out *Package, purl packageurl.PackageURL, _ syftPkg.Package) {
+		if out.Distro == nil {
+			out.Distro = distroFromPURL(purl)
+			if applyChannel(out.Distro) {
+				// the distro was modified by the channel, so we need to update the purl
+				updatePURLWithDistro(out, purl)
```
kzantow: I didn't notice a small detail here when talking through this change last week. This ends up making me lean in the direction that maybe it would be better _not_ to update the per-package distro objects with channels. But instead use a default apply with per-channel overrides, and iterate over the channels directly in the matcher via configuration or simply querying the db to get applicable channels, and test if they are enabled based on `FixChannels` configuration

I think this is going to update a value unexpectedly for end users. If a user supplies an SBOM without `+eus`, and there is a configuration including `eus`, this is going to be added -- but potentially other channels like `+eeus`. I don't think this is expected behavior and users will end up being confused why they started to see modifications to the PURL distro attribute

grype/distro/type.go
wagoodman: windows isn't strictly a valid linux ID (there are no examples of an /etc/os-release  ID with this value... which makes sense), so I kept it only in the alias map

grype/distro/distro.go
westonsteimel: Any channels that were possibly consulted need to be raised here because the configuration can be conditional and we need to know if it may have applied for a particular report or not.  We can't just rely on the configuration because that would mean needing to re-evaluate the conditions outside of grype which we absolutely want to avoid

kzantow: If this information was in the match details, wouldn't that be better -- indicating that we _did_ search EUS data and found a match or identified that something is not yet fixed?

grype/distro/distro.go
wagoodman: > channels are not the same thing as "a distro", but rather a set of fixes available for a distro

yup!

> a user may want to enable multiple channels for a distro...but setting a singular channel here would conflict with that

I don't follow -- the terms "auto" and "enabled" are used somewhat interchangeably in the comment, but they do different things:

- auto: trust the component evidence completely - use the channel if evidence shows it, don't use it if no evidence
- enabled: **always** apply the channel regardless of evidence (override to force inclusion)
- disabled: **never** apply the channel regardless of evidence (override to force exclusion)

<details>
<summary>(see possible states)</summary>

So in the case of one channel the truth table looks like so:

| User Config | Component Evidence | Output (Channel Applied?) |
|-------------|-------------------|---------------------------|
| auto        | yes, channel X    | yes, channel X           |
| auto        | no channel        | no channel               |
| enabled     | yes, channel X    | yes, channel X           |
| enabled     | no channel        | yes, channel X           |
| disabled    | yes, channel X    | no channel               |
| disabled    | no channel        | no channel               |


When you expand this to two channels...:

| EUS Config | EEUS Config | Component Evidence | Output Channels Applied |
|------------|-------------|-------------------|------------------------|
| auto       | auto        | no channels       | -                      |
| auto       | auto        | EUS only          | EUS                    |
| auto       | auto        | EEUS only         | EEUS                   |
| auto       | auto        | EUS + EEUS        | EUS + EEUS             |
| auto       | enabled     | no channels       | EEUS                   |
| auto       | enabled     | EUS only          | EUS + EEUS             |
| auto       | enabled     | EEUS only         | EEUS                   |
| auto       | enabled     | EUS + EEUS        | EUS + EEUS             |
| auto       | disabled    | no channels       | -                      |
| auto       | disabled    | EUS only          | EUS                    |
| auto       | disabled    | EEUS only         | -                      |
| auto       | disabled    | EUS + EEUS        | EUS                    |
| enabled    | auto        | no channels       | EUS                    |
| enabled    | auto        | EUS only          | EUS                    |
| enabled    | auto        | EEUS only         | EUS + EEUS             |
| enabled    | auto        | EUS + EEUS        | EUS + EEUS             |
| enabled    | enabled     | no channels       | EUS + EEUS             |
| enabled    | enabled     | EUS only          | EUS + EEUS             |
| enabled    | enabled     | EEUS only         | EUS + EEUS             |
| enabled    | enabled     | EUS + EEUS        | EUS + EEUS             |
| enabled    | disabled    | no channels       | EUS                    |
| enabled    | disabled    | EUS only          | EUS                    |
| enabled    | disabled    | EEUS only         | EUS                    |
| enabled    | disabled    | EUS + EEUS        | EUS                    |
| disabled   | auto        | no channels       | -                      |
| disabled   | auto        | EUS only          | -                      |
| disabled   | auto        | EEUS only         | EEUS                   |
| disabled   | auto        | EUS + EEUS        | EEUS                   |
| disabled   | enabled     | no channels       | EEUS                   |
| disabled   | enabled     | EUS only          | EEUS                   |
| disabled   | enabled     | EEUS only         | EEUS                   |
| disabled   | enabled     | EUS + EEUS        | EEUS                   |
| disabled   | disabled    | no channels       | -                      |
| disabled   | disabled    | EUS only          | -                      |
| disabled   | disabled    | EEUS only         | -                      |
| disabled   | disabled    | EUS + EEUS        | -                      |
</details>

Where "evidence" can come from:
- SBOM distro block
- PURL qualifier

So...

> What happens if I specify `--distro rhel:8+EEUS` for a scan, does this also have the affect of disabling the other channels the same way as the configuration does?

The `--distro` flag is different from component evidence; this is an override from the user saying "use this exact distro as I've provided it (with or without channel info)". So if EUS information is not in the `--distro` string then it is not applied, no matter the configuration.

grype/pkg/purl_provider.go
westonsteimel: No, because these are going to cause matches to go away, not to appear

grype/distro/fix_channel.go
wagoodman: yeah, we'll defer this issue to the grype 1.0 work (the top level API needs to get designed here)

grype/distro/distro.go
westonsteimel: No, because these are going to cause matches to go away, not to appear.  In other words, EUS is suppressing FPs, not surfacing new matches

grype/pkg/purl_provider.go
wagoodman: I think either way will be confusing: either we're surprising folks by patching the PURL on a package or we're surprising folks by having different answers on the PURL vs the json `.distro` block section (which is always patched). 

I'll remove PURL patching for now.

kzantow: I'm not saying it will result in more matches, but some of the matches will still show up and those matches will have `+eus` or `+eus,eeus` or whatever _in the PURL_, where it didn't have those in the SBOM, which I don't think is what end users will be expecting.

grype/distro/distro.go
westonsteimel: We need to be able to show that EUS matches were considered and therefore some matches that previously appeared will go away.  Doing it any other way would need the concept of negative matches to be implemented and that is a much larger change

kzantow: I don't quite understand. If we patch this object, it's going to show up in the result JSON at the end.

grype/pkg/purl_provider.go
wagoodman: I wasn't trying to say the set of matches would change at all, just the PURL qualifiers would or would not be consistent with the grype `.distro` block. No matter though, since I believe [I've incorporated your ask here](https://github.com/anchore/grype/pull/2782/commits/7f3aaa912fce813c4d15210bfcecee7b83c49a3e): don't patch PURLs 👍

grype/distro/distro.go
westonsteimel: That is what we want

westonsteimel: EUS causes matches to be dropped from the report, you can't annotate a match that doesn't exist in the output

westonsteimel: You need to be able to surface the fact that EUS channels were queried during the scan

kzantow: But the configuration already exists. Also, we could potentially have RHEL:8 and RHEL:9 and the configuration says > 8, so some of the results are filtered by EUS and others aren't, and we only get one of those Distro objects in the final report. I just don't understand how this is better than the configuration.

westonsteimel: The configuration is conditional on the distro though!

westonsteimel: So it may or may not apply depending on the conditions that were evaluated at the time of the scan

westonsteimel: For instance, they may only want EUS to be used for a specific set of minor releases of a distro, we do NOT want to be re-evaluating all of that junk outside of grype to know whether or not EUS data may have applied

kzantow: But [there is only a single distro object output in the final grype report](https://github.com/anchore/grype/pull/2782/files#diff-17bba95fb96250ca676045ca7fb738954f50d05aa2e428490ceb02182b803a3dR80). Let's say we filtered RHEL:8 with EUS but not 9, and we had more RHEL 9 packages. We would get the distro object for RHEL 9 with no indication of EUS, despite filtering some of the results for the RHEL 8 packages based on EUS

westonsteimel: we can't represent packages across multiple distros anyways with the current grype report, much less detect them properly with syft

grype/db/v6/data.go
```diff
@@ -12,15 +12,31 @@ func KnownOperatingSystemSpecifierOverrides() []OperatingSystemSpecifierOverride
...
-		{Alias: "alpine", VersionPattern: ".*_alpha.*", ReplacementLabelVersion: strRef("edge"), Rolling: true},
+
+		// to remain backwards compatible, we need to keep old clients from ignoring EUS data.
+		// we do this by diverting any requests for a specific major.minor version of rhel to only
+		// use the major version. But, this only applies to clients before v6.0.3 DB schema version.
+		// Why 6.0.3? This is when OS channel was introduced, which grype-db will leverage, and add additional
+		// rhel rows to the DB, all which have major.minor versions. This means that any old client (which wont
+		// see the new channel column) will assume during OS resolution that there is major.minor vuln data
+		// that should be used (which is incorrect).
+		{Alias: "rhel", VersionPattern: `^\d+.\d+`, ReplacementMinorVersion: strRef(""), ApplicableClientDBSchemas: "< 6.0.3"},
```
kzantow: Should this be `^\d+\.\d+`?  note the escaping of the dot (same for below)... to account for the unlikely event it would match `123`, for example.

wagoodman: good catch!

PR #2787, opened 2025-07-07
Consider RedHat EUS fix information when matching

### LINKED ISSUE #2446: CVE-2024-53104 (Redhat 9.4)

wagoodman:
This adds support for considering Extended Updated Support (EUS) fix information for RedHat distros when matching. 

Closes #2446

## Changes

- introduces a new `result.Result` object for being able to organize returned vulnerability objects from a provider into pseudo prototype matches, allowing for merging of sets with control over how the merge is done. This enables more flexibility for matchers to combine results (as is needed with EUS)
- splits the RPM matcher into two paths based on the distro channel information: standard matching and EUS matching. 

## TODO
- [x] https://github.com/anchore/vunnel/pull/796 needs to be merged, released, and incorporated into nightly builds
- [x] https://github.com/anchore/grype-db/pull/540 needs to be merged, released, and incorporated into nightly builds


cmd/grype/cli/options/fix_channels.go
```diff
@@ -0,0 +1,51 @@
...
+type FixChannels struct {
+	// EUS is the Extended Update Support channel for RHEL
+	RedHatEUS FixChannel `yaml:"redhat-eus" json:"redhat-eus" mapstructure:"redhat-eus"`
+}
+
+type FixChannel struct {
+	IDs      []string                 `yaml:"ids" json:"ids" mapstructure:"ids"`                // the list of distro release IDs that this channel applies to
+	Apply    distro.FixChannelEnabled `yaml:"apply" json:"apply" mapstructure:"apply"`          // whether this channel should be applied to the distro when matching
+	Versions string                   `yaml:"versions" json:"versions" mapstructure:"versions"` // the version constraint that this channel applies to, e.g. ">= 8.0"
+}
```
wagoodman: note to self: we should probably remove IDs and Versions from the app config -- these will be lower level API configurations, but not available to the CLI. This way we have a chance to make this more data driven later

wagoodman: I've removed IDs and Versions for now from the user config, but it still remains in the API config

cmd/grype/cli/options/fix_channels.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/anchore/grype/grype/distro"
+)
+
+type FixChannelEnabled string
+
+type FixChannels struct {
+	// TODO: in the future we may want to support more channels, as well as have a default-apply value here that can be overridden within each channel configuration
+
+	// EUS is the Extended Update Support channel for RHEL
+	RedHatEUS FixChannel `yaml:"redhat-eus" json:"redhat-eus" mapstructure:"redhat-eus"`
```
kzantow, line 17: I wonder if this would be easy to make a map by channel name, e.g.:
```
fix-channels:
  apply: none
  channels:
    eus:
      apply: auto
      version: > 9
```

Where this configuration is more-or-less what gets sent to the matchers, e.g.:
```
type FixChannels struct {
  Apply FixChannelEnabled
  Channels map[string]FixChannel
}
```
... in some future state where a matcher does something like:
```
for channel in select_applicable_channels() {
  if fixChannels.enabled(channel) {
    // select some things from the special channel records
  }
}
```
... or should the initial configuration not even have redhat, only a global apply option?

grype/matcher/rpm/rhel_eus.go
```diff
@@ -0,0 +1,320 @@
...
+// 1. find disclosures that match the base distro (e.g., '>= 9.0 && < 10').
+// 2. find fixes from the base distro (e.g., '>= 9.0 && < 10') as well as EUS fixes for the specific minor version of the distro (e.g. '9.4+eus').
+//
+// Once searching is complete, we have two collections (matching for each search step above).
+// We then merge these two (disclosure and resolution) collections together, the final result is a collection of
+// prototype matches that the package is vulnerable to that include both the base distro disclosures and the EUS fixes.
+// Any disclosure that does not apply to the original package version (e.g. a fix was found) at this point has been removed.
+//
+// The final step is to render the final matches from the merged collection.
+func redhatEUSMatches(provider result.Provider, searchPkg pkg.Package) ([]match.Match, error) {
```
kzantow, line 65: A few naming suggestions that helped me understand the purpose of this set of functions, including adding 'merge' somewhere to many of them to indicate it is part of the vuln set merge operation:

`redhatEUSMatches` -> `redhatEUSMergeFunc`

grype/matcher/rpm/rhel_eus.go
```diff
@@ -0,0 +1,320 @@
...
+	// c. all fixes from the incoming resolutions are patched onto the disclosures in the returned collection, so the
+	//    final set of vulnerabilities is a fused set of disclosures and fixes together.
+	remaining := disclosures.Merge(resolutions, resolveEUSDisclosures(pkgVersion, false))
+
+	return remaining.ToMatches(), err
+}
+
+// resolveEUSDisclosures returns a function that will filter disclosures based on the provided advisory information (by fix version only).
+// Additionally, this will merge applicable fixes into one vulnerability record, so that the final result contains only one vulnerability record per disclosure.
+func resolveEUSDisclosures(v *version.Version, treatResolutionsAsDisclosures bool) func(disclosures, advisoryOverlays []result.Result) []result.Result {
```
kzantow: Suggestion: `resolveEUSDisclosures` -> `mergeEUSAdvisoriesIntoMainDisclosures`

grype/matcher/rpm/rhel_eus.go
```diff
@@ -0,0 +1,320 @@
...
+				}
+			}
+		}
+
+		return out
+	}
+}
+
+// processDisclosureResult processes a single disclosure Result against its corresponding advisory overlay Results
+func processDisclosureResult(v *version.Version, disclosures result.Result, advisoryOverlays []result.Result) result.Result {
```
kzantow: Suggestion: `processDisclosureResult` -> `mergeEUSAdvisoryIntoMainDisclosure`

grype/matcher/rpm/rhel_eus.go
```diff
@@ -0,0 +1,320 @@
...
+			processedResult.Details = append(processedResult.Details, advisoryDetails...)
+		}
+	}
+
+	finalizeMatchDetails(&processedResult, disclosures.Details, v)
+	return processedResult
+}
+
+// processVulnerabilityWithAdvisories processes a single vulnerability against advisory overlays
+func processVulnerabilityWithAdvisories(v *version.Version, disclosure vulnerability.Vulnerability, advisoryOverlays []result.Result) (*vulnerability.Vulnerability, match.Details) {
```
kzantow: Suggestion: `processVulnerabilityWithAdvisories` -> `mergeEUSAdvisoryIntoSingleDisclosure`

grype/matcher/rpm/rhel_eus.go
```diff
@@ -0,0 +1,320 @@
...
+		// all of the advisories showed we're not vulnerable, so we can skip this disclosure
+		return nil, nil
+	}
+
+	patchedRecord := buildPatchedVulnerabilityRecord(v, disclosure, fixVersions, constraints, state)
+	return &patchedRecord, allAdvisoryDetails
+}
+
+// processAdvisoryVulnerabilities processes vulnerabilities from advisory overlays, applying any new fix versions and updating the given fix state / constraints.
+func processAdvisoryVulnerabilities(v *version.Version, advisories []vulnerability.Vulnerability, fixVersions *version.Set, constraints *[]version.Constraint, state *vulnerability.FixState) {
```
kzantow: Suggestion: `processAdvisoryVulnerabilities` -> `collectMatchingConstraintsDetailsAndFixState`

This function also should accept `details *match.Details` that are set when `*constraints` is set around line 210

grype/matcher/rpm/rhel_eus.go
```diff
@@ -0,0 +1,320 @@
...
+	var allAdvisoryDetails match.Details
+
+	// check if we're vulnerable to the original disclosure
+	if isVulnerableVersion(v, disclosure.Constraint, disclosure.ID) {
+		constraints = append(constraints, disclosure.Constraint)
+	}
+
+	// process advisory overlays, incorporating new fix versions and updating the version constraints
+	for _, advisoryOverlay := range advisoryOverlays {
+		allAdvisoryDetails = append(allAdvisoryDetails, advisoryOverlay.Details...)
```
kzantow: These details being added here might be incorrect: if the advisory doesn't match any constraints, we may have a matching constraint from the disclosure, so we proceed past the `if len(constraints) == 0` below and erroneously include these details. We should instead collect them with the call below, which I've suggested to rename to `collectMatchingConstraintsDetailsAndFixState` in a separate comment with a little more detail below 👇

grype/matcher/rpm/rhel_eus.go
```diff
@@ -0,0 +1,320 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("matcher failed to fetch resolutions for distro=%q pkg=%q: %w", searchPkg.Distro, searchPkg.Name, err)
+	}
+
+	// combine disclosures and fixes so that:
+	// a. disclosures that have EUS fixes that resolve the disclosure for an earlier version of the package (thus we're not vulnerable) are removed.
+	// b. disclosures that have EUS fixes that resolve the disclosure for future versions of the package (thus we're vulnerable) are kept.
+	// c. all fixes from the incoming resolutions are patched onto the disclosures in the returned collection, so the
+	//    final set of vulnerabilities is a fused set of disclosures and fixes together.
+	remaining := disclosures.Merge(resolutions, resolveEUSDisclosures(pkgVersion, false))
```
kzantow: I know I may have suggested this merge operation, but I wonder if there are better options. Reading this function, the two `Find` calls, above are very clear and I can completely understand the logic to this EUS matching to that point but that stops being as true with a Merge call doing some semi-arbitrary logic. I keep wondering if there are some more set-operations that could help with clarity, something like:
```
disclosuresWithAdvisoryFixVersions := copyFixVersions(advisories, disclosures)
vulnerable := disclosuresWithAdvisoryFixVersions.Filter(internal.OnlyVulnerableVersions(searchPkg.Version))
```
... I understand that logic is probably not quite right, but it seems that _something_ might represent the operations in a way that might be simpler to follow, for me at least. Maybe introducing a new criteria type that tests a provided version against fixes and/or filters out records for fix info that doesn't match... or something along those lines.

grype/matcher/internal/result/provider.go
```diff
@@ -0,0 +1,197 @@
...
+
+var _ Provider = (*provider)(nil)
+
+type Provider interface {
+	FindResults(criteria ...vulnerability.Criteria) (Set, error)
+}
+
+type provider struct {
+	vulnProvider vulnerability.Provider
+	catalogedPkg pkg.Package // this is what is passed into the matcher
```
kzantow, line 18: We already talked about this offline, but just noting here that requiring the package and matcher type in the provider prevents this interface from replacing the `vulnerability.Provider` interface, since each matcher would have to instantiate a new instance. I could see a future state where the vulnerability objects returned always have details about how they were fetched, so the current `vulnerability.Provider` would more-or-less just be transitioned to this interface

grype/matcher/internal/result/results.go
```diff
@@ -0,0 +1,186 @@
...
+	out := Set{}
+	if len(mergeFuncs) == 0 {
+		// with no other merge functions specified, append all vulnerability results and details
+		mergeFuncs = []func(existing, incoming []Result) []Result{
+			unionResults,
+		}
+	}
+
+	// det all unique IDs from both sets
+	allIDs := make(map[string]bool)
```
kzantow: nit: `bool` -> `struct{}`

grype/matcher/internal/result/results.go
```diff
@@ -0,0 +1,186 @@
...
+		if len(mergedResults) > 0 {
+			// filter out any results with empty vulnerabilities
+			var validResults []Result
+			for _, result := range mergedResults {
+				if result.ID != "" && len(result.Vulnerabilities) > 0 {
+					validResults = append(validResults, result)
+				}
+			}
+			if len(validResults) > 0 {
+				out[id] = validResults
```
kzantow: These results should use the IDs returned from the results themselves (e.g. `result.ID` above) rather than `id`

grype/matcher/internal/result/results.go
```diff
@@ -0,0 +1,186 @@
...
+			filteredResults = append(filteredResults, Result{
+				ID:              result.ID,
+				Vulnerabilities: vulns,
+				Details:         result.Details,
+			})
+		}
+
+		if len(filteredResults) > 0 {
+			out[id] = filteredResults
+		}
```
kzantow, line 162:
```
} else {
  vulnerability.LogDropped(id, ...
}
```

grype/matcher/rpm/matcher.go
```diff
@@ -23,20 +26,43 @@ func (m *Matcher) Type() match.MatcherType {
...
+	// matching on a source package above where the epoch could be dropped in the
+	// reference data. This means that any missing epoch CAN be assumed to be zero,
+	// as it falls into the case of "the project elected to NOT have an epoch for the
+	// first version scheme" and not into any other case.
+
+	// For this reason match exactly on a package, we should be EXPLICIT about the
+	// epoch (since downstream version comparison logic will strip the epoch during
+	// comparison for the above-mentioned reasons --essentially for the source RPM
+	// case). To do this, we fill in missing epoch values in the package versions with
+	// an explicit 0.
```
kzantow: nit: should this and the other epoch comment below be moved to the `matchPackage` where it's calling a function to set the default epoch? I read this a couple times while trying to understand some of the functionality and was confused why it was talking about epochs here

grype/matcher/rpm/rhel_eus.go
```diff
@@ -0,0 +1,326 @@
...
+				}
+			}
+
+			processedResult := processDisclosureResult(v, ds, as)
+			if len(processedResult.Vulnerabilities) > 0 {
+				out = append(out, processedResult)
+			}
+		}
+
+		if treatResolutionsAsDisclosures {
```
kzantow, line 124: This option is not being used, it should probably be removed

grype/matcher/rpm/rhel_eus.go
```diff
@@ -0,0 +1,326 @@
...
+				if advisory.ID == ds.ID {
+					as = append(as, advisory)
+				}
+			}
+
+			processedResult := processDisclosureResult(v, ds, as)
+			if len(processedResult.Vulnerabilities) > 0 {
+				out = append(out, processedResult)
+			}
+		}
```
kzantow: This is unnecessary, since the merge function is called for the results with the same IDs already.

grype/matcher/rpm/rhel_eus.go
```diff
@@ -0,0 +1,326 @@
...
+
+// buildPatchedVulnerabilityRecord creates the final patched vulnerability record from the original disclosure and fix/constraint information from applicable advisories.
+func buildPatchedVulnerabilityRecord(v *version.Version, disclosure vulnerability.Vulnerability, fixVersions *version.Set, constraints []version.Constraint, state vulnerability.FixState) vulnerability.Vulnerability {
+	patchedRecord := disclosure
+
+	if state == vulnerability.FixStateFixed {
+		patchedRecord.Fix.Versions = nil
+		for _, fixVersion := range fixVersions.Values() {
+			patchedRecord.Fix.Versions = append(patchedRecord.Fix.Versions, fixVersion.Raw)
+			fixConstraint, err := version.GetConstraint(fmt.Sprintf("< %s", fixVersion), v.Format)
```
kzantow: This `fixVersion` is causing the `.String()` to be called which fails to parse due to parenthesis and some entries to be excluded, I think it should be `fixVersions.Raw` here

grype/matcher/rpm/rhel_eus.go
wagoodman: I'm down for ways to improve on the readability of the merge function, right now it's decomposed to oblivion and is itself hard to follow. Sounds like something we can build on here.

grype/matcher/internal/result/provider.go
wagoodman: > I could see a future state where the vulnerability objects returned always have details about how they were fetched...

heavily agreed

> so the current vulnerability.Provider would more-or-less just be transitioned to this interface

I need to think about this more, but at a high level it seems sensible. I think the decoupling in package from the provider hinges on the match.Detail changes / evidence coming back from the provider... without a detailed plan here it's hard to say if this is the final interface.

grype/matcher/rpm/rhel_eus.go
wagoodman: there was some question during development if it should be used (which is still a little up in the air until we can find an example)

grype/matcher/rpm/rhel_eus.go
wagoodman: agreed, but this was a safety measure in case the caller is wired up incorrectly we will not combine results that have different IDs. I'll remove it.

grype/matcher/rpm/rhel_eus.go
wagoodman: I incorporated the others, but this isn't a merge function; there are `standardMatches` and `redhatEUSMatches` calls that are high-level entrypoints to the search, so I'll leave this one as is.

grype/matcher/rpm/rhel_eus.go
wagoodman: ~I'll remove it~ well this is well tested, and could be enabled in the future. I think I will leave this in case it is needed later

grype/distro/distro_test.go
```diff
@@ -14,10 +14,23 @@ import (
 	"github.com/anchore/syft/syft/source/directorysource"
 )
 
+func testFixChannels() []FixChannel {
+	// TODO: change this to DefaultFixChannels() when EUS is configured "auto" by default
+	//return DefaultFixChannels()
```
kzantow: Should this get updated?

grype/matcher/rpm/rhel_eus.go
kzantow: Yes, got my comments in the wrong spot, sorry 🤦

grype/matcher/internal/result/results.go
```diff
@@ -0,0 +1,184 @@
...
+// Result represents a prototype match of a package used to search, a set of vulnerabilities discovered from the search,
+// and match details that describe the search itself. Note that all vulnerabilities in a Result share the same
+// vulnerability ID (in the ID field and `.Vulnerabilities[].ID` fields -- it is invalid to mix vulnerabilities into
+// a Result that have different IDs.
+type Result struct {
+	// ID is the vulnerability ID; all vulnerabilities in this Result share the same ID.
+	ID string
+
+	// Vulnerabilities is a set of vulnerabilities that were discovered from the search.
+	Vulnerabilities []vulnerability.Vulnerability
```
kzantow, line 20: I think we agreed that each Result should reference a single Vulnerability to avoid cases where there are different match details that get merged together, is that right?

grype/search/version_constraint.go
```diff
@@ -20,27 +25,43 @@ func ByConstraintFunc(constraintFunc func(constraint version.Constraint) (bool,
 	return &constraintFuncCriteria{fn: constraintFunc}
 }
 
+type VersionCriteria struct {
+	Version version.Version
+	vulnerability.Criteria
```
kzantow: Does this need to have a Criteria implementation (like `ByConstratntFunc`, below)? Could this struct just implement the functions it needs based on the version in the struct?

grype/version/deprecated.go
```diff
@@ -0,0 +1,10 @@
+package version
+
+// NewVersion creates a new Version instance with the provided raw version string and format.
+// Deprecated: NewVersion is deprecated, use New instead.
+func NewVersion(raw string, format Format) *Version {
+	return &Version{
```
kzantow: nit: `return New(raw, format)`, in case there's some logic that gets added?

grype/distro/distro_test.go
wagoodman: yeah, now that the distro PR is merged I can rebase and update it here

grype/matcher/internal/result/results.go
wagoodman: we did agree, but the main bug was to ensure that the merge function correctly reasoned about these details (which was fixed). I held off incorporating the data shape changes for a few of reasons:
1. this has been tested to show it's working as we'd expect
2. further changes to the Result struct should probably be guided by what we want to do with match details and if we want criteria to take their place in the result struct
3. I though you suggested to hold off on incorporating changes to the Result object based on your prototyping (maybe I misunderstood?)

grype/search/version_constraint.go
wagoodman: need to sync further to understand the motivation behind this comment

PR #2793, opened 2025-07-09
perf: set some mem saving pragmas before vacuum

willmurphyscode:
Previously, building the v5 Vulnerability.db would sometimes fail with OOM errors, mostly because during the database build we trade RAM for speed. However during VACUUM, these settings cause OOM. Therefore, before the vacuum operation, accept more disk I/O in exchange for a smaller RAM footprint.

### DIFF
--- a/grype/db/v5/store/store.go
+++ b/grype/db/v5/store/store.go
@@ -11,6 +11,7 @@ import (
 	"github.com/anchore/grype/grype/db/internal/gormadapter"
 	v5 "github.com/anchore/grype/grype/db/v5"
 	"github.com/anchore/grype/grype/db/v5/store/model"
+	"github.com/anchore/grype/internal/log"
 	"github.com/anchore/grype/internal/stringutil"
 )
 
@@ -98,7 +99,7 @@ func (s *store) GetVulnerability(namespace, id string) ([]v5.Vulnerability, erro
 
 	result := query.Find(&models)
 
-	var vulnerabilities = make([]v5.Vulnerability, len(models))
+	vulnerabilities := make([]v5.Vulnerability, len(models))
 	for idx, m := range models {
 		vulnerability, err := m.Inflate()
 		if err != nil {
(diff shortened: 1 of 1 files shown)


grype/db/v5/store/store.go
```diff
@@ -278,7 +279,27 @@ func (s *store) AddVulnerabilityMatchExclusion(exclusions ...v5.VulnerabilityMat
 }
 
 func (s *store) Close() error {
+	log.Info("optimizing database settings for memory-efficient VACUUM")
```
wagoodman: nit

```suggestion
	log.Debug("optimizing database settings for memory-efficient VACUUM")
```

grype/db/v5/store/store.go
```diff
@@ -278,7 +279,27 @@ func (s *store) AddVulnerabilityMatchExclusion(exclusions ...v5.VulnerabilityMat
...
+
+	for _, stmt := range memoryEfficientStatements {
+		if err := s.db.Exec(stmt).Error; err != nil {
+			log.WithFields("statement", stmt, "error", err).Warn("failed to apply memory optimization")
+		} else {
+			log.WithFields("statement", stmt).Debug("applied memory optimization")
+		}
+	}
+
+	log.Info("starting database VACUUM operation")
```
wagoodman:
```suggestion
	log.Debug("starting database VACUUM operation")
```

grype/db/v5/store/store.go
```diff
@@ -278,7 +279,27 @@ func (s *store) AddVulnerabilityMatchExclusion(exclusions ...v5.VulnerabilityMat
...
 	s.db.Exec("VACUUM;")
+	log.Info("database VACUUM operation completed")
```
wagoodman:
```suggestion
	log.Debug("database VACUUM operation completed")
```

wagoodman: soft suggestions, take it or leave it

PR #2805, opened 2025-07-17
Create ignore regex objects conditionally

wagoodman:
Today a user can configure ignore rules based on package names, which can include regexp patterns. However, if you start including thousands of rules, it can take a lot of time creating those objects (where the bottleneck is the `regexp.Compile`). 

Screenshot 2025-07-17 at 10 35 16 AM

This PR makes creating those objects conditional; if we detect if the package name is likely a regexp expression without compiling with a one-time-compiled regexp, then use that on each example ignore rule to generate either a custom regexp or a simple string comparison. This will help in cases where the user has a set of rules that are simple names instead of expressions.

### DIFF
--- a/grype/db/v6/models_test.go
+++ b/grype/db/v6/models_test.go
@@ -1,9 +1,9 @@
 package v6
 
 import (
-	"github.com/google/go-cmp/cmp"
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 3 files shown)


kzantow (approved):
Today `IgnoreRules` implement `IgnoreFilter`, maybe we could preprocess all `IgnoreRule` input into optimized `Ignorefilter` implementations that have compiled regex/etc.?

wagoodman:
I had a similar thought -- we can iterate towards that 👍 (would need to have ignore-filters-first instead of rules in some of the top-level functionality, which is a larger refactor)

PR #2811, opened 2025-07-19
fix(vex): set default product list

alegrey91:


alegrey91:
Any thoughts about the fix?

PR #2843, opened 2025-08-01
chore: add FP issue template

kzantow:
This PR adds a new issue template to report false positives that both labels them appropriately and encourages providing a PURL to make analysis simpler.

### DIFF
--- a/.github/ISSUE_TEMPLATE/false_positive.md
+++ b/.github/ISSUE_TEMPLATE/false_positive.md
@@ -1,0 +1,32 @@
+---
+name: False Positive
+about: Report an incorrect vulnerability match
+title: ''
+labels: [ bug, false-positive ]
+assignees: ''
+
+---
+
+**Vulnerability ID**:
+
+**Package URL (PURL) or steps to reproduce**:
+<!--
+If possible, please provide a PURL.
+
+If the issue can't be reproduced with a PURL, link to an artifact grype can scan or provide other instructions to reproduce.
+
+Some suggestions:
+1. Link to Dockerhub, GitHub, GitLab, maven central, quay.io, etc to a public
+   artifact we can try scanning
+2. A Dockerfile that we can build and scan
+3. A simple script that creates a directory exhibiting the issue, for example a
+   list of `npm install` commands
+
+Please also include the grype command and any configuration used.
+-->
+
+**Anything else we need to know?**:
+
+**Environment**:
+- Output of `grype version`:
+- OS (e.g: `cat /etc/os-release` or similar):


.github/ISSUE_TEMPLATE/false_positive.md
```diff
@@ -0,0 +1,32 @@
+---
+name: False Positive
+about: Report an incorrect vulnerability match
```
willmurphyscode: This sounds like it could be an FP or an FN. What about something like, "report that Grype detects a vulnerability not actually present"? Not a deal breaker, just trying to be clear about what an FP is since it's not necessarily a term everyone uses every day.

.github/ISSUE_TEMPLATE/false_positive.md
```diff
@@ -0,0 +1,32 @@
...
+labels: [ bug, false-positive ]
+assignees: ''
+
+---
+
+**Vulnerability ID**:
+
+**Package URL (PURL) or steps to reproduce**:
+<!--
+If possible, please provide a PURL.
```
willmurphyscode: This could use an example. Maybe mention of distro flag too.

.github/ISSUE_TEMPLATE/false_positive.md
joshbressers: Will we also have a False Negative template? I think it would make sense to have both

.github/ISSUE_TEMPLATE/false_positive.md
```diff
@@ -0,0 +1,32 @@
...
+1. Link to Dockerhub, GitHub, GitLab, maven central, quay.io, etc to a public
+   artifact we can try scanning
+2. A Dockerfile that we can build and scan
+3. A simple script that creates a directory exhibiting the issue, for example a
+   list of `npm install` commands
+
+Please also include the grype command and any configuration used.
+-->
+
+**Anything else we need to know?**:
```
joshbressers: I would add a section here (or maybe at the top) asking for an explanation. URLs are encouraged. So for example we could get something like "CVE-WHATEVER only affects foobar version 3.10 according to the advisory at URL, but grype is claiming version 3.56 is vulnerable"

.github/ISSUE_TEMPLATE/match_issue.md
```diff
@@ -0,0 +1,39 @@
+---
+name: Vulnerability Match Issue
+about: Report an issue with vulnerability matching
+title: ''
+labels: [ bug, false-positive ]
```
kzantow, line 5: This may, in fact, be a `false-negative`, but it seems much more likely to get FP reports, and having two templates seems like overkill to me. Maybe having separate templates would be better 🤷

willmurphyscode (approved):

PR #2886, opened 2025-08-21
[WIP] feat(grype/matcher): use fix info to add ignore rules

CrosleyZack:
This uses the `vulnerability.Vulnerability.Fix.Versions` to add IgnoreRules. I am not confident this is correct as is. Produces results with [example security feed](https://storage.googleapis.com/chainguard-upstream-mapping/vex/pypi/urllib3.openvex.json):

```
➜ GRYPE_DB_VALIDATE_AGE="false" GRYPE_DB_VALIDATE_BY_HASH_ON_START="false" go run cmd/grype/main.go "pkg:pypi/urllib3@1.26.17"

   ├── by severity: 0 critical, 0 high, 0 medium, 0 low, 0 negligible (1 unknown)
   └── by status:   1 fixed, 0 not-fixed, 0 ignored
NAME     INSTALLED  FIXED IN       TYPE    VULNERABILITY   SEVERITY  EPSS  RISK
urllib3  1.26.17    1.26.16+cgr.1  python  CVE-2023-45803  Unknown   N/A   N/A

➜ GRYPE_DB_VALIDATE_AGE="false" GRYPE_DB_VALIDATE_BY_HASH_ON_START="false" go run cmd/grype/main.go "pkg:pypi/urllib3@1.26.16"

   ├── by severity: 0 critical, 0 high, 0 medium, 0 low, 0 negligible (1 unknown)
   └── by status:   1 fixed, 0 not-fixed, 0 ignored
NAME     INSTALLED  FIXED IN       TYPE    VULNERABILITY   SEVERITY  EPSS  RISK
urllib3  1.26.16    1.26.16+cgr.1  python  CVE-2023-45803  Unknown   N/A   N/A
```


grype/matcher/internal/language.go
```diff
@@ -70,5 +100,27 @@ func MatchPackageByEcosystemPackageName(provider vulnerability.Provider, p pkg.P
...
+			return nil, err
+		}
+		pkgMeta := strings.Split(pkgVersion.Metadata(), ".")[0]
+		fixMeta := strings.Split(fixVersion.Metadata(), ".")[0]
+		// if the fix is unspecified or matches on meta, return
+		if fixMeta == "" || pkgMeta == fixMeta {
+			fixes = append(fixes, fix)
+		}
+	}
+	return fixes, nil
```
crosleyzack: As I understand it, the goal is to only show fix versions that are relevant to the package in question. This can be:
1. Packages with no metadata (IE 1.2.3) which always apply
2. Packages with the same metadata (IE 1.2.3+foo.1 for package with version X.Y.Z+foo.N)

This fulfills these requirements, but happy to know if this is not actually what we want

grype/matcher/internal/language.go
```diff
@@ -33,19 +36,48 @@ func MatchPackageByEcosystemPackageName(provider vulnerability.Provider, p pkg.P
...
+	}
+
 	var matches []match.Match
+	var ignores []match.IgnoreFilter
 	vulns, err := provider.FindVulnerabilities(
 		search.ByEcosystem(p.Language, p.Type),
 		search.ByPackageName(packageName),
 		OnlyQualifiedPackages(p),
-		OnlyVulnerableVersions(version.New(p.Version, pkg.VersionFormat(p))),
+		OnlyVulnerableVersions(version.New(trimVersionMeta(pkgVersion), pkg.VersionFormat(p))),
```
crosleyzack: We need to make sure version metadata is trimmed so it doesn't break this filter

crosleyzack: This makes me think I might be mistaken and metadata will be filtered before this point

grype/matcher/internal/language_test.go
```diff
@@ -109,6 +144,17 @@ func TestFindMatchesByPackageLanguage(t *testing.T) {
...
+		// 	constraint: "< 1.7.6 (gem)",
+		// 	expIgnores: expectedIgnored("nokogiri", "1.7.4+foo.1"),
+		// 	p: pkg.Package{
+		// 		ID:       pkg.ID(uuid.NewString()),
+		// 		Name:     "nokogiri",
+		// 		Version:  "1.7.5+foo.1",
+		// 		Language: syftPkg.Ruby,
+		// 		Type:     syftPkg.GemPkg,
+		// 	},
+		// },
```
crosleyzack: Ruby packages don't appear to support the version metadata, results in an exception

grype/matcher/internal/language.go
crosleyzack: ~~Another option here is to edit https://github.com/anchore/go-version to use the metadata, depending on what we want~~

This is not a valid idea, as its against [semver spec](https://semver.org/#spec-item-10)

grype/matcher/internal/language.go
crosleyzack: It looks like it actually depends on the version type. Ruby specifically doesn't like metadata, but Golang does.

grype/matcher/internal/language.go
```diff
@@ -33,19 +36,48 @@ func MatchPackageByEcosystemPackageName(provider vulnerability.Provider, p pkg.P
 		return nil, nil, nil
 	}
 
+	pkgVersion, err := hashiVer.NewSemver(p.Version)
+	if err != nil {
+		return nil, nil, fmt.Errorf("package contains invalid version %s: %w", p.Version, err)
+	}
```
crosleyzack: This fails with Java tests that allow for versioning 1.2.3.RELEASE

grype/matcher/internal/language.go
```diff
@@ -46,6 +49,28 @@ func MatchPackageByEcosystemPackageName(provider vulnerability.Provider, p pkg.P
 	}
 
 	for _, vuln := range vulns {
+		// filter versions by semver metadata, if this is a semver
+		if filteredVersions, err := filterFixVersions(p.Version, vuln.Fix.Versions); err == nil {
+			vuln.Fix.Versions = filteredVersions
+		}
```
crosleyzack: filter versions by semver metadata, if semver. Else, nothing happens

grype/matcher/internal/language.go
```diff
@@ -70,5 +95,32 @@ func MatchPackageByEcosystemPackageName(provider vulnerability.Provider, p pkg.P
...
+	if err != nil {
+		return nil, fmt.Errorf("pkg is not a valid semver %s: %w", pkgVer, err)
+	}
+	// filter fix versions by semver metadata
+	fixes := make([]string, 0, len(fixVers))
+	for _, fix := range fixVers {
+		fixVersion, err := hashiVer.NewSemver(fix)
+		if err != nil {
+			return nil, err
+		}
```
crosleyzack: If only one version is not a semver, should we just continue?

grype/matcher/internal/language.go
```diff
@@ -70,5 +80,61 @@ func MatchPackageByEcosystemPackageName(provider vulnerability.Provider, p pkg.P
...
+		// TODO: compare to grype/vex/openvex/implementation.go
+		ignores = append(ignores, match.IgnoreRule{
+			Vulnerability: vuln.ID,
+			Reason:        "Fix",
+			Namespace:     vuln.Namespace,
+			FixState:      vulnerability.FixStateFixed.String(),
+			Package: match.IgnoreRulePackage{
+				Name: p.Name,
+				// TODO the fix for 1.2.3+foo.1 should apply to any version 1.2.3+foo.N for n >= 1
+				Version:  fix,
```
crosleyzack: From what I can tell, this won't work as written. So either multiple ignores have to be added, or I have to change the ignore logic. Not sure which is preferable

grype/matcher/internal/language.go
wagoodman: has since been removed

grype/matcher/internal/language.go
wagoodman: this is now format aware

grype/matcher/internal/language.go
wagoodman: this is now format aware

grype/matcher/internal/language.go
wagoodman: removed

PR #2888, opened 2025-08-22
Add unaffected package and CPE stores

wagoodman:
This PR adds a deferred feature from the initial v6 schema work from janurary: the ability to describe packages and CPEs that are explicitly not affected by a vulnerability.

Specifically, this adds:
- `UnaffectedPackageStore`
- `UnaffectedCPEStore`

note: we cannot refactor the existing alpine naks (e.g. `<= 0` version constraint) since there are older clients that depend on that mechanism (unless we decide to duplicate all of that data, which is an option if we wish).

Related PR: https://github.com/anchore/grype-db/pull/648


grype/db/v6/vulnerability_provider.go
```diff
@@ -192,180 +187,156 @@ func (vp vulnerabilityProvider) Close() error {
...
+	}
+	if errors.Is(err, ErrOSNotPresent) {
+		log.WithFields("os", osSpecs).Debug("no OS found in the DB for the given criteria")
+		return nil
+	}
+	return err
+}
+
+// fetchAndProcessPackages fetches and processes both affected and unaffected packages based on the context
+func (vp vulnerabilityProvider) fetchAndProcessPackages(ctx *searchQuery) ([]AffectedPackageHandle, []UnaffectedPackageHandle, error) {
```
spiffcs: from sync review: ctx -> sc or something closer to searchQuery

grype/db/v6/vulnerability_provider.go
```diff
@@ -192,180 +187,156 @@ func (vp vulnerabilityProvider) Close() error {
...
+		if err != nil {
+			return nil, err
 		}
 
-		// if there is an ecosystem provided and a name, we need to make certain that we're using the name normalization
-		// rules specific to the ecosystem before searching.
-		if pkgType != "" && pkgSpec.Name != "" {
-			pkgSpec.Name = name.Normalize(pkgSpec.Name, pkgType)
+		// convert to vulnerabilities
+		vulns, err := vp.toVulnerabilities(affectedPkgs, affectedCPEs, unaffectedPkgs, unaffectedCPEs)
```
spiffcs: from sync review: we might want to do a refactor here where we can just return vulnerabilities from fetchAndProcessPackages as well as fetchAndProcessCPEs and then merge the two lists together rather than  how we have it here where we wait for everything to go into `toVulnerabilities` since `toVulnerabilities` doesn't NEED to have all 4 at the time it's called, it's just a convenience to sum them all into vulnerabilities at once. 

Readers might see `toVulnerabilitiers` and assume `affectedCpeHandles` is needed for something in `affectedPackageHandles` or vise versa for unaffected. This refactor would just separate those concerns

grype/db/v6/vulnerability_provider.go
spiffcs: resolved

grype/db/v6/vulnerability_provider.go
spiffcs: resolved

spiffcs (approved):
No more notes/comments from our previous sync review. I went through each file and didn't see anything else I had questions on. Test coverage is A+ and all the cases seem covered correctly for new unaffected code paths.

👍 LGTM

PR #2904, opened 2025-08-29
feat: add CWEs field to VulnerabilityBlob to support weaknesses data from NVD

hasnatbashir:
Related issue: https://github.com/anchore/grype-db/issues/644


wagoodman (commented):
The goal here is to add a new CWEs table instead of adding CWEs to the vulnerability blob, that way we can use CWEs across multiple providers in theory (same way KEVs are portrayed, as their own table) ... see https://github.com/anchore/grype-db/issues/644#issuecomment-3234703233

hasnatbashir:
@wagoodman Thanks! Could you also clarify the representation? What should the schema look like, right now we just want to store a list of CWEs against CVE IDs Would a simple schema be sufficient? or should we use a blob format, like other handles, to keep it flexible in case we want to store additional fields in the future?

hasnatbashir:
@wagoodman can you please take a look again?

hasnatbashir:
@willmurphyscode Can you please take a look at this PR? Let me know if there are any other changes that are needed?

willmurphyscode:
Hi @hasnatbashir this looks pretty good. There is some busy work around bumping schema version constants and things that the maintainers will do. It looks like I don't have permission to edit this PR (that often happens when the PR author doesn't own the fork / the fork owner is an org, not an individual). The preferred workaround in that case is that I'll merge this into a non-main branch, push some changes to it, and then merge to main. I just wanted to let you know what to expect since that can make a little noise on the PR / issue threads. I'll probably be doing that tomorrow (it's late here).

But if you'd prefer to do this yourself, you can use https://github.com/anchore/grype/compare/add-cwe-ids?expand=1 as a guide for what needs to be updated (basically, rebase onto / merge from main, then get static analysis to pass by incrementing versions of affected schemas).

Thanks very much for this work! It made us realize we didn't have proper validation around changes to the db models, which is what we added in https://github.com/anchore/grype/pull/2962, so basically the changes requested are to pull in the changes from that PR, and then appease the linters it added. But I'll take care of it (I'm assuming the lack of permissions for maintainers to edit the branch is unintentional, but let me know if I'm wrong.)

PR #2917, opened 2025-09-04
Move test data to grype db

willmurphyscode:
**IMPORTANT** this PR needs to be coordinated with https://github.com/anchore/grype-db/pull/592 - it's important that the changes in this PR not be pulled into grype-db ahead of anchore/grype-db#592.

### DIFF
--- a/grype/db/v6/data.go
+++ b/grype/db/v6/data.go
@@ -1,1 +1,0 @@
-package v6


willmurphyscode:
This was part of an abandoned approach to support better Alma matching in grype. What was actually done was supporting alma advisories as unaffectedPackageHandles in the grype db build. This PR is no longer needed.

PR #2939, opened 2025-09-12
Feat alma unaffected

willmurphyscode:
This is a better approach for https://github.com/anchore/grype/issues/2745

### DIFF
--- a/grype/matcher/rpm/almalinux.go
+++ b/grype/matcher/rpm/almalinux.go
@@ -277,8 +277,8 @@ func extractFixVersionFromConstraint(constraint version.Constraint) string {
 
 	// Handle common constraint patterns
 	// ">= version" → "version"
-	if strings.HasPrefix(constraintStr, ">= ") {
-		return strings.TrimPrefix(constraintStr, ">= ")
+	if strings.HasPrefix(constraintStr, ">=") {
+		return strings.TrimSpace(strings.TrimPrefix(constraintStr, ">="))
 	}
 
 	// "= version" → "version"


grype/search/distro.go
```diff
@@ -53,10 +62,50 @@ func (c *DistroCriteria) Summarize() string {
 	return "does not match distro(s): " + strings.Join(distroStrs, ", ")
 }
 
+type ExactDistroCriteria struct {
+	Distros []distro.Distro
+}
+
+func (c *ExactDistroCriteria) MatchesVulnerability(value vulnerability.Vulnerability) (bool, string, error) {
```
wagoodman: Oddly enough we probably don't want to depend on MatchesVulnerability() moving forward. The path for this is to really embed this within the v6 vulnerability provider (in the [query builder](https://github.com/anchore/grype/blob/main/grype/db/v6/search_query.go#L50-L84), we do want to keep this type you made though to type assert with in the vulnerability provider). The reason for this is that there tends not to be enough information in the `vulnerability.Vulnerability` type to make these kinds of decisions without the v6 schema starting to creep through. In this case the code is leveraging namespaces, however, they are also on their way out also, and nothingn functionally depends on these values (they are here as a v5 hold over to keep the json presentation stable until grype v1). The OS information is available within the raw v6 data records, and it a little more natural to implement than with namespaces.

grype/matcher/rpm/almalinux.go
```diff
@@ -0,0 +1,299 @@
...
+
+	// Step 3: Also look for unaffected packages using source/binary RPM relationships
+	relatedUnaffectedResults := findRelatedUnaffectedPackages(provider, searchPkg)
+	// Merge the related results into the main unaffected results
+	for key, results := range relatedUnaffectedResults {
+		unaffectedResults[key] = append(unaffectedResults[key], results...)
+	}
+
+	// Step 4: Update RHEL disclosures with AlmaLinux-specific fix information
+	updatedDisclosures := updateDisclosuresWithAlmaLinuxFixes(disclosures, unaffectedResults, pkgVersion)
```
wagoodman: I suspect (but have not proven this out) that steps 3 and 4 could leverage the result [merge](https://github.com/anchore/grype/blob/main/grype/matcher/internal/result/results.go#L130-L133) and [filter](https://github.com/anchore/grype/blob/main/grype/matcher/internal/result/results.go#L216) functions.

grype/search/distro.go
willmurphyscode: I'm not sure what is being suggested here. Can you elaborate a bit on what a better implementation would look like?

grype/search/distro.go
```diff
@@ -53,10 +62,50 @@ func (c *DistroCriteria) Summarize() string {
 	return "does not match distro(s): " + strings.Join(distroStrs, ", ")
 }
 
+type ExactDistroCriteria struct {
```
kzantow: I instead of adding a new type, would it make the change slightly simpler to add a bool to `DistroCriteria { exact bool ...` and add a bool to the `handleDistro` function?

grype/search/distro.go
wagoodman: ignore my comment it's just, all sorts of wrong

grype/db/v6/search_query.go
```diff
@@ -167,6 +170,38 @@ func (b *searchQueryBuilder) handleDistro(c *search.DistroCriteria) {
 	}
 }
 
+func (b *searchQueryBuilder) handleExactDistro(c *search.ExactDistroCriteria) {
```
kzantow: Essentially the same comment as below: would it be simpler to just add a bool to `handleDistro`?

grype/matcher/rpm/almalinux.go
```diff
@@ -0,0 +1,310 @@
...
+		// to be affected by every CVE that affects their src rpm at any version.
+	}
+	// Create a RHEL-compatible distro for finding base disclosures
+	rhelCompatibleDistro := *searchPkg.Distro
+	rhelCompatibleDistro.Type = distro.RedHat // treat as RHEL for disclosure lookup
+
+	pkgVersion := version.New(searchPkg.Version, pkg.VersionFormat(searchPkg))
+
+	// Step 1: Find RHEL disclosures for the package
+	disclosures, err := provider.FindResults(
```
willmurphyscode: Should additionally check for disclosures on the src rpm.

grype/matcher/internal/only_affected_vulnerabilities.go
```diff
@@ -0,0 +1,16 @@
+package internal
+
+import (
+	"github.com/anchore/grype/grype/search"
+	"github.com/anchore/grype/grype/vulnerability"
+)
+
+// OnlyAffectedVulnerabilities returns a criteria object that filters out unaffected vulnerability records
+func OnlyAffectedVulnerabilities() vulnerability.Criteria {
```
wagoodman: I think [this](https://github.com/anchore/grype/blob/main/grype/search/unaffected.go) overlaps (and might do the same thing)

grype/matcher/rpm/almalinux.go
```diff
@@ -0,0 +1,396 @@
...
+
+// shouldCompletelyFilter determines if a vulnerability should be completely filtered out
+func shouldCompletelyFilter(vuln vulnerability.Vulnerability, pkgVersion *version.Version) bool {
+	if !isVersionUnaffected(pkgVersion, vuln.Constraint, vuln.ID) {
+		return false
+	}
+
+	fixVersion := extractFixVersionFromConstraint(vuln.Constraint)
+	return shouldFilterVulnerability(vuln.Constraint, fixVersion, pkgVersion)
+}
```
wagoodman: having shouldFilterVulnerability vs shouldCompletelyFilter is confusing -- maybe merge them into a single function?

grype/matcher/rpm/almalinux.go
```diff
@@ -0,0 +1,396 @@
...
+
+// extractFixVersionFromConstraint extracts a fix version from a version constraint
+// e.g., ">= 2.4.48" → "2.4.48", "= 1.2.3-4.el8" → "1.2.3-4.el8"
+func extractFixVersionFromConstraint(constraint version.Constraint) string {
+	if constraint == nil {
+		return ""
+	}
+
+	constraintStr := constraint.String()
```
wagoodman: if you use constraint.Value() you can remove the cleanVersionString function (has no parentheticals)

grype/matcher/rpm/almalinux.go
```diff
@@ -0,0 +1,396 @@
...
+
+	// Handle common constraint patterns
+	// ">= version" → "version"
+	if strings.HasPrefix(constraintStr, ">= ") {
+		version := strings.TrimPrefix(constraintStr, ">= ")
```
wagoodman: nit: `version` clashes with import name

grype/matcher/rpm/almalinux.go
```diff
@@ -0,0 +1,396 @@
...
+}
+
+// updateRemainingWithAlmaLinuxFixes updates remaining vulnerabilities with AlmaLinux fix information
+func updateRemainingWithAlmaLinuxFixes(disclosures result.Set, unaffectedResults result.Set) result.Set {
+	almaLinuxFixes := buildAlmaLinuxFixesMap(unaffectedResults)
+
+	log.WithFields("almaLinuxFixesCount", len(almaLinuxFixes), "unaffectedResultsCount", len(unaffectedResults)).Debug("built AlmaLinux fixes map")
+
+	if len(almaLinuxFixes) == 0 {
+		log.Debug("no AlmaLinux fixes found, returning original disclosures")
```
wagoodman: nit: this log line provides little value after initial development

grype/matcher/rpm/almalinux.go
```diff
@@ -0,0 +1,396 @@
...
+							Vulnerabilities: []vulnerability.Vulnerability{{Reference: related}},
+						}}
+					}
+				}
+			}
+		}
+	}
+
+	return toRemove
+}
```
wagoodman: it looks like we are building a set to use later for removal from another set. Could this be better accomplished with the result.Set.Filter() which takes arbitrary criteria functions? That would prevent needing to craft an intermediate set.

grype/matcher/rpm/almalinux.go
```diff
@@ -0,0 +1,396 @@
...
+	if !isVersionUnaffected(pkgVersion, vuln.Constraint, vuln.ID) {
+		return false
+	}
+
+	fixVersion := extractFixVersionFromConstraint(vuln.Constraint)
+	return shouldFilterVulnerability(vuln.Constraint, fixVersion, pkgVersion)
+}
+
+// updateRemainingWithAlmaLinuxFixes updates remaining vulnerabilities with AlmaLinux fix information
+func updateRemainingWithAlmaLinuxFixes(disclosures result.Set, unaffectedResults result.Set) result.Set {
```
wagoodman: this really looks like it should be a result.Set.Merge() with a custom function to control behavior

grype/matcher/rpm/almalinux.go
wagoodman: ah, I see, you're leveraging the alias processing baked into Remove(). I wonder if that could be extracted and used as a filter function? (consider this non-blocking; I'm trying to find ways to make this easier to read and reason about)

grype/matcher/internal/result/results.go
```diff
@@ -213,6 +213,43 @@ func (s Set) Intersection(other Set) Set {
...
 }
 
+// UpdateByIdentity applies an update function to results in the current set when they match
+// (by ID or alias) with entries in the incoming set. This is useful for replacing specific fields
+// (like fix information) while preserving other fields (like match details).
+//
+// For example, if the current set has CVE-2006-20001 and the incoming set has ALSA-2023:0852
+// with alias CVE-2006-20001, the updateFunc will be called to update the CVE entry with data
+// from the ALSA entry.
+func (s Set) UpdateByIdentity(incoming Set, updateFunc func(existing *Result, incoming Result)) Set {
```
wagoodman: from our conversation, I feel like this (one day) could be a little more generic. This is a pretty specific functionality on a generic set implementation. We don't have to block on that now since this is in internal.

From what we chatted about the set intersectoin and getIdentity logic out and passed in as a "key function", thus this really changes from `UpdateByIdentity` to just `Update`. Odds are something like this can be done to `Merge`. TBD on removing of elements, but I feel that could be a separate call to `Filter` or `Remove` before hand.

grype/matcher/rpm/almalinux.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+
+	// constraint.Value() returns raw constraint without format suffix (e.g., ">= 2.4.48" instead of ">= 2.4.48 (rpm)")
+	constraintStr := constraint.Value()
+
+	// Handle common constraint patterns
+	// ">= version" → "version"
+	if strings.HasPrefix(constraintStr, ">= ") {
+		return strings.TrimPrefix(constraintStr, ">= ")
+	}
```
wagoodman:
```suggestion
	if strings.HasPrefix(constraintStr, ">=") {
		return strings.TrimSpace(strings.TrimPrefix(constraintStr, ">="))
	}
```

wagoodman: nit, we can use these https://github.com/anchore/grype/blob/main/grype/version/operator.go

grype/matcher/internal/result/results_test.go
```diff
@@ -522,6 +522,114 @@ func TestSet_Merge(t *testing.T) {
...
+								Advisories: []vulnerability.Advisory{
+									{ID: "ADV-2023-002", Link: "https://example.com/adv-002"},
+								},
+							},
+						},
+						Details: match.Details{{Type: match.ExactIndirectMatch}},
+					},
+				},
+			},
+			mergeFuncs: []func(existing, incoming []Result) []Result{
```
wagoodman: this test could probably be deleted

grype/matcher/internal/result/results_test.go
```diff
@@ -814,3 +922,548 @@ func TestSet_Contains(t *testing.T) {
...
+			)
+
+			// Step 5b: Update remaining with AlmaLinux fix info using UpdateByIdentity
+			got := filtered.UpdateByIdentity(tt.unaffectedResults, func(existing *Result, incoming Result) {
+				for i := range existing.Vulnerabilities {
+					for _, incomingVuln := range incoming.Vulnerabilities {
+						existing.Vulnerabilities[i].Fix = incomingVuln.Fix
+						existing.Vulnerabilities[i].Advisories = incomingVuln.Advisories
+					}
+				}
```
wagoodman: this feels kinda an odd/complicated way to test this

grype/matcher/internal/result/results_test.go
```diff
@@ -814,3 +922,548 @@ func TestSet_Contains(t *testing.T) {
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := tt.base.UpdateByIdentity(tt.incoming, func(existing *Result, incoming Result) {
+				// Update fix information from incoming
+				for i := range existing.Vulnerabilities {
+					for _, incomingVuln := range incoming.Vulnerabilities {
+						existing.Vulnerabilities[i].Fix = incomingVuln.Fix
+						existing.Vulnerabilities[i].Advisories = incomingVuln.Advisories
+					}
```
wagoodman: duplicate test?

grype/matcher/rpm/almalinux_test.go
```diff
@@ -0,0 +1,1689 @@
...
+				},
+			},
+			expectedVulnCount: 1,
+			expectedVulnIDs:   []string{"CVE-2023-5678"}, // CVE-2023-1234 filtered by alias match
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			filtered := tt.disclosures.Remove(tt.unaffectedResults)
```
wagoodman: I think this is already covered by https://github.com/anchore/grype/blob/v0.101.0/grype/matcher/internal/result/results_test.go#L118-L285 in the result package

grype/matcher/rpm/almalinux_test.go
```diff
@@ -0,0 +1,1689 @@
...
+			},
+			expectedCount: 1,
+			expectedIDs:   []string{"CVE-2023-5678"},
+			description:   "ALSA should filter GHSA by transitive alias",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			filtered := tt.disclosures.Remove(tt.unaffected)
```
wagoodman: probably the r=wrong subject under test

PR #2962, opened 2025-09-29
chore: add db schema validation

willmurphyscode:
Add a method for validating when models or blogs change in the v6 schema. Generate two schema files, a JSON schema describing the validation of blobs in the DB, and a SQL schema describing the GORM models.


spiffcs:
We might want to add a `README.md` or some other docs at `schema/grype/db/` explaining:
    - What these files are for
    - How to regenerate them
    - What to do when schema changes
    - Why both versioned and -latest files exist (highlight the drift comparison in ci?)

wagoodman (approved):
🚀 !!

I second what chris is saying around either a readme or more docs in an existing place.

PR #2973, opened 2025-10-03
chore: minor fixes to #2904 before merge

willmurphyscode:
#2904 needed a couple touch ups but wasn't maintainer editable. Pull in those commits and fix linting and handle merge conflict. Primarily this is about pulling in #2962 so that the db schema change is properly tracked and validated.


schema/grype/db/blob/json/schema-latest.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "$id": "anchore.io/schema/grype/db/blob/json/6.1.1",
+  "$id": "anchore.io/schema/grype/db/blob/json/6.1.2",
```
willmurphyscode, line 3: This might be an unintended consequence of the way we're doing schema numbering. The blobs didn't change, but we added a table, so we have a no-op bump in this number. Should we change it while it's still early days on publishing this schema?

grype/db/v6/vulnerability_decorator_store.go
```diff
@@ -14,29 +14,34 @@ import (
...
 	kevEnabled  bool
 	epssEnabled bool
+	cweEnabled  bool
 	epssDate    *time.Time
 }
 
 func newVulnerabilityDecoratorStore(db *gorm.DB, bs *blobStore, dbVersion schemaver.SchemaVer) *vulnerabilityDecoratorStore {
 	minSupportedKEVClientVersion := schemaver.New(6, 0, 1)
 	minSupportedEPSSClientVersion := schemaver.New(6, 0, 2)
+	minSupportedCWEClientVersion := schemaver.New(6, 0, 0)
```
wagoodman: we should put this under test similar to https://github.com/anchore/grype/blob/ff8a98609a3dd67fbab9231a28d428a9c97fdc44/grype/db/v6/vulnerability_decorator_store_test.go#L139

wagoodman: if we're interested in completion testing we could do some refactoring (this is not a blocking comment).

schema/grype/db/blob/json/schema-latest.json
wagoodman: after talking offline, versioning the schema + blobs together makes the most sense

wagoodman (approved):

PR #2976, opened 2025-10-03
Feat config missing epoch

### LINKED ISSUE #2964: Configurable option to ignore RPM epochs in version comparisons
I would like to be able to tell Grype to just ignore the epoch component of both the SBOM package and vuln records so that I can get fewer false positives for SBOMs that omit epochs from RPM package versions. This is primarily focused at using Grype to scan SBOMs in other formats and/or from other tools. This should not be the default behavior and the user should opt-in understanding the risks if used in contexts that are not appropriate.

**Why is this needed**:
When vulnerability scanning an SBOM generated by another tool (not Syft), the epoch component of RPM versions is often left off even when it shouldn't be (the actual package has epoch > 1). This makes it very difficult to get a good vulnerability scan result when I don't have the ability to influence the SBOM creation tool/process.

**Additional context**:

willmurphyscode:
fixes #2964

### DIFF
--- a/grype/matcher/rpm/rhel_eus_test.go
+++ b/grype/matcher/rpm/rhel_eus_test.go
@@ -1212,7 +1212,7 @@ func TestRedhatEUSMatches(t *testing.T) {
 
 			resultProvider := result.NewProvider(vulnProvider, tt.catalogPkg, match.RpmMatcher)
 
-			got, err := redhatEUSMatches(resultProvider, *tt.searchPkg)
+			got, err := redhatEUSMatches(resultProvider, *tt.searchPkg, "zero")
 			tt.wantErr(t, err)
 
 			if err != nil {
--- a/grype/version/comparator.go
+++ b/grype/version/comparator.go
@@ -1,5 +1,16 @@
 package version
 
+// ComparisonConfig contains configuration for version comparison behavior.
+type ComparisonConfig struct {
+	// MissingEpochStrategy controls how missing epochs in package versions are handled
+	// during vulnerability matching.
+	//
+	// Valid values:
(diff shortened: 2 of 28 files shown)


grype/matcher/rpm/rhel_eus_test.go
```diff
@@ -1212,7 +1212,7 @@ func TestRedhatEUSMatches(t *testing.T) {
 
 			resultProvider := result.NewProvider(vulnProvider, tt.catalogPkg, match.RpmMatcher)
 
-			got, err := redhatEUSMatches(resultProvider, *tt.searchPkg)
+			got, err := redhatEUSMatches(resultProvider, *tt.searchPkg, "zero")
```
wagoodman, line 1215: what is the reason for the test change here?

wagoodman: nvm

grype/version/comparator.go
```diff
@@ -1,5 +1,16 @@
...
 
+// ComparisonConfig contains configuration for version comparison behavior.
+type ComparisonConfig struct {
+	// MissingEpochStrategy controls how missing epochs in package versions are handled
+	// during vulnerability matching.
+	//
+	// Valid values:
+	//   - "zero": Treat missing epochs as 0 (default, backward compatible)
+	//   - "auto": Assume missing epoch matches the constraint's epoch
+	MissingEpochStrategy string
```
wagoodman: probably should be it's own type + enum if we're going this way

spiffcs:
Topic number 1 for livestream on Thursday of this week

PR #2980, opened 2025-10-07
feat: Multiple templated outputs

asgeirn:
Support multiple templated outputs by specifying template as part of the
--output argument.

To invoke this, use the following --output format:

--output template=tmpl=out

Signed-off-by: Asgeir Nilsen <asgeir@twingine.no>


asgeirn:
@westonsteimel care to have a look?

PR #3019, opened 2025-10-28
fix: junit template use CDATA block to prevent XML parse errors

nvtkaszpir:
Sometimes CVE descriptions contain chars that end up in being an error in XML parser, for example `<` and `>`.
Adding description in CDATA block prevents processing of such content and thus no more errors.

ref https://en.wikipedia.org/wiki/CDATA

### DIFF
--- a/templates/junit.tmpl
+++ b/templates/junit.tmpl
@@ -11,10 +11,14 @@
 <![CDATA[
 {{ html .Vulnerability.Description }}
 
-{{ .Artifact.CPEs }}
+CPEs:
+{{- range .Artifact.CPEs }}
+{{ . }}
+{{- end }}
 
+DataSource:
 {{ html .Vulnerability.DataSource }}
-            
+
 ]]></failure>
         </testcase>
         {{- end }}


templates/junit.tmpl
```diff
@@ -7,7 +7,15 @@
         </properties>
         {{- range .Matches }}
         <testcase classname="{{ .Artifact.Name }}-{{ .Artifact.Version }} ({{ .Artifact.Type }})" name="[{{ .Vulnerability.Severity }}] {{ .Vulnerability.ID }}">
-            <failure message="{{ .Artifact.Name }}: {{ .Vulnerability.ID }}" type="description">{{html .Vulnerability.Description }} {{ .Artifact.CPEs }} {{ .Vulnerability.DataSource }}</failure>
+            <failure message="{{ .Artifact.Name }}: {{ .Vulnerability.ID }}" type="description">
+<![CDATA[
+{{ html .Vulnerability.Description }}
+
+{{ .Artifact.CPEs }}
```
willmurphyscode: Does it make sense to loop over the CPEs and put a newline? Right now it looks like this:

``` xml
        </testcase>
        <testcase classname="libcrypto3-3.5.0-r0 (apk)" name="[Medium] CVE-2025-4575">
            <failure message="libcrypto3: CVE-2025-4575" type="description">
<![CDATA[


[cpe:2.3:a:libcrypto3:libcrypto3:3.5.0-r0:*:*:*:*:*:*:* cpe:2.3:a:libcrypto3:libcrypto:3.5.0-r0:*:*:*:*:*:*:* cpe:2.3:a:libcrypto:libcrypto3:3.5.0-r0:*:*:*:*:*:*:* cpe:2.3:a:libcrypto:libcrypto:3.5.0-r0:*:*:*:*:*:*:*]

https://security.alpinelinux.org/vuln/CVE-2025-4575

]]></failure>
        </testcase>
```

which makes the CPEs a little hard to read.

nvtkaszpir: Oh right, good catch, fixed in the next commit. If it looks good, then please resolve the conversation.

PR #3035, opened 2025-11-06
Add `--from` flag

wagoodman:
This brings grype up to parity with syft's `--from` flag so users do not need to specify a scheme in the user input and instead have the additional option to provide the same scheme value as a separate flag:

```
grype --from docker alpine:latest
```

This has the same semantics as in syft, so multiple values can be provided, allowing for multiple sources to be attempted in order:

```
# try podman first, if unsuccessful try docker
grype --from podman,docker alpine:latest
```

### DIFF
--- a/cmd/grype/cli/options/grype.go
+++ b/cmd/grype/cli/options/grype.go
@@ -2,7 +2,6 @@ package options
 
 import (
 	"fmt"
-	"sort"
 	"strings"
 
 	"github.com/anchore/clio"
@@ -218,7 +217,7 @@ func (o Grype) FailOnSeverity() *vulnerability.Severity {
 	return &severity
 }
 
-// flatten takes a list of comma-separated entries and returns a flattened list of trimmed values
+// flatten takes a list of comma-separated entries and returns a flattened list of trimmed values (preserving order)
 func flatten(commaSeparatedEntries []string) []string {
 	var out []string
 	for _, v := range commaSeparatedEntries {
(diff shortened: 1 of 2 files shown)


spiffcs (approved):

cmd/grype/cli/options/grype.go
```diff
@@ -207,3 +217,15 @@ func (o Grype) FailOnSeverity() *vulnerability.Severity {
...
+
+// flatten takes a list of comma-separated entries and returns a flattened list of trimmed values
+func flatten(commaSeparatedEntries []string) []string {
+	var out []string
+	for _, v := range commaSeparatedEntries {
+		for _, s := range strings.Split(v, ",") {
+			out = append(out, strings.TrimSpace(s))
+		}
+	}
+	sort.Strings(out)
```
willmurphyscode: I don't understand why we're sorting. Isn't the order of user input meaningful, for example try Podman, then try Docker? If so, shouldn't we preserve the input order rather than sort?

willmurphyscode (changes requested):

cmd/grype/cli/options/grype.go
wagoodman: You are correct -- I ported this as-is from syft (should have looked closer). I've got a [fix in syft](https://github.com/anchore/syft/pull/4350), I'll push a fix here.

wagoodman: pushed!

PR #3057, opened 2025-11-18
use syft for matching CPE fields to package type

chovanecadam:
Minor change here. `grype/internal/cpe_target_software_to_pkg_type.go` was moved to Syft [in this MR](https://github.com/anchore/syft/pull/4207) so here we make use of this.

PR #3058, opened 2025-11-18
use Syft for decoding CPEs

chovanecadam:
Makes use of new API in Syft for decoding CPEs and enables reading a file with CPEs for scan.

I deleted tests in `grype/pkg/cpe_provider_test.go` as they are the same as in syft (see [here](https://github.com/chovanecadam/syft/blob/main/syft/format/cpes/decoder_test.go)). It doesn't make sense to me to have the same test the same functionality in two places, but let me know if you want it some other way.

### DIFF
--- a/grype/pkg/cpe_provider_test.go
+++ b/grype/pkg/cpe_provider_test.go
@@ -84,6 +84,37 @@ func Test_CPEProvider(t *testing.T) {
 			},
 		},
 		{
+			name:      "takes CPE 2.3 format",
+			userInput: "cpe:2.3:a:apache:log4j:2.14.1:*:*:*:*:*:*:*",
+			context: Context{
+				Source: &source.Description{
+					Metadata: CPELiteralMetadata{
+						CPE: "cpe:2.3:a:apache:log4j:2.14.1:*:*:*:*:*:*:*",
+					},
+				},
+			},
+			pkgs: []Package{
+				{
+					Name:    "log4j",
+					Version: "2.14.1",
+					CPEs: []cpe.CPE{
+						cpe.Must("cpe:2.3:a:apache:log4j:2.14.1:*:*:*:*:*:*:*", ""),
+					},
+				},
+			},
+			sbom: &sbom.SBOM{
+				Artifacts: sbom.Artifacts{
+					Packages: pkg.NewCollection(pkg.Package{
+						Name:    "log4j",
+						Version: "2.14.1",
+						CPEs: []cpe.CPE{
+							cpe.Must("cpe:2.3:a:apache:log4j:2.14.1:*:*:*:*:*:*:*", ""),
+						},
+					}),
+				},
+			},
+		},
+		{
 			name: "takes multiple CPEs",
 			userInput: `cpe:/a:apache:log4j:2.14.1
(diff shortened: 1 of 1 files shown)


grype/pkg/cpe_provider.go
```diff
@@ -39,61 +42,3 @@ func getCPEReader(userInput string) (r io.Reader, ctx Context, err error) {
...
-	c, err := cpe.New(rawLine, "")
-	if err != nil {
-		return nil, nil, fmt.Errorf("unable to decode cpe %q: %w", rawLine, err)
-	}
-
-	syftPkg := pkg.Package{
-		Name:    c.Attributes.Product,
-		Version: c.Attributes.Version,
-		CPEs:    []cpe.CPE{c},
-		Type:    internal.CPETargetSoftwareToPackageType(c.Attributes.TargetSW),
```
kzantow, line 86: Should the [CPETargetSoftwareToPackageType](https://github.com/anchore/grype/blob/b1706ca14e88428cddb30ef72a09867b9df09851/grype/internal/cpe_target_software_to_pkg_type.go#L11) get removed now, too since it's in syft?

chovanecadam: yeah, I did that in a separate MR for some reason https://github.com/anchore/grype/pull/3057

kzantow: Sorry it's taken a little while to get back to you. Would you mind combining these two pull requests? I think they should go in together, in the same commit. And there's one small outstanding request on the other, have you had a chance to look at that?

chovanecadam: sorry this took that long. I reflected the requests in 3057 here, put it into the first chore commit, and rebated onto main.

grype/pkg/cpe_provider_test.go
```diff
@@ -1,169 +0,0 @@
...
-		name      string
-		userInput string
-		context   Context
-		pkgs      []Package
-		sbom      *sbom.SBOM
-		wantErr   require.ErrorAssertionFunc
-	}{
-		{
-			name:      "takes a single cpe",
-			userInput: "cpe:/a:apache:log4j:2.14.1",
```
kzantow, line 28: I think most of this test can be removed, but we should leave at least one `cpe:` example to validate that it is correctly returning the expected package; maybe one with each CPE format `cpe:2.3:a:apache:log4j:2.14.1:*:*:*:*:*:*:*` and `cpe:/a:apache:log4j:2.14.1`. I think this single-CPE case is not tested if this file is removed, is it tested elsewhere?

chovanecadam: The file with test cases is in Syft already (see [here](https://github.com/chovanecadam/syft/blob/main/syft/format/cpes/decoder_test.go)), so I removed it to not replicate code. I can add a few sanity-check test cases for a single CPEs in both formats and multiple CPEs. Sounds good?

kzantow: Right, acknowledge that Syft testing is important and handles most of the cases, but by removing this test file we don't have anything directly testing this `cpeProvider`, right?

chovanecadam: agreed, will do

chovanecadam:
How can we move this forward? Is there anything I can do? @kzantow

PR #3137, opened 2025-12-30
Add Root IO vulnerability qualifier support

chait-slim:
Adds support for Root IO package qualification to enable the NAK
  (Negative Acknowledgment) pattern, where vulnerabilities with Root IO
  qualifier only match Root IO patched packages.

  Implementation:
  - Add RootIO field to PackageQualifiers in grype/db/v6/blobs.go
  - Increment DB schema Addition to 6.1.4 with changelog
  - Add rootio qualifier implementation in grype/pkg/qualifier/rootio/
  - Wire up qualifier in vulnerability.go toPackageQualifiers()

  The Root IO qualifier ensures that Root IO vulnerabilities (from
  api.root.io/external/osv) only match packages with rootio- prefix,
  preventing false matches against standard upstream packages.


grype/db/v6/build/transformers/osv/transform_test.go
```diff
@@ -712,14 +712,151 @@ func Test_getPackageQualifiers(t *testing.T) {
...
+					require.Equal(t, "osv", relatedEntries.VulnerabilityHandle.ProviderID)
+					require.NotNil(t, relatedEntries.VulnerabilityHandle.BlobValue)
+
+					// Verify related packages
+					require.NotEmpty(t, relatedEntries.Related, "should have at least one related package")
+
+					for _, rel := range relatedEntries.Related {
+						aph, ok := rel.(db.AffectedPackageHandle)
+						require.True(t, ok, "related entry should be AffectedPackageHandle")
+						require.NotNil(t, aph.Package, "Package should not be nil")
```
willmurphyscode: Can you make these assertions specific? "not empty" type tests are easy to pass with an incorrect value.

willmurphyscode:
Hi @chait-slim I spent some time testing this out this week. I think it's missing changes to actual matching behavior. In other words, this change puts new rootio records in the database, but nothing queries for them.

I think probably what we want to:

1. At the top of https://github.com/anchore/grype/blob/main/grype/matcher/dpkg/matcher.go#L39 , ask, "are we scanning a rootio package" and if so jump into a new matcher that
2. Trims the rootio-prefix from the package name and searches the appropriate ubuntu vulns
3. Searches rootio for appropriate unaffected packages, and subtracts fixes
4. Trims the rootio-prefix from the _upstream_ package's name and search the appropriate ubuntu vulns
5. Search rootio for appropriate unaffected packages on the upstream, and subtracts fixes
6. Returns the matching results of these.

We will probably need to do something similar in the APK matcher and other supported rootio ecosystems.

grype/db/v6/build/transformers/osv/transform_test.go
willmurphyscode: Also, I expected that rootio data would make an unaffected package handle, not an affected package handle.

grype/matcher/apk/matcher.go
```diff
@@ -34,34 +35,59 @@ func (m *Matcher) Type() match.MatcherType {
 }
 
 func (m *Matcher) Match(store vulnerability.Provider, p pkg.Package) ([]match.Match, []match.IgnoreFilter, error) {
+	if rootio.IsRootIOPackage(p) {
```
willmurphyscode, line 38: We don't have adequate test coverage for this change. 

1. Can we get an Alpine-based test image and labels added to https://github.com/anchore/vulnerability-match-labels/pull/167 ?
2. This needs a to be well unit tested.

grype/matcher/dpkg/matcher.go
```diff
@@ -37,6 +38,10 @@ func (m *Matcher) Type() match.MatcherType {
 }
 
 func (m *Matcher) Match(store vulnerability.Provider, p pkg.Package) ([]match.Match, []match.IgnoreFilter, error) {
+	if rootio.IsRootIOPackage(p) {
```
willmurphyscode, line 41: Changes to matchers need good unit tests.

grype/matcher/internal/distro.go
```diff
@@ -147,3 +96,72 @@ func distroMatchDetails(upstreamMatcher match.MatcherType, searchPkg pkg.Package
...
+		OnlyVulnerableVersions(pkgVersion),
+		OnlyNonWithdrawnVulnerabilities(),
+		search.ForUnaffected(),
+	}
+	naks, err := provider.FindResults(nakCriteria...)
+	if err != nil {
+		return nil, nil, fmt.Errorf("matcher failed to fetch rootio naks distro=%q pkg=%q: %w", p.Distro, p.Name, err)
+	}
+
+	remaining := disclosures.Remove(naks)
```
willmurphyscode, line 164: This depends on aliases, but in my testing the aliases seem to be missing from the rootio unaffected package handles, so nothing gets removed. For example, in the ubuntu 22 test images from the vuln match labels PR has `rootio-libgcrypt20` at  `1.9.4-3ubuntu3.root.io.2` in it, which is listed as fixed at  https://osv.dev/vulnerability/ROOT-OS-UBUNTU-2204-CVE-2024-2236, but grype built from this branch finds `CVE-2024-2236` on that package anyway.

grype/db/v6/build/transformers/osv/transform.go
```diff
@@ -364,24 +366,45 @@ func getPackage(p models.Package) *db.Package {
...
 // getPackageTypeFromEcosystem determines package type from OSV ecosystem
-// Currently only supports AlmaLinux; other ecosystems use PURL-based detection
+// Supports AlmaLinux and Root IO OS ecosystems (Alpine, Debian, Ubuntu)
+// Also supports Root IO language ecosystems (npm, pypi, maven)
 func getPackageTypeFromEcosystem(ecosystem string) pkg.Type {
 	if ecosystem == "" {
 		return ""
 	}
 
+	ecosystemLower := strings.ToLower(ecosystem)
```
willmurphyscode, line 376: I think I'd prefer to use a narrower change here. There are a lot of changes in this file that affect all OSV providers, but are implemented for RootIO. I think it makes sense to instead try to have something like:
``` go
if isRootIO {
    return handleRootIO(...)
}
```

pretty early in the OSV transformers. This helps everyone out: other OSV providers' bugs can't break RootIO (and vice versa) and reviewers don't have to spend much focus on whether changes to shared parts of the transformer will break other providers.

grype/db/v6/build/transformers/osv/transform.go
```diff
@@ -57,14 +58,11 @@ func Transform(vulnerability unmarshal.OSVVulnerability, state provider.State) (
 		},
 	}
 
-	// Check if this is an advisory record
-	if isAdvisory {
-		// For advisory records, emit unaffected packages
+	if isAdvisory || isRootIO {
```
willmurphyscode, line 61: https://github.com/anchore/vunnel/pull/963/changes#diff-e3183aa5564d1be5757075e9114b9ca21572612f5cb34afdb6a08f0091dae127R132 should be marking this as advisory type records. I think maybe your test data in this repo is missing that normalization.

grype/db/v6/build/transformers/osv/transform.go
willmurphyscode, line 39: I think the origin of the bug causing FPs (see https://github.com/anchore/grype/pull/3137/changes#r3065870827 ) is that the Upstream and Related vulnerabilities on your OSV data aren't being added to the aliases here, so the resultSet.Remove call is skipping them (Because `ROOT-OS-UBUNTU-2204-CVE-2024-2236` is not the same as `CVE-2024-2236` and there's no entry added to the alias table).

The fix for this should probably be basically to around line 36 say, "this is a rootio advisory and we are going to go transform it in handleRootIO(...)". I think that will make this change safer and easier to reason about.

grype/matcher/internal/language.go
```diff
@@ -17,7 +18,20 @@ func MatchPackageByLanguage(store vulnerability.Provider, p pkg.Package, matcher
 	var matches []match.Match
 	var ignored []match.IgnoreFilter
 
-	for _, name := range store.PackageSearchNames(p) {
+	searchNames := store.PackageSearchNames(p)
```
willmurphyscode, line 21: Changes to matchers need good unit tests.

grype/db/v6/build/transformers/osv/test-fixtures/ROOT-APP-NPM-CVE-2022-25883.json
```diff
@@ -0,0 +1,52 @@
+{
+  "schema_version": "1.6.1",
+  "id": "ROOT-APP-NPM-CVE-2022-25883",
```
willmurphyscode, line 3: It would be wonderful to get npm and pypi test cases added to https://github.com/anchore/vulnerability-match-labels/pull/167 and we need npm and pypi records at least handled by unit tests in this repo.

willmurphyscode (commented):
Thanks for putting some more work into this! It's much closer. I have quite a few inline comments, but we're moving in the right direction.

I do think that https://github.com/anchore/vulnerability-match-labels/pull/167 could use some more images _and_ a lot more labels. I left instructions at https://github.com/anchore/vunnel/pull/963/changes#r3065955122 about why some matches aren't showing up in the diff to be labeled.

chait-slim:
@willmurphyscode
1. Fixed all comments
2. Added a bunch of images and Root FPs as labels for different ecosystems (npm,alpine,python)

grype/db/v6/build/transformers/osv/transform_rootio.go
```diff
@@ -0,0 +1,240 @@
...
+	var uphs []db.UnaffectedPackageHandle
+	for _, affected := range vuln.Affected {
+		ecosystem := string(affected.Package.Ecosystem)
+		pkgType := rootioPackageType(ecosystem)
+		if pkgType == "" {
+			// rootio ships occasional records for ecosystems the transformer
+			// hasn't been taught about (Go, NuGet, the synthetic
+			// Root:Ubuntu:plucky test record). Storing them would emit DB
+			// entries the matcher can't use; skip with a warning so the
+			// drift is visible.
```
wagoodman, line 112: I think I might be misunderstanding something here but this comment suggests that we should be skipping ubuntu:plucky, but I think that is not the case. That is, `rootioPackageType("Ubuntu:plucky")` returns `"deb"`, not `""`.

grype/internal/rootio/rootio.go
```diff
@@ -0,0 +1,150 @@
...
+	if len(rev) < 5 {
+		return false
+	}
+	for i := 0; i < len(rev); i++ {
+		if rev[i] < '0' || rev[i] > '9' {
+			return false
+		}
+	}
+	return true
+}
```
wagoodman, line 236: I think we've chatted about this but I want to ask again since it's been a while: is the heuristic of there being 5 or more digits a safe one? That is, will adding the 5-digit apk rev heuristic cause false-positives non-rootio package... e.g. `util-linux@2.38.1-r10000` and `musl@1.2.5-r100001` (both made up) are both flagged as `rootio` with no `rootio` name prefix and no `root.io` marker anywhere.

grype/db/v6/build/transformers/osv/transform_rootio.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	osVersion := parts[1]
+	versionFields := strings.Split(osVersion, ".")
+	if len(versionFields) == 0 || versionFields[0] == "" {
+		return nil
+	}
+
+	major := versionFields[0]
+	if _, err := strconv.Atoi(major[:1]); err != nil {
```
wagoodman: Checking `unicode.IsDigit(rune(major[0]))` would be a tad safer in case we ever encounter non-ascii sequence one day (this is super contrived I know, so feel free to ignore)

grype/db/v6/build/transformers/osv/transform_rootio.go
willmurphyscode: The expected namespaces look like `Root:Ubuntu:24.04` for Ubuntu, but there were 1 or two records with "plucky" in them that seemed like test data. I'll probably just filter them (or normalize codename to version) in vunnel.

grype/internal/rootio/rootio.go
wagoodman: answer: yes

wagoodman (approved):

wagoodman (approved):

PR #3146, opened 2026-01-06
fix: preserve local version segment in constraints for PEP 440 comparison

willmurphyscode:
PR #3121 fixed an issue where 6.4+cgr.1 wasn't matching <= 6.4.0 by stripping local version segments during comparison. However, this introduced a regression: constraints with local segments (like = 2.0.0+cgr.1) incorrectly matched versions without that local segment (like 2.0.0).

This caused false negatives where upstream packages were incorrectly marked as "unaffected" by downstream distributor patches. For example, flask==2.0.0 from PyPI was not flagged for GHSA-m2qf-hxjv-5gpq because Chainguard's unaffected entry for = 2.0.0+cgr.1 was incorrectly matching it.

The fix:
- Package versions: local segment is still ignored (preserving #3121 fix)
- Constraint versions: local segment is now preserved and compared using PEP 440 rules

This ensures:
- 6.4+cgr.1 matches <= 6.4.0 ✓ (original fix preserved)
- 2.0.0 does NOT match = 2.0.0+cgr.1 ✓ (regression fixed)
- 2.0.0+cgr.1 matches = 2.0.0+cgr.1 ✓ (exact match works)
- 2.0.0+cgr.12 > 2.0.0+cgr.2 ✓ (proper numeric comparison per PEP 440)

### DIFF
--- a/grype/version/pep440_version.go
+++ b/grype/version/pep440_version.go
@@ -7,7 +7,10 @@ import (
 var _ Comparator = (*pep440Version)(nil)
 
 type pep440Version struct {
-	obj goPepVersion.Version
+	// public is the public portion of the version (without local segment), used for most comparisons
+	public goPepVersion.Version
+	// full is the complete parsed version including local segment, used when constraint has local
+	full goPepVersion.Version
 }
 
 func newPep440Version(raw string) (pep440Version, error) {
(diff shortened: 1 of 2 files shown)


wagoodman (approved):

mosabua:
Thank you for fixing this so quickly!

Issue #1409, opened 2023-07-31
Should we combine the grype and grype-db projects?


wagoodman:
This has come up a couple of times in ad-hoc conversations so I wanted to try and get this in a ticket for more feedback. Today we have a project for `grype` (the application that matches packages and vulnerabilities) and `grype-db` (the application that builds the DB of vulnerabilities that grype uses). 

We could consider combining these projects (where the grype-db codebase is merged into the grype repo). There are at least a couple ways this can go:
- Add another entrypoint (build two binaries): ./cmd/grype (today's) and ./cmd/grype-db 
- Update the `grype` application with more subcommands: `grype db build ...`

It might mean that the workflow that uses grype / grype-db to build OSS databases nightly remains where it is or is also migrated... this would be TBD.

To be clear: this is purely speculative. At the current time there is no plan to make this change (as it is a lot of work), but again, since it's been asked a couple times I wanted to see if there was anyone out there with strong opinions about this and gather as much feedback as possible.

willmurphyscode:
What are the advantages of combining them? When the question of combining them comes up, why does it come up?

wagoodman:
Today the DB definitions live in grype. But it's grype DB that needs these for all schema versions. One conclusion is to move DB definitions into grype-db, however, DB concerns like namespace are tightly coupled to the definition of a grype package... so if we move these definitions we end up with a go module cycle, which is technically allowed, but highly discouraged.

We also have the problem of when making specific updates exactly what needs to be merged and released first before being able to continue. This is especially true with new DB schemas. Merging these repos does not eliminate the issue, but does alleviate it some: it's one less repo to coordinate such changes.

wagoodman:
My latest thoughts is that combining these projects is the way to go. Specifically:
- Migrate all golang code to the grype repo -- adding either a new binary or new subcommand in grype
- Keep all python (`manager`) code and github workflows for operationalizing the DB nightly in the grype-db repo. Probably would move the manager code up to the root of the repo. This also means keeping all operations cache (OCI images with data sync results) in the grype-db repo.

This way:
- there is one place where DB readers and writers are defined and used (not cross repo). This allows us to internalize some (most? all?) of those concerns, simplifying the public API in grype (something needed for 1.0)
- there are no grype updates that could break grype-db, since all PR validations for both tools would run at the same time (in the grype repo)
[...]

PR #3149, opened 2026-01-07
Port grype-db lib to grype

wagoodman:
This partially addresses [1409](https://github.com/anchore/grype/issues/1409). 

This is taking the majority of the `pkg/` package from `anchore/grype-db` and migrating it to here. This is mainly a lift-and-shift of the `build` and `package` aspects of grype-db (`pull` will remain in grype-db).

Here is the [docs update](https://github.com/anchore/oss-docs/pull/136) for this change.

## What's being added to grype

- **Transformers/Unmarshalers/Processors** for all the vulnerability data sources we ingest (NVD, OSV, OS advisories, GitHub, MSRC, OpenVEX, etc.)
- **DB writers** and build infrastructure
- **Codename lookup** generation (the debian/ubuntu version-to-codename mapping) + weekly workflow to keep codename data up to date via automated PRs.

A similar question, what is NOT being added to grype:
- anything related to running vunnel or the `pull` process in grype-db
- DB acceptance tests, these will remain in grype-db for the meantime

## Why do this?

1. The grype-db lib has a lot of code that is responsible for taking data and writing it to the DB. Today the DB schema and surrounding code lives in the grype repo. This can lead to odd situations where you've made a change to the DB code in grype and released it but it doesn't function practically in the grype-db repo, what is responsible for making production DBs. This is the first step in moving all of the logic needed to write to a DB in one spot, so that we can catch these kinds of issues as early as possible.

2. Unit/Integration tests could benefit from being able to build realistic DBs from test fixture data in the same manner that we build DBs in production. Today we utilize a mocking layer, which is very cumbersome and not providing a lot of value. By cumbersome, I mean that it takes many lines of code to express a test + test-fixture (which needs to be manually written). By "not providing a lot of value" I mean that it is easy to write a test that will pass but does not reflect what will happen in production (thus is not a good safety net to continue to develop against).

3. Simplifies development workflow. Instead of needing to implement logic across two repos, it's implemented in one (grype). Grype-db still needs to pull in the changes, but only via a go-module bump, not with additional code/test review.

## Questions

- **Will we migrate the remaining grype-db functionality (commands) into grype?**... Maybe. There are good reasons for and against this. The most minimal/impactful action is to get the underlying DB writing functionality migrated to grype in order to improve our test posture.

- **Does this refactor tests with a new DB build infrastructure?** Not yet, that will be a follow up PR(s).

- **Does this internalize the new API?** Not yet, only a few areas have been put into `internal` packages. This will be a focus of the grype 1.0 work as a whole (more things will be put into `internal` over time).

- **Will this new code be removed from grype-db?** [Yes](https://github.com/anchore/grype-db/pull/794).

### DIFF
--- a/.github/workflows/validations.yaml
+++ b/.github/workflows/validations.yaml
@@ -281,6 +281,9 @@ jobs:
 
       - name: Run CLI Tests (Linux)
         run: make cli
+        env:
+          GRYPE_SNAPSHOT_PREBUILT: "true"
+          GRYPE_BINARY_LOCATION: ${{ github.workspace }}/snapshot/linux-build_linux_amd64_v1/grype
 
   Cleanup-Cache:
     name: "Cleanup snapshot cache"
--- a/Taskfile.yaml
+++ b/Taskfile.yaml
@@ -221,12 +221,20 @@ tasks:
       # exercise most of the CLI with the data race detector enabled
       - "go run -race cmd/{{ .PROJECT }}/main.go alpine:latest"
 
+  _ensure-snapshot:
+    # Ensure the snapshot binary is available, building only if needed.
+    # In CI, set GRYPE_SNAPSHOT_PREBUILT=true to skip the rebuild when the binary was restored from cache.
+    # Locally, this falls through to the snapshot task which uses checksum-based staleness detection
+    # so that code changes always produce a fresh binary.
+    internal: true
+    status:
+      - test -n "$GRYPE_SNAPSHOT_PREBUILT" -a -f "{{ .SNAPSHOT_BIN }}"
+    cmds:
+      - task: snapshot
+
   cli:
     desc: Run CLI tests
-    # note: we don't want to regenerate the snapshot unless we have to. In CI it's probable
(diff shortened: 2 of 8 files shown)


.github/workflows/update-generated-code.yml
```diff
@@ -0,0 +1,47 @@
...
+      - uses: peter-evans/create-pull-request@98357b18bf14b5342f975ff684046ec3b2a07725 #v8.0.0
+        with:
+          signoff: true
+          delete-branch: true
+          branch: auto/latest-codename-data
+          labels: dependencies
+          commit-message: "chore(deps): update OS codename data"
+          title: "chore(deps): update OS codename data"
+          body: |
+            Update OS codename data from endoflife.date
```
kzantow: Is this using endoflife.date?

grype/db/internal/tarutil/file_entry.go
```diff
@@ -0,0 +1,37 @@
+package tarutil
```
kzantow, line 1: It's unfortunate this became duplicated code. Note: populate is no longer used and could be removed from grype-db, I think.

PR #3154, opened 2026-01-09
Add archlinux matcher to grype

willmurphyscode:


grype/db/v6/data.go
```diff
@@ -40,9 +40,9 @@ func KnownOperatingSystemSpecifierOverrides() []OperatingSystemSpecifierOverride
 		{Alias: "chainguard", Rolling: true},
 
 		// others
-		{Alias: "arch", Rolling: true},
+		{Alias: "archlinux", Rolling: true},
 		{Alias: "minimos", Rolling: true},
-		{Alias: "archlinux", ReplacementName: strRef("arch"), Rolling: true}, // non-standard, but common (dockerhub uses "archlinux")
+		{Alias: "arch", ReplacementName: strRef("archlinux"), Rolling: true}, // os-release ID=arch, but namespace uses archlinux
```
willmurphyscode, line 45: To avoid changing https://github.com/anchore/grype/blob/56752cec9f012d9987a0e680c816fad85a061af8/grype/distro/type.go#L22

kzantow: Right, the operating system in the db is `archlinux` 👍

kzantow (approved):
LGTM

PR #3156, opened 2026-01-11
Fix vulnerability ignore matching to be case-insensitive

RohitYandigeri:
### Summary
Vulnerability identifiers such as CVE and GHSA are case-insensitive, but ignore
rules were previously matched using exact string comparison. This caused ignore
rules to fail when casing differed between sources.

### Root Cause
Ignore rule evaluation compared vulnerability IDs using strict string equality,
while vulnerability sources (e.g., Grype data, GHSA) may report identifiers in
mixed or lowercase formats.

### Changes
- Make vulnerability ID matching case-insensitive when applying ignore rules
- Add a unit test validating that GHSA identifiers are ignored regardless of casing

### Impact
- Fixes ignore rules not applying to GHSA vulnerabilities with lowercase IDs
- Backward compatible; does not affect existing ignore behavior

### Notes
Local full test execution could not complete due to disk space constraints on the
development machine. The change is isolated to `grype/match`, and CI will validate
the full test suite.

### DIFF
--- a/grype/match/ignore.go
+++ b/grype/match/ignore.go
@@ -2,6 +2,7 @@ package match
 
 import (
 	"regexp"
+	"strings"
 
 	"github.com/bmatcuk/doublestar/v2"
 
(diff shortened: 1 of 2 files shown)


willmurphyscode:
Hi @RohitYandigeri thanks for the PR! Would you mind signing off your commits? See https://oss.anchore.com/docs/contributing/sign-off/#how-to-sign-off

Also, since you're pushing anyway, it would help if you merged / rebased on main to resolve conflicts. Thanks!

Issue #3167, opened 2026-01-15
VEX Documents still not working with syft sbom


dariozachow:
**What happened**:

When I was trying out to use a OpenVEX Document with a sbom created by syft. The same error as here https://github.com/anchore/grype/issues/2471 occurs. I get an error saying VEX is not supported.

**What you expected to happen**:

This should be working because it is stated, that grype would support VEX and especially with syft it should work without any Problems.
With some time and digging into the code changing one line of the implementation.go (https://github.com/anchore/grype/blob/main/grype/vex/openvex/implementation.go#L70) to dont throw an error it works perfectly fine. 
Is it just that easy or I am overlooking something here?

**How to reproduce it (as minimally and precisely as possible)**:

same steps as in the mentioned Issue.

dariozachow:
I created a PR to address this Issue. Maybe this is enough to support VEX properly: https://github.com/anchore/grype/pull/3168

kzantow:
Hey @dariozachow 👋 good to meet you during office hours!

We definitely want to get VEX working, too!

I have to preface this with I don't fully understand OpenVEX, but from reading the spec, it seems to me that a _product_ could be an _image_ or a _package_. The spec has this description: _"The product is a piece of software that can be correlated to an entry in an SBOM"_, which sounds like directly describing a package is a perfectly valid use. We have a few cases that I think should match grype results:

- if _product_ is an image _and there are no subcomponents_ and _the image matches the SBOM image source_, it seems like the statements apply to the entire scan
- if _product_ is an image _and there are subcomponents_ and _the image matches the SBOM image source_, it seems like we should add rules to make the statements only apply to the specific packages that match the subcomponents (I believe this is the current behavior)
- if product _is not an image_ and _has no subcomponents_, it should apply to any matching packages during the scan

There's at least one more case that we could support in the future but I don't think we need to support right away: if a non-image product is defined _with subcomponents_, we would need to look at relationships to other packages to determine this, but the Grype data model doesn't have an easy way to do this, so I wouldn't expect this to be done right away.

Without taking into account package-to-package relationships, it seems to me like we should do something like:
```
sbomIdentifier = // get the SBOM's identifier, image ID or named identifier (what happens today, but should not return an error)

// iterate over all statements
if product == sbom identifier:
  if no subcomponents:
    // make rule that applies to all packages
  else
    // make individual rules for each subcomponent to apply to the specific packages
else
  // make rule that applies to the specific packages matching the product
```

[...]

PR #3168, opened 2026-01-15
fix: improve VEX product and subcomponent matching

dariozachow:


kzantow:
Hey @dariozachow, I left [a comment on the issue](https://github.com/anchore/grype/issues/3167#issuecomment-3756169786) about how I think OpenVEX should behave. Does this match your understanding? If you feel like making some changes to support this, that would be great; I can also push some changes to your PR here if you like

dariozachow:
@kzantow forgot to sign it off but should be working now

spiffcs:
Thanks @dariozachow for the implementation, following the feedback from the maintainers, and coming to office hours to help improve how grype handles vex documents.

I took some time to try and add tests around the cases that have been roughly outlined:

1) Image product, no subcomponents → statement applies to entire scan
2) Image product, with subcomponents → statement applies only to listed subcomponent packages
3) Package product, no subcomponents → statement applies to any matching packages

Case 1 and 2 worked very well with this branch.

I saw some bugs when trying out 3 and wanted to add test cases to validate what I think is the correct behavior.

First, when scanning an image with a VEX doc listing a package as the product (no subcomponents), the statement was silently ignored. The  previous `productIdentifierFromVEX` fallback would never trigger, and the package PURL never gots matched.

Second, when scanning a directory with a VEX doc listing a package as the product we would see over-matching. `productIdentifierFromVEX` would pass the PURL as both the Vex Product ID and the product argument when doing `doc.Matches()`

`go-vex's` `Product.Matches` ignores the subIdentifier when the VEX product has no subcomponents causing ALL matches for that CVE to be filtered.

I replaced `productIdentifierFromVEX` with a two pass matching strategy:
1) This is what currently existed. Try SBOM/context identifiers as the product (handles image-as-product cases)

2) Second pass which is new behavior. If no match, try the package PURL as the product argument to `doc.Matches()` with nil subcomponents (handles package-as-product cases).

`go-vex's` `Component.Matches` checks if the VEX product's ID matches the passed identifier. 

When we pass the package PURL, only VEX products whose ID matches that specific package will match. Other packages for the same CVE won't match which prevents the over-matching.

I'm unsure if I should also update `processor_test.go` or if the changes to `implementation_test.go` are enough to cover the cases we outlined in the issue and during our discussion of this.

dariozachow:
@kzantow yes perfect works as intended. I was opening an Issue https://github.com/anchore/grype/issues/3212 on Friday because I saw this bug aswell. But this works now too. Thanks a lot for your help :)

PR #3171, opened 2026-01-16
feat: warn about packages from EOL distros

willmurphyscode:
``` sh
$ grype debian:buster-slim
....
84 packages from EOL distro "debian 10" - vulnerability data may be incomplete or outdated; consider upgrading to a supported version
```

See https://github.com/anchore/vunnel/pull/973 and https://github.com/anchore/grype-db/pull/817

### DIFF
--- a/cmd/grype/cli/commands/root.go
+++ b/cmd/grype/cli/commands/root.go
@@ -294,7 +294,6 @@ func warnDistroAlerts(data *models.DistroAlertData) {
 	for distroName, count := range countPackagesByDistro(data.EOLDistroPackages) {
 		msg := fmt.Sprintf("%d packages from EOL distro %q - vulnerability data may be incomplete or outdated; consider upgrading to a supported version", count, distroName)
 		bus.Notify(msg)
-		log.Warn(msg)
 	}
 }
 


grype/presenter/models/document.go
```diff
@@ -84,20 +77,97 @@ func NewDocument(id clio.Identification, packages []pkg.Package, context pkg.Con
...
+		addAlert(p, AlertTypeDistroEOL, fmt.Sprintf("Package is from end-of-life distro: %s", distroString(p)), distroMetadata(p))
+	}
+
+	// convert map to slice
+	if len(alertsByPkg) == 0 {
+		return nil
+	}
+
+	result := make([]PackageAlerts, 0, len(alertsByPkg))
+	for _, pa := range alertsByPkg {
```
spiffcs, line 166: I think this should be sorted so that when we get the results from the range over the map they are consistent in the output. Right now the order read from `map` is non deterministic.

spiffcs:
🟢 with rebase and determinism comment addressed

grype/presenter/models/document.go
willmurphyscode: It is sorted now.

PR #3176, opened 2026-01-21
Add CycloneDX v1.5 writing

RossComputerGuy:
It is useful to be able to write CycloneDX v1.5 VEX since the CycloneDX Rust crate does not support v1.6 yet (ref: https://github.com/CycloneDX/cyclonedx-rust-cargo/issues/769). Rather than waiting for the crate to support that, or spending the time to implement it, it is simpler to just make Grype support it.

### DIFF
--- a/internal/format/format.go
+++ b/internal/format/format.go
@@ -4,54 +4,74 @@ import (
 	"strings"
 )
 
-const (
-	UnknownFormat   Format = "unknown"
-	JSONFormat      Format = "json"
-	TableFormat     Format = "table"
-	CycloneDXFormat Format = "cyclonedx"
-	CycloneDXJSON   Format = "cyclonedx-json"
-	CycloneDXXML    Format = "cyclonedx-xml"
-	SarifFormat     Format = "sarif"
-	TemplateFormat  Format = "template"
+var (
+	UnknownFormat       = Format{name: "unknown", version: ""}
+	JSONFormat          = Format{name: "json", version: ""}
+	TableFormat         = Format{name: "table", version: ""}
+	CycloneDXFormat     = Format{name: "cyclonedx", version: ""}
+	CycloneDXJSON       = Format{name: "cyclonedx-json", version: ""}
+	CycloneDXXML        = Format{name: "cyclonedx-xml", version: ""}
+	CycloneDXFormatv1_5 = Format{name: "cyclonedx", version: "1.5"}
+	CycloneDXJSONv1_5   = Format{name: "cyclonedx-json", version: "1.5"}
+	CycloneDXXMLv1_5    = Format{name: "cyclonedx-xml", version: "1.5"}
+	SarifFormat         = Format{name: "sarif", version: ""}
+	TemplateFormat      = Format{name: "template", version: ""}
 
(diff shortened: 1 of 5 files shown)


internal/format/format.go
```diff
@@ -5,18 +5,21 @@ import (
...
-	CycloneDXXML    Format = "cyclonedx-xml"
-	SarifFormat     Format = "sarif"
-	TemplateFormat  Format = "template"
+	UnknownFormat   		Format = "unknown"
+	JSONFormat      		Format = "json"
+	TableFormat     		Format = "table"
+	CycloneDXFormat 		Format = "cyclonedx"
+	CycloneDXJSON   		Format = "cyclonedx-json"
+	CycloneDXXML    		Format = "cyclonedx-xml"
+	CycloneDXFormatv1_5	Format = "cyclonedx-v1.5"
```
kzantow: I don't think we should use this format exactly. We have handling in Syft for formats in different versions with the syntax: `<format>@<version>`, for CycloneDX, it would be `cyclonedx-json@1.5`. There is some logic to handle versions nicely including finding the latest version where only a major version is specified, for example `spdx-json@2` results in `spdx-json` version `2.3`. It would be nice to reuse this logic, [but it looks like it's somewhat specific to an `sbom.FormatEncoder`](https://github.com/anchore/syft/blob/795432441754b910d4456fa982eb7df64c98b432/syft/format/encoders_collection.go#L105). In the meantime, we should use an identical syntax in Grype, e.g.: `cyclonedx@1.5`, `cyclonedx-json@1.5`, and `cyclonedx-xml@1.5`

RossComputerGuy: I wasn't aware of there being a specific format to the versioning. Would it be good to change `Format` into a struct and use some of the logic from `FormatEncoder`?

kzantow: We want to update the Grype presenters to be more like Syft formats, but I wouldn't ask you to do this as part of your PR here unless you felt like it. The main thing is the version strings specified should be the same between the tools

RossComputerGuy: Yep, that makes total sense to me. I'm not fully familiar with Go so I'll probably just update the formatting stuff to handle the version.

RossComputerGuy: Ok, I've gotten it to use the Syft formatted version string. Not the prettiest implementation but it's straightforward and it works.

kzantow:
Apologies for not running the tests earlier -- it looks like there are [some static analysis failures](https://github.com/anchore/grype/actions/runs/21224874100/job/61897000484?pr=3176#step:4:85) -- these seem straightforward: unnecessary conversion and `Format` specifier.  And some [test failures](https://github.com/anchore/grype/actions/runs/21224874100/job/61897000435?pr=3176#step:4:88); are you able to get these cleaned up?

RossComputerGuy:
Great, I believe those should be fixed now. I tried running `make unit` locally but it fails because it doesn't like my systems being aarch64 so Docker things fail.

spiffcs (approved):

RossComputerGuy:
Since this is being pulled into #3477, going to close this PR.

Issue #2847, opened 2025-08-04
EUS not affected packages are showing as affected (0.97.1)


pkeecom:
This relates to recent release of grype 0.97.1 to address https://github.com/anchore/grype/issues/2840. I tried the latest release but it is showing lots (22000+) of vulnerabilities related to higher versions which are not affecting the EUS 9.4 system.

**Examples**:

```
kernel                      5.14.0-427.79.1.el9_4  0:5.14.0-503.11.1.el9_5  rpm     CVE-2020-10135       Medium    14.9% (94th)   7.7
kernel-core                 5.14.0-427.79.1.el9_4  0:5.14.0-503.11.1.el9_5  rpm     CVE-2020-10135       Medium    14.9% (94th)   7.7
requests                    2.25.1                 2.31.0                   python  GHSA-j8r2-6x86-q33q  Medium    6.3% (90th)    3.5
rhc                         1:0.2.4-5.el9          1:0.2.5-1.el9_5          rpm     CVE-2022-3064        High      1.9% (82nd)    1.4
kernel-core                 5.14.0-427.79.1.el9_4  0:5.14.0-570.12.1.el9_6  rpm     CVE-2024-27398       Low       0.9% (74th)    0.4
python3-perf                5.14.0-427.79.1.el9_4  0:5.14.0-570.12.1.el9_6  rpm     CVE-2025-1272        Medium    N/A            N/A
python3-perf                5.14.0-427.79.1.el9_4  0:5.14.0-570.16.1.el9_6  rpm     CVE-2025-21633       Medium    N/A            N/A
kernel                      5.14.0-427.79.1.el9_4  0:5.14.0-503.11.1.el9_5  rpm     CVE-2023-52758       Low       N/A            N/A
kernel                      5.14.0-427.79.1.el9_4  0:5.14.0-570.12.1.el9_6  rpm     CVE-2023-52917       Low       N/A            N/A
kernel                      5.14.0-427.79.1.el9_4  0:5.14.0-503.11.1.el9_5  rpm     CVE-2024-26650       Low       N/A            N/A
kernel                      5.14.0-427.79.1.el9_4  0:5.14.0-503.11.1.el9_5  rpm     CVE-2024-35923       Low       N/A            N/A
kernel                      5.14.0-427.79.1.el9_4  0:5.14.0-503.11.1.el9_5  rpm     CVE-2024-35928       Low       N/A            N/A
```

[...]

PR #3181, opened 2026-01-23
fix(rpm): respect EUS limitation in --only-fixed

willmurphyscode:
When using --only-fixed with RHEL EUS, fixes targeting a higher minor version (e.g., el9_5) were incorrectly shown as "fixed" for users on lower EUS versions (e.g., 9.4 EUS). EUS users cannot upgrade to a higher minor version, so these fixes are not actually reachable.

Now grype parses the elX_Y suffix from fix versions and compares against the user's EUS minor version. Fixes for higher minor versions are marked as "not-fixed" instead of "fixed".

- Add isFixReachableForEUS to check fix version against EUS distro
- Mark unreachable fixes as NotFixed instead of Fixed
- Treat missing minor version as 0 (e.g., "9+eus" = "9.0+eus")
- Remove unused treatResolutionsAsDisclosures parameter

Fixes [#2847](https://github.com/anchore/grype/issues/2847)

### DIFF
--- a/grype/matcher/rpm/rhel_eus.go
+++ b/grype/matcher/rpm/rhel_eus.go
@@ -2,6 +2,8 @@ package rpm
 
 import (
 	"fmt"
+	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/anchore/grype/grype/distro"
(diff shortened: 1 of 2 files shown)


grype/matcher/rpm/rhel_eus.go
```diff
@@ -202,17 +285,29 @@ func mergeEUSAdvisoryIntoSingleDisclosure(v *version.Version, disclosure vulnera
...
+		// Get all fixes greater than current version (parses versions once)
+		allFixes := neededFixes(v, advisory.Fix.Versions, advisory.Constraint.Format(), advisory.ID)
+
+		// Filter to only those reachable for EUS (quick string-based check)
+		applicableFixes := filterFixesForEUS(allFixes, eusDistro, advisory.ID)
+
 		if len(applicableFixes) == 0 {
-			// none of the fixes on this advisory are greater than the current version, so we can skip this advisory
+			// If there were fixes but none are reachable for EUS, mark as NotFixed
+			if len(allFixes) > 0 && eusDistro != nil && *state != vulnerability.FixStateFixed {
```
spiffcs, line 304: The comment here is doing a lot of work and I appreciate it's help in grokking this conditional.
One thing that seems important is the 3rd case:
`*state != vulnerability.FixStateFixed`

If it's NOT FixStateFixed it's one of these:
`FixStateWontFix  FixState = "wont-fix"`
`FixStateNotFixed FixState = "not-fixed"`
`FixStateUnknown  FixState = "unknown"`

Is it correct below to modify `*state` for all those cases so they converge on `FixStateNotFixed`?

I think yes, since we confied they are not reachable for US to won't/unknown don't apply. It's just confirmed outside the distro so we give it the NotFixed 👍 ok answered my own comment 😭

PR #3182, opened 2026-01-24
feat(postmarketos): support more accurate scanning for postmarketos

westonsteimel:
Add the proper postmarketos release version to underlying alpine release version mappings from https://wiki.postmarketos.org/wiki/Releases to allow more accurate scanning for postmarketos. Since the versions do not align, grype was matching against all possible versions of alpine therefore causing many false positives.

There are images at https://codeberg.org/achill/docker-postmarketos to try it on

### DIFF
--- a/test/quality/.yardstick.yaml
+++ b/test/quality/.yardstick.yaml
@@ -132,9 +132,9 @@ result-sets:
     description: "same as 'pr_vs_latest_via_sbom', but includes vulnerabilities from 2022 and before, instead of 2021 and before"
     max_year: 2022
     validations:
-      - max-f1-regression: 0.1 # allowed to regress 0.1 on f1 score
-        max-new-false-negatives: 10
-        max-unlabeled-percent: 0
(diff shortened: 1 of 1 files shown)


westonsteimel:
This almost works, but not for edge.  The issue is that for postmarketos, the VERSION_ID in the os-release file is set to `edge` (there are no CODENAME entries in any of the postmarketos release files), and our grype OS search code can't currently handle an OSSpecifier lookup on just plain version string that isn't representable by a semantic like version (despite the override table supporting this since the lookup key is wisely just name, version, codename).  So need to find a way to make this work first.  

The issue from what I have found so far is that we use https://github.com/anchore/grype/blob/58368e32032464ef3cd73df710d7234a9a1bdf55/grype/db/v6/operating_system_store.go#L19-L40 when building the search query in https://github.com/anchore/grype/blob/58368e32032464ef3cd73df710d7234a9a1bdf55/grype/db/v6/search_query.go#L140 and this has no way of representing just a raw string version

PR #3200, opened 2026-02-01
feat: support context cancellation while finding vuln matches

luhring:
Today Grype's library entrypoint for vulnerability matching `FindMatches` doesn't accept a `context.Context`.

In idiomatic Go, function calls that either perform I/O or do nontrivial processing (`FindMatches` does both) accept a `ctx context.Context` as the first parameter and check for context cancellation at appropriate points during processing to allow the caller to stop long-running work early as needed.

This PR adds support for context cancellation in a non-breaking way:

- The implementation of `FindMatches` moves to a new variant, `FindMatchesContext`.
- `FindMatchesContext` checks for context errors (e.g. cancellation) at each iteration of the inner loop right before a specific matcher is called for a specific package. Non-nil context errors cause the function to return immediately.
- The signature of `FindMatches` doesn't change, and it's now a thin wrapper that passes `context.Background()` to `FindMatchesContext`.
- Cobra's command context is now able to be plumbed to `FindMatchesContext` (via `runGrype`).

**Note:** If there's other process cleanup or CLI UX changes needed as a result of this that aren't handled yet, just let me know!

### DIFF
--- a/cmd/grype/cli/commands/root.go
+++ b/cmd/grype/cli/commands/root.go
@@ -1,6 +1,7 @@
 package commands
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"strings"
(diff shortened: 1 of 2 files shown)


kzantow:
I'm not opposed to this change as-is possibly, but I would note that we have also been interested in parallelizing the searches, and the [parallelization library we're using in Syft](https://github.com/anchore/go-sync) supports context cancellation already; it might be worth adding this or following on this PR; is a part of the cancellation due to length of scans?

spiffcs:
@kzantow I added a follow up issue here: https://github.com/anchore/grype/issues/3202

spiffcs (approved):

luhring:
Thanks to both of you!!

PR #3213, opened 2026-02-06
fix(eol-warnings): fix fallback on major only distro

willmurphyscode:
Previously, there was a bug where rolling distros like alpine:edge would
incorrectly fallback to an EOLed version of alpine, resulting in
spurious EOL warnings. This was fixed by disabling the fallback in the
EOL query. However, this fix introduced a bug where major only distros,
like "Debian 8", would not fall back correctly. Therefore, when fallback
is disabled on the distro, and the minor version is missing, treat
"blank minor version" as part of the query, instead of unintentionally
excluding based on it.

Add a couple more unit tests to OS querying.


kzantow (approved):
LGTM

PR #3224, opened 2026-02-10
fix(db search vuln): missing EPSS/KEV should not be fatal error

willmurphyscode:
Previously, when running 'grype db search vuln ...', especially in a dev environment with only a subset of the real database available, grype would treat missing EPSS or KEV for the vuln as a fatal error. However, this missing auxiliary information should not prevent grype from displaying the vuln at all.

### DIFF
--- a/cmd/grype/cli/commands/internal/dbsearch/affected_packages.go
+++ b/cmd/grype/cli/commands/internal/dbsearch/affected_packages.go
@@ -232,15 +232,11 @@ func findAffectedPackages(reader interface { //nolint:funlen,gocognit
 	// ensures that all paths are handled the same way.
 	defer func() {
 		for i := range allAffectedPkgs {
-			if err := decorateVulnerabilities(reader, &allAffectedPkgs[i]); err != nil {
-				log.WithFields("error", err).Debug("unable to decorate vulnerability on affected package")
-			}
+			decorateVulnerabilities(reader, &allAffectedPkgs[i])
 		}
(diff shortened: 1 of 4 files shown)


cmd/grype/cli/commands/internal/dbsearch/affected_packages.go
```diff
@@ -232,15 +232,11 @@ func findAffectedPackages(reader interface { //nolint:funlen,gocognit
 	// ensures that all paths are handled the same way.
 	defer func() {
 		for i := range allAffectedPkgs {
-			if err := decorateVulnerabilities(reader, &allAffectedPkgs[i]); err != nil {
-				log.WithFields("error", err).Debug("unable to decorate vulnerability on affected package")
-			}
+			decorateVulnerabilities(reader, &allAffectedPkgs[i])
```
spiffcs, line 235: nit: since this is the caller should we have kept `decorateVulnerabilities` signature to return the error and logged here with a warn or non fatal error?

I noticed `findAffectedPackages` has other logging responsibilities and this feels like a message that affected packages should be in control of:

```
unable to decorate vulnerability on affected package: err unable to get EPSS scores
unable to decorate vulnerability on affected package: err unable to get known exploited vulnerabilities
```

The two error messages in decorateVulnerabilities doesn't tell me the impact about what failing to fetch these values actually has on the dbsearch or `findAffectedPackages` execution path

PR #3225, opened 2026-02-10
ci(snapshots): update CLI tests to exercise existing snapshots

willmurphyscode:
Previously, the CI validations workflow would make build the snapshots, then invoke a Taskfile target that would build the snapshot again, then invoke a go test that would `go build` the CLI again, and then exercise that CLI. Instead, update it so that the snapshot built by the snapshot CI job is used in tests and not rebuilt unnecessarily.

### DIFF
--- a/.github/workflows/validations.yaml
+++ b/.github/workflows/validations.yaml
@@ -281,6 +281,9 @@ jobs:
 
       - name: Run CLI Tests (Linux)
         run: make cli
+        env:
+          GRYPE_SNAPSHOT_PREBUILT: "true"
+          GRYPE_BINARY_LOCATION: ${{ github.workspace }}/snapshot/linux-build_linux_amd64_v1/grype
 
   Cleanup-Cache:
     name: "Cleanup snapshot cache"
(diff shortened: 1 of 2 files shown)


test/cli/utils_test.go
```diff
@@ -83,6 +83,14 @@ func getGrypeBinaryLocationByOS(t testing.TB, goOS string) string {
...
 
 func buildBinary(t testing.TB, loc string) {
+	if _, err := os.Stat(loc); err == nil {
+		// Use the existing binary (e.g. a GoReleaser snapshot). Freshness is
+		// managed by the snapshot task's checksum-based staleness detection;
+		// this function is only a fallback for when no binary exists at all
+		// (e.g. running "go test ./test/cli" without building first).
+		t.Logf("Using existing binary: %s", loc)
+		return
+	}
```
kzantow: This should be removed. This only runs when a binary isn't specified from an environment variable -- e.g. if you debug a test locally. See this above:
```
func getGrypeSnapshotLocation(t testing.TB, goOS string) string {
	// GRYPE_BINARY_LOCATION is the absolute path to the snapshot binary
	const envKey = "GRYPE_BINARY_LOCATION"
	if os.Getenv(envKey) != "" {
		return os.Getenv(envKey)
	}
```

willmurphyscode: Nice catch! Fixed.

PR #3255, opened 2026-02-28
feat: Handle Chainguard OSV used for the grypedb build.

vaikas:
As part of moving to OSV for Chainguard feed, this adds the transformer for it. There is another PR that I'll link in vunnel for it shortly and x-link them.

I think we need this: https://github.com/anchore/vunnel/pull/1082 to land first

### DIFF
--- a/grype/db/v6/build/transformers/osv/transform.go
+++ b/grype/db/v6/build/transformers/osv/transform.go
@@ -18,6 +18,8 @@ import (
 	"github.com/anchore/grype/grype/db/v6/build/transformers"
 	"github.com/anchore/grype/grype/db/v6/build/transformers/internal"
 	"github.com/anchore/grype/grype/db/v6/name"
+	"github.com/anchore/grype/internal/stringutil"
+	"github.com/anchore/packageurl-go"
 	"github.com/anchore/syft/syft/pkg"
 )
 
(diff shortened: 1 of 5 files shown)


grype/db/v6/build/transformers/osv/transform.go
```diff
@@ -36,6 +36,9 @@ func Transform(vulnerability unmarshal.OSVVulnerability, state provider.State) (
 
 	if isAdvisory {
 		aliases = append(aliases, vulnerability.Related...)
+	} else if strings.HasPrefix(vulnerability.ID, "CGA-") {
+		// Chainguard CGA records put CVE/GHSA IDs in "related" rather than "aliases"
```
vaikas: This is what we do today (related), I'd like to make sure this is correct.

willmurphyscode: I think you want records in Grype DB that mean, for example, https://osv.dev/vulnerability/CGA-46f4-r7mm-232v fixes `CVE-2026-31988` and `GHSA-gmq8-994r-jv83`?

That's what the "isAdvisory" is checking for, because that's the semantics AlmaLinux advisories have, see for example https://osv.dev/vulnerability/ALSA-2026:4443 claims to fix 2 CVEs.

Generally for transformer changes we want to work backwards from the matching semantics we're trying to make Grype do.

crosleyzack:
This does not rebase anymore, as files like `grype/db/v6/build/transformers/osv/transform.go` no longer exist. This might be easiest to close and create a new PR

PR #3256, opened 2026-02-28
Apk upstream matcher

vaikas:
Once these land:
 * https://github.com/anchore/grype/pull/3255
 * https://github.com/anchore/vunnel/pull/1082

And everything gets validated, then we want to be able to start taking advantage of more granular advisories. Flag controlled, so we can roll out safely.

### DIFF
--- a/grype/matcher/apk/matcher.go
+++ b/grype/matcher/apk/matcher.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/anchore/grype/grype/distro"
 	"github.com/anchore/grype/grype/match"
 	"github.com/anchore/grype/grype/matcher/internal"
 	"github.com/anchore/grype/grype/pkg"
@@ -35,6 +36,22 @@ func NewApkMatcher(cfg MatcherConfig) *Matcher {
 	return &Matcher{cfg: cfg}
 }
 
+// useUpstreamForPackage returns whether origin/upstream lookups should be performed
+// for this package. Alpine always requires upstream lookups regardless of the config flag —
+// it uses secdb-style advisories keyed by origin package name, so disabling lookups would
+// silently miss vulnerabilities. OSV-based distros with per-sub-package entries (e.g.
+// Chainguard, Wolfi) can disable upstream lookups via UseUpstreamMatcher=false to avoid
+// false positives from origin-level entries applying to unaffected sub-packages.
+//
+// TODO: if Alpine ever publishes per-sub-package OSV advisories, this hardcoded override
+// should be removed and Alpine should respect the flag like other distros.
+func (m *Matcher) useUpstreamForPackage(p pkg.Package) bool {
+	if p.Distro != nil && p.Distro.Type == distro.Alpine {
+		return true
+	}
+	return m.cfg.UseUpstreamMatcher
+}
+
 func (m *Matcher) PackageTypes() []syftPkg.Type {
 	return []syftPkg.Type{syftPkg.ApkPkg}
 }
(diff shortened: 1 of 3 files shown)


grype/matcher/apk/matcher.go
```diff
@@ -23,7 +24,33 @@ var (
...
+// useUpstreamForPackage returns whether origin/upstream lookups should be performed
+// for this package. Alpine always requires upstream lookups regardless of the config flag —
+// it uses secdb-style advisories keyed by origin package name, so disabling lookups would
+// silently miss vulnerabilities. OSV-based distros with per-sub-package entries (e.g.
+// Chainguard, Wolfi) can disable upstream lookups via UseUpstreamMatcher=false to avoid
+// false positives from origin-level entries applying to unaffected sub-packages.
+//
+// TODO: if Alpine ever publishes per-sub-package OSV advisories, this hardcoded override
+// should be removed and Alpine should respect the flag like other distros.
+func (m *Matcher) useUpstreamForPackage(p pkg.Package) bool {
```
willmurphyscode, line 48: I'd love to make this more data driven. It seems like the actual logic we want is closer to, "If `p` is from a distro where there is fix info for binary APKs (if that's the right term) and not just origin/upstream APKs (if that's the right term), then we don't need to search upstream."

There's some subtlety here too:

* a vulnerability can be about p or p's upstream
* a vulnerability can come from distro data or NVD data
* a fix or NAK can be about p or p's upstream, but always comes from distro data
* after https://github.com/anchore/grype/pull/3282 lands, we can emit ignore rules to suppress aliases of the vuln on owned paths, either about p or p's upstream or both

Is there an approach we could do where we have a precedence where evidence of fix takes precedence over evidence of vulnerability which takes precedence over a search miss, and if there's data about both p and p's upstream/origin, the specific data (p) takes precedence. For example:

1. NVD has a vuln on the upstream, Wolfi OSV has a fix on p -> package is not vulnerable, specific fix
2. NVD has a vuln on p, Wolfi OSV has a fix on p's upstream -> package is not vulnerable; distro claims it's generally patched
3. NVD has a vuln on upstream, Wolfi OSV has not info on this CVE -> package is vulnerable
4. etc

That way Grype does the best matching job it can given the data it has, rather than hard-coding heuristics for different distros (which is something we've done a lot of and would love to stop doing). What do you think @vaikas ?

vaikas: This does not directly answer your question, but hopefully gives a bit more context on what problem the change is meant to address.

So the current use of 'origin' (and I think in this context == 'upstream') is imho a bit wonky in this context. In the current context, it says that pkg-a and pkg-a-sub were built from the same source code. But for the vuln identification purposes I fail to see why they should be treated identically (today, IIUC), in that if I see a pkg-a-sub I automatically 'inherit' all the vulns for pkg-a. 
Some examples where this is especially wonky, when there are things like:
pkg
pkg-dev [headers for example]
pkg-doc [docs for example]
pkg-compat [shell scripts for example]

In today's world, if we see pkg-doc it's treated as pkg-a which doesn't make any sense. For Chainguard, we track occurances of vulnerabilities at the artifact level (APK) vs. origin because it's lossy.

And sub-packages can have vulns that do not appear in the origin package, so need to surface that as well.

I'm happy to hop on a call since that might be a higher BW conversation 🤣

So, that was the driving motivation for this change.

willmurphyscode: That makes a lot of sense. I agree that saying pkg-docs is vulnerable because pkg is vulnerable is silly. The reason this practice exists in Grype is that historically, distros and CVEs are published against the source/upstream/origin package, and we're using the granularity we have.

The thing I don't love about the config option approach is that it makes it the _user's_ responsibility to figure out whether they're scanning an image whose distro's publisher publishes granular vulnerability and fix data, but that's a question about the state of grype's database more than its a question about the user's preferences. I'm not opposed to having a config (we often put in configs with behavior changes so that people can pin back to older grype behavior), but I would love for the default behavior to be smarter.

Maybe this is something that should (eventually?) be on Grype's metadata about vulnerability providers? Like the matcher asks the database, "Hey, Is Wolfi one of the ones where binary APKs have fix data, or one of the ones where everything is stuck to the source/upstream/origin package?" and behaves differently based on the answer?

I'd be happy to jump on a call. We have a [Vuln Data Office Hours](https://calendar.app.google/zyQCfYTM4gqsHEfu8) appointment link if you find a convenient time on there.

vaikas:
Fantastic! Yeah, I totes get why it was the way it was, and like I said, we're trying to find a way to make it better and not break everything in the process. I was trying to make the most surgical / least invasive approach to validate the approach, and the flag option was so that I could start testing things on my end since there's few other pieces in play (vunnel, transformer).

I like the idea about the metadata provider as a solution.

Thanks for the offer to chat, I booked time on Friday (earliest that works from our side, aren't meetings fun?!?!) Looking forward to chatting more about this.

PR #3264, opened 2026-03-04
Draft: add BellSoft family of OS

i-bs:
coupled with https://github.com/anchore/vunnel/pull/924


i-bs:
coupled with https://github.com/anchore/vunnel/pull/924

grype/db/internal/versionutil/constraint.go
```diff
@@ -24,7 +24,7 @@ func EnforceSemVerConstraint(constraint string) string {
 }
 
 func AndConstraints(c ...string) string {
-	return strings.Join(c, " ")
```
willmurphyscode, line 27: This change isn't correct, for example you can see it breaks unit tests for another OSV provider: https://github.com/anchore/grype/actions/runs/25853467114/job/76529191535?pr=3264#step:4:108

I think the arguments passed to `AndConstraint` are wrong in the BellSoft case?

grype/distro/type.go
```diff
@@ -16,6 +16,8 @@ const (
 	CentOS       Type = "centos"
 	Fedora       Type = "fedora"
 	Alpine       Type = "alpine"
+	Alpaquita    Type = "alpaquita"
```
willmurphyscode: Can you help me understand the relationship between alpaquita and bellsoft hardened containers? It looks like they're both in OSV but with overlap and the same vendor prefix? https://osv.dev/list?ecosystem=Alpaquita and https://osv.dev/list?ecosystem=BellSoft+Hardened+Containers ?

It looks like their two different distros and we should just model them as two different distros, and their relationship is just that the same company works on them?

grype/db/v6/build/transformers/osv/transform_bellsoft.go
```diff
@@ -0,0 +1,117 @@
...
+//     dropped: the bellsoft matcher never queries by CPE, the platform CPE
+//     qualifier is a runtime no-op for application CPEs, and CPEs without
+//     ecosystem context produce noisy matches.
+type bellsoftStrategy struct{}
+
+func (bellsoftStrategy) Matches(id string) bool {
+	return strings.HasPrefix(id, "BELL-")
+}
+
+func (bellsoftStrategy) Transform(vuln unmarshal.OSVVulnerability, state provider.State) ([]data.Entry, error) {
```
willmurphyscode, line 42: This needs some unit tests asserting what the results of this transformation are. `transform_bellsoft_test.go` should be added.

grype/db/v6/build/transformers/osv/transform_bellsoft.go
```diff
@@ -0,0 +1,117 @@
...
+	var aphs []db.AffectedPackageHandle
+	for _, affected := range vuln.Affected {
+		var ranges []db.Range
+		for _, r := range affected.Ranges {
+			ranges = append(ranges, getGrypeRangesFromRange(r, bellsoftRangeType(r.Type))...)
+		}
+		aphs = append(aphs, db.AffectedPackageHandle{
+			Package: bellsoftPackage(affected.Package),
+			BlobValue: &db.PackageBlob{
+				CVEs:   vuln.Aliases, // FIXME: should be `vuln.Upstream`
```
i-bs: `upstream` field is not supported in the "github.com/google/osv-scanner/pkg/models"

willmurphyscode: I don't understand this comment. Can you elaborate a bit?

willmurphyscode (commented):
I had a few inline comments.

This needs some `dbtest` style matching behavior unit tests before it can be merged. https://github.com/anchore/grype/pull/3418 has some examples, and some other recent PRs. These unit tests are also a great way to get a sense of whether matching works end to end or not.

grype/db/v6/build/transformers/osv/transform_bellsoft.go
i-bs: sure.
https://ossf.github.io/osv-schema/#upstream-field
it is advised to use it for a "root" CVE. But the "github.com/google/osv-scanner/pkg/models" structures do not have it (yet).

grype/db/internal/versionutil/constraint.go
i-bs: this is to fix https://github.com/anchore/vunnel/pull/924#issuecomment-3799736849

The fix is incomplete/incorrect and need to be fixed in its order.

grype/distro/type.go
i-bs: https://github.com/anchore/vunnel/pull/924#issuecomment-3927091773

PR #3269, opened 2026-03-06
feat: generate json-schema for -o json

### LINKED ISSUE #214: Add json schema
It would be ideal to add a JSON schema to validate grype JSON output, ideally in a similar way to https://github.com/anchore/syft/pull/276 .

axel7083:
Fixes https://github.com/anchore/grype/issues/214

### DIFF
--- a/grype/presenter/models/document.go
+++ b/grype/presenter/models/document.go
@@ -13,6 +13,10 @@ import (
 	"github.com/anchore/grype/grype/vulnerability"
 )
 
+const (
+	DocumentSchemaVersion = "1.0.0"
+)
+
 // Document represents the JSON document to be presented
 type Document struct {
 	Matches         []Match         `json:"matches"`
--- a/cmd/grype/cli/commands/internal/jsonschema/main.go
+++ b/cmd/grype/cli/commands/internal/jsonschema/main.go
@@ -16,9 +16,10 @@ import (
 	"golang.org/x/tools/go/packages"
 
 	"github.com/anchore/grype/cmd/grype/cli/commands/internal/dbsearch"
+	"github.com/anchore/grype/grype/presenter/models"
 )
 
-func main() {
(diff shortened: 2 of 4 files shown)


grype/presenter/models/document.go
```diff
@@ -13,6 +13,10 @@ import (
 	"github.com/anchore/grype/grype/vulnerability"
 )
 
+const (
+	DocumentSchemaVersion = "1.0.0"
```
kzantow, line 17: This should definitely not be `1.0`

axel7083: As this seems for the other json-schemas to be unrelated to the version of grype, and seems to increase when the schema changes, what would be the expected _initial value_?

E.g. version for the dbsearch schema

https://github.com/anchore/grype/blob/9d54499415ad6f34426ce82082fa205a2e6bc45a/cmd/grype/cli/commands/internal/dbsearch/versions.go#L5

https://github.com/anchore/grype/blob/9d54499415ad6f34426ce82082fa205a2e6bc45a/schema/grype/db-search/json/schema-latest.json#L3

kzantow: Sorry I started typing something else up and got sidetracked... we are definitely going to be iterating on this model as we work towards 1.0. We will probably what we did with Syft and have a legacy schema for pre-1.0 Grype, essentially the one that we have today. It's a great question what that number should be! I would think `0.9` indicates it's close to `1.0`, but I think it is going to change significantly (e.g. possibly have matches only with references to packages and vulns). Maybe a `0.5` is something that indicates "significantly before 1.0". I'll bring this question to the team...

axel7083: Thanks for the details, keep me updated so I can update the PR!

axel7083:
Hey @kzantow may I ask if there were any updates on this topic? Thanks in advance!

PR #3276, opened 2026-03-11
chore(units): use synthesized grype db rather than mock vuln provider

willmurphyscode:
Previously, Grype used a mock implementation of the vulnerability provider in its unit test, meaning that bugs in SQL queries, unexpected case sensitivity or insensitivity on queries or indexes, etc, might go undetected in the test suite and crop up at runtime.

This limitation was mostly due to the fact that there was no way to build a grype database because that code lived in grype-db. Now that the database builder's core code lives in this repo, we can build test databases to have better testing of matchers.

To address these issues:

1. Create a helper that can dump JSON test fixtures of actual vunnel results to use as input
2. Create some helpers to build a test database
3. Replace a lot of matcher unit tests with unit tests that use committed fixtures (from the tool in step 1) to exercise the matcher, rather than typing up lists of Vulnerabilty structs or whatever in the test case.


willmurphyscode:
Closing in favor of https://github.com/anchore/grype/pull/3284

PR #3277, opened 2026-03-11
feat: db diff for v6

kzantow:
This PR adds a `db diff` command that implements package-based diffing between two Grype v6 databases.

The `grype db diff <db1> <db2>` command in this PR operates by comparing two grype databases from the perspective of "what changes have affected matching of specific packages" -- in other words: any data change that will result in a vulnerability being returned differently for a specific package should be returned as either an addition, modification, or removal. For example: if a vulnerability is added, it should would expectedly show up in the `added` section, whereas if a vulnerability has had the version constraints modified, it would be considered `modified`, and similarly if a vulnerability for a specific package in a specific ecosystem / provider has been removed, it would show up in the `removed` section.

The output is currently defined in the [db-diff schema](https://github.com/anchore/grype/pull/3277/changes#diff-acb31aeb39d0b29fc35e02bcb0c28c777b3913d64c79839eca0a8c1d825613dd), for example:
```
{
  "schema": "anchore.io/schema/grype/db-diff/json/0.5.0/results",
  "databases": {
    "before": {
      "buildTimestamp": "2026-02-07T06:24:08Z",
      "modelVersion": "v6.1.4",
      "revision": 1,
      "checksum": "xxh64:a2fac5f546c7b9b7"
    },
    "after": {
      "buildTimestamp": "2026-02-09T06:37:39Z",
      "modelVersion": "v6.1.4",
      "revision": 1,
      "checksum": "xxh64:3ea413099fbdbe37"
    }
  },
  "packages": [
    {
      "ecosystem": "apk",
      "name": "cloudflared-fips",
      "vulnerabilities": {
        "added": [
          {
            "provider": "chainguard",
            "id": "CVE-2025-58063"
          },
        ]
      }
    },
    {
      "ecosystem": "apk",
      "name": "glibc-2.41",
      "vulnerabilities": {
        "modified": [
          {
            "provider": "secureos",
            "id": "CVE-2025-5702"
          },
...
```

In order to get reasonable performance , a series of temp tables and views are created in the "old" database. These can be written to the database directly using the environment variable `GRYPE_DEV_DB_DEBUG=true` for further investigation about the results.


grype/db/v6/diff/package_tables.go
```diff
@@ -0,0 +1,253 @@
...
+		// package + language/os + vulnerability
+		`CREATE TEMP TABLE {prefix}_pkgs AS
+		 SELECT p.ecosystem,
+			p.name AS pkg_name,
+			-- get blob with information removed that does not affect matching
+			json_set(pb.value,'$.ranges',(SELECT json_group_array(json_remove(value, '$.fix.detail.available')) FROM json_each(pb.value, '$.ranges')))
+				AS pkg_blob,
+			vh.provider_id,
+			vh.name AS vuln_name,
+			vh.id AS vulnerability_id,
```
willmurphyscode, line 25: Does this need `vh.status` in case a matcher is using https://github.com/anchore/grype/blob/dee8de483dfba5b4e0bc0aa8e4ab2ce52137e490/grype/matcher/internal/only_non_withdrawn_vulnerabilities.go#L9 ?

grype/db/v6/diff/vuln_tables.go
```diff
@@ -0,0 +1,86 @@
...
+		LEFT JOIN diff_new_vulns n
+			ON n.provider_id = o.provider_id
+			AND n.name = o.name
+		WHERE n.name IS NULL`,
+	}
+
+	return d.executeTemplates(templates, r)
+}
+
+func (d *DBDiffer) createDiffTablesEPSS(prefix, schema string) error {
```
wagoodman: EPSS will always be different as these are recalculated daily and should be considered really separate from the current affected package and vulnerability information. Downstream it is most correct to use the latest EPSS values even when considering older vulnerability scan results. 

Is this affecting diff results broadly (it should be)? Is there a way to exclude EPSS from the diff?

kzantow: EPSS is excluded by default. If a user wants to get a list of EPSS differences beyond a threshold, they are able to via some configuration.

PR #3284, opened 2026-03-13
Add DB test harness

wagoodman:
This works towards replacing the existing test mocks that sit above the DB layer with a new test harness that supports:

- building a DB for one or more schemas from vunnel test fixture data
- managing DB caches from previous test runs
- be able to update vunnel test fixture material from vunnel caches
- test assertion helpers, pkg builder helpers, and distro constants to keep ver to keep the verbosity of test code to a minimum
- test helpers have completion capabilities on by default (so we check we assert on the entire result) and this does not require the test to be written with the knowledge of the data shapes in the top-level API (which will help with future refactors).
- manage local testdata fixtures or shared fixtures (across packages) to reuse vunnel material
- sub-select vunnel material as input in to the DB build (allowing for fewer duplicate fixtures)

The nice thing about the testdata directories used as input to the builds is that they are simply vunnel workspaces, so should still be 100% interoperable with the vunnel CLI if need be (pending a small vunnel fix inbound for windows compatibility).

This PR starts with replacing only the `dpkg` matcher tests and mocks with this new approach, however, these refactors will be rolled through the remaining matcher and integration tests.

In terms of writing tests, the helpers keep the verbosity down a lot, while giving us a nice chokepoint for future refactors to take place (instead of full test rewrites). Here's an example of writing a test using the new helpers:

```golang
func TestMatcherDpkg_DirectMatch(t *testing.T) {
	dbtest.SharedDBs(t, "all").
		SelectOnly("debian:11/CVE-2024-0727").
		Run(func(t *testing.T, db *dbtest.DB) {
			matcher := Matcher{}

			p := dbtest.NewPackage("openssl", "1.1.1k-1", syftPkg.DebPkg). // vulnerable (< 1.1.1w-0+deb11u2)
											WithDistro(dbtest.Debian11).
											Build()

			db.Match(t, &matcher, p).
				SelectMatch("CVE-2024-0727").
				SelectDetailByType(match.ExactDirectMatch).
				AsDistroSearch()
		})
}
```

I feel that as we port more matcher tests over we will refine the helper functions considerably. 

> [!NOTE]
> Reviewers might want to take a look at the [dbtest readme](https://github.com/anchore/grype/blob/db-test-harness/internal/dbtest/README.md) that is on the branch. This explains a lot of how this would be used.

## TODO
- [x] update CI workflows to upload DB caches


grype/matcher/dpkg/testdata/eol-debian8/debian/results/debian@8/cve-1999-1332.json
```diff
@@ -0,0 +1 @@
+{"schema":"https://raw.githubusercontent.com/anchore/vunnel/main/schema/vulnerability/os/schema-1.1.0.json","identifier":"debian:8/cve-1999-1332","item":{"Vulnerability":{"CVSS":[],"Description":"","FixedIn":[{"Name":"gzip","NamespaceName":"debian:8","VendorAdvisory":{"AdvisorySummary":[],"NoAdvisory":false},"Version":"1.3.5-6","VersionFormat":"dpkg","Available":{"Date":"2026-02-27","Kind":"first-observed"}}],"Link":"https://security-tracker.debian.org/tracker/CVE-1999-1332","Metadata":{"NVD":{"CVSSv2":{"Score":2.1,"Vectors":"AV:L/AC:L/Au:N/C:N/I:P/A:N"}}},"Name":"CVE-1999-1332","NamespaceName":"debian:8","Severity":"Low"}}}
```
willmurphyscode: Is it possible to get indents and newlines in these? I hate trying to read compact JSON on a diff.

internal/dbtest/README.md
```diff
@@ -0,0 +1,457 @@
...
+    // use a fixture from your package's testdata directory
+    dbtest.DBs(t, "my-fixture").Run(func(t *testing.T, db *dbtest.DB) {
+        // db implements vulnerability.Provider
+        vulns, err := db.FindVulnerabilities(...)
+    })
+}
+
+func TestWithSharedFixture(t *testing.T) {
+    // use a fixture from internal/dbtest/testdata/shared/
+    dbtest.SharedDBs(t, "all").Run(func(t *testing.T, db *dbtest.DB) {
```
kzantow, line 20: Is there a benefit of having the Run callback instead of just something like:
```
db := dbtest.SharedDBs(t, "all").Build()
```

grype/db/provider/entry/sqlite.go
```diff
@@ -27,6 +27,45 @@ type results struct {
...
+		return nil, err
+	}
+
+	query := db.Model(&results{})
+
+	if len(patterns) > 0 {
+		var conditions []string
+		var args []interface{}
+		for _, p := range patterns {
+			conditions = append(conditions, "id LIKE ?")
```
kzantow, line 51: This looks like patterns will always query as part of the ID, not the contents? Is this for exploring the results or something else?

internal/dbtest/cmd/manager/extract.go
```diff
@@ -0,0 +1,147 @@
...
+	fs.StringVar(&vunnelData, "vunnel-data", "", "path to vunnel data directory (required)")
+	fs.Var(&providers, "provider", "provider name to extract from (can be specified multiple times)")
+	fs.Var(&selects, "select", "pattern for record selection using SQL LIKE matching (can be specified multiple times)")
+	fs.StringVar(&outputDir, "output", "", "path for new fixture directory (mutually exclusive with --append)")
+	fs.StringVar(&appendDir, "append", "", "path for existing fixture directory to append to (mutually exclusive with --output)")
+
+	fs.Usage = extractUsage(fs)
+
+	if err := fs.Parse(args); err != nil {
+		os.Exit(1)
```
kzantow, line 32: nit: `panic(err)` or some other way to see the error?

internal/dbtest/README.md
willmurphyscode: I had the same question.

wagoodman: There is a `Build()` method similar to what you have there, only it returns a slice of DBs. One thing baked into this is the ability to build/test with multiple schemas. This allows a single chokepoint where we can control which schemas as used by default for testing (or can be overridden by arguments). Having the `Run()` method allows us to not really change the test body but we can test multiple subjects -- it's only saving test writers from boilerplate of `for each db build call t.Run with the DB schema name and test name...`.

grype/db/provider/entry/sqlite.go
wagoodman: not against adding that -- lets do that in a follow up PR

grype/matcher/dpkg/testdata/eol-debian8/debian/results/debian@8/cve-1999-1332.json
kzantow: Agree with this, we should not be including single-line JSON, any changes in the future become hard to review

kzantow (approved):

Issue #3267, opened 2026-03-05
Update APK NAK handling to be based on ownership-by-file-overlap relationship


kzantow:
Today, the [APK NAK handling is based entirely on file location](https://github.com/anchore/grype/blob/main/grype/matcher/apk/matcher.go#L265). This should be able to be applied to packages having no location information but instead the `ownership-by-file-overlap` relationship.

willmurphyscode:
@kzantow do you have some examples of incorrect matching that this results in?

PR #3286, opened 2026-03-16
feat: ignore by relationship

kzantow:
This PR adds relationship information to Grype packages and the ability for APK ignore rules by package overlap to also apply if that location information is missing, but relationship information is present.

Here are some handcrafted SBOMs (copied and reduced in size from larger SBOMs) that illustrate differences with files only, with relationships only, with both files and relationships, and with no files or relationships. A vulnerability to focus on with these SBOMs is `GHSA-xjjg-vmw6-c2p9`, which is an explicit Chainguard NAK record.

[python-overlap-no-files-or-relationships.json](https://github.com/user-attachments/files/26035601/python-overlap-no-files-or-relationships.json)
```
 ✔ Scanned for vulnerabilities     [8 vulnerability matches]  
   ├── by severity: 0 critical, 3 high, 4 medium, 0 low, 0 negligible (1 unknown)
NAME    INSTALLED  FIXED IN  TYPE    VULNERABILITY        SEVERITY  EPSS         RISK   
httpie  1.0.2      1.0.3     python  GHSA-xjjg-vmw6-c2p9  High      0.5% (65th)  0.4    
```

And the remaining 3 permutations of this SBOM correctly filter the results:

[python-overlap-with-files-and-relationships.json](https://github.com/user-attachments/files/26035620/python-overlap-with-files-and-relationships.json)
[python-overlap-with-files-only.json](https://github.com/user-attachments/files/26035622/python-overlap-with-files-only.json)
[python-overlap-with-relationships-only.json](https://github.com/user-attachments/files/26035623/python-overlap-with-relationships-only.json)
```
 ✔ Scanned for vulnerabilities     [7 vulnerability matches]  
   ├── by severity: 0 critical, 3 high, 4 medium, 0 low, 0 negligible (1 unknown)
NAME    INSTALLED  FIXED IN  TYPE    VULNERABILITY        SEVERITY  EPSS         RISK   
```

Fixes: #3267

### DIFF
--- a/grype/vulnerability_matcher.go
+++ b/grype/vulnerability_matcher.go
@@ -251,8 +251,10 @@ func (m *VulnerabilityMatcher) searchDBForMatches(
 	}
 
 	// apply ignores based on matchers returning ignore rules
+	startTime := time.Now()
 	filtered, dropped := match.ApplyIgnoreFilters(allMatches, ignoredMatchFilter(allIgnorers))
 	logIgnoredMatches(dropped)
+	log.Debugf("took %v to process %v vulns with %v ignores", time.Since(startTime), len(allMatches), len(allIgnorers))
 
 	// get deduplicated set of matches
 	res := match.NewMatches(filtered...)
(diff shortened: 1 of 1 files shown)


grype/match/ignore.go
```diff
@@ -49,6 +52,54 @@ type IgnoreRulePackage struct {
...
+			overlaps = true
+			break
+		}
+	}
+	if !overlaps {
+		return nil
+	}
+	return []IgnoreRule{
+		{
+			Vulnerability: i.VulnerabilityID,
```
willmurphyscode: Should this set `IncludeAliases`? Otherwise it won't suppress a GHSA on the owned package from a distro-based CVE fix.

grype/pkg/package.go
```diff
@@ -106,7 +111,30 @@ func FromPackages(syftPkgs []syftPkg.Package, config SynthesisConfig, enhancers
 			}
 		}
 
-		pkgs = append(pkgs, New(p, enhancers...))
+		grypePkg := New(p, enhancers...)
+		pkgIdx[grypePkg.ID] = len(pkgs)
+		pkgs = append(pkgs, grypePkg)
+	}
+
+	for _, r := range relationships {
```
willmurphyscode, line 119: I think if we wanted to avoid breaking library callers who use this directly, we could put it outside this function up in https://github.com/kzantow-anchore/grype/blob/main/grype/pkg/syft_sbom_provider.go#L48 or https://github.com/kzantow-anchore/grype/blob/main/grype/pkg/syft_provider.go#L40 . Do you think that's worth doing?

kzantow: If we wanted to avoid breaking consumers, I think we would add another function `FromRelatedPackages` or something and just call it with nil as the only thing `FromPackages` is doing -- we need it at least in both of those places you noted. I think in this case we might actually want to break consumers, if we make this change: they would not be aware of a necessary change to handle what I think is going to be an increasing number of cases for accuracy. I'm on the fence about the breakage here, but I think it might actually be better than silently letting consumers do the wrong thing. Besides, it's just a `, nil` addition if they don't have relationships.

PR #3316, opened 2026-03-25
perf: replace sort-based epssPercentile with O(n) linear scan

matiasinsaurralde:
## Summary

`epssPercentile` was calling `sort.Slice` to find the maximum value in a slice — O(n log n) with 4 heap allocations — when a single linear pass suffices.

This replaces it with an O(n), zero-allocation loop.

**Bonus fix**: `sort.Slice` sorted the caller's slice in-place, silently mutating `a.Vulnerability.EPSS` during output rendering as a sideeffect. The linear scan is pure (read-only).

## Benchmarks (Apple M4 Pro)

| Input      | Before       | After      | Speedup | Allocs before | Allocs after |
|------------|-------------|------------|---------|---------------|--------------|
| 3 items    | 100.6 ns/op | 2.2 ns/op  | 45x     | 4             | 0            |
| 100 items  | 1131 ns/op  | 62 ns/op   | 18x     | 4             | 0            |

In practice, EPSS slices are length 0–1 per CVE (both implementations are already O(1) there), so the primary value of this change is the removal of the mutation side-effect and the allocation savings on any multi-entry path

## Test plan
- [x] Existing `TestEPSSPercentile` unit tests pass unchanged
- [x] All tests in `grype/presenter/models` pass
- [x] Benchmark file added: `epss_bench_test.go`

Feel free to leave the benchmark out from the PR if it's not relevant

### DIFF
--- a/grype/presenter/models/epss_bench_test.go
+++ b/grype/presenter/models/epss_bench_test.go
@@ -1,0 +1,94 @@
+package models
+
+import (
+	"sort"
+	"testing"
+)
+
+// --- two implementations under test ---
+
+// current: full sort to find maximum
+func epssPercentileCurrent(es []EPSS) float64 {
+	switch len(es) {
+	case 0:
+		return 0.0
+	case 1:
+		return es[0].Percentile
+	}
+	sort.Slice(es, func(i, j int) bool {
+		return es[i].Percentile > es[j].Percentile
+	})
+	return es[0].Percentile
+}
+
+// proposed: linear scan to find maximum
+func epssPercentileLinear(es []EPSS) float64 {
+	if len(es) == 0 {
+		return 0.0
+	}
+	max := es[0].Percentile
+	for _, e := range es[1:] {
+		if e.Percentile > max {
+			max = e.Percentile
+		}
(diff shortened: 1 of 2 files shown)


willmurphyscode (approved):

willmurphyscode:
Thanks for including benchmarks and remembering to sign off!

willmurphyscode:
Hi @matiasinsaurralde,

There's a failing static analysis check, just because `max` is the name of a builtin function in go now, and there's a local variable shadowing it:

```
task: [lint] .tool/golangci-lint run --tests=false
Error: grype/presenter/models/sort.go:267:2: redefines-builtin-id: redefinition of the built-in function max (revive)
	max := es[0].Percentile
	^
Error: grype/presenter/models/sort.go:270:4: redefines-builtin-id: redefinition of the built-in function max (revive)
			max = e.Percentile
			^
2 issues:
* revive: 2
```

matiasinsaurralde:
@willmurphyscode Updated the code, thanks!

willmurphyscode (approved):

PR #3327, opened 2026-03-31
feat: integrate RapidFort advisory matching for their curated images

vaibhav-rf:
## Summary

Adds a dedicated RapidFort matcher that activates when the scanned image carries
a `maintainer=RapidFort*` Docker label. RapidFort advisories are stored under
`rapidfort-{os}:{version}` namespaces (e.g. `rapidfort-ubuntu:20.04`) — fully
isolated from standard distro scans so non-RF images are never affected.

Supported base distros: **Ubuntu** (dpkg), **Alpine** (apk), **Red Hat** (rpm).

### New: `grype/matcher/rapidfort/`

| File | Purpose |
|---|---|
| `matcher.go` | Core matcher: upstream + binary lookups, RPM release-identifier filtering, already-fixed suppression, dedup |
| `label.go` | `IsRapidFortImage` — detects RF-curated images via maintainer label |
| `matcher_test.go` | Unit tests: release identifier detection, RPM release matching, distro version mapping, already-fixed filter |
| `label_test.go` | `IsRapidFortImage` edge cases: nil source, non-image, case-insensitive key/value |

### Modified: existing packages

| File | Change |
|---|---|
| `grype/matcher/matchers.go` | `ApplySelectionPolicy` replaces dpkg/apk/rpm matchers with `RapidFortMatcher` for RF images; Java/stock matchers are preserved |
| `grype/matcher/matchers_test.go` | Tests for selection policy: RF image, non-RF image, no source |
| `grype/match/matcher_type.go` | Register `RapidFortMatcher` type (excluded from `AllMatcherTypes` — it is a conditional override) |
| `grype/distro/type.go` | Add `RapidFortUbuntu`, `RapidFortAlpine`, `RapidFortRedHat` distro types with `IDMapping` entries |
| `grype/db/internal/provider/unmarshal/os_vulnerability.go` | Add `Identifier` field to `OSFixedIn` for per-release RPM advisory routing |
| `grype/db/v6/build/transformers/os/transform.go` | `getPackageType` maps RF distros; `getFix` emits `release-identifier` advisory references from `Identifier` field |

### Test fixtures

| File | Purpose |
|---|---|
| `test-fixtures/rapidfort-ubuntu-20.04.json` | Transform test: Ubuntu dpkg advisory with VulnerableRange and fix availability |
| `test-fixtures/rapidfort-redhat-9.json` | Transform test: Red Hat rpm advisory with el9 (unfixed) and fc36 (fixed) entries |

## Design notes

**Namespace isolation** — The RF matcher queries `rapidfort-{os}:{version}` namespaces
that vunnel emits exclusively for RF advisories. Standard distro matchers (dpkg/apk/rpm)
never see these namespaces, so there is zero cross-contamination.

**RPM release-identifier routing** — Red Hat advisories carry per-release `Identifier`
fields (e.g. `el9`, `fc36`). The matcher derives the installed release from the package
version suffix (`.el9`, `.fc41`, `.rf`) and filters to advisories with a matching
`release-identifier:*` advisory reference. Unknown release identifiers fall back to
matching `el*` advisories.

**Matcher activation** — `ApplySelectionPolicy` in `matchers.go` checks
`IsRapidFortImage` and, when true, replaces dpkg/apk/rpm matchers with a single
`RapidFortMatcher`. This avoids double-reporting and keeps RF advisory logic
self-contained.

## Test plan

- [ ] `go test ./grype/matcher/rapidfort/...` — all matcher and label tests pass
- [ ] `go test ./grype/matcher/...` — selection policy tests pass
- [ ] `go test ./grype/db/v6/build/transformers/os/...` — transform tests pass for both RF fixtures
- [ ] `go test ./grype/distro/...` — distro type registration passes
- [ ] Scan a RapidFort-curated image: verify `rapidfort-matcher` appears in match details
- [ ] Scan a non-RapidFort image: verify `rapidfort-matcher` is absent from match details


willmurphyscode:
Hi @vaibhav-rf is there a publicly pullable test image I can try this out on?

ravinder-rapidfort:
Hi @willmurphyscode,

Thank you for reaching out regarding the test images. We are currently in the process of creating them and will share them with you shortly.

These images will include coverage across all relevant operating systems: Alpine, Ubuntu, and Red Hat to support comprehensive end-to-end testing of the PRs.

Appreciate your patience.

Thanks Again.

ravinder-rapidfort:
Hi @willmurphyscode,

Follow-up on my earlier message - the test images have been created and pushed. You can pull them using:

```
docker pull quay.io/rapidfort/rf-advisory/test1:20260521
docker pull quay.io/rapidfort/rf-advisory/test2:20260521
docker pull quay.io/rapidfort/rf-advisory/test3:20260521
```

These cover Alpine, Ubuntu, and Red Hat for end-to-end testing of the PRs. Let me know if you run into any issues pulling them.

Thanks!

vaibhav-rf:
Hi @willmurphyscode 

Just checking in on the test images we shared earlier. Were you able to pull them and validate the PRs?

Please let us know if you encountered any issues during testing or if there's any additional information we can provide. Looking forward to hearing your feedback.

Thanks!

ravinder-rapidfort:
Hi @willmurphyscode,

Circling back on this, we haven't heard back since sharing the test images and PRs for review. Is there anything we can do on our end to help move this forward, or any additional information you need?

Happy to assist however we can.

Thanks!

willmurphyscode:
Hi @ravinder-rapidfort thanks for the ping! I hope to take a look at this pretty soon.

ravinder-rapidfort:
Hi @willmurphyscode 

Thanks for the update! Just checking in to see if you've had a chance to take a look. Let us know if anything comes up happy to help.

ravinder-rapidfort:
Hi @willmurphyscode ,

Just checking in for any updates on this. Let me know if you've had a chance to take a look. Thanks!

ravinder-rapidfort:
Hi @willmurphyscode , any updates on this? Thanks!

ravinder-rapidfort:
Hi @willmurphyscode  , any updates on this? Thanks!

grype/db/internal/provider/unmarshal/os_vulnerability.go
```diff
@@ -14,6 +14,7 @@ type OSFixedIn struct {
 	Module         *string `json:"Module,omitempty"`
 	Name           string  `json:"Name"`
 	NamespaceName  string  `json:"NamespaceName"`
+	Identifier     string  `json:"Identifier,omitempty"`
```
kzantow: It looks like this field is the namespace + vulnerability ID -- is it necessary to add this instead of just using the existing `NamespaceName` and `Name` fields in conjunction with a transform?

grype/db/v6/build/transformers/os/transform.go
```diff
@@ -211,6 +226,10 @@ func getFixAvailability(fixedInEntry unmarshal.OSFixedIn) *db.FixAvailability {
 	}
 }
 
+func releaseIdentifierReferenceID(identifier string) string {
+	return "release-identifier:" + strings.ToLower(strings.TrimSpace(identifier))
```
kzantow, line 230: I'm a little unclear what this release-identifier is: is this an internal vulnerability ID? an advisory ID? If I better understood this perhaps I could understand if there is an existing data field for it.

kzantow (commented):
Hi, @vaibhav-rf 👋 — sorry for the delay getting back to you. I’ve run the full end-to-end test here and I can see results from the sample images you provided. I think there are some details we need to work through on this PR and by association [the change in vunnel](https://github.com/anchore/vunnel/pull/1114).

I see a crux of the issue is the fact that RapidFort images seem to have the original `/etc/os-release` from the upstream source\*\* instead of information indicating these came from RapidFort. I see the function in this PR that looks at _image labels_ to determine if there is a maintainer with the text rapidfort in it. This is likely to be a problem in a number of scenarios: when we download the images, there are times when we only have bytes on disk and do not necessarily have the same labels that a registry has, which would result in incorrect matching. Is there anything _in the bytes on disk_ that indicates the image came from RapidFort?

A quick look at one of the images shows at least one file:

```
/usr/share/rapidfort/curated.md:# RapidFort Alpine 3.23 curated image
```

What if we were to classify these images with a slightly different distro when we find evidence of being from RapidFort? That would have the benefit of potentially making the SBOM somewhat better when users scan them, too. I find it a little confusing that the `os-release` doesn’t indicate this in some way. Is there some better indicator than a markdown file?

In terms of the changes to Grype in this PR, I think we should try to find a better way to classify these packages as having come from RapidFort than the image labels. The easiest way would be to just have an alternate distro that more accurately reflects this as noted earlier. If we had this, the matching would simply use that and there would be no need to have some hardcoded logic switching the set of matchers being used and no need to have a custom matcher at all, I believe. We definitely want to avoid writing custom code to support each vendor’s specifics, if possible.

Lastly, the versions I see look somewhat like standard Alpine, Debian, etc. — some vendors use their own version suffix such as `+ubuntu.12` or similar at the end of the version string, this seems like a much more clear indicator that an organization has patched and rebuilt something, have I missed something? Are there any indicators in the version strings or package names or package manager configuration that might have an indication these were from RapidFort?

One other suggestion: I believe generally your images will have the same vulnerability exposure as the upstream source, e.g. Ubuntu or Alpine until you’ve issued patches for you packages, is that right? Are you issuing _vulnerability records without fix information_? E.g. if curl at v7.2 is vulnerable, but you have not fixed it, does your data include a record indicating curl v7.2 is _vulnerable_ or does a record only get included when you have issued a fix? If it’s the latter, and given that it looks like you might have some unique versions, e.g. `7.78.0-4.fc36` that wouldn’t match any other released versions of the software exactly, there might be another simpler option: we have the ability to mark packages with ranges or exact versions as “unaffected”. Perhaps this entire data set could output the `unaffected_package_handles` records with constraints of `= <version>` and use the existing upstream `operating_system` such that these specific versions would use the same upstream data with your unaffected fix infomration overlayed. With this approach, there may be fewer changes required and no hardcoded behavior specific to RapidFort.

Sorry for this rather long comment, but I’m very interested to get this moved forward, let me know how I can help!


\*\* the distro for these sample images has no RapidFort indicator from what I can tell:

```
% syft quay.io/rapidfort/rf-advisory/test2:20260521 -o json | jq .distro
{
  "prettyName": "Red Hat Enterprise Linux 9.7 (Plow)",
  "name": "Red Hat Enterprise Linux",
  "id": "rhel",
  "idLike": [
    "fedora"
  ],
  "version": "9.7 (Plow)",
  "versionID": "9.7",
  "homeURL": "https://www.redhat.com/",
  "bugReportURL": "https://issues.redhat.com/",
  "cpeName": "cpe:/o:redhat:enterprise_linux:9::baseos"
}

% syft quay.io/rapidfort/rf-advisory/test1:20260521 -o json | jq .distro
{
  "prettyName": "Alpine Linux v3.23",
  "name": "Alpine Linux",
  "id": "alpine",
  "versionID": "3.23.3",
  "homeURL": "https://alpinelinux.org/",
  "bugReportURL": "https://gitlab.alpinelinux.org/alpine/aports/-/issues"
}

% syft quay.io/rapidfort/rf-advisory/test3:20260521 -o json | jq .distro                                                   
{
  "prettyName": "Ubuntu 22.04.5 LTS",
  "name": "Ubuntu",
  "id": "ubuntu",
  "idLike": [
    "debian"
  ],
  "version": "22.04.5 LTS (Jammy Jellyfish)",
  "versionID": "22.04",
  "versionCodename": "jammy",
  "homeURL": "https://www.ubuntu.com/",
  "supportURL": "https://help.ubuntu.com/",
  "bugReportURL": "https://bugs.launchpad.net/ubuntu/",
  "privacyPolicyURL": "https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
}
```

grype/db/internal/provider/unmarshal/os_vulnerability.go
rfJanit: This field is not NamespaceName + Name. It's a per-fix release-variant tag (values like el9, fc36, rf) that scopes an advisory to a specific base OS release. Downstream, transform.go (line 229) turns it into a release-identifier:<X> reference, which the RapidFort matcher's byRPMReleaseIdentifier uses to filter advisories per installed package's release (see grype/matcher/rapidfort/matcher.go). Without this field, an el8-scoped fix would incorrectly match an el9 package.

I'll rename the field to ReleaseIdentifier and add a docstring to prevent this confusion in future reads. Sending that as a follow-up commit.

grype/db/v6/build/transformers/os/transform.go
rfJanit: This is the downstream side of the Identifier discussion above — same concept at a different lifecycle stage. 

`"release-identifier:<X>"` is not a vulnerability ID or an advisory ID. It's a synthetic reference that carries the per-fix release-variant tag (el9, fc36, rf, ubuntu, etc.) through the DB so the RapidFort matcher can filter by it at scan time. The release-identifier: prefix acts as a namespace marker so these synthetic entries stay distinguishable from real advisory reference IDs (RHSA-, GHSA-, etc.) sharing the same db.Reference.ID slot.

The DB v6 schema doesn't currently model per-fix release scoping as a first-class column, so we piggyback on the references list. I've already made changes to handle : the source field was renamed to ReleaseIdentifier with a docstring pointing at the transformer and matcher, and the transformer function now has a matching docstring plus a named releaseIdentifierRefPrefix constant — so both ends of the chain document what the value is and where it's consumed.

rfJanit:
> Hi, @vaibhav-rf 👋 — sorry for the delay getting back to you. I’ve run the full end-to-end test here and I can see results from the sample images you provided. I think there are some details we need to work through on this PR and by association [the change in vunnel](https://github.com/anchore/vunnel/pull/1114).
> 
> I see a crux of the issue is the fact that RapidFort images seem to have the original `/etc/os-release` from the upstream source** instead of information indicating these came from RapidFort. I see the function in this PR that looks at _image labels_ to determine if there is a maintainer with the text rapidfort in it. This is likely to be a problem in a number of scenarios: when we download the images, there are times when we only have bytes on disk and do not necessarily have the same labels that a registry has, which would result in incorrect matching. Is there anything _in the bytes on disk_ that indicates the image came from RapidFort?
> 
> A quick look at one of the images shows at least one file:
> 
> ```
> /usr/share/rapidfort/curated.md:# RapidFort Alpine 3.23 curated image
> ```
> 
> What if we were to classify these images with a slightly different distro when we find evidence of being from RapidFort? That would have the benefit of potentially making the SBOM somewhat better when users scan them, too. I find it a little confusing that the `os-release` doesn’t indicate this in some way. Is there some better indicator than a markdown file?
> 
> In terms of the changes to Grype in this PR, I think we should try to find a better way to classify these packages as having come from RapidFort than the image labels. The easiest way would be to just have an alternate distro that more accurately reflects this as noted earlier. If we had this, the matching would simply use that and there would be no need to have some hardcoded logic switching the set of matchers being used and no need to have a custom matcher at all, I believe. We definitely want to avoid writing custom code to support each vendor’s specifics, if possible.
> 
> Lastly, the versions I see look somewhat like standard Alpine, Debian, etc. — some vendors use their own version suffix such as `+ubuntu.12` or similar at the end of the version string, this seems like a much more clear indicator that an organization has patched and rebuilt something, have I missed something? Are there any indicators in the version strings or package names or package manager configuration that might have an indication these were from RapidFort?
> 
> One other suggestion: I believe generally your images will have the same vulnerability exposure as the upstream source, e.g. Ubuntu or Alpine until you’ve issued patches for you packages, is that right? Are you issuing _vulnerability records without fix information_? E.g. if curl at v7.2 is vulnerable, but you have not fixed it, does your data include a record indicating curl v7.2 is _vulnerable_ or does a record only get included when you have issued a fix? If it’s the latter, and given that it looks like you might have some unique versions, e.g. `7.78.0-4.fc36` that wouldn’t match any other released versions of the software exactly, there might be another simpler option: we have the ability to mark packages with ranges or exact versions as “unaffected”. Perhaps this entire data set could output the `unaffected_package_handles` records with constraints of `= <version>` and use the existing upstream `operating_system` such that these specific versions would use the same upstream data with your unaffected fix infomration overlayed. With this approach, there may be fewer changes required and no hardcoded behavior specific to RapidFort.
> 
> Sorry for this rather long comment, but I’m very interested to get this moved forward, let me know how I can help!
> 
> ** the distro for these sample images has no RapidFort indicator from what I can tell:
> 
> ```
> % syft quay.io/rapidfort/rf-advisory/test2:20260521 -o json | jq .distro
> {
>   "prettyName": "Red Hat Enterprise Linux 9.7 (Plow)",
>   "name": "Red Hat Enterprise Linux",
>   "id": "rhel",
>   "idLike": [
>     "fedora"
>   ],
>   "version": "9.7 (Plow)",
>   "versionID": "9.7",
>   "homeURL": "https://www.redhat.com/",
>   "bugReportURL": "https://issues.redhat.com/",
>   "cpeName": "cpe:/o:redhat:enterprise_linux:9::baseos"
> }
> 
> % syft quay.io/rapidfort/rf-advisory/test1:20260521 -o json | jq .distro
> {
>   "prettyName": "Alpine Linux v3.23",
>   "name": "Alpine Linux",
>   "id": "alpine",
>   "versionID": "3.23.3",
>   "homeURL": "https://alpinelinux.org/",
>   "bugReportURL": "https://gitlab.alpinelinux.org/alpine/aports/-/issues"
> }
> 
> % syft quay.io/rapidfort/rf-advisory/test3:20260521 -o json | jq .distro                                                   
> {
>   "prettyName": "Ubuntu 22.04.5 LTS",
>   "name": "Ubuntu",
>   "id": "ubuntu",
>   "idLike": [
>     "debian"
>   ],
>   "version": "22.04.5 LTS (Jammy Jellyfish)",
>   "versionID": "22.04",
>   "versionCodename": "jammy",
>   "homeURL": "https://www.ubuntu.com/",
>   "supportURL": "https://help.ubuntu.com/",
>   "bugReportURL": "https://bugs.launchpad.net/ubuntu/",
>   "privacyPolicyURL": "https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
> }
> ```

Thanks @kzantow . Let me take your points in order.

**On a reliable RapidFort indicator**

Agreed that image labels are the wrong signal — your point about only having bytes on disk is fair and we'd hit the same mismatch. The file marker is the right thing to key off: every RapidFort image ships `/usr/share/rapidfort/curated.md` and `/usr/share/rapidfort/curated.json`. This is guaranteed — we don't publish an image without it, so it's present regardless of how the image is obtained. There is nothing else on disk beyond those two files, so if we're picking one durable signal, that's it.

**On modifying `/etc/os-release` / classifying as a different distro**

We deliberately leave the OS identity untouched. In an RF image the base OS genuinely is the upstream OS — Alpine 3.23 is Alpine 3.23. What differs is that the package set is curated, not that the distro changed. Rewriting `os-release` would make the image misreport itself to every other tool in the ecosystem, not just Grype, so we'd rather not go down that path.

**On version suffixes**

Partly yes. Our Ubuntu and RHEL builds do carry an `rf` marker in the package names and versions, so your instinct there is right. Alpine is the exception: those packages have to be built to match stock versioning exactly, so we can't append or rename anything there. That's why version strings can't serve as the universal detection mechanism even though they work for most of our catalog.

Worth noting: you don't need to special-case these exceptions on your side. The advisory data we emit already accounts for them, so matching against our records should be correct without Grype knowing which builds carry a suffix and which don't.

**On the `unaffected_package_handles` overlay approach**

This one doesn't quite fit, unfortunately. Our database includes open CVEs, not only fixed ones — we don't wait for a fix to publish a record. So an unaffected-only overlay would represent a subset of our dataset rather than all of it.

On the related assumption: RF images do share upstream exposure in some cases, but not universally, and there's currently no way for Grype to determine from the artifact whether a given package has been patched or not. That's really the reason we're emitting a dedicated dataset for our versions instead of layering onto upstream — our package versions receive updates continuously from the team, and re-emitting the full upstream version space to keep pace with that is considerably more work on both sides than shipping events scoped to our own versions.

**Where that leaves us**

Proposed shape, tell me if this sounds reasonable:

1. Drop label-based detection entirely, key off `/usr/share/rapidfort/curated.json` (with `curated.md` as a fallback).
2. Grype derives its RapidFort classification from that marker, so no per-vendor hardcoding in the matcher path.
3. Matching runs against our advisory data, which already handles the Alpine version-parity exceptions.

Thanks!

xnox (changes requested):
The proprietary forks of community and commercial distributions that rapidfort creates intentionally obfuscate the overlapping origin of files. In effect the artifacts being scanned are multivendor without cryptographically traceble provenance of which bits came from which vendor and what the interaction between them is. This is done with explicit intention to supress findings without any source based remediations.

As a maintainer in UAPI group, Systemd and Freedesktop the implementation and changes performed by Rapidfort are in violation of the https://www.freedesktop.org/software/systemd/man/latest/os-release.html specification. When proprietary derivatives of community distributions are created one must update the ID= field, and use ID_LIKE= to indicate derivative and where it came from. There are many examples of Fedora / Debian / Ubuntu derivatives that do this correctly. As well more container focused projects like distroless, azure linux, amazon linux. Concrete examples are various SIG spin ups in Fedora family, and Linux Mint editions. One good option is to have parent distribution os-release file in /usr/lib/os-release whilst the /etc/os-release has modified ID= with parent distro in ID_LIKE=

Labels on containers are not cryptographically secure - thus such labels can redirect and silence vulnerability findings, that are applicable to parent distribution, yet rapidford chooses to silence, whilst software remains still vulnerable and unpatched. Worse this can be applied accidentally to parent distributions.

From trademark perspective Debian (Software in Public Interest), RedHat (IBM), and Ubuntu (Canonical) require to strip and remove all branding precisely to protect any confusion if one is using genuine product, or something that appears counterfeited or tampered with.

It is true that containers often contain artifacts from multiple vendors, as many end users use dockerfiles and copy from functionality to deploy binaries from different vendors onto multiple bases. This can often be seen in container deployments and grype being used to scan Kubernetes hosts or pods - which happen to mount and copy CNI plugins and sidecars from different vendors. An Ubuntu EKS node, has Amazon Linux binaries on the host from EKS addons, and can also have Fedora flatpacks, in addition to hand installed packages from Debian. But in such multivendor situation - whole scan results must not be match to a single vendor; but each binary/package must be attributed to correct vendor feeds without allowing one vendor to override feess of another one. Even if it is a correct false positive determination. This is the base orinciple of the CVE & OCV programmes - which rapidfort chooses not to participate in and/or didn't meet their basic quality requirements to join.

Rapidfort creates proprietary forks of community distributions, similar to more mature players such as Echo and Root. However in case of Echo and Root both of them correctly identify themselves in os-release file, change branding throughout and participate in the osv.dev programme to publish accurate advisories for the components they produced, signed and distributed. And all that integrates not just with grype but pleora of other scanners such as dependency track, trivy and many more.

Thus, if it is trivial to provide proprietary derivatives, with standards based open vulnerability data, that is already well supported by all the tools - why is this proprietary extensions in grype needed at all? This code only risks masking and interfering with findings of parent distributions for the users of the genuine products untampered by rapidfort.

My recommendation is to focus efforts in improving syft cataloger overlapping evidence enumeration and enable multi-vendor binaries co-existence and vulnerability attribution - in a way that doesn't supress one from another. Such that accurate package database metadata is used to correctly atteibute things - note rapidfort often modifies and removes files from parent distribution without any attribution or cryptographic signatures or presenting any coherent information as to which version of removals was applied. Which for ourposes of findings atteibutions must be discoverable in an unambigious way.

Merging this will lead to exponential increse of false positives and false negatives across all parent distributions that rapidfort created proprietary forks of. There is also legal risk of trademark violations as well, because trademark protected products are being used to reffer to 3rd party proprietary forks.

ps. I am Debian Developer, Ubuntu Core Developer, and ex board director of SPI Inc.

PR #3366, opened 2026-04-13
feat(scan): add support for a Zarf scan target

### LINKED ISSUE #3329: Proposal: zarf package scan target

This proposal introduces a `zarf:` scan target scheme that allows Grype to directly scan [Zarf](https://zarf.dev) package archives (`.tar.zst`) for vulnerabilities. Rather than requiring users to manually extract SBOMs from a Zarf package and feed them individually to Grype, a single command scans all bundled SBOMs and produces a unified vulnerability report.

```
grype zarf:path/to/package.tar.zst
```

Zarf is an [OpenSSF Sandbox Project](https://openssf.org/projects/zarf/) which declaratively packages and deploys artifacts. `syft` is used to create an SBOM for all included files and container images which are then stored in an `sboms.tar` file within the package and can be extracted. 

The nuance here is that Zarf is providing multiple sboms. I was reading the [syft issue](https://github.com/anchore/syft/issues/562) that relates to multiple sbom scanning targets (and also mentions Zarf). 

**Why is this needed**:

We're looking to expand recommendations around vulnerability scanning and general threshold creation for zarf packages in upstream open source software. Zarf is coordinating a few different efforts around how we can help secure open source software (in alignment with OpenSSF objectives) and additionally how to increase insights available for downstream consumers of packages. 

This option being available will provide package producers and consumers with fast feedback optionality around total vulnerability surface. Users would use this as a way to establish a threshold for review at more depth on a per-sbom level. 

**Additional context**:

In the initial model - I imagine that grype would produce a single report. This can be noisy and require mapping back to individual SBOMs/artifacts.

Zarf is working towards a v1.0.0 currently without any plans to change the underlying package structure around how SBOMs are stored in the package. Given the zstd archive of the package the the tar archive of the sboms collection - I believe this can be implemented without incurring any additional dependencies on the project. 

Future work could include adding VEX documents to the packaging schema such that they can be ingested during that scan natively. 

I would be more than willing to draft a Pull Request for review.

brandtkeller:
## Summary

Adds a `zarf:` scan target so grype can scan a Zarf package archive (.tar.zst) directly, read every bundled SBOM, and produce a single vulnerability report where each finding is traceable back to the originating image/file/binary.

```
grype zarf:path/to/package.tar.zst -o json
```

Resolves anchore/grype#3329.

## Design decisions

Discussion on the livestream and issue thread landed on a grype-only implementation, conditional on findings being resolvable back to a specific artifact inside the Zarf package. The decisions reflected in this PR:

- Grype-only, not a syft source/cataloger. The general multi-source problem (anchore/syft#562) is a broader effort and this PR solves the concrete Zarf use case without blocking on it.
- Per-finding provenance via package annotations. A new `Annotations map[string][]string` field on `pkg.Package` carries provider-supplied per-package metadata. The Zarf provider populates `zarf-sbom-source` with the SBOM's declared source name (or tar entry filename as a fallback). Values are stored as a sorted, deduplicated slice so shared base-image packages retain the full list of bundled images they came from. 
  - Trade-off: annotations surface in JSON output, not in the default table format.
- Dedupe-by-ID at provider boundary. When the same package ID appears across multiple bundled SBOMs (typical for shared base-image content), the provider folds them onto a single pkg.Package and unions the source list. This avoids the existing pkg.ByID(...) "first-wins" behavior dropping later sources before they ever reach a match.
- Prefer the SBOM's declared source name (typically an image reference, file path, or binary path); fall back to the tar entry filename when the source name is empty.

## Resolving a finding back to its source

Every match in the JSON output carries:

```json
{
  "artifact": {
    "name": "libxml2",
    "version": "2.13.8-r0",
    "locations": [
      { "path": "/lib/apk/db/installed" }
    ],
    "annotations": {
      "zarf-sbom-source": [
        "ghcr.io/go-gitea/gitea",
        "ghcr.io/zarf-dev/agent"
      ]
    }
  },
  "vulnerability": { "id": "CVE-2025-49796" }
}
```
The locations array is unchanged from a normal scan — it shows where in the container the package was discovered.
annotations["zarf-sbom-source"] lists every bundled SBOM in the Zarf archive

### Common queries

List every bundled source in the Zarf package (output of grype zarf:zarf-package.tar.zst -o json >> findings.json):

```bash
jq -r '[.matches[].artifact.annotations["zarf-sbom-source"][]?] | unique[]' findings.json
```
Count findings per source:

```bash
jq -r '[.matches[] | .artifact.annotations["zarf-sbom-source"][]? as $src | $src]
       | group_by(.) | map({source: .[0], count: length}) | sort_by(-.count)' findings.json
```

Filter findings that touch a specific image:

```bash
jq '[.matches[] | select(.artifact.annotations["zarf-sbom-source"]? | index("ghcr.io/go-gitea/gitea"))]' findings.json
```

The quality of each zarf-sbom-source identifier reflects the source name recorded in the underlying SBOM by Zarf's SBOM generator.

### Core UX

I think a consideration for this change is how we assume this provider will be used:
- table output provides optionality for quality gates - typically by severity
- when additional granularity is required the json output serves as a mechanism for filtering by artifact/image. 

Inherently table output won't show annotations but I believe this is okay. 

### Additional Context

Today the nested `sboms.tar` does contain both sboms and html that we used historically for sbom comparison. We are evaluating the removal of those files but currently the output will contain errors for non-sbom artifacts and continues to process - I think this is okay. 

An older example zarf package with findings for testing: https://github.com/orgs/zarf-dev/packages/container/package/packages%2Fhelm-charts

Open to any and all feedback on where I may have misinterpreted any discussions and any gaps that we need to resolve.


grype/pkg/zarf_provider.go
```diff
@@ -0,0 +1,188 @@
...
+		}
+
+		packages := FromCollection(catalog, s.Relationships, config.SynthesisConfig, enhancers...)
+
+		// annotate each package with provenance back to the originating SBOM, and
+		// propagate the per-SBOM distro to packages that don't already have one
+		// (e.g. syft-JSON SBOMs, where distro is at the SBOM level rather than per-PURL).
+		annotatePackagesFromZarfSBOM(packages, s, hdr.Name, d)
+
+		allPackages = append(allPackages, packages...)
```
kzantow, line 162: Just noting that there's an issue [I'm currently fixing](https://github.com/anchore/grype/pull/3363) we'll want to get sorted out before merging this regarding parent package references.

grype/pkg/zarf_provider.go
```diff
@@ -0,0 +1,188 @@
...
+// annotatePackagesFromZarfSBOM adds a synthetic Location to each package pointing
+// back to the originating SBOM inside the Zarf archive (preferring the SBOM's
+// declared source name, falling back to the tar entry name), and propagates the
+// per-SBOM distro to any package that does not already carry one.
+func annotatePackagesFromZarfSBOM(packages []Package, s *sbom.SBOM, entryName string, d *distro.Distro) {
+	identifier := s.Source.Name
+	if identifier == "" {
+		identifier = entryName
+	}
+	loc := file.NewLocation(zarfLocationPrefix + identifier)
```
kzantow, line 199: This is related to what we were discussing during office hours, if I'm reading this right, the result is that there would be a `zarf:<source-name>` location added and a secondary location that came from whatever's in the SBOM. I think this satisfies the "can I find what I'm looking for" aspect.

grype/pkg/zarf_provider.go
brandtkeller: Tracking! I gave it a read and should be able to rebase with any changes required after that merges.

grype/pkg/zarf_provider.go
brandtkeller: Correct!

kzantow: Sorry for the delay getting back to you on this. I was finally able to talk through the immediate options to get this PR moved forward with the team. Ultimately, I don't think adding synthetic location like this is exactly what we want to do. Instead a reasonably simple solution is to annotate the `Package` with custom annotations holding the image source identifier.

The annotations should end up in the JSON as an object to an array of strings. For Zarf specifically you could probably do something like `"zarf-image-source": [ "<identifier>" ]` or maybe you have a better naming idea, since naming is hard. The reason for the array is that there is a complication: [Grype will deduplicate results](https://github.com/anchore/grype/blob/main/grype/match/match.go#L40) based on [a fingerprint](https://github.com/anchore/grype/blob/main/grype/match/match.go#L34) but we don't want to lose the indicators about which images in the Zarf package have the reported vulnerabilities. It seems better to have a list of strings here to account for it from the start, since we know it's likely to have this scenario (with base images, for example).

In terms of implementation, the main concern is that we probably want to sort the array for consistent output, we could probably just use a `map[string][]string` for now, I don't know if having any custom data types here really help except maybe to automatically sort the list in JSON output.

brandtkeller: acknowledged! let me take a look and see if I follow the recommendation as I trace the source.

brandtkeller: Updated! Handled the dedup concern in the provider rather than in Match.Merge, since the JSON presenter reads via pkg.ByID against the package slice rather than from m.Package. Open to moving the union into Match.Merge if you'd prefer that location.

Updated the PR description to reflect the changes if you'd like to review before looking at the code.

brandtkeller:
rebased with main - I'll try and make the next community call in the event there are any items we want to discuss.

grype/pkg/zarf_provider.go
```diff
@@ -0,0 +1,213 @@
...
+// or binary it describes) inside a Zarf archive a package came from. Surfaces
+// in JSON output as `artifact.annotations["zarf-sbom-source"]`.
+const zarfSBOMSourceAnnotation = "zarf-sbom-source"
+
+const zarfInputPrefix = "zarf:"
+
+// maxSBOMEntryBytes is the maximum size of a single SBOM entry in sboms.tar.
+// This is arbitrarily selected and could be improved with empirical data from
+// large production Zarf packages.
+const maxSBOMEntryBytes = 256 << 20 // 256 MB
```
kzantow, line 31: nit: `256 * 1024 * 1024` might be a little more clear?

grype/pkg/provider.go
```diff
@@ -227,6 +227,19 @@ func provide(userInput string, config ProviderConfig, applyChannel func(d *distr
 		return packages, ctx, s, err
 	}
 
+	packages, ctx, s, err = zarfProvider(userInput, config, applyChannel)
```
kzantow, line 230: ~~if a user scans a zarf archive without the `zarf:` prefix, would syft open and treat it as an archive that it doesn't find packages from? e.g. should this check get moved _before_ syftSBOMProvider so a user who doesn't specify `zarf:` will still get zarf results?~~ this is after `syftSBOMprovider, the syftProvider is _after_ this.

brandtkeller:
cleaned up linting findings

kzantow (approved):
Thanks for being responsive to the updates @brandtkeller; this looks good!

Issue #3368, opened 2026-04-14
Expand ignore rules to owned sub packages of distro packages


willmurphyscode:
**Background**:

In https://github.com/anchore/grype/pull/3282 (which had terrible perf and was fixed by @kzantow in https://github.com/anchore/grype/pull/3355 ) we introduced behavior where:

1. An APK package owns a language package (npm, PyPI, etc)
2. The Alpine secDB or similar source for the distro explicitly says that the APK is either "fixed" or "not affected" for CVE-1234-1234
3. Then the APK matcher should emit an ignore rule that says the language package is not affected by GHSAs whose alias is CVE-1234-1234.

**What would you like to be added**:

This behavior should be expanded to:
- [ ] rpm matcher - https://github.com/anchore/grype/pull/3326
- [ ] dpkg matcher - also https://github.com/anchore/grype/pull/3326
- [ ] alpm matcher?
- [ ] grype db v6 OS schema transformer: https://github.com/anchore/grype/pull/3370
- [ ] RHEL vunnel provider: https://github.com/anchore/vunnel/pull/1134
- [ ] Ubuntu vunnel provider: https://github.com/anchore/vunnel/pull/1136
- [ ] Debian vunnel provider: https://github.com/anchore/vunnel/pull/1137
- [ ] (maybe additional vunnel providers)

**Why is this needed**:

1. Right now there are numerous false positives in these situations.
2. The current heuristic approach for "comprehensive distros" has some false positives and false negatives

**Additional context**:

- This should fix many of the numerous FPs against SLES that grype currently has.
- This should unblock the removal of the [current heuristic](https://github.com/anchore/grype/blob/475e3bad8f9b5f92d98f1d17c41d18823471b638/grype/pkg/package.go#L195) which applies only to certain distros and has both fps and fns.
- There will be vunnel work required (issue there forthcoming)

PR #3326, opened 2026-03-30
feat: ignore known not-vulnerable records across related packages

kzantow:
This PR is a follow-on to apply changes from #3286 and #3282 across other matchers. This PR (as well as the alternate implementation in #3304) have the result of fixing [quite a lot of the reported SLES issues](https://github.com/anchore/grype/issues?q=is%3Aissue%20state%3Aopen%20label%3Apackage-overlap) where we have _fixed_ information for specific RPM packages but the Python or other language packages erroneously result in the same vulnerability.

An example using `SLES 15.6` and `urllib3` (with `docker build -t suse-overlap:latest`):
```dockerfile
FROM registry.suse.com/suse/sle15:15.6
RUN zypper in -y --no-recommends python311-urllib3=2.0.7-150400.7.21.1
```

Based on the vulnerabilities in the Grype database for `15.6`:
```
$ grype db search python311-urllib3 --distro sles:15.6   
VULNERABILITY   PACKAGE            ECOSYSTEM  NAMESPACE              VERSION CONSTRAINT       
CVE-2016-9015   python311-urllib3  rpm        sles:distro:sles:15.6  < 0:2.0.7-150400.7.11.1  
CVE-2018-20060  python311-urllib3  rpm        sles:distro:sles:15.6  < 0:2.0.7-150400.7.11.1  
CVE-2019-11236  python311-urllib3  rpm        sles:distro:sles:15.6  < 0:2.0.7-150400.7.11.1  
CVE-2019-11324  python311-urllib3  rpm        sles:distro:sles:15.6  < 0:2.0.7-150400.7.11.1  
CVE-2019-9740   python311-urllib3  rpm        sles:distro:sles:15.6  < 0:2.0.7-150400.7.11.1  
CVE-2020-26137  python311-urllib3  rpm        sles:distro:sles:15.6  < 0:2.0.7-150400.7.11.1  
CVE-2021-33503  python311-urllib3  rpm        sles:distro:sles:15.6  < 0:2.0.7-150400.7.11.1  
CVE-2023-43804  python311-urllib3  rpm        sles:distro:sles:15.6  < 0:2.0.7-150400.7.11.1  
CVE-2023-45803  python311-urllib3  rpm        sles:distro:sles:15.6  < 0:2.0.7-150400.7.11.1  
CVE-2024-37891  python311-urllib3  rpm        sles:distro:sles:15.6  < 0:2.0.7-150400.7.18.1  
[...]

PR #3370, opened 2026-04-15
feat(os transformer): unaffected package handles

willmurphyscode:
Previously, vunnel dropped records in OS schema providers when the record indicated a package was not affected, since that could not result in a match. However, in order to suppress false positives in the case where a distro has recorded a package is not affected, and that package brings in a language package (npm, PyPI, etc), these records are forwarded by vunnel as if fixed at "version 0". In Grype, take version 0 FixedIn records and transform them into Unaffected Packages in the database.

Notably, skip doing this for APKs, since Alpine already has a separate system of NAKs built around forwarding the APK "< 0" versions.

This is the database transform component of https://github.com/anchore/grype/issues/3368

### DIFF
--- a/grype/db/v6/build/transformers/os/testdata/rhel-8-not-affected.json
+++ b/grype/db/v6/build/transformers/os/testdata/rhel-8-not-affected.json
@@ -1,0 +1,52 @@
+[
+  {
+    "Vulnerability": {
+      "CVSS": [
+        {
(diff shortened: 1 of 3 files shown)


willmurphyscode:
Was pulled into https://github.com/anchore/grype/pull/3326/ - no longer needed.

PR #3372, opened 2026-04-16
fix: compare mismatched package / db versions

kzantow:
This PR corrects an issue where explicitly set database constraint version types that did not match package version types were dropping matches, for example: an APK package and a Go version type in the db.

NOTE: the [version compare function](https://github.com/anchore/grype/blob/main/grype/version/version.go#L96-L126) already accounts for package and db version differences, this check must have been missed when the latest refactorings were made to the version package.

### DIFF
--- a/grype/db/v6/vulnerability_provider_mocks_test.go
+++ b/grype/db/v6/vulnerability_provider_mocks_test.go
@@ -65,7 +65,7 @@ func testVulnerabilityProvider(t *testing.T) vulnerability.Provider {
 		{
 			PackageName:       "neutron",
 			Namespace:         "debian:distro:debian:8",
-			VersionConstraint: "< 70.3.0-rc0", // intentionally bad value
+			VersionConstraint: "< fred70.3.0-rc0", // intentionally bad value
 			ID:                "CVE-2014-fake-3",
 			VersionFormat:     "apk",
 		},
--- a/grype/matcher/apk/matcher_test.go
+++ b/grype/matcher/apk/matcher_test.go
@@ -84,6 +84,41 @@ func TestSecDBOnlyMatch(t *testing.T) {
 	assertMatches(t, expected, actual)
 }
 
+func TestCPEConstraintVersionsOtherThanAPK(t *testing.T) {
+	nvdVuln := vulnerability.Vulnerability{
+		Reference: vulnerability.Reference{
+			ID:        "CVE-2020-1234",
+			Namespace: "nvd:cpe",
+		},
+		PackageName: "libvncserver",
+		Constraint:  version.MustGetConstraint("<= 0.9.11", version.GolangFormat),
+		CPEs: []cpe.CPE{
+			cpe.Must(`cpe:2.3:a:lib_vnc_project-\(server\):libvncserver:*:*:*:*:*:*:*:*`, ""),
+		},
+	}
+
+	vp := mock.VulnerabilityProvider(nvdVuln)
+
+	m := Matcher{}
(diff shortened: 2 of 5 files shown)


philroche:
I have verified these changes locally with a known (private to Chainguard, unfortunately) APK package and image SBOM, which was previously missing detections of type APK. With this proposed change here, these detections of type APK have been restored. +1

grype/db/v6/vulnerability_provider_mocks_test.go
```diff
@@ -65,7 +65,7 @@ func testVulnerabilityProvider(t *testing.T) vulnerability.Provider {
 		{
 			PackageName:       "neutron",
 			Namespace:         "debian:distro:debian:8",
```
willmurphyscode, line 67: I don't think it matters for this PR but it's a confusing piece of test data to have a debian 8 apk package. Is this change part of this PR?

grype/version/portage_version_test.go
```diff
@@ -146,19 +146,6 @@ func TestPortageConstraint_Constraint_NilVersion(t *testing.T) {
 	}
 }
 
-func TestPortageVersion_Constraint_UnsupportedFormat(t *testing.T) {
```
willmurphyscode, line 149: Just to make sure I understand: this test is being deleted because grype should instead try to parse the portage version as a semantic format, which will succeed, so the test doesn't mean anything now?

Would it be worth keeping the test and changing the assertion?

grype/matcher/apk/matcher_test.go
```diff
@@ -84,6 +84,41 @@ func TestSecDBOnlyMatch(t *testing.T) {
...
 }
 
+func TestCPEConstraintVersionsOtherThanAPK(t *testing.T) {
+	nvdVuln := vulnerability.Vulnerability{
+		Reference: vulnerability.Reference{
+			ID:        "CVE-2020-1234",
+			Namespace: "nvd:cpe",
+		},
+		PackageName: "libvncserver",
+		Constraint:  version.MustGetConstraint("<= 0.9.11", version.GolangFormat),
```
willmurphyscode, line 94: Let me make sure I understand: the grype db vulnerability record (affected CPE blob) has a version format of golang. Is that because the target software field in the ingested CPE had Golang in it?

So before the removal at https://github.com/anchore/grype/pull/3372/changes#diff-134ffc77b2487bfe3c694605a63e0763fa602000969a07c9771bdfe40d622400L59 , this wouldn't have matched, because the version comparison would have had an error?

willmurphyscode (commented):
I think this looks good but I had a couple of questions in line.

grype/db/v6/vulnerability_provider_mocks_test.go
kzantow: see the comment: this is supposed to be a unparseable version. I just changed it so it's unparseable in all senarios

grype/matcher/apk/matcher_test.go
kzantow: That's exactly right. This is the reported problem. We [made a change](https://github.com/anchore/grype/pull/3328) so these CPE version types are being set from the TargetSW field. These _definitely do not match APK_.

grype/version/portage_version_test.go
kzantow: This test is just testing the bad assumption of the mismatch of version types. I don't see a need to replace it with something else

willmurphyscode (approved):

PR #3089, opened 2025-12-03
version comparison uppercase V prefixes fix

wasup-yash:
updated for Uppercase V based on https://github.com/anchore/grype/issues/3037#issuecomment-3512586943 

fixes: #3037

PR #3385, opened 2026-04-21
fix(version): accept uppercase V prefix in fuzzy version comparison

### LINKED ISSUE #3037: Treat uppercase V prefixes the same as lowercase v prefixes in version comparison
**Vulnerability IDs**:

**Problem**:

**What you need to know**:

**Environment**:
- `grype version`: 0.102.0

Thank you for your support!

Dashtid:
## Summary

- Fuzzy version comparison had two hardcoded lowercase `v` prefix strippers — uppercase `V` versions like `V1.2.3` or `V1500-SP2` were not normalized and compared as unrelated strings against CVE ranges
- Result: silent false negatives when a CPE or SBOM package carried a `V`-prefixed version (e.g. the Bosch `V1500-SP2` case from the issue)
- Fix both locations to accept `[vV]`, matching the existing `trimLeadingV` helper in `apk_version.go`

Fixes #3037

## Details

`pseudoSemverPattern` at [fuzzy_constraint.go:16](grype/version/fuzzy_constraint.go#L16) used `^v?` — only lowercase. `V`-prefixed versions failed the gate and fell through to the fuzzy path where `stripLeadingV` at [fuzzy_constraint.go:213](grype/version/fuzzy_constraint.go#L213) then did `strings.TrimPrefix(ver, \"v\")` — also lowercase. Both paths left the `V` in place, producing wrong comparisons.

Changes:
1. Regex: `^v?` -> `^[vV]?` — strictly more permissive, no behavior change for existing `v`/no-prefix inputs
2. `stripLeadingV`: explicit `ver[0] == 'v' || ver[0] == 'V'` check, with empty-string guard to preserve `TrimPrefix`'s safety for empty inputs

## Scope note

There's a related third location at [semantic_version.go:22-32](grype/version/semantic_version.go#L22-L32) (`versionStartsWithV` regex and a lowercase-only `TrimPrefix`) that would also benefit from this change. I intentionally left it out of scope — @wagoodman's comment on #3037 named exactly the two locations in this PR, and `newSemanticVersion` has callers outside the fuzzy path that I haven't audited. Happy to file a follow-up if maintainers want it addressed.

## Test plan

- [x] `TestPseudoSemverPattern`: 2 new cases (V prefix accepted, v prefix as baseline)
- [x] `TestFuzzyVersionComparison`: 4 new cases exercising `stripLeadingV` through the comparator — V vs v equal, V vs no-prefix equal, V ordering correctness, mixed-case cross-version ordering
- [x] `TestFuzzyVersion_Constraint`: 3 end-to-end cases — V on version satisfies range, V on both sides, V out of range returns false (guard against false positives)
- [x] Full `grype/version/...` test suite passes with no regressions

### DIFF
--- a/grype/version/fuzzy_constraint.go
+++ b/grype/version/fuzzy_constraint.go
@@ -12,8 +12,8 @@ import (
 
 // derived from https://semver.org/, but additionally matches:
 // - partial versions (e.g. "2.0")
-// - optional prefix "v" (e.g. "v1.0.0")
-var pseudoSemverPattern = regexp.MustCompile(`^v?(0|[1-9]\d*)(\.(0|[1-9]\d*))?(\.(0|[1-9]\d*))?(?:(-|alpha|beta|rc)((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$`)
+// - optional prefix "v" or "V" (e.g. "v1.0.0", "V1.0.0")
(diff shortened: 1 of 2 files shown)


kzantow:

PR #3394, opened 2026-04-24
Add support for CycloneDX Vex

### LINKED ISSUE #2462: why VEX fail with grype?
Hi everyone, I recently uploaded a sbom and then scanned it but when i scanned with VEX it gives an error which is this: 

**without vex flag**
asad@asad-Lenovo-V330-15IKB:~/Downloads/wwe$ grype sbom:0vo0efli.cdx.json 
   ├── by severity: 0 critical, 0 high, 1 medium, 0 low, 0 negligible
   └── by status:   1 fixed, 0 not-fixed, 0 ignored 
[0000]  WARN attempted CPE search on OWASP Dependency-Track, which has no CPEs. Consider re-running with --add-cpes-if-none
NAME    INSTALLED  FIXED-IN  TYPE          VULNERABILITY        SEVERITY 
lucene  8.11.4     9.12.0    java-archive  GHSA-g643-xq6w-r67c  Medium

**with vex flag**
asad@asad-Lenovo-V330-15IKB:~/Downloads/wwe$ grype sbom:0vo0efli.cdx.json --vex 0vo0efli.cdx.json 
   ├── by severity: 0 critical, 0 high, 1 medium, 0 low, 0 negligible
   └── by status:   1 fixed, 0 not-fixed, 0 ignored 
[0000]  WARN attempted CPE search on OWASP Dependency-Track, which has no CPEs. Consider re-running with --add-cpes-if-none
[0000] ERROR unable to find matches against VEX sources: unable to find matches against VEX documents: parsing vex document: merging vex documents: opening 0vo0efli.cdx.json: unable to detect document format reading 0vo0efli.cdx.json

THE SBOM FILE IS ATTACHED, any reason for this? 

[0vo0efli.cdx.json](https://github.com/user-attachments/files/18866336/0vo0efli.cdx.json)

bennetleins:
## Summary

These changes add the capability to use a CycloneDX Vex file. CycloneDX Vex files don’t have a designated PURL or CPE field. Because of that only the CVE ID is matched to ignore a vulnerability right now.

## Motivation

Some applications like Dependency Track provide the option to export a Vex file. The only available format in this case is CycloneDX.

## Changes

- Added CycloneDX Vex implementation
- Added CycloneDX Vex tests
- Added CycloneDX Vex matcher unit test

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## Checklist

Closes #2462


noqt (changes requested):
Reviewing this for NOQT. This needs component scoping before it is safe to merge.

`FilterMatches` calls `findMatchingVulnerability(boms, m.Vulnerability.ID)`, and the lookup stops at the first matching CVE. Neither that lookup nor `matchingRule` checks `vuln.Affects` against `m.Package`. A CycloneDX statement that marks CVE-X `not_affected` for `pkg:npm/package-a` will therefore suppress CVE-X on `pkg:npm/package-b` too. The bundled `cdx-alpine.json` already contains `affects[].ref` PURLs, so the data needed to avoid that cross-package suppression is present in the format.

Please scope each statement to the matched package: handle a direct PURL ref and, where the ref is a component `bom-ref`, resolve it through the BOM component before comparing package identity. If a statement cannot be scoped safely, keep the match rather than applying a global suppression. A regression with two packages sharing one CVE, but only one listed in `affects`, should leave the other package active.

I checked exact head `7a10b1625d3a268cabb95e1f231beea3b3795326`. It merges cleanly with current main `f6f2ce0a2293e65c97bf5468926cb0c9ddad822f`, and the exact-range diff passes `git diff --check`. Go isn't available in this environment, so I'm not claiming a local test pass.

PR #3396, opened 2026-04-25
fix(registry): warn when insecure transport is configured

### LINKED ISSUE #3101: Add Runtime Warnings When TLS Verification Is Disabled or HTTP Is Enabled
grype currently allows registry communication using:

* `insecure-skip-tls-verify`
* `insecure-use-http`
* corresponding environment variables

These options fully disable TLS verification or fall back to plaintext HTTP, which can expose registry credentials and enable MITM in CI/CD environments.

Even when these insecure modes are enabled, grype does not show a **runtime warning**, including when running with `--log-level debug`.

This makes it difficult for users to realize that they are operating with insecure transport.

---

## **Problem Details**

Examples:

```yaml
registry:
  insecure-skip-tls-verify: true
  insecure-use-http: true
```

Running:

```sh
grype --config grype-insecure.yaml docker.io/library/nginx:latest
```

Produces no warning such as:

* “TLS validation has been disabled”
* “Using insecure HTTP transport”

In CI/CD environments, insecure configuration may be automatically injected via environment variables, without direct user knowledge.

---

## **Impact**

When transport security is disabled without clear warnings:

* registry credentials may be exposed in plaintext
* MITM attacks may modify OCI image layers
* false vulnerability results may be produced
* pipelines might unknowingly operate insecurely

This directly affects software supply-chain integrity when grype is used as a security gate.

---

## **Related Example**

GitGuardian’s **GGShield** recently addressed similar concerns by adding warnings and improving secure-by-default behavior when TLS validation was disabled.

This shows that explicit runtime warnings are considered best practice for DevSecOps tooling.

---

## **Suggested Improvements**

When insecure transport is enabled:

* log a warning at `INFO` or higher level
* optionally show a CLI message the first time
* document secure alternatives (CA bundles, self-signed handling)

Optional UX additions:

* require explicit “I understand the risk” confirmation
* display a startup banner when insecure settings are active

---

## **Why this is useful**

Users may intentionally want to disable TLS for testing, but they should never lose transport security **without knowing it happened**.

Warnings would improve visibility without removing existing functionality.

---

### Thanks! Happy to provide additional reproduction details or open a PR if desired.

---

Dashtid:
## Summary

- Emit a one-time warning during config load when `insecure-skip-tls-verify` or `insecure-use-http` is set
- These flags can be silently picked up from a config file or env var, leaving registry traffic unprotected with no indication in normal CLI output
- The warning lists exactly which insecure flag(s) are enabled so the user can find and address them

Fixes #3101 (grype CLI half — see scope note below).

## Details

The warning is emitted from the existing `PostLoad()` hook on the registry config struct, which is the natural single point where merged config (flag + env + YAML) is available exactly once per invocation. There is precedent for `log.Warnf` from inside `PostLoad()` in the same package (see [database_search_packages.go](cmd/grype/cli/options/database_search_packages.go) lines 43, 62).

Implementation:

- `insecureTransportWarning() string` is a pure helper that returns the message, or empty if neither flag is set. Splitting message construction from the log call lets the branching logic be unit-tested deterministically without touching the logger singleton.
- `PostLoad()` calls the helper and emits `log.Warn(msg)` only when non-empty.
- No behavior change beyond the new log line.

Sample output when both flags are set:

```
[WARN] registry communication is insecure: insecure-skip-tls-verify, insecure-use-http enabled
```

## Scope note

#3101 outlined two changes: (1) a debug log inside stereoscope at the actual TLS/HTTP application sites, and (2) a warning at the grype CLI level. This PR addresses only (2), which @Tanish-26 indicated was the higher priority (\"If I had to choose only one, I'd prioritize the grype CLI warning\"). The stereoscope debug log can follow as a separate, smaller PR to anchore/stereoscope.

## Test plan

- [x] `Test_registry_insecureTransportWarning`: 5 cases covering neither set, each alone, both set, and credentials-without-insecure-flags (regression guard)
- [x] `Test_registry_PostLoad_returnsNoError`: verifies all 4 flag combinations don't return an error from PostLoad (warning is observability, must never block config load)
- [x] Full `cmd/grype/cli/options/...` test suite passes — no regressions

### DIFF
--- a/cmd/grype/cli/options/registry.go
+++ b/cmd/grype/cli/options/registry.go
@@ -1,9 +1,12 @@
 package options
 
 import (
+	"fmt"
 	"os"
+	"strings"
 
 	"github.com/anchore/clio"
+	"github.com/anchore/grype/internal/log"
 	"github.com/anchore/stereoscope/pkg/image"
 )
 
@@ -53,7 +56,30 @@ func (cfg *registry) PostLoad() error {
 			},
 		}, cfg.Auth...)
 	}
+
+	if msg := cfg.insecureTransportWarning(); msg != "" {
+		log.Warn(msg)
+	}
+
 	return nil
+}
+
+// insecureTransportWarning returns a warning message listing any insecure registry transport
+// flags that are enabled, or an empty string if none are. These flags may be picked up from a
+// config file or environment variable without the user realizing, so surfacing them in CI/CD
+// output makes the risk visible.
+func (cfg *registry) insecureTransportWarning() string {
+	var flags []string
+	if cfg.InsecureSkipTLSVerify {
+		flags = append(flags, "insecure-skip-tls-verify")
+	}
+	if cfg.InsecureUseHTTP {
+		flags = append(flags, "insecure-use-http")
+	}
+	if len(flags) == 0 {
+		return ""
+	}
+	return fmt.Sprintf("registry communication is insecure: %s enabled", strings.Join(flags, ", "))
 }
 
 func (cfg *registry) DescribeFields(descriptions clio.FieldDescriptionSet) {
--- a/cmd/grype/cli/options/registry.go
+++ b/cmd/grype/cli/options/registry.go
@@ -1,7 +1,6 @@
 package options
 
 import (
-	"fmt"
 	"os"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


cmd/grype/cli/options/registry.go
```diff
@@ -53,9 +56,32 @@ func (cfg *registry) PostLoad() error {
...
+	if cfg.InsecureSkipTLSVerify {
+		flags = append(flags, "insecure-skip-tls-verify")
+	}
+	if cfg.InsecureUseHTTP {
+		flags = append(flags, "insecure-use-http")
+	}
+	if len(flags) == 0 {
+		return ""
+	}
+	return fmt.Sprintf("registry communication is insecure: %s enabled", strings.Join(flags, ", "))
```
kzantow: seems like this could just log here instead of returning a warning message

Tanish-26 (commented):
Thanks for implementing this — this looks like a very reasonable and practical approach.

I especially like that the warning is emitted from `PostLoad()`, since that ensures merged configuration from flags, env vars, and YAML is evaluated once in a central location.

The warning text is also clear and actionable:

> `registry communication is insecure: ... enabled`

which should help users quickly identify unexpected insecure transport configuration in CI/CD logs.

I also appreciate the separation between:

* the grype CLI warning layer (this PR)
* the lower-level stereoscope debug logging

That split makes sense architecturally.

Thanks again for taking the concern seriously and improving visibility around insecure transport usage.

cmd/grype/cli/options/registry.go
Dashtid: Good call — inlined it to match the `database_search_packages.go` pattern (`log.Warnf` directly in `PostLoad`, no string-returning helper). Done in a458d0d0.

Dashtid:
Friendly bump on this — addressed your comment in `a458d0d0` about 9 days ago. For context, the stereoscope-side companion ([stereoscope#587](https://github.com/anchore/stereoscope/pull/587)) just merged, so this grype-side PR is the last open piece of #3101. Happy to address anything else when you get a chance.

spiffcs (approved):

PR #3406, opened 2026-04-30
chore(rpm-unit-tests): migrate to dbtest pattern

willmurphyscode:


grype/matcher/rpm/almalinux_test.go
```diff
@@ -78,1954 +56,380 @@ func TestShouldUseAlmaLinuxMatching(t *testing.T) {
...
+					ModularityLabel: strPtr("httpd:2.4:1234:5678"),
+				}).
+				Build()
+
+			findings := db.Match(t, &matcher, p)
+			findings.IsEmpty()
+			// the rhel disclosure path produces "Distro Fixed" since the binary
+			// package version is past the RHEL fix
+			findings.Ignores().SkipCompleteness().
+				SelectRelatedPackageIgnore("Distro Fixed", "CVE-2021-40438").
```
wagoodman: we should start using constants for all ignore reasons

grype/matcher/rpm/almalinux_test.go
```diff
@@ -78,1954 +56,380 @@ func TestShouldUseAlmaLinuxMatching(t *testing.T) {
...
+					Epoch:           intPtr(0),
+					ModularityLabel: strPtr("httpd:2.4:1234:5678"),
+				}).
+				Build()
+
+			findings := db.Match(t, &matcher, p)
+			findings.IsEmpty()
+			// the rhel disclosure path produces "Distro Fixed" since the binary
+			// package version is past the RHEL fix
+			findings.Ignores().SkipCompleteness().
```
wagoodman: We should use skip completeness sparingly

grype/matcher/rpm/almalinux_test.go
```diff
@@ -78,1954 +56,380 @@ func TestShouldUseAlmaLinuxMatching(t *testing.T) {
...
+			require.Len(t, matches, 1, "expected one match")
+
+			m := matches[0]
+			assert.Equal(t, "CVE-2021-40438", m.Vulnerability.ID)
+			// fix info should be from AlmaLinux ALSA, not RHEL
+			require.Equal(t, vulnerability.FixStateFixed, m.Vulnerability.Fix.State)
+			require.Equal(t, []string{"2.4.37-43.module_el8.5.0+2597+c4b14997.alma"}, m.Vulnerability.Fix.Versions)
+			// advisory should reference ALSA-2021:4537
+			require.Len(t, m.Vulnerability.Advisories, 1)
+			assert.Equal(t, "ALSA-2021:4537", m.Vulnerability.Advisories[0].ID)
```
wagoodman: I want to say we should find common assertion cases like this and introduce helpers over asserting fields on the struct. It's not like this is a bad thing or we should stop doing it, but when we refactor the top-level API objects we'll thank ourselves later for helpers in more (non-contrived) places.

This feels right here (assert a vuln ID with fix version) ... questioning the advisories, but honestly, I can see how that would help with some of the awkwardness here too.

grype/matcher/rpm/almalinux_test.go
```diff
@@ -78,1954 +56,380 @@ func TestShouldUseAlmaLinuxMatching(t *testing.T) {
...
+			matcher := Matcher{}
+			pkgID := pkg.ID("patch-fixed")
+			p := dbtest.NewPackage("patch", "2.7.6-11.el8", syftPkg.RpmPkg).
+				WithID(pkgID).
+				WithDistro(dbtest.AlmaLinux8).
+				Build()
+
+			findings := db.Match(t, &matcher, p)
+			findings.IsEmpty()
+			findings.Ignores().SkipCompleteness().
```
wagoodman: I'm having a hard time understanding if skip completeness is being used correctly or not. There are a couple of ways to use it wrong:
- we skip on multiple findings which we should have had some assertion for (don't know how to know automatically detect this case)
- we skip on things that were asserted anyway (we can detect this)

maybe we should fail the test if skip completeness was specified but we're in the latter case.

internal/dbtest/package.go
```diff
@@ -59,6 +62,14 @@ func (b *PackageBuilder) WithType(t syftPkg.Type) *PackageBuilder {
 	return b
 }
 
+// WithID overrides the auto-generated package ID. Use this when a test needs
+// a stable, comparable ID (e.g., to assert against an IgnoreFilter that
+// references the package by ID).
+func (b *PackageBuilder) WithID(id pkg.ID) *PackageBuilder {
```
wagoodman, line 68: ~this is good, slightly related... we should remove Build() entirely~ oversight on my part: we dont get the package until we call build

wagoodman: maybe in the future we'd consider make a package embedding in the builder, I bed we'd run into problems so not a comment on this pr

grype/matcher/rpm/almalinux_test.go
```diff
@@ -78,1954 +56,380 @@ func TestShouldUseAlmaLinuxMatching(t *testing.T) {
...
+			// whose RHEL fix the package is past
+			igs.SelectRelatedPackageIgnore("Distro Fixed", "CVE-2021-40438").ForPackage(pkgID)
+			igs.SelectRelatedPackageIgnore("Distro Fixed", "CVE-2021-26691").ForPackage(pkgID)
+
+			// the alma matcher emits "Alma Unaffected" for the ALSA itself
+			// AND for each CVE the ALSA references (alias unwind)
+			igs.SelectRelatedPackageIgnore("Alma Unaffected", "ALSA-2021:4537").ForPackage(pkgID)
+			igs.SelectRelatedPackageIgnore("Alma Unaffected", "CVE-2021-40438").ForPackage(pkgID)
+			igs.SelectRelatedPackageIgnore("Alma Unaffected", "CVE-2021-26691").ForPackage(pkgID)
+			igs.SelectRelatedPackageIgnore("Alma Unaffected", "CVE-2021-20325").ForPackage(pkgID)
```
wagoodman: we should accept variadic IDs on this method and similar methods

grype/matcher/rpm/rhel_eus_test.go
```diff
@@ -769,900 +772,203 @@ func TestResolveEUSDisclosures(t *testing.T) {
...
+				SkipCompleteness().
+				HasCount(1).
+				ContainsVulnerabilities("CVE-2024-0340")
+			findings.Ignores().IsEmpty()
+
+			m := findings.Matches()[0]
+			// fix info should reflect the EUS-reachable fix
+			assert.Equal(t, vulnerability.FixStateFixed, m.Vulnerability.Fix.State)
+			assert.Equal(t, []string{"0:5.14.0-427.68.1.el9_4"}, m.Vulnerability.Fix.Versions)
+			assertEUSDualDistroDetails(t, m, "9.4")
```
wagoodman: This is functional, but we could make this more readable. What seems to be happening everywhere is grouping assertions into two parts:
- assert on the group/universe of findings, followed by... 
- assert elements on specific findings

Nothing wrong with this, but the helpers make this pattern irrelevant (the AI is orienting around how testify assertions tend to be, which we've obviated some).

So in this case something like:
```
db.Match(t, &matcher, p).
	SelectMatch("CVE-2024-0340").
    .MoreAssertions(...)  # made up...
```

Or even without the MoreAssertions part to just get the match out:

```
m := db.Match(t, &matcher, p).
	SelectMatch("CVE-2024-0340").
    .Match()  # made up...
```

which would return the raw match itself to do normal assertions on.

grype/matcher/rpm/almalinux_test.go
```diff
@@ -78,1954 +55,368 @@ func TestShouldUseAlmaLinuxMatching(t *testing.T) {
...
+				Build()
+
+			findings := db.Match(t, &matcher, p)
+			sf := findings.SelectMatch("CVE-2021-40438")
+			// fix info should be from AlmaLinux ALSA, not RHEL
+			sf.HasFix(vulnerability.FixStateFixed, "2.4.37-43.module_el8.5.0+2597+c4b14997.alma").
+				HasAdvisories("ALSA-2021:4537")
+			sf.SelectDetailByDistro("redhat", "8"). // alma matching queries the rhel namespace
+								HasMatchType(match.ExactDirectMatch)
+			findings.Ignores().IsEmpty()
```
wagoodman: this line (the findings ignores is empty) should not be required. That is, if we are doing completeness testing, the nice thing about it is that if we don't assert on a finding we get an error, so asserting it's empty is redundant. 

My proposal here is if it's not working quite like I mentioned above we should probably change it (what do you think about this?). The main motivation is to reduce the number of boilerplate assertions by making the assertion helpers smarter. This kind of feature requires to know the full set of ignores, which is already a requirement in completness testing, which makes me feel it's possible/the right direction.

grype/matcher/rpm/almalinux_test.go
```diff
@@ -78,1954 +55,368 @@ func TestShouldUseAlmaLinuxMatching(t *testing.T) {
...
+					ModularityLabel: strPtr("httpd:2.4:1234:5678"),
+				}).
+				Build()
+
+			findings := db.Match(t, &matcher, p)
+			findings.IsEmpty()
+			// the rhel disclosure path produces "Distro Fixed" since the binary
+			// package version is past the RHEL fix
+			findings.Ignores().
+				HasCount(1).
```
wagoodman: maybe we should remove the HasCount helper entirely? I feel like its a zero-force member here and in most places I see it used.

grype/matcher/rpm/almalinux_test.go
wagoodman: For clarity: The IsEmpty directly on findings I think is helpful / useful, it makes it clear that they whole result should be empty. Where as within findings if we look at only matches or only ignores, I feel that they are mostly noise that dont really clarify anything.

grype/matcher/rpm/almalinux_test.go
```diff
@@ -78,1954 +55,368 @@ func TestShouldUseAlmaLinuxMatching(t *testing.T) {
...
+				WithDistro(dbtest.AlmaLinux8).
+				Build()
+
+			findings := db.Match(t, &matcher, p)
+			findings.IsEmpty()
+			findings.Ignores().
+				HasCount(1).
+				SelectRelatedPackageIgnore(IgnoreReasonDistroFixed, "CVE-2019-13636").
+				ForPackage(pkgID)
+		})
```
wagoodman: ah, amending my thoughts ... I think the awkwardness I'm finding is that there is a top level IsEmpty() which is ONLY about matches, and the ignores helper that is only for ignores... they are really siblings but the hierarchy in the api here that's not the case.

wagoodman: I think the conclusion from the earlier comment still stands, but hints that the top level IsEmpty could use more smartness being encoded.

grype/matcher/rpm/almalinux_test.go
```diff
@@ -78,1954 +55,368 @@ func TestShouldUseAlmaLinuxMatching(t *testing.T) {
...
```
wagoodman: yeah there is a smell here in the helpers... reading the english here is confusing, because we're stating we "have no findings" but "assert things on specific findings".

grype/matcher/rpm/almalinux_test.go
```diff
@@ -78,1954 +55,368 @@ func TestShouldUseAlmaLinuxMatching(t *testing.T) {
...
+				WithMetadata(pkg.RpmMetadata{
+					Epoch:           intPtr(0),
+					ModularityLabel: strPtr("httpd:2.4:1234:5678"),
+				}).
+				Build()
+
+			findings := db.Match(t, &matcher, p)
+			findings.IsEmpty()
+
+			igs := findings.Ignores().HasCount(6)
```
wagoodman: more ammo for deleting HasCount helper methods: this provides no value since completeness testing ensures this test will fail if we dont have the below assertions. Thus this count line is not needed and will cause more issues in debugging anyway (and is more to read).

grype/matcher/rpm/almalinux_test.go
```diff
@@ -78,1954 +55,368 @@ func TestShouldUseAlmaLinuxMatching(t *testing.T) {
...
+			p := dbtest.NewPackage("tar", "2:1.30-5.el8", syftPkg.RpmPkg).
+				WithDistro(dbtest.AlmaLinux8).
+				WithMetadata(pkg.RpmMetadata{Epoch: intPtr(2)}).
+				Build()
+
+			findings := db.Match(t, &matcher, p)
+			sf := findings.SelectMatch("CVE-2005-2541")
+			sf.HasFix(vulnerability.FixStateWontFix)
+			sf.SelectDetailByDistro("redhat", "8").
+				HasMatchType(match.ExactDirectMatch)
```
wagoodman: thinking about ergonomics here:
```
findings.SelectMatch("CVE-...").AssertMatch(
  func(m *dbtest.SingleFindingAssertion) {
     m.HasFix(vulnerability.FixStateWontFix).
     SelectDetailByDistro(
       func(d *dbtest.DistroDetailAssertion) {
         d.HasMatchType(match.ExactDirectMatch)
      }
    )
  }
)
```

I'm putting this out there as an alternative... preventing the need for local variables in the stack.

Now that I've written this comment I think I hate it, but I'll put it out there to at least call out "is there a better way to do this without making `sf` local variables everywhere?" and put forth at least something.

I'm also ok with "this is not a problem" and moving on.

grype/matcher/rpm/almalinux_test.go
```diff
@@ -78,1954 +55,351 @@ func TestShouldUseAlmaLinuxMatching(t *testing.T) {
...
-			VersionConstraint: vuln.Constraint.String(),
-		},
-		Confidence: 1.0,
-	}}
+// TestAlmaLinuxMatching_NonModularVulnerable verifies that a non-modular
+// vulnerable package matches via AlmaLinux logic and the fix info is preserved
+// (the AlmaLinux fix here matches the RHEL fix exactly, so no replacement).
+func TestAlmaLinuxMatching_NonModularVulnerable(t *testing.T) {
+	dbtest.DBs(t, "alma8").
+		SelectOnly("rhel:8/cve-2019-13636", "almalinux8/alsa-2020:1852").
```
wagoodman, line 140: what can we do in the future for us to know which qualities really matter in these files and assert they are still there over time

wagoodman (approved):

PR #3413, opened 2026-05-02
version: accept uppercase V prefix the same as lowercase v

### LINKED ISSUE #3037: Treat uppercase V prefixes the same as lowercase v prefixes in version comparison

ChrisJr404:
Closes #3037.

Both `pseudoSemverPattern` (`^v?…`) and `stripLeadingV` only treated a lowercase `v` prefix as significant, so a Bosch industrial-firmware version like `V1500-SP2` (NEXO-OS) never matched a CPE without the prefix, silently dropping vulnerability hits like BOSCH-SA-711465.

@kzantow and @wagoodman both confirmed the bug and pointed at the exact spots to fix:
> the semver check is only looking for lowercase v within the [fuzzy version constraint](https://github.com/anchore/grype/blob/ee4941541a39a35c2bddf533927341b135456b65/grype/version/fuzzy_constraint.go#L16) , this could probably be relaxed to look for both upper and lower
> when we fallthrough to comparing non-semver cases, we should also allow for uppercase Vs as well

### Changes
- `grype/version/fuzzy_constraint.go`:
  - `pseudoSemverPattern` regex: `^v?…` → `^[vV]?…`
  - `stripLeadingV`: trim either `v` or `V`
- `grype/version/fuzzy_constraint_test.go`: extend `TestPseudoSemverPattern` with explicit `v1.2.3` and `V1.2.3` cases; add `TestFuzzyVersionComparison` cases for `V1500-SP2` vs `1500-SP2` (equal), `V1500-SP2` vs `v1500-SP2` (equal), and `V1500-SP1` vs `V1500-SP2` (`-1`).

### Verification
```
$ go test -count=1 ./grype/version/...
ok  	github.com/anchore/grype/grype/version	0.041s
```

### DIFF
--- a/grype/version/fuzzy_constraint.go
+++ b/grype/version/fuzzy_constraint.go
@@ -12,8 +12,8 @@ import (
 
 // derived from https://semver.org/, but additionally matches:
 // - partial versions (e.g. "2.0")
-// - optional prefix "v" (e.g. "v1.0.0")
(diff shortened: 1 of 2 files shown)


kzantow:

PR #3414, opened 2026-05-02
feat(matcher/java): add per-request timeout for Maven SHA1 lookups (#1624)

### LINKED ISSUE #1624: Allow configurting timeout for external-sources

ChrisJr404:
## Description
Closes #1624.

Adds a new `external-sources.maven.timeout` config field that is wired through to the HTTP client used by the Java matcher's Maven SHA1 upstream search.

```yaml
external-sources:
  enable: true
  maven:
    search-upstream: true
    timeout: 5s   # new — per-request limit for the SHA1 lookup
```

The default is `10s`, and a value of `0` disables the limit (preserving the previous behavior of `http.DefaultClient` with no timeout).

## Why this fixes the reported pain
The issue describes the scenario where Maven Central (or any configured `base-url`) is unresponsive: grype falls back to a SHA1 lookup, the request hangs indefinitely, and the user sees a stuck scan with no console signal until they enable `--debug`. With this change, a per-request timeout fires, the lookup returns an error, and the scan continues with the remaining matchers.

## Per-request vs cumulative time
@tgerla suggested in the issue that "total time would be a good starting point" with cascading global + per-source overrides. This PR ships the smallest correct unit first — a per-request timeout — because a single hung request was the actual failure mode in the report. A cumulative time budget can build on top of this once a second external source lands; right now there is only one (Maven), so a global `abort-after` would be functionally identical to the per-source one.

If the project would prefer a different shape (cumulative-only, or `abort-after` instead of `timeout`), happy to revise — small enough patch.

## Implementation notes
- New `Timeout time.Duration` field on the `maven` struct in `cmd/grype/cli/options/datasources.go`, default `10s`, mapped through `ToJavaMatcherConfig()` as `MavenTimeout`.
- New `MavenTimeout time.Duration` field on `java.ExternalSearchConfig`.
- `NewJavaMatcher` constructs a dedicated `*http.Client` with the configured timeout when it is non-zero, so the timeout does not leak onto unrelated callers of `http.DefaultClient`. When the timeout is zero, the existing `http.DefaultClient` is used unchanged.

## Tests
Two new sub-tests in `TestNewJavaMatcherTimeout`:
- `timeout aborts hanging upstream` — boots an `httptest.Server` whose handler blocks on a never-closed channel, configures a `MavenTimeout` of 100ms, and asserts that `GetMavenPackageBySha` returns an error within ~2s.
- `zero timeout disables the per-request limit` — asserts the underlying `http.Client.Timeout` is `0` when no `MavenTimeout` is configured (so default behavior is preserved).

## Test plan
- [x] `go build ./...` clean
- [x] `go test ./grype/matcher/java/ -run TestNewJavaMatcherTimeout` passes
- [x] `go test ./grype/matcher/java/ ./cmd/grype/cli/options/` — no regressions
- [x] `grype --help` and YAML config still load with the new field set, default value populated

### DIFF
--- a/cmd/grype/cli/options/datasources.go
+++ b/cmd/grype/cli/options/datasources.go
@@ -24,6 +24,7 @@ type maven struct {
 	SearchUpstreamBySha1 bool          `yaml:"search-upstream" json:"searchUpstreamBySha1" mapstructure:"search-maven-upstream"`
 	BaseURL              string        `yaml:"base-url" json:"baseUrl" mapstructure:"base-url"`
 	RateLimit            time.Duration `yaml:"rate-limit" json:"rateLimit" mapstructure:"rate-limit"`
+	Timeout              time.Duration `yaml:"timeout" json:"timeout" mapstructure:"timeout"`
 }
 
 func defaultExternalSources() externalSources {
@@ -32,6 +33,7 @@ func defaultExternalSources() externalSources {
 			SearchUpstreamBySha1: true,
 			BaseURL:              defaultMavenBaseURL,
 			RateLimit:            300 * time.Millisecond,
+			Timeout:              10 * time.Second,
 		},
 	}
 }
@@ -46,6 +48,7 @@ func (cfg externalSources) ToJavaMatcherConfig() java.ExternalSearchConfig {
 		SearchMavenUpstream: smu,
 		MavenBaseURL:        cfg.Maven.BaseURL,
 		MavenRateLimit:      cfg.Maven.RateLimit,
+		MavenTimeout:        cfg.Maven.Timeout,
 	}
 }
 
@@ -53,4 +56,5 @@ func (cfg *externalSources) DescribeFields(descriptions clio.FieldDescriptionSet
 	descriptions.Add(&cfg.Enable, `enable Grype searching network source for additional information`)
 	descriptions.Add(&cfg.Maven.SearchUpstreamBySha1, `search for Maven artifacts by SHA1`)
 	descriptions.Add(&cfg.Maven.BaseURL, `base URL of the Maven repository to search`)
+	descriptions.Add(&cfg.Maven.Timeout, `per-request timeout for Maven SHA1 lookups; 0 disables the timeout`)
 }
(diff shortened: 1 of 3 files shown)


noqt (changes requested):
I’m reviewing this for NOQT. The timeout direction makes sense, but the default path changes more than the timeout.

`defaultExternalSources()` sets `MavenTimeout` to 10 seconds, so `NewJavaMatcher` now replaces `http.DefaultClient` with `&http.Client{Timeout: ...}` on every normal run. That drops any `Transport`, `CheckRedirect`, or `Jar` already configured on the default client. Existing embedders that rely on a custom transport for authentication, proxying, or test hooks will silently bypass it when this lands.

Please shallow-copy the existing client and change only its timeout, for example `clientCopy := *http.DefaultClient; clientCopy.Timeout = cfg.MavenTimeout`, then pass `&clientCopy`. A regression test with a custom `RoundTripper` should show it is still called when the timeout is enabled.

I checked exact head `857fcd8958fdc318b498158d3c740584c448c89d`; it merges cleanly with current main. Only DCO is currently available as a hosted check, so I’m not claiming a hosted test pass.

ChrisJr404:
Good point, thanks. Switched to a shallow copy of `http.DefaultClient` and set `Timeout` on the copy, so any Transport, CheckRedirect, or Jar configured on the default client is preserved instead of being dropped. Only the copy carries the per-request timeout, so unrelated callers are unaffected.

PR #3421, opened 2026-05-05
chore: exercise SLES NAKs via dbtest style tests in RPM matcher

### LINKED ISSUE #2061: False Positive: GHSA-jfmj-5v4g-7637 (CVE-2024-5569) python3-zipp due to Syft noise and mismatch of package name
**What happened**:

Scan on image that has  python3-zipp-0.6.0-150100.3.5.1.noarch installed.
It generates vulnerabilities:

NAME                INSTALLED              FIXED-IN                 TYPE       VULNERABILITY              SEVERITY
zipp                    0.6.0                       3.19.1                       python     GHSA-jfmj-5v4g-7637  Medium

There are 2 possibilities of error detection: 

1)  Issue from Syfy noise
The package install in the container is:  python3-zipp-0.6.0-150100.3.5.1.noarch
When run Syft,  we can see there are 2 outputs related to *zipp,  e.g.

python3-zipp                    0.6.0-150100.3.5.1               rpm
zipp                                   0.6.0                                     python

The zipp 0.6.0 output by Syft may contribute to trigger vulnerability detection.
But it is not what is installed in the container.

2) Mismatch package  

According SUSE Advisory: 
https://www.suse.com/security/cve/CVE-2024-5569.html

This package python311-zipp has CVE-2024-5569 fixed from version 3.15.0-150400.10.10.1

And python3-zipp has no impact on the CVE-2024-5569.

They are 2 different packages:        python311  3.xx.yy-nnn       ( this package in not used in the container)
                                                        python3-zip   0.6.0-xx.yy.zz-nnn   (This is the one install in the container)

Probably Grype see version 0.6.0 is lower than 3.xx.yy, then issue vulnerability detection.

You can reproduce step by step,  with below instruction.

1) Create the Dockerfile with this content:

FROM registry.suse.com/suse/sle15:15.5
RUN zypper in -y --no-recommends  python3-zipp=0.6.0-150100.3.5.1
ENTRYPOINT [""]
CMD ["bash"]

2) Build an image from Dockerfile

$ docker build -t "suse15.5_python3-zipp:v1" .

3) Verify package in the container

$ docker run -it suse15.5_python3-zipp:v1 bash
#  rpm -qa | grep zipp
   python3-zipp-0.6.0-150100.3.5.1.noarch

4) Run Syft
$  syft suse15.5_python3-zipp:v1 | grep zipp

python3-zipp                    0.6.0-150100.3.5.1                rpm
zipp                                   0.6.0                                     python

5) Test with Grype

$ grype --distro sles:15.5 suse15.5_python3-zipp:v1 | grep zipp

zipp                0.6.0                  3.19.1                   python     GHSA-jfmj-5v4g-7637  Medium

**Environment**:

$ grype --version
grype 0.79.4

In container image eco-system:
-----------------------------
bash-4.4$ cat /etc/*release*
NAME="SLES"
VERSION="15-SP5"
VERSION_ID="15.5"
PRETTY_NAME="SUSE Linux Enterprise Server 15 SP5"
ID="sles"
ID_LIKE="suse"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:suse:sles:15:sp5"
DOCUMENTATION_URL="https://documentation.suse.com/"

willmurphyscode:
~~Normal querying falls back to broader and broader distro searches, e.g. major and minor match, matcher matches and minor is blank, then major matches and minor does not. So searching for "rhel 9.3" would match "rhel 9.3", then if that's empty, "rhel 9" and if that's empty "rhel 9.2". This makes sense, because if the last time a CVE was mentioned for RHEL 9 it was a disclosure on RHEL 9.2, it probably still applies on RHEL 9.3. The same is not necessarily true of NAKs and fixes.~~

~~In order to be more conservative with the UnaffectedPackageHandle records from distro NAKs, searches for unaffected packages should be strict by default (they should not fall back to mismatched major/minor distro version rows). In other words, when no records match a distro, by default expand the search for _disclosures_ that may match the distro, but do not expand the search for _unaffected records_ that match may match the distro.~~

The reason it looked like a query fallback change was needed was that sparse test data, e.g. a test with a SLES 15.7 package and only SLES 15.6 NAKs, triggered a fallback path that's not expected in production code.

Addressing differences between SLES and RHEL on RPM matching in particular is out of scope of this PR, and requires some caution. This PR now just adds a bunch of test cases to confirm how SLES matching behaves today (and after https://github.com/anchore/vunnel/pull/1153 is released).


wagoodman (dismissed):

grype/db/v6/operating_system_store.go
```diff
@@ -42,6 +42,16 @@ type OSSpecifier struct {
...
+
+	// DisableCrossMinorFallback blocks the loose "any-minor-with-this-major"
+	// fallback while still allowing the major-only-empty-minor fallback. Used
+	// for unaffected/NAK lookups: a major-only record (e.g. rhel:8) is still
+	// applied to a minor-versioned scan (e.g. rhel 8.4) because the publisher
+	// intentionally chose not to be more specific, but a sibling-minor record
+	// (e.g. sles:15.6) is NOT applied to a different minor (sles:15.7) because
+	// the publisher said something specific about SP6, not SP7. See
+	// applyUnaffectedOSStrictness in search_query.go.
+	DisableCrossMinorFallback bool
```
kzantow: I'm having a little bit of a hard time understanding the need for another flag here: if I search for a MAJOR and MINOR version, it should only match that exactly or records _without_ a minor version, right? When would there ever be different behavior? If a matcher wants something different, it should include more criteria.

willmurphyscode: So we want an unaffected package for RHEL 9 to match a RHEL 9.4 package.

But we want an unaffected package for SLES 15.6 _not_ to match a SLES 15.7 package.

kzantow (approved):

grype/db/v6/operating_system_store.go
willmurphyscode: This change was reverted. See updated description for details.

PR #3424, opened 2026-05-06
fix(matcher/golang): surface advisories pinned at import-path granularity (#3398)

### LINKED ISSUE #3398: Go matcher silently drops advisories pinned at import-path granularity
**What happened**:

Go advisories filed against an import path *inside* a larger module are silently missed unless the GHSA is amended to also list the containing module. Go binaries only carry module-granularity build info (via `debug/buildinfo`), so syft cannot emit the import-path entries the advisory expects. The Go matcher does case-insensitive exact equality on package name with no module↔import-path reconciliation, so the match never lands.

Recent in-the-wild instance: GHSA-mh2q-q3fh-2475 / CVE-2026-29181, originally pinned at `go.opentelemetry.io/otel/baggage` and `go.opentelemetry.io/otel/propagation`. Both are packages inside the `go.opentelemetry.io/otel` module — same `go.mod`, shared version stream, never separately distributed. Grype reported nothing for binaries linking `go.opentelemetry.io/otel v1.40.0`. The sibling advisory GHSA-hfvc-g4fc-pqhx (`go.opentelemetry.io/otel/sdk`, an independent module) was detected, purely because upstream chose to split that one into its own `go.mod`.

The CVE-2026-29181 case has since been resolved by amending the GHSA to also list `go.opentelemetry.io/otel` as affected: github/advisory-database#7490. A similar precedent exists for `gofiber`: GHSA-98j2-3j3p-fw2v, where the parent module was added alongside the import path. The pattern works, but it is a per-advisory manual workaround for what is structurally a matcher gap. Every future Go advisory pinned at import-path granularity will require the same amendment dance before grype can see it.

Relevant code:

- `grype/matcher/golang/matcher.go:63` — `p.Name` passed straight to the generic pipeline.
- `grype/db/v6/name/resolver.go:13-22` — Python and Java have name resolvers; Go falls through to `nil`.
- `grype/search/package_name.go:22` — `strings.EqualFold(vuln.PackageName, v.PackageName)`.
- `grype/db/v6/package_store.go:305` — `WHERE packages.name = ? collate nocase`.

The Go matcher should treat a cataloged module name as also matching advisories whose affected name is an import path under that module, so that import-path-granularity advisories land without depending on the upstream advisory listing every containing module explicitly. Trade-off acknowledged: this over-reports for consumers that don't import the vulnerable subpackages — the correct direction for a vulnerability scanner, with reachability triage delegated to govulncheck or the user.

The original CVE-2026-29181 reproducer no longer demonstrates the gap, since github/advisory-database#7490 amended the advisory in-place. Pre-amendment behavior was:

```
$ grype 'pkg:golang/go.opentelemetry.io/otel@v1.40.0'
No vulnerabilities found
```

OSV at the time confirmed the data was reachable, just not at the module name grype was querying:

```
$ curl -s -X POST https://api.osv.dev/v1/query -H 'Content-Type: application/json' \
    -d '{"package":{"name":"go.opentelemetry.io/otel/baggage","ecosystem":"Go"},"version":"1.40.0"}' \
  | jq -r '.vulns[].id'
GHSA-mh2q-q3fh-2475

$ curl -s -X POST https://api.osv.dev/v1/query -H 'Content-Type: application/json' \
    -d '{"package":{"name":"go.opentelemetry.io/otel","ecosystem":"Go"},"version":"1.40.0"}'
{}
```

The same pattern will resurface any time a Go advisory is filed at import-path granularity without listing the containing module. Filing this issue ahead of the next instance so there is a single place to attribute it.

**Environment**:
- Output of `grype version`: 
```
Version:             0.110.0
BuildDate:           2026-03-19T17:37:05Z
GitCommit:           Homebrew
```

- OS (e.g: `cat /etc/os-release` or similar): Mac OS

ChrisJr404:
Closes #3398.

## What

The Go matcher silently drops advisories pinned at an import path inside a larger module. Go binaries only retain module-granularity build info via \`debug/buildinfo\`, so syft emits the module (\`golang.org/x/crypto\`) but the advisory's package name (\`golang.org/x/crypto/ssh\`) is what GitHub Advisories / OSV record. The matcher's case-insensitive exact-name lookup never lands.

The issue calls out GHSA-mh2q-q3fh-2475 / CVE-2026-29181 as a recent example: pinned at \`go.opentelemetry.io/otel/baggage\` and \`/propagation\`, missed by grype against binaries linking \`go.opentelemetry.io/otel\` v1.40.0 until the advisory was manually amended to also list the parent module. Every future advisory filed at import-path granularity hits the same wall.

## Fix

The Go matcher now supplements the existing exact-name lookup with a prefix-search keyed on the SBOM package name plus a \`/\` segment break. So when the SBOM has \`golang.org/x/crypto\`, advisories pinned at \`golang.org/x/crypto/ssh\` (or any deeper sub-path) surface alongside ones pinned at the module name itself.

Path-segment safety: the prefix is \`p.Name + "/"\`, not just \`p.Name\`. \`golang.org/x/crypto\` matches \`golang.org/x/crypto/ssh\` but does NOT match \`golang.org/x/cryptographer\`. The \`/\` is enforced in two places: the SQL \`LIKE\` pattern, and the criterion's \`MatchesVulnerability\` (defense-in-depth). LIKE wildcards in the prefix are neutralized via SQLite \`ESCAPE\` clause.

The prefix path runs only when:
- the package type is \`syftPkg.GoModulePkg\`, and
- the name contains at least one \`/\`.

Both conditions together prevent the synthetic \`stdlib\` entry or other single-segment names from fanning out across the entire Go advisory corpus.

The trade-off (over-reporting when a binary doesn't import a vulnerable sub-package of a module it depends on) is called out in the issue and is the desired direction for a vulnerability scanner; downstream reachability triage belongs to tools like govulncheck.

## Implementation notes

- New \`search.ByPackageNamePrefix\` criterion. \`MatchesVulnerability\` enforces the \`prefix + "/"\` boundary case-insensitively.
- \`PackageSpecifier\` gains a \`NamePrefix\` field; the v6 store translates it into \`packages.name LIKE ? ESCAPE '\\\\' collate nocase\`. The escape clause + replacer handle \`%\`, \`_\`, and \`\\\\\` in user-supplied prefixes safely.
- Search query builder dispatches on the new criterion type.
- A shared internal helper (\`matchPackageByEcosystem\`) factors out the disclosure / unaffected / version / qualified-package pipeline so the prefix path reuses it unchanged. Prefix-matched advisories participate in version filtering and ignore-rule construction the same way exact-matched ones do — including unaffected naks.
- CPEs are not re-queried under the prefix path; the exact-name path above already covered them for \`p.Name\`.

## Test plan

- [x] \`Test_ByPackageNamePrefix\` covers the criterion: matches under boundary, deep sub-paths, case-insensitive, sibling-substring-without-segment-break does not match, exact name does not match, empty prefix returns false.
- [x] \`TestMatcher_ImportPathGranularityAdvisories\` exercises the matcher end-to-end via the mock vulnerability provider:
  - \`golang.org/x/crypto\` SBOM surfaces both an exact \`golang.org/x/crypto\` advisory AND an import-path \`golang.org/x/crypto/ssh\` advisory, while a sibling \`golang.org/x/cryptographer\` advisory does NOT match (boundary trap).
  - \`go.opentelemetry.io/otel\` SBOM surfaces sibling import-path advisories under it (the original issue's example).
  - SBOM directly matching an import-path advisory still hits via the exact path (no regression).
  - Single-segment package name (\`stdlib\`) does NOT fan out to the corpus.
  - Non-Go package types are unaffected.
- [x] \`TestNewSearchCriteria\` gains a case for the new prefix criterion / specifier wiring.
- [x] Existing \`grype/matcher/golang\` and \`grype/matcher/internal\` tests continue to pass.
- [x] \`go build ./...\`, \`go vet ./grype/...\` (no new warnings), \`task lint\` (\`golangci-lint\` clean), \`task format\` (gofmt + gosimports + go mod tidy clean).
- [x] Full \`go test ./grype/matcher/... ./grype/search/... ./grype/db/v6/...\` suite green.

### DIFF
--- a/grype/db/v6/package_store.go
+++ b/grype/db/v6/package_store.go
@@ -41,9 +41,14 @@ type GetPackageOptions struct {
 type PackageSpecifiers []*PackageSpecifier
 
 type PackageSpecifier struct {
-	Name      string
-	Ecosystem string
-	CPE       *cpe.Attributes
+	Name string
+	// NamePrefix, when set, restricts results to packages whose name begins with the prefix
+	// followed by a "/" path-segment boundary (case-insensitive). It is mutually exclusive
+	// with Name. This is used to surface advisories pinned at sub-paths of a module name
+	// (e.g. Go import-path granularity advisories under a containing module).
+	NamePrefix string
+	Ecosystem  string
+	CPE        *cpe.Attributes
 }
 
 func (p *PackageSpecifier) String() string {
@@ -54,6 +59,10 @@ func (p *PackageSpecifier) String() string {
 	var args []string
 	if p.Name != "" {
 		args = append(args, fmt.Sprintf("name=%s", p.Name))
+	}
+
+	if p.NamePrefix != "" {
+		args = append(args, fmt.Sprintf("name-prefix=%s", p.NamePrefix))
 	}
 
 	if p.Ecosystem != "" {
@@ -303,6 +312,13 @@ func (s *packageStore) handlePackage(query *gorm.DB, p *PackageSpecifier, allowB
 
 	if p.Name != "" {
 		query = query.Where("packages.name = ? collate nocase", p.Name)
+	} else if p.NamePrefix != "" {
+		// match any package whose name begins with NamePrefix followed by a "/" path-segment
+		// boundary, e.g. NamePrefix "golang.org/x/crypto" -> "golang.org/x/crypto/ssh" but not
+		// "golang.org/x/cryptographer". The "/" requirement is enforced both here (so the SQL
+		// returns no false positives) and again at the criteria level for safety. The ESCAPE
+		// clause neutralizes any wildcard metacharacters that may appear in the prefix.
+		query = query.Where(`packages.name LIKE ? ESCAPE '\' collate nocase`, escapeLikePattern(p.NamePrefix)+"/%")
 	}
 	if p.Ecosystem != "" {
 		query = query.Where("packages.ecosystem = ? collate nocase", p.Ecosystem)
@@ -315,6 +331,14 @@ func (s *packageStore) handlePackage(query *gorm.DB, p *PackageSpecifier, allowB
 	}
 
 	return query
+}
+
+// escapeLikePattern escapes SQLite LIKE wildcard metacharacters ("%", "_") and the escape
+// character itself ("\") so a user-supplied substring can be safely embedded into a LIKE
+// pattern. Use together with `LIKE ? ESCAPE '\'` in the query.
+func escapeLikePattern(s string) string {
+	r := strings.NewReplacer(`\`, `\\`, `%`, `\%`, `_`, `\_`)
+	return r.Replace(s)
 }
 
 func (s *packageStore) handleVulnerabilityOptions(query *gorm.DB, configs []VulnerabilitySpecifier, tableName string) (*gorm.DB, error) {
--- a/grype/db/v6/search_query.go
+++ b/grype/db/v6/search_query.go
@@ -55,6 +55,9 @@ func (b *searchQueryBuilder) ApplyCriteria(criteriaSet []vulnerability.Criteria)
 		case *search.PackageNameCriteria:
 			b.handlePackageName(c)
 			applied = true
+		case *search.PackageNamePrefixCriteria:
+			b.handlePackageNamePrefix(c)
+			applied = true
 		case *search.UnaffectedCriteria:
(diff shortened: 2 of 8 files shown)


kzantow (commented):
The example in the issue indicates this was a data problem that has now been fixed. Are there other examples that show we should actually make a change here?

PR #3428, opened 2026-05-07
feat: warn when picking up .grype.yaml from CWD

### LINKED ISSUE #3427: Grype reads .grype.yaml from the CWD
Grype will read a `.grype.yaml` in the CWD when it is run. There is nothing exactly wrong with this, but I think it might be unexpected for users of Grype.

This was reported to us as a potential vulnerability, but given you need a certain level of access to create this file, this does not cross a trust boundary and is not a vulnerability.

```
➜  grype alpine:latest
 ✔ Parsed image                                        sha256:3cb067eab609612d81b4d82ff8ad71d73482bb3059a87b642d7e14f0ed659cde 
 ✔ Cataloged contents                                         4d889c14e7d5a73929ab00be2ef8ff22437e7cbc545931e52554a7b00e123d8b 
   ├── ✔ Packages                        [16 packages]  
   ├── ✔ Executables                     [17 executables]  
   ├── ✔ File digests                    [79 files]  
   └── ✔ File metadata                   [79 locations]  
 ✔ Scanned for vulnerabilities     [3 vulnerability matches]  
   ├── by severity: 0 critical, 0 high, 3 medium, 0 low, 0 negligible
   └── by status:   0 fixed, 3 not-fixed, 0 ignored 
NAME           INSTALLED   TYPE  VULNERABILITY   SEVERITY  EPSS           RISK   
busybox        1.37.0-r30  apk   CVE-2025-60876  Medium    < 0.1% (15th)  < 0.1  
busybox-binsh  1.37.0-r30  apk   CVE-2025-60876  Medium    < 0.1% (15th)  < 0.1  
ssl_client     1.37.0-r30  apk   CVE-2025-60876  Medium    < 0.1% (15th)  < 0.1

➜  cat > .grype.yaml
ignore:
  - package:
      name: ".*"

➜  grype alpine:latest
 ✔ Parsed image                                        sha256:3cb067eab609612d81b4d82ff8ad71d73482bb3059a87b642d7e14f0ed659cde 
 ✔ Cataloged contents                                         4d889c14e7d5a73929ab00be2ef8ff22437e7cbc545931e52554a7b00e123d8b 
   ├── ✔ Packages                        [16 packages]  
   ├── ✔ File metadata                   [79 locations]  
   ├── ✔ Executables                     [17 executables]  
   └── ✔ File digests                    [79 files]  
   ├── by severity: 0 critical, 0 high, 3 medium, 0 low, 0 negligible
   └── by status:   0 fixed, 3 not-fixed, 3 ignored 
No vulnerabilities found

```

There are two instances where Grype is run against `alpine:latest`. In the second run there are no vulnerabilities found because the CWD has a `.grype.yaml` written to it.

ChrisJr404:
Closes #3427.

Grype's default fangs config search walks the current working directory before `~/.grype.yaml` and the XDG locations, so a hidden `.grype.yaml` (or `.grype/config.yaml`) sitting in CWD is silently applied. @joshbressers flagged this on the issue: "I'm not entirely comfortable having Grype read a hidden file in the CWD without any sort of warning." This change adds the warning.

## What this does

On startup, if grype is about to pick up a config file from CWD via the implicit finder, it logs:

```
[0000]  WARN using configuration file from current directory: .grype.yaml
```

The warning suppresses itself when the file came from anywhere explicit:

- `--config` / `-c <path>` (and the `--config=path`, `-c=path`, `-cpath` short forms)
- `GRYPE_CONFIG` environment variable
- `~/.grype.yaml` or any XDG config location

It also stays quiet when there's no CWD config to warn about.

## Where it lives

A small `cmd/grype/cli/internal/configwarn` package owns the detection so the logic is unit-testable without the whole config loader stack. The detector uses the same extension list viper supports (`yaml`, `yml`, `json`, `toml`, `properties`, etc.) and checks both finder layouts that fangs uses by default: `./.grype.<ext>` and `./.grype/config.<ext>`.

The hook fires from the existing `WithInitializers` block in `cmd/grype/cli/cli.go`, right after the logger is hoisted into place, so the warning goes through the regular logging pipeline (respects `--quiet`, log file routing, redaction, etc.).

## Manual verification

Built locally and verified the four cases:

```
$ cd /tmp/has-cwd-cfg && grype purl
[0000]  WARN using configuration file from current directory: .grype.yaml
...

$ cd /tmp/has-cwd-cfg && grype -c .grype.yaml purl
[0000] (no warn)
...

$ cd /tmp/has-cwd-cfg && GRYPE_CONFIG=.grype.yaml grype purl
[0000] (no warn)
...

$ cd /tmp/no-cwd-cfg && grype purl
[0000] (no warn)
...
```

Subdir form (`.grype/config.yaml`) also fires correctly.

## Tests

- `cmd/grype/cli/internal/configwarn/configwarn_test.go` covers: no file present, each supported extension, subdir layout, all `--config` / `-c` / `--config=` / `-c=` / `-cpath` forms, `GRYPE_CONFIG` env, and a trailing-`-c`-with-no-value edge case (treated as not explicit, so the warning still fires).
- `go test ./...` passes for the whole tree minus pre-existing docker-daemon-required integration tests in this environment (`test/integration`, `test/cli`, `grype/pkg` image fixtures), which fail with "could not build docker image" unrelated to this change.
- `go vet ./cmd/...` is clean. The pre-existing vet warnings on `grype/db/v5/differ`, `grype/db/v6/installation`, and `grype/version` are present on `main` already.

## Notes

- I picked WARN over INFO because the default log level is warn, which matches the spirit of the issue (the user has to actually see it for it to help).
- Wording: "using configuration file from current directory: <path>" is short and tells the user both that something happened and where to look.


cmd/grype/cli/internal/configwarn/configwarn.go
```diff
@@ -0,0 +1,85 @@
...
+//
+// It returns "" (no warning) when:
+//   - the user passed an explicit -c / --config / --config=... flag
+//   - GRYPE_CONFIG (or the equivalent <appName>_CONFIG) is set
+//   - no recognised config file exists in the current working directory
+//
+// It looks for both finder layouts that fangs uses by default:
+//   - ./.<appName>.<ext>
+//   - ./.<appName>/config.<ext>
+func Detect(appName string, args []string, env func(string) string) string {
```
willmurphyscode: This just checks whether a file exists, not whether content is loaded from it. Do we want to log.Warn based on file names, or based on unexpectedly loading configs?

cmd/grype/cli/internal/configwarn/configwarn.go
```diff
@@ -0,0 +1,85 @@
...
+import (
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+// supportedExts mirrors viper.SupportedExts. We hard-code the list rather
+// than depend on viper here so this package can be used without pulling the
+// configuration loader stack into tests.
+var supportedExts = []string{
```
willmurphyscode: There are formats in here grype doesn't currently support (`.ini` for one), and we don't want to have to keep this in sync with the formats we support.

cmd/grype/cli/internal/configwarn/configwarn.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	return ""
+}
+
+// explicitConfigFlag returns true if a -c / --config flag appears in args
+// (in any of: "-c file", "--config file", "-c=file", "--config=file", or
+// the bundled short form "-cfoo"). We intentionally accept false positives
+// here over false negatives: if anything that looks like a config flag is
+// present, suppress the warning.
+func explicitConfigFlag(args []string) bool {
```
willmurphyscode: This duplicates flag parsing elsewhere. If we added `--configuration` as a valid option in Grype (or Fangs/Clio) we would have a bug here.

cmd/grype/cli/internal/configwarn/configwarn.go
```diff
@@ -0,0 +1,85 @@
...
+		switch {
+		case a == "-c", a == "--config":
+			if i+1 < len(args) {
+				return true
+			}
+		case strings.HasPrefix(a, "--config="):
+			return true
+		case strings.HasPrefix(a, "-c=") && len(a) > 3:
+			return true
+		case strings.HasPrefix(a, "-c") && len(a) > 2 && !strings.HasPrefix(a, "--"):
```
willmurphyscode: What is this case for? when does `a` have the prefix `-c` and the prefix `--`? I feel like I'm missing something here.

willmurphyscode (commented):
I am not sure this is the right approach. Does it make sense instead to hook into actual config parsing in Fangs or Clio and update the behavior there to return a warning if config values are changed?

That would have a couple of advantages:
1. It would address the issue in other tools, not just grype
2. If we changed how config file parsing works (e.g. add or remove support for toml, support `--configuration-file` as an alias for `--config`, etc.) we wouldn't need to coordinate across repositories.

kzantow:
fangs already includes the list of config files being loaded:
```
% touch .syft.yaml ~/.syft.yaml && go run ./cmd/syft alpine:latest -vv
[0000]  INFO syft version: [not provided]
[0000] DEBUG config:
...
  config: .syft.yaml,/Users/kzantow/.syft.yaml
```
... so a tool can detect if a config file in the current directory was used fairly easily. [Syft has an explicit spot for this config to land](https://github.com/anchore/syft/blob/main/cmd/syft/internal/options/config.go#L7) that probably needs to get added to grype and then read. I'm pretty sure it can be a `[]string` but isn't in syft for historical reasons and instead concats the files together.

This PR should be adjusted to use this ☝️

ChrisJr404:
Makes sense, reworking it to use the loaded config list instead of re-deriving discovery here. Added the same config field syft has in options/config.go so fangs populates the paths, then checking in PostLoad whether any of them resolved out of the current directory. Dropping the configwarn package and the -c parsing since fangs already knows what it loaded.

PR #3429, opened 2026-05-07
Handle uppercase V prefix in semantic version comparison

### LINKED ISSUE #3037: Treat uppercase V prefixes the same as lowercase v prefixes in version comparison

tjhub1983:
## Summary

Treat both `v1.5.0` and `V1.5.0` equivalently for version matching.

## Fixes #3037

SBOM files with uppercase V prefix (e.g., `V1500-SP2`) were not being matched against CVE vulnerabilities because the version comparison logic only recognized lowercase `v` prefix.

## Changes

- Update regex from `^v\d+` to `^[vV]\d+` in `semantic_version.go`
- Use direct slice `[1:]` instead of `strings.TrimPrefix` for consistency

## Testing

Added test cases:
- `uppercase V prefix comparison`: `1.5.0` vs `V1.5.0`
- `uppercase V prefix on this version`: `V1.5.0` vs `1.5.0`

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/grype/version/semantic_version.go
+++ b/grype/version/semantic_version.go
@@ -19,7 +19,7 @@ type semanticVersion struct {
 	obj *hashiVer.Version
 }
 
-var versionStartsWithV = regexp.MustCompile(`^v\d+`)
+var versionStartsWithV = regexp.MustCompile(`^[vV]\d+`)
 
 func newSemanticVersion(raw string, strict bool) (semanticVersion, error) {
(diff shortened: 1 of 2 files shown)


kzantow:

PR #3433, opened 2026-05-08
fix(cli): fix autocompletion for zsh

### LINKED ISSUE #2933: zsh completion fails
**What happened**:
I tried <tab> for the commands which fails

No failing but instead have completion provided

- `/usr/bin/grype completion zsh > /usr/share/zsh/site-functions/_grype`
- reload zsh `exec zsh`
- `grype <tab>`
- see nothing happens

Output of the debugging log;
```
========= starting completion logic ==========
CURRENT: 2, words[*]: grype 
Truncated words[*]: grype ,
lastParam: , lastChar: 
Adding extra empty parameter
About to call: eval grype __complete  ""
completion output: completion	Generate a shell completion for Grype (listing local docker images)
config	show the grype configuration
db	vulnerability database operations
explain	Ask grype to explain a set of findings
help	Help about any command
version	show version information
completion failed
:1
last line: :1
directive: 1
completions: completion	Generate a shell completion for Grype (listing local docker images)
config	show the grype configuration
db	vulnerability database operations
explain	Ask grype to explain a set of findings
help	Help about any command
version	show version information
completion failed
flagPrefix: 
Completion received error. Ignoring completions.
```

having already completed typing something whole completion works a little bit;
```
========= starting completion logic ==========
CURRENT: 3, words[*]: grype db 
Truncated words[*]: grype db ,
lastParam: , lastChar: 
Adding extra empty parameter
About to call: eval grype __complete db  ""
completion output: check	Check to see if there is a database update available
delete	Delete the vulnerability database
import	Import a vulnerability database or archive from a local file or URL
list	List all DBs available according to the listing URL
providers	List vulnerability providers that are in the database
search	Search the DB for vulnerabilities or affected packages
status	Display database status and metadata
update	Download and install the latest vulnerability database
:4
last line: :4
directive: 4
completions: check	Check to see if there is a database update available
delete	Delete the vulnerability database
import	Import a vulnerability database or archive from a local file or URL
list	List all DBs available according to the listing URL
providers	List vulnerability providers that are in the database
search	Search the DB for vulnerabilities or affected packages
status	Display database status and metadata
update	Download and install the latest vulnerability database
flagPrefix: 
Adding completion: check:Check to see if there is a database update available
Adding completion: delete:Delete the vulnerability database
Adding completion: import:Import a vulnerability database or archive from a local file or URL
Adding completion: list:List all DBs available according to the listing URL
Adding completion: providers:List vulnerability providers that are in the database
Adding completion: search:Search the DB for vulnerabilities or affected packages
Adding completion: status:Display database status and metadata
Adding completion: update:Download and install the latest vulnerability database
Calling _describe
_describe found some completions
```

**Environment**:
- Output of `grype version`: 0.99.1
- OS (e.g: `cat /etc/os-release` or similar): Arch Linux

brandtkeller:
## Summary

`grype <TAB>` in zsh produces no completions on hosts where the Docker daemon isn't reachable from the Docker Go SDK. The root command's ValidArgsFunction (dockerImageValidArgsFunction) returned `cobra.ShellCompDirectiveError` whenever ImageList failed or yielded no matches. Zsh treats that directive as "discard everything," so it threw away not just the (empty) image list but also the subcommand completions Cobra auto-emits, leaving the user with nothing.

Both failure paths now return (nil, cobra.ShellCompDirectiveDefault) instead of an error directive with placeholder strings. Subcommand completion (db, config, explain, …) flows through cleanly, and the shell falls back to filename completion for non-image scan targets like dir:, sbom:, file:. 

The Docker happy path is unchanged: when ImageList returns tags matching the prefix, they're returned with ShellCompDirectiveDefault as before.

Fixes #2933

## Other

I noticed another subtle issue as well. The completer uses `client.NewClientWithOpts(client.FromEnv, ...)`, which only honors DOCKER_HOST and does not read Docker contexts from ~/.docker/config.json. On Colima, Docker Desktop, Rancher Desktop, Podman, and rootless Docker setups, the SDK falls back to unix:///var/run/docker.sock (which doesn't exist) and silently fails to find the daemon. So even with this PR, image suggestions won't appear for those users; they'll just get the subcommand list and shell file completion. That's an improvement over the current "completely silent" state. I'll open a separate issue for context-aware client construction.

### DIFF
--- a/cmd/grype/cli/commands/completion.go
+++ b/cmd/grype/cli/commands/completion.go
@@ -96,16 +96,13 @@ func listLocalDockerImages(prefix string) ([]string, error) {
 }
 
 func dockerImageValidArgsFunction(_ *cobra.Command, _ []string, toComplete string) ([]string, cobra.ShellCompDirective) {
-	// Since we use ValidArgsFunction, Cobra will call this AFTER having parsed all flags and arguments provided
+	// Since we use ValidArgsFunction, Cobra will call this AFTER having parsed all flags and arguments provided.
+	// When the docker daemon is unavailable (or has no images), fall through to the shell's default behavior so
+	// that subcommand completions and filename completion still work — returning ShellCompDirectiveError here
+	// causes shells (notably zsh) to discard all completions, including the auto-generated subcommand list.
 	dockerImageRepoTags, err := listLocalDockerImages(toComplete)
-	if err != nil {
-		// Indicates that an error occurred and completions should be ignored
-		return []string{"completion failed"}, cobra.ShellCompDirectiveError
+	if err != nil || len(dockerImageRepoTags) == 0 {
+		return nil, cobra.ShellCompDirectiveDefault
 	}
-	if len(dockerImageRepoTags) == 0 {
-		return []string{"no docker images found"}, cobra.ShellCompDirectiveError
-	}
-	// ShellCompDirectiveDefault indicates that the shell will perform its default behavior after completions have
-	// been provided (without implying other possible directives)
 	return dockerImageRepoTags, cobra.ShellCompDirectiveDefault
 }


kzantow (approved):
LGTM

PR #3437, opened 2026-05-10
feat: scheme-aware shell completion for scan targets

### LINKED ISSUE #3432: Extend shell autocompletion to all scan target schemes

Grype currently provides dynamic shell completion only for the default (unprefixed) scan target, where dockerImageValidArgsFunction in cmd/grype/cli/commands/completion.go:98-111 enumerates local Docker daemon images. Every other supported scheme either falls back to the shell's filesystem path completion or gets no completion at all. A user typing podman:<TAB> sees nothing despite Podman being a first-class source, and sbom:<TAB> doesn't narrow to plausible SBOM files.

The proposal is to make completion scheme-aware. The ValidArgsFunction would inspect the user's partial input, detect a known scheme prefix, and dispatch to a completer appropriate to that source. The unprefixed default would continue to suggest Docker images as it does today.

**Why is this needed**:

Low criticality but provides a nice user experience for autocompletion.

ChrisJr404:
Closes #3432.

## Problem

The root command's `ValidArgsFunction` (`dockerImageValidArgsFunction` in `cmd/grype/cli/commands/completion.go`) has only ever enumerated local Docker daemon images. Every other documented target scheme (`registry:`, `oci-archive:`, `oci-dir:`, `singularity:`, `dir:`, `file:`, `sbom:`, `purl:`, `cpes:`, `podman:`, `docker-archive:`) silently fell through to the shell, so a user typing `<TAB>` with no input sees nothing useful and a user typing `oci-<TAB>` gets no hint that `oci-archive:` or `oci-dir:` is even a legal prefix.

## Change

`dockerImageValidArgsFunction` is now scheme-aware:

- On an empty input, or a partial input that prefix-matches one or more known schemes, the documented scheme prefixes are returned alongside the docker-image suggestions. Each prefix carries its trailing `:` and the directive includes `ShellCompDirectiveNoSpace` so the user can keep typing the rest of the target (e.g. `dir:./project`) without the shell jumping to a fresh token.
- For `docker:` and `podman:` inputs, the local docker daemon is queried with the scheme stripped, and the scheme is re-prepended onto each tag so the completion is a complete, valid grype argument.
- For every other scheme prefix (file- or directory-backed: `sbom:`, `dir:`, `oci-archive:`, etc.), the function returns `ShellCompDirectiveDefault` so the shell does its own filename completion. That preserves the historical fallback rather than trying to second-guess what file extensions or directory shapes are valid for each scheme.

The scheme list lives in a single `targetSchemePrefixes` slice next to the function, and a coverage test pins it to every prefix documented in the root command's long-help text so the two won't drift.

## Verification

- `go test ./cmd/grype/cli/commands/` passes.
- Manual smoke against `grype __complete ""`:
  - empty input lists all 12 scheme prefixes (and any local docker tags) with `ShellCompDirectiveNoSpace`
  - `oci-` narrows to `oci-archive:` and `oci-dir:`
  - `dir:/tmp/` returns no suggestions and `ShellCompDirectiveDefault`, so the shell does path completion
  - `docker:` with no daemon falls through to the shell's default
- `go vet ./...` is clean.

## Notes for reviewers

I kept the function name `dockerImageValidArgsFunction` to minimize diff; happy to rename it to something like `scanTargetValidArgsFunction` if you'd prefer. I also resisted the urge to add per-scheme file-extension filters (`sbom:` to `.json`, `oci-archive:` to `.tar`) since several of those formats are accepted by extension and bare-name both, and a wrong filter would hurt UX more than it'd help.

### DIFF
--- a/cmd/grype/cli/commands/completion.go
+++ b/cmd/grype/cli/commands/completion.go
@@ -68,6 +68,31 @@ $ grype completion fish > ~/.config/fish/completions/grype.fish
 	}
 }
 
+// targetSchemePrefixes is the set of scheme prefixes documented in the root command's long help text.
+// Each entry includes the trailing ":" so that completion produces a value the user can keep typing
+// after without having to add the separator themselves.
+var targetSchemePrefixes = []string{
+	"docker:",
+	"podman:",
+	"docker-archive:",
+	"oci-archive:",
+	"oci-dir:",
+	"singularity:",
+	"registry:",
+	"dir:",
+	"file:",
+	"sbom:",
+	"purl:",
+	"cpes:",
+}
+
+// imageSchemePrefixes is the subset of scheme prefixes for which we can usefully enumerate local
+// Docker daemon images (the docker SDK speaks to both Docker and Podman daemons via DOCKER_HOST).
+var imageSchemePrefixes = []string{
+	"docker:",
+	"podman:",
+}
+
 func listLocalDockerImages(prefix string) ([]string, error) {
 	var repoTags = make([]string, 0)
 	ctx := context.Background()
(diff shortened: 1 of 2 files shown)


cmd/grype/cli/commands/completion.go
```diff
@@ -68,6 +68,31 @@ $ grype completion fish > ~/.config/fish/completions/grype.fish
 	}
 }
 
+// targetSchemePrefixes is the set of scheme prefixes documented in the root command's long help text.
+// Each entry includes the trailing ":" so that completion produces a value the user can keep typing
+// after without having to add the separator themselves.
+var targetSchemePrefixes = []string{
```
kzantow, line 74: Thanks for this PR. I think we will need to experiment with it to see if it "feels good" for real world use, but I would note that we have a `--from` flag which is also intended to replace the scheme. I'm not sure if autocompleting the scheme is ideal. We would also want to make sure the behavior is the same with `--from` specified.

PR #3441, opened 2026-05-12
OSV strategy pattern

### LINKED ISSUE #3294: Numerous OSV transformer improvements proposed simultaneously
This issue is to help figure out what changes are needed to the [osv transformer](https://github.com/anchore/grype/blob/main/grype/db/v6/build/transformers/osv/transform.go) in Grype. There are a number of proposed changes that are needed to support various data sources. I'm trying to get all of them linked to this issue so that we can figure out how to proceed with getting them in (without contributors rebasing on one another's OSV transformer fixes half a dozen times.)

Here are some PRs that are proposing to change the Grype OSV transformer:

- RootIO: https://github.com/anchore/grype/pull/3137
- Bellsoft / Alpaquita Linux: https://github.com/anchore/grype/pull/3264
- CleanStart: https://github.com/anchore/grype/pull/3281
- Chainguard / Wolfi: https://github.com/anchore/grype/pull/3255

And some Vunnel PRs that will end up wanting to change the OSV transformer:

- openEuler https://github.com/anchore/vunnel/pull/839 (I plan to revise that to use OSV data)
- CRAN (R package) https://github.com/anchore/vunnel/pull/1043 

My goal in this issue is to get all of this work merged or unblocked. If you're the author of one of these PRs, please be patient with me while I figure out a path forward for our suddenly very crowded OSV transformer; I've let a little bit of a traffic jam develop and need to figure it out, and I thought I'd figure it out on this issue so you all aren't guessing why I haven't merged your PR.

willmurphyscode:
Delete the generic OSV transformer in favor of per-provider OSV transformers.

Fixes https://github.com/anchore/grype/issues/3294

## Why?

1. The OSV spec has large and widely used escape hatches (ecosystem specific, database specific)
2. Some OSV feeds publish disclosures and advisories, others only advisories
3. A large number of OSV transformers are about to be added, and this makes it obvious how to add one and greatly reduces the probability that they will break one another

## Key decisions

1. Identify strategy based on ID prefix, e.g. Alma records start with AL(S|B|E)A-, Bitnami records start with BIT-*. You can see this by clicking around https://osv.dev/
2. Log a warning a skip if no transformer matches a record. Because every provider means specific things by their data, it doesn't make sense to add a generic provider. Folks adding a vunnel OSV provider should also add a transformer.
3. Put each transformer in its own file but not its own package.
4. Add a `helpers.go` for parts of the OSV spec that are common.
5. Intentionally drop bitnami CPEs. Previously they were always dropped by an impossible type check, and they have never been used for matching.
6. Includes backfilling bitnami "dbtest" style matchers since they were missing.


grype/db/v6/build/transformers/osv/helpers.go
```diff
@@ -0,0 +1,317 @@
...
+		Version: fixedInVersion,
+		Detail:  detail,
+	}
+}
+
+// normalizeRangeType maps an OSV range type to the format string used by the
+// grype version matcher. The ecosystem-specific overrides (Bitnami → "bitnami",
+// AlmaLinux ECOSYSTEM → "rpm") are kept here as shared logic; new strategies
+// that need similar overrides can extend the switch.
+func normalizeRangeType(t models.RangeType, ecosystem string) string {
```
willmurphyscode: Shouldn't be a helper ; collapse into specific transformers.

spiffcs (approved):

PR #3461, opened 2026-05-25
fix: normalize uppercase V in semantic version comparison

immanuwell:
tiny follow-up to #3089.

fuzzy path was fixed, semantic path still trips on uppercase `V`, so `V1.0.0` still blows up on `main`. a bit sneaky tbh.
this is not a made-up edge case either. #3037 reports a real version string with an uppercase `V` prefix.

repro:
```go
c, _ := version.GetConstraint("= V1.0.0", version.SemanticFormat)
ok, err := c.Satisfied(version.New("1.0.0", version.SemanticFormat))
// before: ok=false, err contains "Malformed version: V1.0.0"

ok, err = version.New("V1.0.0", version.SemanticFormat).Is(version.EQ, version.New("1.0.0", version.SemanticFormat))
// before: ok=false, err contains "Malformed version: V1.0.0"
```

this patch normalizes uppercase `V` in the semantic path too, and adds regression tests.

checked:
- `go test ./grype/version/... -count=1`
- `go test ./grype/search -count=1`
- `go test ./grype/pkg -count=1`

refs #3037


kzantow (approved):
LGTM 👍

Issue #1365, opened 2023-06-28
Add support for using VEX statements to filter/enrich match results


wagoodman:
Given a set of [VEX statements](https://github.com/openvex/spec/blob/main/OPENVEX-SPEC.md#the-vex-statement), which represents status assessments relative to a vulnerability matched with a product, it would be ideal to filter grype results down to useful or novel results (removing results that have `not_affected` status values in a VEX statement). The primary motivator behind this is to attempt to reduce the result size when possible to help the user focus on the results that have a practical impact (and not spend time attempting to remediate non-issues).

One question might be: where should these VEX documents come from? There is a bit of a spectrum here, and I think that motivates a possible implementation path:

Enable grype to be able to...
1. take a single document that may have one or more VEX statements and filter the results. This is a good first step since we could blindly take statements and apply them to all artifacts regardless to the scan target.
2. take multiple VEX documents and filter the results. This has an added challenge of determining which documents apply to the artifact scanned. There are also secondary problems such as if given a directory with a large set of documents, are there cheap and easy ways to filter down to the correct set of files (do we need an index? or to build an index?)?
3. pull down remote VEX documents given an explicit reference (e.g. a git repo URL) and filter the results using only applicable documents. The reference should not be vague, so the added challenge is less about discovery and more about authentication and caching concerns.
4. discover VEX documents for a given input reference (e.g. `alpine:latest`). There are a lot of added challenges here but ultimately this would be the most "magical", requiring the least amount of user input to leverage vex documents. This added automagic-ness should not sacrifice security concerns to achieve this and ideally would require no additional configuration (not a requirement though). There are a lot of directions this could go in, so I'll leave this speculation for later.

The nice thing about this path is that we defer decisions about where these documents come from while working on the logistics of lining up the existing OpenVEX spec.

For `1` something like this could be the input:

```
$ grype myimage:tag --vex ./path/to/vexdoc.json
```

The same input could work for `2` and `3`, where the argument might be a directory and we look for vex documents, or the argument could be a remote resource such as `git@github.com:myorg/myvexrepo.git`... I'm softly suggesting this initially to help set an initial direction, but consider none of this set in stone. 

_Side note: I think we should focus initial conversations and efforts just on `1` for now, but I wanted to at least get a vision going for later._

[...]

Issue #2112, opened 2024-09-13
Matcher dispatch should be based on any package quality


wagoodman:
Today we dispatch to the correct matcher(s) based on the `Type()` returned from any given matcher. This isn't very flexible, since it doesn't allow for multiple types, to be able to introspect package values, or metadata to understand if it could make a match.

We should make a (breaking) change to the `Matcher` interface to allow for more flexibility here.

willmurphyscode:
@wagoodman I think this is a fast follow for schema v6, since schema v6 will add more flexibility to how the database is queried, as well.

Do you agree? Do you want to track it as a part of the v6 milestone or as part grype 1.0?

PR #3464, opened 2026-05-26
feat(vex): synthesize matches from SBOM for affected VEX statements

### LINKED ISSUE #3145: vex document of golang subcomponents failed to be matched
**What happened**:

With a set of exclusions of the golang core modules in a vex generated by govulchecker, grype failed to match the submodules.

```
$ grype -vvv mariadb:lts-ubi9   --vex .vex/gosu-v1.19.0.vex.json --output=table --show-suppressed 

[0003] TRACE searching for vulnerability matches package=pkg:golang/github.com/tianon/gosu@v1.19.0
[0003] TRACE fetched package record distro=none duration=101.42µs pkg=package(name=github.com/tianon/gosu, ecosystem=go-module) records=0 vulns=any
[0003] TRACE attached blob values count=0 duration=164ns
[0003] TRACE attached blob values count=0 duration=38ns
[0003] TRACE fetching all provider records
[0003] TRACE fetched package record distro=none duration=71.165µs pkg=package(name=github.com/tianon/gosu, ecosystem=go-module) records=0 vulns=any
[0003] TRACE attached blob values count=0 duration=66ns
[0003] TRACE attached blob values count=0 duration=30ns
[0003] TRACE fetching all provider records
...
[0003] TRACE searching for vulnerability matches package=pkg:golang/github.com/moby/sys/user@v0.1.0
[0003] TRACE fetched package record distro=none duration=189.789µs pkg=package(name=github.com/moby/sys/user, ecosystem=go-module) records=0 vulns=any
[0003] TRACE attached blob values count=0 duration=97ns
[0003] TRACE attached blob values count=0 duration=46ns
[0003] TRACE fetching all provider records
[0003] TRACE fetched package record distro=none duration=158.295µs pkg=package(name=github.com/moby/sys/user, ecosystem=go-module) records=0 vulns=any
[0003] TRACE attached blob values count=0 duration=86ns
[0003] TRACE attached blob values count=0 duration=39ns
[0003] TRACE fetching all provider records
[0003] TRACE searching for vulnerability matches package=pkg:golang/github.com/tianon/gosu@v1.19.0
[0003] TRACE fetched package record distro=none duration=101.42µs pkg=package(name=github.com/tianon/gosu, ecosystem=go-module) records=0 vulns=any
[0003] TRACE attached blob values count=0 duration=164ns
[0003] TRACE attached blob values count=0 duration=38ns
[0003] TRACE fetching all provider records
[0003] TRACE fetched package record distro=none duration=71.165µs pkg=package(name=github.com/tianon/gosu, ecosystem=go-module) records=0 vulns=any
[0003] TRACE attached blob values count=0 duration=66ns
[0003] TRACE attached blob values count=0 duration=30ns


stdlib            go1.24.6             *1.24.8, 1.25.2   go-module  CVE-2025-61723  High      < 0.1% (7th)   < 0.1  
stdlib            go1.24.6             *1.24.8, 1.25.2   go-module  CVE-2025-61725  High      < 0.1% (6th)   < 0.1  
stdlib            go1.24.6             *1.24.8, 1.25.2   go-module  CVE-2025-58185  Medium    < 0.1% (9th)   < 0.1  
stdlib            go1.24.6             *1.24.8, 1.25.2   go-module  CVE-2025-58186  Medium    < 0.1% (7th)   < 0.1  
stdlib            go1.24.6             *1.24.8, 1.25.2   go-module  CVE-2025-61724  Medium    < 0.1% (6th)   < 0.1  
stdlib            go1.24.6             *1.24.8, 1.25.2   go-module  CVE-2025-47912  Medium    < 0.1% (6th)   < 0.1  
stdlib            go1.24.6             *1.24.9, 1.25.3   go-module  CVE-2025-58187  High      < 0.1% (2nd)   < 0.1  
stdlib            go1.24.6             *1.24.8, 1.25.2   go-module  CVE-2025-58188  High      < 0.1% (2nd)   < 0.1  
stdlib            go1.24.6             *1.24.11, 1.25.5  go-module  CVE-2025-61729  High      < 0.1% (2nd)   < 0.1  
stdlib            go1.24.6             *1.24.8, 1.25.2   go-module  CVE-2025-58189  Medium    < 0.1% (4th)   < 0.1  
stdlib            go1.24.6             *1.24.8, 1.25.2   go-module  CVE-2025-58183  Medium    < 0.1% (1st)   < 0.1  
stdlib            go1.24.6             *1.24.11, 1.25.5  go-module  CVE-2025-61727  Medium    < 0.1% (0th)   < 0.1  
```

With vex document https://github.com/grooverdan/mariadb-docker/commit/17be165e5fce852492a4cd0f451639c1c7674a98

We've got a VEX document with right purl package `pkg:golang/github.com/tianon/gosu@v1.19.0` that has a govulchecker generated vex documented listing that its subcomponents that we not affected, bu `grype` doesn't match these to exclude them.

I'm attempting to make it useful for golang based products to release a VEX document - https://github.com/tianon/gosu/issues/157.

No go-module listing in the output for the `gosu` binary, the only go-module in the MariaDB container, as all vulnerability reports have been addressed by the VEX document.

**Environment**:
- Output of `grype version`:

grype 0.104.3

```
$ cat /etc/os-release 
NAME="Fedora Linux"
VERSION="43 (Workstation Edition)"
RELEASE_TYPE=stable
ID=fedora
VERSION_ID=43
```

xnox:
## Summary

Lets VEX `affected` / `under_investigation` statements add findings for
packages present in the SBOM but absent from grype's vulnerability
database — the gap that makes govulncheck-style VEX docs round-trip
into grype output today.

- **OpenVEX**: after the existing ignored-match promotion loop, walk
  the package catalog and synthesize a `match.Match` for each statement
  whose product (or product+subcomponent) purl names a package. Version
  matching is exact, or wildcard when the statement omits a version —
  no implicit ranges, matching the OpenVEX spec.
- **CSAF**: same synthesis, with status-aware version semantics from
  the spec:
    - `last_affected` → pkg.version `<=` stmt.version (ceiling)
    - `first_affected` → pkg.version `>=` stmt.version (floor)
    - `known_affected` / `recommended` / `under_investigation` → exact
    - `fixed` / `known_not_affected` → never synthesize
  Comparisons use `grype/version` via `pkg.VersionFormat`, so they are
  ecosystem-aware (semver, deb, rpm, apk, go-module, …). Statement
  qualifiers must be a subset of the package's qualifiers; type,
  namespace and name must match exactly.
- The `vexProcessorImplementation` interface and `ApplyVEX` now take
  `[]pkg.Package`, plumbed through `findVEXMatches`.
- Synthesis keys on `(vuln ID, package purl)` and skips pairs already
  present in remaining or ignored matches, so the path never duplicates
  a DB-backed finding.
- Behavior is gated by existing config (`vex-add: [affected,
  under_investigation]` + a matching ignore rule), so default scans are
  unchanged.

Closes #3145, completes the augment phase from #1365.

## Motivation

Stock grype:

```
$ grype ./step
No vulnerabilities found
```

govulncheck against the same binary:

```
$ govulncheck -mode=binary ./step
Vulnerability #1: GO-2026-5030
    Found in: golang.org/x/net@v0.53.0
    Fixed in: golang.org/x/net@v0.55.0
… (and 5 more in x/net)
```

With this change plus an OpenVEX doc declaring those vulns `affected`:

```
$ grype ./step --vex affected.vex.json -c grype.yaml
NAME              INSTALLED  TYPE       VULNERABILITY  SEVERITY
golang.org/x/net  v0.53.0    go-module  GO-2026-5025   Unknown
golang.org/x/net  v0.53.0    go-module  GO-2026-5026   Unknown
… (one row per VEX statement)
```

The same machinery works for CSAF documents, where `last_affected`
produces a ceiling match (e.g. SBOM `v0.50.0` matches `last_affected
v0.99.0` but is excluded by `last_affected v0.10.0`).

## Test plan

- [x] `go test ./grype/vex/... ./grype/` — passes, including the new
  unit tests
- [x] `go build ./...` and `go vet ./grype/...` clean
- [x] End-to-end: built grype with these changes, ran against a real
  Go binary (`step`) with both OpenVEX and CSAF documents, confirmed
  synthesized findings appear in `table` and `json` output formats
- [x] `Test_UnaffectedFiltering` still passes (verified independently;
  any local breakage was caused by a stale auto-generated
  `listing.xxh64` fixture, not by this change)
- [x] Coverage on touched packages: `grype/vex/csaf` 46.3% → 62.8%,
  `grype/vex/openvex` 63.4% → 77.1%; no regression elsewhere

### New tests

`grype/vex/openvex/implementation_test.go`:
- `TestAugmentMatches_SynthesizesFromPackageCatalog` — 7 cases:
  affected synthesizes; under_investigation synthesizes; not_affected /
  fixed do not; purl mismatch does not; empty catalog does not;
  non-matching vulnerability in ignore rule does not.
- `TestAugmentMatches_DoesNotDuplicateExistingMatches` — dedup against
  an existing DB-backed match.

`grype/vex/csaf/implementation_test.go`:
- `TestPackageMatchesStatement` — 16 cases covering ceiling/floor/
  exact/wildcard plus name/namespace/type mismatches across go-module
  versions.
- `TestAugmentMatches_SynthesizesFromPackageCatalog` — 9 cases covering
  each affected-like status (`last_affected`, `first_affected`,
  `known_affected`) against lower/equal/higher SBOM versions, plus
  `fixed` and `known_not_affected` negatives.
- `TestAugmentMatches_DoesNotDuplicateExistingMatches_CSAF` — dedup.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/grype/vex/csaf/csaf.go
+++ b/grype/vex/csaf/csaf.go
@@ -142,6 +142,69 @@ type synthesisCandidate struct {
 	Package       *pkg.Package
 }
 
+// indexedPackage is a package whose purl has been parsed once and whose
+// ecosystem version format has been resolved, so synthesis does not re-parse
+// the same package on every statement comparison.
+type indexedPackage struct {
+	pkg    *pkg.Package
+	purl   packageurl.PackageURL
+	format version.Format
+}
+
+// purlIdentityKey returns the (type, namespace, name) identity of a purl.
+// packageMatchesStatement requires these three to be equal, so a package can
+// only ever match a statement that shares this key. Indexing packages by it
+// lets synthesis compare each statement against the handful of packages with a
+// matching identity instead of the whole catalog.
+func purlIdentityKey(p packageurl.PackageURL) string {
+	return p.Type + "\x00" + p.Namespace + "\x00" + p.Name
+}
+
+// buildPackageIndex parses every package purl once and buckets the packages by
+// their (type, namespace, name) identity.
+func buildPackageIndex(pkgs []pkg.Package) map[string][]indexedPackage {
+	index := make(map[string][]indexedPackage)
+	for i := range pkgs {
+		if pkgs[i].PURL == "" {
+			continue
+		}
+		parsed, err := packageurl.FromString(pkgs[i].PURL)
+		if err != nil {
+			continue
+		}
+		key := purlIdentityKey(parsed)
+		index[key] = append(index[key], indexedPackage{
+			pkg:    &pkgs[i],
+			purl:   parsed,
+			format: pkg.VersionFormat(pkgs[i]),
+		})
+	}
+	return index
+}
+
+// statusProducts pairs a CSAF product-status slice with the synthesis status it
+// maps to. Using a fixed slice avoids allocating a map per vulnerability.
+type statusProducts struct {
+	status   status
+	products *csaf.Products
+}
+
+// synthesisStatuses returns the affected-like product-status buckets that are
+// eligible for synthesis. fixed and known_not_affected are intentionally
+// excluded.
+func synthesisStatuses(ps *csaf.ProductStatus) []statusProducts {
+	if ps == nil {
+		return nil
+	}
+	return []statusProducts{
+		{firstAffected, ps.FirstAffected},
+		{knownAffected, ps.KnownAffected},
+		{lastAffected, ps.LastAffected},
+		{recommended, ps.Recommended},
+		{underInvestigation, ps.UnderInvestigation},
+	}
+}
+
 // findSynthesisCandidates walks every advisory and yields (vuln, package)
 // pairs eligible for synthesis. Range semantics are applied per status:
 //   - last_affected: pkg.version <= stmt.version (ceiling)
@@ -152,10 +215,20 @@ type synthesisCandidate struct {
 // Statuses that are not "affected-like" (fixed, known_not_affected) never
 // trigger synthesis.
 //
+// Packages are pre-parsed and indexed by purl identity so each statement purl
+// is matched against only the packages that share its (type, namespace, name)
+// rather than the entire catalog. Per-advisory product purls are cached so the
+// product tree is walked once per product instead of once per package.
+//
 //nolint:gocognit
 func (advisories advisories) findSynthesisCandidates(pkgs []pkg.Package) []synthesisCandidate {
 	var out []synthesisCandidate
 	if len(pkgs) == 0 {
+		return out
+	}
+
+	index := buildPackageIndex(pkgs)
+	if len(index) == 0 {
 		return out
 	}
 
@@ -164,43 +237,46 @@ func (advisories advisories) findSynthesisCandidates(pkgs []pkg.Package) []synth
 			continue
 		}
 
+		// Cache product purls per advisory so CollectProductIdentificationHelpers
+		// (which walks the whole product tree) runs once per product ID.
+		helpersCache := map[csaf.ProductID][]string{}
+		purlsForProduct := func(productID csaf.ProductID) []string {
+			if cached, ok := helpersCache[productID]; ok {
+				return cached
+			}
+			purls := purlsFromProductIdentificationHelpers(adv.ProductTree.CollectProductIdentificationHelpers(productID))
+			helpersCache[productID] = purls
+			return purls
+		}
+
 		for _, vuln := range adv.Vulnerabilities {
 			if vuln == nil || vuln.CVE == nil {
 				continue
 			}
 
-			productsByStatus := map[status]*csaf.Products{
-				firstAffected:      vuln.ProductStatus.FirstAffected,
-				knownAffected:      vuln.ProductStatus.KnownAffected,
-				lastAffected:       vuln.ProductStatus.LastAffected,
-				recommended:        vuln.ProductStatus.Recommended,
-				underInvestigation: vuln.ProductStatus.UnderInvestigation,
-			}
-
-			for st, products := range productsByStatus {
-				if products == nil {
+			for _, sp := range synthesisStatuses(vuln.ProductStatus) {
+				if sp.products == nil {
 					continue
 				}
-				for _, productIDPtr := range *products {
+				for _, productIDPtr := range *sp.products {
 					if productIDPtr == nil {
 						continue
 					}
 					productID := *productIDPtr
-					helpers := adv.ProductTree.CollectProductIdentificationHelpers(productID)
-					for _, stmtPURL := range purlsFromProductIdentificationHelpers(helpers) {
-						for i := range pkgs {
-							p := &pkgs[i]
-							if p.PURL == "" {
-								continue
-							}
-							if !packageMatchesStatement(stmtPURL, p, st) {
+					for _, stmtPURL := range purlsForProduct(productID) {
+						stmt, err := packageurl.FromString(stmtPURL)
+						if err != nil {
+							continue
+						}
+						for _, cand := range index[purlIdentityKey(stmt)] {
+							if !packageMatchesParsed(stmt, cand.purl, cand.format, sp.status) {
 								continue
 							}
 							out = append(out, synthesisCandidate{
 								Vulnerability: vuln,
-								Status:        st,
+								Status:        sp.status,
 								ProductID:     productID,
-								Package:       p,
+								Package:       cand.pkg,
 							})
 						}
 					}
@@ -225,7 +301,13 @@ func packageMatchesStatement(stmtPURL string, p *pkg.Package, st status) bool {
 	if err != nil {
 		return false
 	}
+	return packageMatchesParsed(stmt, pkgPURL, pkg.VersionFormat(*p), st)
+}
 
+// packageMatchesParsed is packageMatchesStatement operating on already-parsed
+// purls and a resolved version format, so the hot synthesis loop does not
+// re-parse purls it has already seen.
+func packageMatchesParsed(stmt, pkgPURL packageurl.PackageURL, format version.Format, st status) bool {
 	if stmt.Type != pkgPURL.Type || stmt.Namespace != pkgPURL.Namespace || stmt.Name != pkgPURL.Name {
 		return false
 	}
@@ -241,8 +323,6 @@ func packageMatchesStatement(stmtPURL string, p *pkg.Package, st status) bool {
 		// Statement is version-specific but the package's purl has none.
 		return false
 	}
-
-	format := pkg.VersionFormat(*p)
 
 	switch st {
 	case lastAffected:
--- a/grype/vex/csaf/csaf.go
+++ b/grype/vex/csaf/csaf.go
@@ -4,6 +4,10 @@ import (
 	"slices"
 
 	"github.com/gocsaf/csaf/v3/csaf"
+
+	"github.com/anchore/grype/grype/pkg"
+	"github.com/anchore/grype/grype/version"
+	"github.com/anchore/packageurl-go"
 )
 
 // advisoryMatch captures the criteria that caused a vulnerability to match a CSAF advisory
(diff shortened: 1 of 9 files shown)


grype/vex/csaf/csaf.go
```diff
@@ -127,3 +131,157 @@ func purlsFromProductIdentificationHelpers(helpers []*csaf.ProductIdentification
...
+				}
+				for _, productIDPtr := range *products {
+					if productIDPtr == nil {
+						continue
+					}
+					productID := *productIDPtr
+					helpers := adv.ProductTree.CollectProductIdentificationHelpers(productID)
+					for _, stmtPURL := range purlsFromProductIdentificationHelpers(helpers) {
+						for i := range pkgs {
+							p := &pkgs[i]
```
kzantow: I think this will end up being a performance issue. There will be _lots_ of packages. Probably best to build some sort of index for all the known vulnerabilities at least by package name or simplified PURL?

xnox: @kzantow i am also pondering if this should be some sort of vunnel datafeed; as in something similar to hummingbird; but allow one to point at a directory of self-created CSAF for internal (or public) things; and build a db and use a db. Because yes this will perform very poorly compared to prebuilding custom CSAF vex into a custom grype db.

Or like import vunnel and generate vunnel db on the fly.

kzantow: Today the db building functionality isn't included in grype itself, but that may change in the not too distant future. Maybe there could be a way to build an [in-memory db](https://sqlite.org/inmemorydb.html); I don't think this would work without some changes across the board from where things are today, but it isn't out of the question. I think the approach you have here is probably fine generally speaking. I do wonder if the VEX functionality should just be another matcher if matchers [were no longer based on package types](https://github.com/anchore/grype/issues/2112). I think the main thing to do here is just make a lookup based on some package information like the name or PURL so the iteration isn't O(n^6)

xnox: @kzantow this is now optimized; it is more lines of code; but the scan times are now back to ms and increase linearly with sbom amount of packages.

xnox:
Pushed a follow-up (`2c9f9d52`) addressing the performance concerns on the affected/under_investigation synthesis path.

The original synthesis walked the whole package catalog for every VEX statement and re-parsed PURLs on each comparison — `O(statements × packages)` with as many PURL parses, i.e. quadratic in catalog size. Packages are now parsed once and bucketed by `(type, namespace, name)` identity (which `PurlMatches`/`packageMatchesStatement` already require to be equal), so each statement only looks at the handful of packages that share its identity. CSAF additionally caches per-advisory product purls so the product tree is walked once per product instead of once per package. Semantics are unchanged (image-wide statements still fall back to the full catalog) and all existing `grype/vex` tests pass unchanged.

Benchmark (one affected statement per package, driving `AugmentMatches`):

| catalog | before | after | speedup |
|--:|--:|--:|--:|
| OpenVEX 1000 pkgs | 1.31 s | 7.1 ms | ~184× |
| OpenVEX 2000 pkgs | 5.80 s | 14 ms | ~412× |
| CSAF 1000 pkgs | 1.09 s | 14 ms | ~78× |
| CSAF 2000 pkgs | 5.28 s | 36 ms | ~148× |

Scaling is now linear instead of quadratic (doubling the catalog ~doubles the time rather than ~quadrupling it). At ~1000 packages the synthesis step is single-digit-to-low-double-digit milliseconds, so the affected-VEX supplement is now effectively negligible on top of a normal scan (which is dominated by SBOM cataloging and DB matching that take seconds).

grype/vex/csaf/csaf.go
```diff
@@ -127,3 +131,237 @@ func purlsFromProductIdentificationHelpers(helpers []*csaf.ProductIdentification
...
+	purl   packageurl.PackageURL
+	format version.Format
+}
+
+// purlIdentityKey returns the (type, namespace, name) identity of a purl.
+// packageMatchesStatement requires these three to be equal, so a package can
+// only ever match a statement that shares this key. Indexing packages by it
+// lets synthesis compare each statement against the handful of packages with a
+// matching identity instead of the whole catalog.
+func purlIdentityKey(p packageurl.PackageURL) string {
```
kzantow, line 159: we should probably use comparable types instead of concat'd strings as map keys -- these can definitely add up to noticable time with the scale we have sometimes, e.g.:
```
type purlKey struct {
  typ, namespace, name string
}
```

(same comment everywhere we are making strings as map keys)

grype/vex/csaf/implementation_test.go
```diff
@@ -198,3 +198,219 @@ func Test_matchingRule(t *testing.T) {
...
+
+	tests := []struct {
+		name     string
+		stmtPURL string
+		pkgPURL  string
+		status   status
+		want     bool
+	}{
+		// last_affected: ceiling
+		{"last_affected matches lower pkg version", "pkg:golang/golang.org/x/net@v0.54.0", "pkg:golang/golang.org/x/net@v0.53.0", lastAffected, true},
```
kzantow, line 218: these would be more clear if they used the property name form, e.g.:
```
{
  name: ...
  stmtPURL: ...
  ...
```

grype/vex/csaf/csaf.go
```diff
@@ -127,3 +131,237 @@ func purlsFromProductIdentificationHelpers(helpers []*csaf.ProductIdentification
 	}
 	return purls
 }
+
+// synthesisCandidate describes a (vulnerability, package) pair that should be
+// added to grype's results based on a CSAF advisory, when no DB-backed match
+// already exists.
+type synthesisCandidate struct {
```
kzantow, line 138: it looks like we now have a `synthesisCandidate` which is converted to an `advisoryMatch` which is converted to a `match.Match`... could we avoid the middlemen on these and just directly create `Vulnerability`, `IgnoreRule`/`IgnoreFilter`, and `Match` objects or similar? We could move the [IgnoreFilter indexing](https://github.com/anchore/grype/blob/main/grype/vulnerability_matcher.go#L409) to some shared location

grype/vex/csaf/csaf.go
```diff
@@ -127,3 +131,237 @@ func purlsFromProductIdentificationHelpers(helpers []*csaf.ProductIdentification
...
+// only ever match a statement that shares this key. Indexing packages by it
+// lets synthesis compare each statement against the handful of packages with a
+// matching identity instead of the whole catalog.
+func purlIdentityKey(p packageurl.PackageURL) string {
+	return p.Type + "\x00" + p.Namespace + "\x00" + p.Name
+}
+
+// buildPackageIndex parses every package purl once and buckets the packages by
+// their (type, namespace, name) identity.
+func buildPackageIndex(pkgs []pkg.Package) map[string][]indexedPackage {
```
kzantow, line 165: It looks like both of these implementations have similar `buildPackageIndex` functions that operate on the full set of packages. I think we should flip this to instead build indexes for VEX rules. It's hard to say which would be a smaller set (definitely VEX rules with no vex files), but we operate a single-package at a time in the matcher world and already have indexes for `IgnoreRule`s and other `IgnoreFilter`s; I see a lot of similarity to matchers here and I think a future refactoring is likely to introduce per-package streaming, which this would be incompatible with.

kzantow (commented):
Sorry a bit of this feedback might be a bit abstract, ping me if you have questions.

I think we can probably move forward with this if we could at least flip the package indexes to vulnerability indexes and avoid passing a package slice to the implementations. It would be great to maybe move some of the duplicated logic into the processor: e.g. add a new interface function for each implementation to return a list of `Vulnerability` objects (the affected records) instead of passing `[]pkg.Package` into each implementation, the processor handles iterating the packages in a single spot and matching against indexed vulnerabilities. I think this will fit better into a future state. If you take away the _augmentation_, each vex file is effectively a [VulnerabilityProvider](https://github.com/anchore/grype/blob/7bf3e633b7f5be278e3b21121a703b89dd7371a7/grype/vulnerability/provider.go#L29) and the VEX processor is effectively a [Matcher](https://github.com/anchore/grype/blob/7bf3e633b7f5be278e3b21121a703b89dd7371a7/grype/match/matcher.go#L13) that [handles all package types](https://github.com/anchore/grype/issues/2112). We already want to merge vulnerabilities, so I think the aforementioned changes will require less refactoring later.

xnox:
@kzantow thank you! yes, I agree that my current design is clunky and doesn't fit. And yet it should be designed better to integrate into Vulnerability, IgnoreRule/IgnoreFilter, and Match. Because yes it should be all of these things like a "normal" vuln source / matcher / ignore-rule / ignore-filter. Sure a human thinks it is a single vex statement, but it actually is a syntactic sugar of custom vulnerability source + match + ignore.

PR #3468, opened 2026-05-28
Add OSV transformer for Ubuntu records

willmurphyscode:
Includes moving on from the google osv-scanner model to our own, since the osv-scanner models are no longer maintained.

### DIFF
--- a/grype/db/internal/provider/unmarshal/osvmodel/anchore.go
+++ b/grype/db/internal/provider/unmarshal/osvmodel/anchore.go
@@ -1,0 +1,81 @@
+package osvmodel
+
+import "encoding/json"
+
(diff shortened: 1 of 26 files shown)


wagoodman (approved):

grype/db/internal/provider/unmarshal/osvmodel/anchore.go
wagoodman, line 1: This looks to be pattern setting for any extension, we should sync a little more on this

wagoodman:
I converted this to draft since the provider took a different turn and this is no longer needed in the short term, though ideally, this should be merged and be how the provider writes out entries in the long term.

Issue #2838, opened 2025-07-31
False positive scanning purl pkg:golang/github.com/hashicorp/vault@v0.9.0#api/auth/kubernetes


goatwu1993:
**What happened**:

```bash
grype 'pkg:golang/github.com/hashicorp/vault@v0.9.0#api/auth/kubernetes'
```

It does not handle the subpath, interpreted it as name `github.com/hashicorp/vault` and version `v0.9.0`, therefore report vulnerabilities such as CVE-2020-16250.

However https://pkg.go.dev/github.com/hashicorp/vault/api/auth/kubernetes@v0.9.0 is a seperate module from https://pkg.go.dev/github.com/hashicorp/vault@v0.9.0 and has no vulnerability.

**What you expected to happen**:

no cve found

**How to reproduce it (as minimally and precisely as possible)**:

```bash
# init a minimal go dir
go mod init app
go get github.com/hashicorp/vault/api/auth/kubernetes@v0.9.0

# use syft to scan the dir
syft . -o json | jq  '.artifacts[].purl' -r | grep 'pkg:golang/github.com/hashicorp/vault'
 ✔ Indexed file system                                                                                                                                                   .
 ✔ Cataloged contents                                                                                     cdb4ee2aea69cc6a83331bbe96dc2caa9a299d21329efb0336fc02a82e1839a8
   ├── ✔ Packages                        [20 packages]
   ├── ✔ Executables                     [0 executables]
   ├── ✔ File metadata                   [1 locations]
   └── ✔ File digests                    [1 files]
[0000]  WARN no explicit name and version provided for directory source, deriving artifact ID from the given path (which is not ideal)
pkg:golang/github.com/hashicorp/vault@v1.16.0#api
pkg:golang/github.com/hashicorp/vault@v0.9.0#api/auth/kubernetes

# use grype to scan the purl
grype 'pkg:golang/github.com/hashicorp/vault@v0.9.0#api/auth/kubernetes'
 ✔ Scanned for vulnerabilities     [0 vulnerability matches]
   ├── by severity: 0 critical, 7 high, 10 medium, 1 low, 0 negligible
   └── by status:   18 fixed, 0 not-fixed, 0 ignored
NAME                        INSTALLED  FIXED IN  TYPE       VULNERABILITY        SEVERITY  EPSS           RISK
github.com/hashicorp/vault  v0.9.0     1.2.5     go-module  GHSA-fp52-qw33-mfmw  High      1.0% (76th)    0.8
[...]

PR #3469, opened 2026-05-29
Fix false positives for literal Go PURLs with subpaths

immanuwell:
fixes the literal `pkg:golang/...#...` scan path from #2838.

right now grype builds `pkg:golang/github.com/hashicorp/vault@v0.9.0#api/auth/kubernetes` as `github.com/hashicorp/vault`, so it can match the wrong module. thats the bug.

this keeps the Go subpath when grype builds a package from a literal PURL. tiny change, no funny business.

repro:
1. run `grype 'pkg:golang/github.com/hashicorp/vault@v0.9.0#api/auth/kubernetes'`
2. before this patch, the package name becomes `github.com/hashicorp/vault`
3. with this patch, it becomes `github.com/hashicorp/vault/api/auth/kubernetes`

checks:
- `go test ./grype/pkg -run Test_PurlProvider`
- `go test ./...` in this env still trips in `test/integration` because `.tool/skopeo` is missing

related:
- #2838
- anchore/syft#4395

### DIFF
--- a/grype/pkg/purl_provider.go
+++ b/grype/pkg/purl_provider.go
@@ -24,7 +24,7 @@ type PURLLiteralMetadata struct {
 }
 
 func purlEnhancers(applyChannel func(*distro.Distro) bool) []Enhancer {
-	return []Enhancer{setUpstreamsFromPURL, setDistroFromPURL(applyChannel)}
+	return []Enhancer{setNameFromPURL, setUpstreamsFromPURL, setDistroFromPURL(applyChannel)}
 }
 
 func purlProvider(userInput string, config ProviderConfig, applyChannel func(*distro.Distro) bool) ([]*Package, Context, *sbom.SBOM, error) {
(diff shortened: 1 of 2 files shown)


noqt (changes requested):
I reviewed this for NOQT at `160b004e2e79ab6238807864fc1800d33ba1b410`.

The literal-PURL regression passes, but this needs one more turn before it solves #2838 across Grype's accepted input paths.

The patch adds `setNameFromPURL` through `purlEnhancers`. Native Syft JSON deliberately skips those enhancers: `syft_sbom_provider.go` applies them only when the format isn't `syftjson.ID`. I added a reviewer-only regression based on the existing valid Syft JSON fixture, with `name=github.com/hashicorp/vault` and PURL `pkg:golang/github.com/hashicorp/vault@v0.9.0#api/auth/kubernetes`. At this head it still produces the root name, so the same false-positive path remains for native Syft JSON input.

Current `main` also conflicts in `purlEnhancers` because the `applyChannel` signature changed. More importantly, the maintainer conclusion in #2838 was to add the subpath as an additional Go-matcher search rather than rewriting package identity only in selected providers. Please rebase and cover that matcher-level fanout for both literal PURL and native Syft JSON input.

The PR's focused test passes. Package-wide testing only hit local Docker-fixture failures, and vet passed.

PR #3474, opened 2026-05-29
feat(cg/transformers): add chainguard osv transformer

crosleyzack:
### What

Add Chainguard transformer for new OSV feed. This also sets the architecture on packages so vuln matches can be filtered by architecture. Test images from Chainguard are currently being reviewed and should be in our GCR bucket soon.

### Why

Allows grype to convert the vunnel output into vuln matches

### Notes

Similar but updated version of https://github.com/anchore/grype/pull/3255/

Part of Chainguard issue INT-520


grype/db/v6/blobs.go
```diff
@@ -159,6 +159,10 @@ type PackageQualifiers struct {
 	// RootIO indicates that the vulnerability applies only to Root IO packages (packages with Root IO fixes).
 	// When true, standard packages will not match this vulnerability (NAK pattern).
 	RootIO *bool `json:"rootio,omitempty"`
+
+	// Architecture specifies the CPU architecture this vulnerability applies to (e.g., "aarch64", "x86_64").
+	// If nil, the vulnerability applies to all architectures.
+	Architecture *string `json:"architecture,omitempty"`
```
crosleyzack: Can we just re-use the RpmArch package qualifier? Or collapse these into one field?

grype/db/v6/vulnerability.go
```diff
@@ -317,6 +317,9 @@ func toPackageQualifiers(qualifiers *PackageQualifiers) []qualifier.Qualifier {
 	if qualifiers.RootIO != nil && *qualifiers.RootIO {
 		out = append(out, rootio.New())
 	}
+	if qualifiers.Architecture != nil && *qualifiers.Architecture != "" {
+		// TODO can we reuse RpmArch, or should this be a new qualifier separate from RPM?
+	}
```
crosleyzack: Can we just re-use the RpmArch package qualifier? Or collapse these into one field?

grype/db/v6/vulnerability_test.go
```diff
@@ -1128,6 +1133,118 @@ func TestToFix(t *testing.T) {
...
+		{
+			name: "architecture only — not yet wired into matcher",
+			input: &PackageQualifiers{
+				Architecture: asPtr("x86_64"),
+			},
+			// Architecture is persisted on the blob but toPackageQualifiers does
+			// not currently translate it. When CG advisories rely on this field
+			// alone, the matcher sees an empty qualifier slice.
+			want: nil,
+		},
```
crosleyzack: Not hooked up yet, want feedback on what to do with arch first

grype/db/v6/build/transformers/osv/transform_cg.go
```diff
@@ -0,0 +1,144 @@
...
+// types fall back to the default mapping. CG OSV records should all use
+// ECOSYSTEM, but the fallback is here so unexpected shapes don't silently
+// become "unknown".
+func cgRangeType(t models.RangeType) string {
+	if t == models.RangeEcosystem {
+		// TODO is this correct? We do use APK I believe
+		return pkg.ApkPkg.String()
+	}
+	return defaultRangeType(t)
+}
```
crosleyzack: I believe our ECOSYSTEM ranges are apk format? @tdunlap607

crosleyzack: I don't see anything in our documentation indicating the version format specifically

grype/db/internal/provider/unmarshal/osv_vulnerability.go
```diff
@@ -6,7 +6,18 @@ import (
...
+// fields used by Chainguard/Wolfi advisories that aren't yet in the osv-scanner library.
+type OSVVulnerability struct {
+	models.Vulnerability
+
+	// TODO should this be put into upstream osv-scanner?
+	// Upstream contains IDs of upstream vulnerabilities that this advisory addresses.
+	// Per OSV spec, this is semantically correct for distro advisories that reference
+	// upstream CVEs. Included alongside "related" for backwards compatibility.
+	// See: https://ossf.github.io/osv-schema/
+	Upstream []string `json:"upstream,omitempty"`
```
crosleyzack: Should this be put into the upstream models.Vulnerability?

willmurphyscode: `"github.com/google/osv-scanner/pkg/models"` is deprecated and not getting updates any more. We might need to implement our own structs in this repo soon. I think this approach is ok for this PR.

grype/pkg/syft_sbom_provider.go
```diff
@@ -38,9 +38,12 @@ func syftSBOMProvider(userInput string, config ProviderConfig, applyChannel func
...
 	d, distroDetectionFailed := distroFromSBOM(s, config, applyChannel)
 
-	var enhancers []Enhancer
+	// setArchFromPURL runs for every format: no SBOM decoder populates grype's
+	// pkg.Arch field directly, so without this the architecture qualifier on
+	// distro-keyed vulnerabilities can't filter by package arch.
+	enhancers := []Enhancer{setArchFromPURL}
 	if fmtID != syftjson.ID {
-		enhancers = purlEnhancers(applyChannel)
+		enhancers = append(enhancers, setUpstreamsFromPURL, setDistroFromPURL(applyChannel))
```
crosleyzack: we want normal scans to also include the arch information if available

grype/pkg/syft_provider.go
```diff
@@ -35,7 +35,10 @@ func syftProvider(userInput string, config ProviderConfig, applyChannel func(*di
 
 	d, distroDetectionFailed := distroFromSBOM(s, config, applyChannel)
 
-	packages := FromCollection(s.Artifacts.Packages, s.Relationships, config.SynthesisConfig)
+	// setArchFromPURL is applied for image scans because syft does not populate
+	// grype's pkg.Arch field directly — without this the architecture qualifier
+	// on distro-keyed vulnerabilities can't filter by package arch.
+	packages := FromCollection(s.Artifacts.Packages, s.Relationships, config.SynthesisConfig, setArchFromPURL)
```
crosleyzack: include arch information in sbom scans

grype/pkg/purl_provider.go
```diff
@@ -100,3 +100,19 @@ func distroFromPURL(purl packageurl.PackageURL) (d *distro.Distro) {
...
 }
+
+func setArchFromPURL(out *Package, purl packageurl.PackageURL, _ syftPkg.Package) {
+	if out.Arch == "" || out.PURL == "" {
+		out.Arch = archFromPURL(purl)
+	}
+}
+
+// archFromPURL reads arch data for Grype can use, which is ignored by Syft's PURL conversion
+func archFromPURL(purl packageurl.PackageURL) (arch string) {
```
willmurphyscode: I think this is unnecessary now that https://github.com/anchore/syft/pull/4987 is in Syft.

grype/pkg/package.go
```diff
@@ -49,6 +49,8 @@ type Package struct {
 	Locations file.LocationSet // the locations that lead to the discovery of this package (note: this is not necessarily the locations that make up this package)
 	Language  syftPkg.Language // the language ecosystem this package belongs to (e.g. JavaScript, Python, etc)
 	Distro    *distro.Distro   // a specific distro this package originated from
+	// TODO: should this be an enum to avoid mismatches like "amd64" vs "x86_64"?
+	Arch      string // the architecture of the package (e.g. "amd64", "arm64", etc)
```
willmurphyscode: This field should be on the ApkMetadata struct, not the top level package. Specifically here: https://github.com/anchore/grype/blob/main/grype/pkg/apk_metadata.go#L12 and set here: https://github.com/anchore/grype/blob/main/grype/pkg/package.go#L381

The package types in Grype are a narrowing of the Syft package interface. For example, APKs in Grype have owned file records because grype uses that for matching. This PR adds the ability for APKs to filter matching based on architecture.

We don't want Architecture on the top level package struct because it doesn't make sense for all package types. For example, a binary package could be a macOS multiarch binary, which would give us nothing good to put in a string field here.

grype/pkg/purl_provider.go
crosleyzack: removed in 0c9f5f08f5841a5bed0bc9c49805c9b47a362c04. Local testing by generating sbom with syft main branch and scanning the sbom shows it still works without this code.

grype/pkg/package.go
crosleyzack: Added to the ApkMetadata in 0c9f5f08f5841a5bed0bc9c49805c9b47a362c04 and used that value for arch filtering.

grype/pkg/qualifier/architecture/qualifier.go
```diff
@@ -34,10 +34,13 @@ func (r architectureQualifier) Arch() string {
...
-// operate on the vulnerability side (see internal.SourceOrUnspecifiedArch), not by
-// per-package qualifier evaluation. Direct hits on a binary-tagged entry must still match
-// by name, so this always returns true.
-func (r architectureQualifier) Satisfied(_ pkg.Package) (bool, error) {
+// Satisfied checks whether the package's architecture matches the qualifier's architecture.
+// If the package does not have an architecture specified, the qualifier is inert
+// (i.e., it does not filter out packages without architecture information).
+func (r architectureQualifier) Satisfied(p pkg.Package) (bool, error) {
+	if meta, ok := p.Metadata.(pkg.ApkMetadata); ok {
+		return meta.Arch == r.arch, nil
```
willmurphyscode: > If the package does not have an architecture specified, the qualifier is inert.

This should be true, but doesn't seem to be true of the current implementation. What if the packager has an ApkMetadata set, but the metadata does not have an architecture? What if the vuln record has no architecture.

grype/pkg/apk_metadata.go
```diff
@@ -10,6 +10,8 @@ var _ FileOwner = (*ApkMetadata)(nil)
 
 type ApkMetadata struct {
 	Files []ApkFileRecord `json:"files"`
+	// TODO: should this be an enum to avoid mismatches like "amd64" vs "x86_64"?
```
willmurphyscode: That's a good question. In https://github.com/anchore/grype/pull/3504 we decided to try to wire up the list of architecture aliases into the database. I'll see if I can get that one in quickly and then you can rebase on it.

grype/pkg/qualifier/architecture/qualifier.go
crosleyzack: ah, seem to have lost this when I switched to ApkMetadata

crosleyzack: fixed in eb34c6e162e5f51a7296ccff3c620412a56b9d90

grype/pkg/syft_sbom_provider.go
```diff
@@ -38,9 +38,9 @@ func syftSBOMProvider(userInput string, config ProviderConfig, applyChannel func
 
 	d, distroDetectionFailed := distroFromSBOM(s, config, applyChannel)
 
-	var enhancers []Enhancer
```
willmurphyscode, line 41: I don't think this change is needed anymore. Have you caught syft not doing this?

willmurphyscode (approved):

grype/pkg/syft_sbom_provider.go
crosleyzack: I did not, so I think reverting this is fine

PR #3482, opened 2026-06-01
chore: rename rpm_arch to architecture

willmurphyscode:
This field was initially added to support RPM architecture distinctions (especially arch=src), but multiple package types will eventually want to have fixes or vulnerabilities that apply to only one architecture. Since the field hasn't been released yet, rename it to architecture from rpm_arch, rather than having different fields for different package types.

In order to land this, some supporting changes are included:

1. DBTest cache is invalidated if db schemas change
2. DBTest cache in invalidated if source code of db build package changes
3. Centralize a RepoRoot function in internal/testutils, because it was already implemented separately in 3 places and I needed a fourth :)


grype/internal/packagemetadata/discover_type_names.go
```diff
@@ -22,7 +23,7 @@ var metadataExceptions = strset.New(
 )
 
 func DiscoverTypeNames() ([]string, error) {
-	root, err := RepoRoot()
+	root, err := testutils.RepoRoot()
```
wagoodman: should not be importing `testutils` in a non-test file

cmd/grype/cli/commands/internal/jsonschema/main.go
```diff
@@ -251,13 +251,9 @@ func schemaID(component, version string) jsonschema.ID {
 }
 
 func repoRoot() string {
-	root, err := exec.Command("git", "rev-parse", "--show-toplevel").Output()
+	root, err := testutils.RepoRoot()
```
wagoodman: same story

grype/db/v6/schema/main.go
```diff
@@ -460,13 +460,9 @@ func getJSONTag(field *ast.Field) string {
 }
 
 func repoRoot() string {
-	root, err := exec.Command("git", "rev-parse", "--show-toplevel").Output()
+	root, err := testutils.RepoRoot()
```
wagoodman: same story

wagoodman (approved):

PR #3497, opened 2026-06-05
fix: pin google-gocontainerregistry back to v0.21.5

### LINKED ISSUE #3496: Regression in v0.113.0: grype hangs indefinitely (no timeout) on registry image pull — works in v0.112.0
### Summary

**Regression in v0.113.0:** `grype` intermittently **hangs forever (no timeout)** while pulling an image directly from a registry. The same command works reliably in **v0.112.0 and all earlier versions**. In CI this burns the full 6-hour GitHub-runner limit.

The root cause appears to be the `google/go-containerregistry` bump from `v0.21.5` → `v0.21.6` (pulled in via the `anchore/stereoscope` `v0.1.23` → `v0.2.0` bump). See "Root cause" below.

### Environment

- grype: **0.113.0** (hangs) vs **0.112.0** (works) — installed via `curl -sSfL https://get.anchore.io/grype`
- DB schema: v6.1.7
- OS: macOS (darwin 24.6.0, local repro) and Ubuntu 24.04 (GitHub-hosted runner, CI)
- Registry: private **Azure Container Registry** (`*.azurecr.io`)
- Image media type: `application/vnd.oci.image.manifest.v1+json` (single platform, multi-layer)
- For comparison: **syft 1.38.2 and syft 1.45.0 both succeed** on the same image/registry (see below).

### What happens

With `-vv`, the last line before the hang is consistently the image-read step, right after the manifest downloads successfully:

```
DEBUG pulling image info directly from registry image="<registry>/<repo>:<tag>" from=stereoscope
DEBUG no registry credentials configured for "<registry>", using the default keychain from=stereoscope
INFO  completed downloading manifest from=stereoscope image=<registry>/<repo>:<tag> time=2.26s
DEBUG reading image digest=sha256:... from=stereoscope mediaType=application/vnd.oci.image.manifest.v1+json tags=[]
   <-- hangs here indefinitely, no further output, no timeout, no error -->
```

Auth succeeds, the manifest downloads in ~2s. The stall is during the subsequent **layer/blob fetch**.

### Strongly reproducible under parallel scans

We run 4 `grype` processes in parallel over 16 images (by digest) against ACR. In a representative failed CI run, 2 of the first batch of 4 finished in ~8s and the **other 2 hung forever** — still alive as orphan processes when the runner was force-killed at 6h:

```
21:18:22  Starting Grype scan for 16 images using 4 threads...
21:18:22  Scanning batch of 4 images...
21:18:29  spire-controller-manager@sha256:... completed in 7.43s
21:18:31  bash@sha256:...                      completed in 9.38s
                      -- 5h 51m of silence --
03:09:20  ##[error]The operation was canceled.
03:09:20  Terminate orphan process: pid (3304) (grype)
03:09:20  Terminate orphan process: pid (3312) (grype)
```

### Steps to reproduce

1. With grype **v0.113.0**: `grype registry:<your-registry>.azurecr.io/<repo>:<tag>` against a private ACR (auth via keychain / `docker login`). Single runs usually complete (slowly); intermittently they hang at "reading image" and never return.
2. Reproduces far more reliably with concurrency — e.g. 4 concurrent `grype registry:...` invocations against ACR; typically 1–2 hang indefinitely.
3. On a fresh CI runner (no local image cache), even the bare `grype <ref>` form hits the registry-pull path and hangs the same way.
4. Downgrade to grype **v0.112.0** with the identical command/images → no hangs.

### Root cause (suspected)

The dependency deltas between the two grype tags:

| dependency | v0.112.0 | v0.113.0 |
|---|---|---|
| `github.com/anchore/stereoscope` | v0.1.23 | v0.2.0 |
| `github.com/google/go-containerregistry` | v0.21.5 | **v0.21.6** |

go-containerregistry **v0.21.6** includes [`remote: limit concurrent layer pulls (#2271)`](https://github.com/google/go-containerregistry/pull/2271) (fixing #2140), which adds a pull limiter that **holds a token until each blob's response body is closed**. This introduced a deadlock on multi-layer pulls, already reported upstream as [**google/go-containerregistry#2309 — "crane pull hangs on multi-layer images (v0.21.6 regression)"**](https://github.com/google/go-containerregistry/issues/2309).

This matches every symptom here: intermittent, multi-layer images, worse under concurrency, hangs at the layer-read step with no timeout.

That `syft` does **not** hang on the same images — with both **1.38.2** and the latest **1.45.0** — is consistent with this: those syft builds depend on a go-containerregistry version outside the regressed `v0.21.6`.

### Expected behavior

`grype` should not regress from v0.112.0's reliable behavior. Picking up the upstream fix for go-containerregistry#2309 (bump go-containerregistry, and/or stereoscope once it does) should resolve it. A configurable overall pull timeout would also let CI fail fast and retry instead of hanging indefinitely.

### Possibly related

- #492, #1051 (grype hangs in CI)

### Current workarounds

- Pin grype to **v0.112.0**.
- Pre-`docker pull` then scan the local daemon image (avoids the direct-registry pull path).
- Wrap grype in an external `timeout` + retry.

willmurphyscode:
There is a regression in v0.21.6 causing Syft and Grype to hang sometimes on --from registry pulls.

Fixes https://github.com/anchore/grype/issues/3496

Testing:

Running `go run ./cmd/grype openjdk:27-ea-trixie --from registry` hangs on main but not on this branch.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -37,7 +37,7 @@ require (
 	github.com/gocsaf/csaf/v3 v3.5.1
 	github.com/gohugoio/hashstructure v0.6.0
 	github.com/google/go-cmp v0.7.0
-	github.com/google/go-containerregistry v0.21.6
+	github.com/google/go-containerregistry v0.21.5
 	github.com/google/osv-scanner v1.9.2
 	github.com/google/shlex v0.0.0-20191202100458-e7afc7fbc510
 	github.com/google/uuid v1.6.0
@@ -165,6 +165,7 @@ require (
 	github.com/containerd/log v0.1.0 // indirect
 	github.com/containerd/platforms v1.0.0-rc.4 // indirect
 	github.com/containerd/plugin v1.1.0 // indirect
+	github.com/containerd/stargz-snapshotter/estargz v0.18.2 // indirect
 	github.com/containerd/ttrpc v1.2.8 // indirect
 	github.com/containerd/typeurl/v2 v2.2.3 // indirect
 	github.com/cyphar/filepath-securejoin v0.6.1 // indirect
@@ -302,6 +303,7 @@ require (
 	github.com/tidwall/pretty v1.2.1 // indirect
 	github.com/tidwall/sjson v1.2.5 // indirect
 	github.com/vbatts/go-mtree v0.7.0 // indirect
+	github.com/vbatts/tar-split v0.12.2 // indirect
 	github.com/vifraa/gopom v1.0.0 // indirect
 	github.com/xanzy/ssh-agent v0.3.3 // indirect
 	github.com/xo/terminfo v0.0.0-20220910002029-abceb7e1c41e // indirect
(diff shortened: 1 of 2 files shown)


kzantow:
We decided to move forward without the need to roll back

PR #3499, opened 2026-06-08
Add OSV Model generator

willmurphyscode:
Previously, there was an OSV scanner repository that exported a models.Vulnerability struct we could use. That repository deprecated their exported models and the models in grype are falling behind. Therefore, add a generator target that generates our own models based on the latest v1 JSON scehma from the official OSV spec repository. In this way automation can PR new models in.

# Alternatives

Just write the models directly and keep them up to date by hand: https://github.com/anchore/grype/pull/3500

# Rejected alternatives

## Alternative considered: depend on github.com/ossf/osv-schema/bindings/go/osvschema directly

Rejected. Those bindings are protobuf-generated, which forces:

  - Pointer types everywhere (*Package, []*Range, []*Event) — every affected.Package.Ecosystem becomes affected.GetPackage().GetEcosystem() across every strategy.
  - *timestamppb.Timestamp instead of time.Time — .AsTime() at every timestamp read site (every strategy uses &vuln.Modified today).
  - Enums typed as int32 (Range_SEMVER = 2) — the existing string switches in helpers.go and per-strategy rangeType() funcs all rewrite, and encoding/json can't unmarshal "type": "SEMVER" into an int enum, so parsing needs protojson and the
  google.golang.org/protobuf runtime as a direct dep.
  - database_specific / ecosystem_specific become *structpb.Struct — current m["anchore"]["fixes"]-style walks become GetFields() / GetStructValue() chains (or an .AsMap() call per access).

  Large mechanical churn through every strategy file plus a heavy new runtime dep, in exchange for "model auto-tracks upstream" — a property the JSON-schema generator gives us without the protobuf tax.

## Why not use a JSON-Schema → Go generator library?

Tried github.com/atombender/go-jsonschema (the most-maintained option in the Go ecosystem — omissis/go-jsonschema is a fork with a colliding module path; quicktype is Node-based; everything else is OpenAPI-flavored). Off-the-shelf output against the OSV schema is unusable:

  - top-level interface{} (the schema's allOf defeats the unifier)
  - *string/*T for every optional field
  - anonymous []struct{...} slice elements (no parameter type for helpers)
  - map[string]interface{} for Event (the schema's oneOf bails out)
  - typed-string aliases like CurrentlySupportedHomeDatabaseIdentifierPrefixes
  - a bare Type enum that collides with anything else of that name
  - no UnmarshalJSON for the defined Timestamp alias

Making the library produce usable output requires ~250 lines of schema preprocessing — strip oneOf/allOf, flatten Event's oneOf into properties, lift inline items/enums into named $defs, mark every property required, denull union types, inline simple-string refs, delete orphan defs. That's almost as much code as a hand-rolled emitter and exists purely to bridge the library's opinions to ours.

Trade-off: own ~660 lines of focused emission code we wrote ourselves, vs. own ~250 lines of schema preprocessing plus the library's quirks plus an adapter file (constant aliases, Timestamp.UnmarshalJSON) plus strategy adjustments for the residual library-shaped fields. The hand-rolled generator reads top-to-bottom in a single file; the library route reads as a sequence of workarounds for somebody else's defaults.


grype/db/internal/provider/unmarshal/osvmodel/generate/main.go
```diff
@@ -0,0 +1,657 @@
...
+	"strings"
+	"time"
+
+	"github.com/dave/jennifer/jen"
+)
+
+const (
+	releasesAPI   = "https://api.github.com/repos/ossf/osv-schema/releases/latest"
+	schemaURLFmt  = "https://raw.githubusercontent.com/ossf/osv-schema/%s/validation/schema.json"
+	pinnedFile    = "schema-v1.json"
```
kzantow, line 61: It looks like this is made to account for eventually having a v2 (-v1.*, _v1_generated) but these are saved with v1 in a filenames instead of a package name. Would it make sense to output these to `osvmodel/v1` so we could end up with 2 separate models in use concurrently when a v2 lands? I guess I'm just confused about using "v1" in filenames where we could not also add a v2 in the same directory since presumably there would be conflicting Vulnerability types, etc.

wagoodman (approved):

wagoodman:
If we end up not liking the new generated approach over time, the simple fallback is to remove the `main.go` and curate the model by hand like in #3500

grype/db/internal/provider/unmarshal/osvmodel/generate/main.go
```diff
@@ -0,0 +1,814 @@
...
+	}
+	return io.ReadAll(resp.Body)
+}
+
+func atomicWrite(path string, data []byte) error {
+	tmp := path + ".tmp"
+	if err := os.WriteFile(tmp, data, 0o600); err != nil {
+		return err
+	}
+	return os.Rename(tmp, path)
```
kzantow, line 259: Will this succeed if the file already exists?

kzantow (approved):

grype/db/internal/provider/unmarshal/osvmodel/generate/main.go
willmurphyscode: That's a good point!

grype/db/internal/provider/unmarshal/osvmodel/generate/main.go
willmurphyscode: I think so, but it will probably fail on a cross-volume rename, which comes up with temp files in CI a lot. Let me take a look here.

grype/db/internal/provider/unmarshal/osvmodel/generate/main.go
willmurphyscode: I think we should probably just remove `v1` from the filename, and then if we need osv models v2, we just make a new package with v2 in the path? @kzantow what do you think?

kzantow: yeah, just drop the v1 to avoid confusion

grype/db/internal/provider/unmarshal/osvmodel/generate/main.go
willmurphyscode: Since it's just adding `.tmp` to the path, I think it will always be on the same volume and `os.Rename` is correct.

grype/db/internal/provider/unmarshal/osvmodel/generate/main.go
willmurphyscode: Renamed to lack `_v1_`.

wagoodman (approved):

PR #3500, opened 2026-06-09
Add hand-written OSV model package

willmurphyscode:
The osv-scanner project deprecated the exported models.Vulnerability struct grype was using, so the existing alias broke against newer OSV schema versions. Replace it with a hand-maintained osvmodel package covering the OSV v1 schema, swap all OSV strategies and helpers off osv-scanner, and drop the dependency.

New fields land by editing osvmodel/vulnerability.go directly; the strategies that need a field add it in the same PR.

This is a different approach to solve the same issues as https://github.com/anchore/grype/pull/3499 - that PR uses an in-repo _generator_ to make structs like these. This just commits the structs and we have to remember to update them. It's less code but we don't get automatic updates.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -38,7 +38,6 @@ require (
 	github.com/gohugoio/hashstructure v0.6.0
 	github.com/google/go-cmp v0.7.0
 	github.com/google/go-containerregistry v0.21.6
-	github.com/google/osv-scanner v1.9.2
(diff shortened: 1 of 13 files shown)


willmurphyscode:
Closed in favor of https://github.com/anchore/grype/pull/3499

PR #3504, opened 2026-06-12
feat: populate package architecture for matching

willmurphyscode:
Enable matchers to filter on architecture-specific vulnerabilities or fixes.

There are basically three changes here: First, wire up the package collection creator to populate architecture on the packages when an image is scanned or an sbom is loaded in Grype. Second, wire up the architecture package qualifier so that if a vulnerability or fix specifies that it is only applicable to a particular architecture, the matchers will respect this, and third, in the RPM matcher, specify `arch=src` when synthesizing the upstream package to search by (rather than letting it inherit the binary package's architecture).

The qualifier logic is like this: If either the package or the DB record has a blank architecture, the filtering is a noop. If the package has a specific architecture and the DB record has 'src', the package is qualified, because a disclosure or fix in a src RPM for example is assumed to affect things built from it. If the DB record and the package both have specific binary architectures, but they don't match, e.g. x86_64 != aarch64, the package is not qualified and filters out. Architecture qualifiers are canonicalized at match time so, for example, an RPM that has x86_64 architecture is a qualified package on a vuln that affects "amd64" processors (assuming distro and everything else line up). This canonicalization is probably more important for future expansion since a given distro will probably use the same architecture strings as their own package manager.

Hummingbird also has a special "binary-no-arch-specified" architecture that means, "this is not a src RPM". That is so that in Hummingbird data, if we have a package that says, 'upstream=glibc' in its PURL, and we have a glibc vuln against the binary but not the src RPM, we can fitler out this false positive. If a DB record has 'binary-no-arch-specified' as its architecture, any package that has any non-src architecture (or a blank architecture) is qualified, but src architecture is not.

Note that this could use https://github.com/anchore/syft/pull/4977 but there's no need to couple them.


grype/matcher/rpm/matcher.go
```diff
@@ -215,7 +216,12 @@ func (m *Matcher) matchUpstreamPackages(vp vulnerability.Provider, p pkg.Package
 	var ignored []match.IgnoreFilter
 
 	for _, indirectPackage := range pkg.UpstreamPackages(p) {
```
willmurphyscode, line 217: from @wagoodman : Setting the arch to `src` should happen in here instead remembering everywhere.

grype/pkg/purl_provider.go
```diff
@@ -24,7 +24,7 @@ type PURLLiteralMetadata struct {
 }
 
 func purlEnhancers(applyChannel func(*distro.Distro) bool) []Enhancer {
-	return []Enhancer{setUpstreamsFromPURL, setDistroFromPURL(applyChannel)}
+	return []Enhancer{setUpstreamsFromPURL, setArchFromPURL, setDistroFromPURL(applyChannel)}
```
willmurphyscode: From @wagoodman this should be added to types where it matters as well, like https://github.com/anchore/grype/blob/f54a905415fb95fa47274f33241d63f4ca55d351/grype/pkg/rpm_metadata.go#L3 explicitly _instead of_ https://github.com/anchore/syft/pull/4977/changes which is messy because Syft packages don't have necessarily exactly one architecture (e.g. macho binaries).

grype/pkg/purl_provider.go
```diff
@@ -100,3 +100,19 @@ func distroFromPURL(purl packageurl.PackageURL) (d *distro.Distro) {
 
 	return d
 }
+
+func setArchFromPURL(out *Package, purl packageurl.PackageURL, _ syftPkg.Package) {
```
willmurphyscode: Additionally, we should set this in metadata things: https://github.com/anchore/grype/blob/main/grype/pkg/apk_metadata.go#L12 for example written to here (and analogous changes for the RPM stuff.)

We _might_ not need a PURL enhancer if we do it that way.

willmurphyscode: Might be worth setting arch in here: https://github.com/anchore/syft/blob/main/syft/format/internal/backfill.go#L37 - otherwise we do need the enhancer in grype.

grype/pkg/qualifier/architecture/qualifier.go
```diff
@@ -34,10 +34,63 @@ func (r architectureQualifier) Arch() string {
...
+// differently by different tooling: RPM/GNU/Alpine say "x86_64"/"aarch64", while Debian/Go/OCI
+// say "amd64"/"arm64" (and Windows "x64"). Since a package carries its package manager's
+// spelling and an advisory carries its source's spelling, an exact string compare would
+// wrongly drop a real match when those dialects differ.
+//
+// Only genuine CPU architectures are folded. The role markers ("src", binary-no-arch) and
+// arch-independent values (rpm "noarch", deb "all") are not architectures and pass through
+// unchanged. The canonical token chosen is arbitrary (we use the Go/OCI spelling); only
+// equivalence matters.
+func canonicalArch(a string) string {
```
willmurphyscode: We should consider putting this in the database instead of the code.

grype/pkg/qualifier/architecture/qualifier.go
```diff
@@ -34,10 +34,63 @@ func (r architectureQualifier) Arch() string {
...
+//     keeps a binary record from matching through the synthesized "src" upstream package,
+//     so a sibling binary built from the same source isn't falsely matched.
+//
+// A package with no recorded arch is inert (older DBs / inputs without arch), as is an entry
+// with no arch (treated as "any"); neither direction filters when we cannot decide.
+//
+// Concrete arches are compared in canonical form (see canonicalArch) so a package and an
+// entry that name the same architecture in different ecosystem dialects still match — e.g.
+// an "amd64" deb package against an "x86_64" RPM-dialect advisory.
+func (r architectureQualifier) Satisfied(p pkg.Package) (bool, error) {
```
willmurphyscode, line 71: @wagoodman Do you think this should have been a set contains operation? Like the architecture qualifier has a `Set[string]` of aliases and the package is true if it's satisfied? I could go either way (and there are some databases built today with the string field, so we may be stuck with it, or with adopting a custom deserializer or something ...)

wagoodman: after discussing, lets not do this since there isn't a strict need now and this means we need to re-coordinate the writers to write something semantically different into the DB and think through how old clients would behave.

grype/pkg/qualifier/architecture/qualifier.go
```diff
@@ -18,26 +18,185 @@ import (
...
 }
 
-// Arch returns the stored architecture value (e.g. "src", "x86_64",
-// "binary-no-arch-specified", or "" if unset).
-func (r architectureQualifier) Arch() string {
-	return r.arch
+// defaultAliases maps each non-canonical architecture spelling to its canonical token. The
+// canonical token is arbitrary (we use the Go/OCI spelling); only equivalence matters, so a
+// canonical spelling (e.g. "amd64") is intentionally absent and resolves to itself.
+var defaultAliases = map[string]string{
```
willmurphyscode, line 182: Does `"all" : "noarch"` belong in here?

grype/matcher/rpm/almalinux.go
```diff
@@ -89,7 +89,6 @@ func almaLinuxMatchesWithUpstreams(provider result.Provider, binaryPkg pkg.Packa
 			search.ByPackageName(upstreamPkg.Name),
 			search.ByDistro(rhelCompatibleDistro),
 			internal.OnlyQualifiedPackages(upstreamPkg),
-			internal.SourceOrUnspecifiedArch(),
```
willmurphyscode, line 92: I thought this helper was used only in hummingbird. Why the alma diff?

grype/matcher/rpm/matcher.go
```diff
@@ -215,7 +215,11 @@ func (m *Matcher) matchUpstreamPackages(vp vulnerability.Provider, p pkg.Package
...
-		indirectMatches, ignores, err := m.findMatches(provider, indirectPackage, internal.SourceOrUnspecifiedArch())
+		// An rpm's upstream is its source rpm, so tag the synthesized package "src". The
```
willmurphyscode, line 218: Clean up this comment as this is now done in the helper.

grype/db/v6/blobs.go
```diff
@@ -150,10 +150,11 @@ type PackageQualifiers struct {
...
-	// arch. Upstream-indirected RPM matches drop entries tagged with any arch other than
-	// "src", so providers that disclose at binary granularity (e.g. hummingbird CSAF VEX)
-	// avoid FP-matching unrelated sibling binaries built from the same source. See
-	// pkg/qualifier/architecture for the constants.
+	// arch. At match time the architecture qualifier compares this value against the scanned
+	// package's arch (see pkg/qualifier/architecture, Satisfied): a "binary-no-arch-specified"
+	// entry matches any binary package but rejects the rpm matcher's synthesized "src" upstream,
+	// so providers that disclose at binary granularity (e.g. hummingbird CSAF VEX) avoid
+	// FP-matching unrelated sibling binaries built from the same source.
 	Architecture *string `json:"architecture,omitempty"`
```
willmurphyscode, line 158: This already shipped as a string, but maybe missed opportunity to do a `Set[string]` so that we could check "intersection of package arch with vuln arch is not empty" instead of equality.

grype/pkg/qualifier/architecture/qualifier.go
```diff
@@ -18,26 +18,185 @@ import (
...
+	"x86_64":  "amd64",
+	"x64":     "amd64",
+	"aarch64": "arm64",
```
willmurphyscode, line 185: What about like `arm64/v8`? That's a thing that can be passed to `--platform` on docker CLI for example.

Also "architecture" and "platform" are not exactly this same thing and this conflates them?

grype/db/v6/models.go
```diff
@@ -533,6 +534,22 @@ type PackageSpecifierOverride struct {
 	ReplacementEcosystem *string `gorm:"column:replacement_ecosystem;primaryKey"`
 }
 
+// ArchitectureAlias folds an architecture spelling onto a canonical token shared by every
+// dialect spelling of the same CPU architecture (e.g. rpm "x86_64" and deb/OCI "amd64" both
+// fold to "amd64"). The architecture qualifier reads this table at match time so cross-dialect
+// arch comparisons don't drop real matches; the data lives in the DB rather than the grype
+// binary so the folding can change without a release. See pkg/qualifier/architecture.
+type ArchitectureAlias struct {
```
willmurphyscode, line 542: Is this complete? Do we need an "is platform independent" bool column or something? Right now the source code knows that `noarch` and `all` mean "this thing applies across all architectures" but the database doesn't?

We think that this case be added later if need be, and that it is probably not necessary.

grype/pkg/upstream_package.go
```diff
@@ -49,7 +49,21 @@ func UpstreamPackages(p Package) (pkgs []Package) {
 		}
 		tmp.CPEs = updatedCPEs
 
+		setSrcArchOnUpstream(&tmp)
```
wagoodman, line 52: This overrides any architectures that could be explicitly provided. I see that this is intentional, but does this consider that there might be instances where the vuln data is more specific than `src`? (say with hummingbird) .... there is a good chance this is a non-issue

wagoodman: Well actually, doesn't oracle provider per-binary-per-arch NERVA's?

grype/pkg/qualifier/architecture/qualifier.go
wagoodman, line 182: just want to state clearly but this isn't a comment to address, but instead a specific feature designed in: the DB is authoritative over clients, so newer clients that have old DBs might run into issues, but in these cases we expect folks to update their dbs. This is a nice feature of using the DB since DBs ship nightly instead of needing to upgrade the client.

grype/pkg/rpm_metadata.go
```diff
@@ -3,4 +3,10 @@ package pkg
 type RpmMetadata struct {
 	Epoch           *int    `json:"epoch" cyclonedx:"epoch"`
 	ModularityLabel *string `json:"modularityLabel" cyclonedx:"modularityLabel"`
+	// Arch is the package's architecture in rpm's native spelling ("x86_64", "aarch64",
+	// "noarch", ...), copied verbatim from the syft metadata. It is read at match time by
+	// pkg/qualifier/architecture to match a package against the architecture a vulnerability
+	// entry applies to. Cross-dialect aliases (amd64↔x86_64, ...) are reconciled there, so
+	// this is intentionally a raw string rather than an enum.
+	Arch string `json:"architecture" cyclonedx:"architecture"`
```
wagoodman, line 11: nit, this could have `omitempty`

grype/pkg/upstream_package.go
willmurphyscode: So the idea here is that the _upstream_ package is always a source RPM. This actually fixes false positives in Hummingbird and is an intentional and necessary part of this PR.

Consider this scan:

```
syft -q -o json quay.io/hummingbird/postgresql:latest@sha256:6eefb14bced346a8583011f84dd71de8249af74c060369be6ad73891b2a768f0 | jq '.artifacts[] | .purl' | rg glibc
```

which produces:

```
"pkg:rpm/redhat/glibc@2.42-10.1.hum1?arch=aarch64&distro=hummingbird-20251124&upstream=glibc-2.42-10.1.hum1.src.rpm"
"pkg:rpm/redhat/glibc-common@2.42-10.1.hum1?arch=aarch64&distro=hummingbird-20251124&upstream=glibc-2.42-10.1.hum1.src.rpm"
"pkg:rpm/redhat/glibc-langpack-en@2.42-10.1.hum1?arch=aarch64&distro=hummingbird-20251124&upstream=glibc-2.42-10.1.hum1.src.rpm"
```

And this [CSAF VEX](https://security.access.redhat.com/data/csaf/v2/vex-feed/2026/cve-2026-5450.json):

``` json
      "product_status": {
        "fixed": [
          "hummingbird-1:glibc-0:2.42-12.hum1", // binary rpm
          "hummingbird-1:glibc-0:2.42-12.hum1.src", // src rpm, removed intentionally in datat processing
         ...
```

In this case, `glibc-langpack-en` has the upstream `glibc`, and the _binary_ glibc is vulnerable. However, before this change, `glibc-langpack-en` could be reported as vulnerable because its upstream has **the same name** as a vulnerable binary. However, that would be considered an FP. This is a change to allow us to consider upstream matching still, but only if the upstream package is a src rpm, so that marking a _binary_ rpm that has the same name as a bunch of source rpm's upstream (like `glibc`) doesn't produce a wave of FPs.

willmurphyscode: > Well actually, doesn't oracle provider per-binary-per-arch NERVA's?

I don't think anyone has architecture on their src RPMs, and upstream will always be a src rpm? But I confess I don't completely understand the comment.

willmurphyscode: @wagoodman I don't think this matters on oracle. the OVAL files seem to list every binary RPM anyway, and I don't think the notion of a non src `upstream` makes sense. This only sets arch on upstreams.

wagoodman (approved):

PR #3509, opened 2026-06-17
feat: add go symbol matching qualifier for better go vuln matching

### LINKED ISSUE #2960: Feature Request: Lightweight Reachability Analysis to Reduce False Positives

I propose adding an optional, lightweight static analysis feature directly into Grype.

This feature would work as follows:
1. After matching a vulnerability, Grype would identify the specific function(s) or method(s) known to be vulnerable for that CVE.
2. It would then perform a simple, text-based search (effectively a grep) across the scanned source code to determine if these vulnerable functions are actually called.
3. If no calls to the vulnerable functions are found, the vulnerability would be either suppressed or flagged with a specific tag like unreachable.
This could be controlled by a new flag, for instance, `--enable-reachability-analysis`.

**Why is this needed**:

The primary goal is to drastically reduce the high number of false positive alerts generated by unreachable vulnerabilities, which is a major source of alert fatigue for developers.

My research on this topic shows that this simple, low-cost static analysis method is remarkably effective. By checking for actual function calls, it's possible to filter out approximately 60% of the vulnerabilities typically reported by SBOM-based scanners.

This provides immense value by:

1. Dramatically improving the signal-to-noise ratio of scan results.
2. Allowing development and security teams to focus their limited time on vulnerabilities that pose a tangible risk.
3. Making Grype's results significantly more actionable out of the box, without requiring complex external tools or manual verification for a large portion of findings.

While this method isn't a perfect replacement for deep call-graph analysis (e.g., it may miss dynamic calls), a `~60%` reduction in noise represents a massive and practical improvement for the user experience.

**Additional context**:

Here is a conceptual example to illustrate the feature:
1. A Python project uses the library `requests==2.25.0`.
2. Grype correctly identifies that this version is vulnerable to a CVE related to the `requests.auth.HTTPDigestAuth` function.
3. However, the project's source code only ever calls `requests.get()` and never imports or uses the vulnerable HTTPDigestAuth function.
4. With the proposed --enable-reachability-analysis flag, Grype would scan the source, see that HTTPDigestAuth is not called, and automatically suppress this finding or mark it as "unreachable".

This would prevent a developer from spending time investigating a vulnerability that poses no real threat to their specific application.

Thank you for considering this feature. I believe it would be a powerful enhancement to an already excellent tool. I'm happy to discuss my research findings in more detail.

### LINKED ISSUE #3511: Deduplicate Go matches that are aliases of each other (same CVE reported under both a `govulndb` GO-* ID and its GHSA)
### What happened

When a Go module is affected by a vulnerability that exists in the DB under multiple provider IDs
that are aliases of one another (a `govulndb` GO-* ID and its `github` GHSA, both mapping to the
same CVE), grype reports it as **two separate matches** for the one underlying vulnerability.

Example: [grafana](https://pkg.go.dev/github.com/grafana/grafana) `@v12.0.0` is genuinely affected
by [CVE-2025-41115](https://nvd.nist.gov/vuln/detail/CVE-2025-41115), and grype reports it twice:

```sh
grype pkg:golang/github.com/grafana/grafana@v12.0.0
```
```
NAME                        INSTALLED  FIXED IN  TYPE       VULNERABILITY        SEVERITY
github.com/grafana/grafana  v12.0.0    12.0.7    go-module  GHSA-w62r-7c53-fmc5  Critical
github.com/grafana/grafana  v12.0.0              go-module  GO-2025-4153         Critical
```

Both rows are the same vulnerability, and grype itself records the alias link:

- [GHSA-w62r-7c53-fmc5](https://github.com/advisories/GHSA-w62r-7c53-fmc5) -> `relatedVulnerabilities: [CVE-2025-41115]`, fixedIn `12.0.7`
- [GO-2025-4153](https://pkg.go.dev/vuln/GO-2025-4153) -> `relatedVulnerabilities: [CVE-2025-41115, GHSA-w62r-7c53-fmc5]`, fixedIn (none)

So a single CVE ([CVE-2025-41115](https://nvd.nist.gov/vuln/detail/CVE-2025-41115)) surfaces as two
findings, and the two rows even **disagree on the fix**: the GHSA row says fixed in `12.0.7`, the GO
row says nothing. This is confirmed at the DB layer with `grype db search` (public Anchore DB):

```sh
grype db search --pkg github.com/grafana/grafana --vuln GHSA-w62r-7c53-fmc5
```
```
VULNERABILITY        PACKAGE                     NAMESPACE           VERSION CONSTRAINT
GHSA-w62r-7c53-fmc5  github.com/grafana/grafana  github:language:go  >=12.0.0,<12.0.7   (fix 12.0.7)
GHSA-w62r-7c53-fmc5  github.com/grafana/grafana  github:language:go  >=12.1.0,<12.1.4   (fix 12.1.4)
GHSA-w62r-7c53-fmc5  github.com/grafana/grafana  github:language:go  >=12.2.0,<12.2.2   (fix 12.2.2)
```
```sh
grype db search --pkg github.com/grafana/grafana --vuln GO-2025-4153
```
```
VULNERABILITY  PACKAGE                     NAMESPACE             VERSION CONSTRAINT
GO-2025-4153   github.com/grafana/grafana  govulndb:language:go  >=1.9.2-0.20250310110405-e6fdb746f235   (no fix)
```

### What you expected to happen

The same underlying vulnerability should be reported **once**, not as two separate findings under
two aliased IDs.

### How to reproduce it

Public Anchore DB, no auth:

```sh
grype pkg:golang/github.com/grafana/grafana@v12.0.0 -o json \
  | jq '[.matches[] | select([.relatedVulnerabilities[].id] | index("CVE-2025-41115"))]
        | map({id: .vulnerability.id, fixedIn: .vulnerability.fix.versions})'
```

```json
[
  {
    "id": "GHSA-w62r-7c53-fmc5",
    "fixedIn": [
      "12.0.7"
    ]
  },
  {
    "id": "GO-2025-4153",
    "fixedIn": []
  }
]
```

Two entries ([GHSA-w62r-7c53-fmc5](https://github.com/advisories/GHSA-w62r-7c53-fmc5) and
[GO-2025-4153](https://pkg.go.dev/vuln/GO-2025-4153)) for the one
[CVE-2025-41115](https://nvd.nist.gov/vuln/detail/CVE-2025-41115), disagreeing on the fix: the GHSA
reports `12.0.7`, the GO record reports nothing.

### Why this matters / impact

- **Inflated and inconsistent counts:** one vulnerability counted twice, with conflicting fix
  metadata across the two rows (one says fixed in `12.0.7`, the other says no fix), which is
  confusing to triage and skews totals.

### Related

Sibling issue #3510 covers a separate symptom of the same underlying `govulndb` records: a rangeless
record reporting a finding as unfixed even though the aliased GHSA carries the fix.

### Additional context

- The duplication became visible after the recently added `govulndb` provider (anchore/vunnel#1172,
  merged 2026-06-03) introduced a second set of IDs (GO-*) for Go advisories already present as
  GHSAs.

### Environment

- grype: v0.114.0-5-g55b3304b (commit 55b3304b, built from source, main, 2026-06-16)
- DB: public Anchore database, schema v6.1.7 (e.g. `vulnerability-db_v6.1.7_2026-06-16T01:05:09Z`,
  from https://grype.anchore.io/databases/v6/latest.json)
- Aliases referenced: [CVE-2025-41115](https://nvd.nist.gov/vuln/detail/CVE-2025-41115) / [GHSA-w62r-7c53-fmc5](https://github.com/advisories/GHSA-w62r-7c53-fmc5) / [GO-2025-4153](https://pkg.go.dev/vuln/GO-2025-4153)

spiffcs:
## Go symbol matching: stdlib and golang.org/x advisories match correctly based on new symbol data; merge govulndb symbol data onto GHSA records

The [Go Vuln DB] (https://vuln.go.dev/) is the only source for Go standard library advisories. It is also an authoritative source for a share of golang.org/x/* advisories.

Grype recently started hosting GoVulnDB records in its DB (#3527, #3534, #3527).

Keeping these records exposed without symbol matching lead to matching inaccuracies. We also would see inaccuracies from cases in GHSA where we did not have symbol data and only matched at the module/version range level.

## PR Features                                                                                                                                                                                                   
- stdlib and golang.org/x/* records now match on symbols
- grype catalogs Go binary function symbols by default (syft symbol capture enabled in getProviderConfig),
- govulndb's per-symbol reachability is stored as a go_imports qualifier, and only matches when a vulnerable symbol is actually present in the binary  
- Stdlib packages always stay on the GO records (nothing else covers them); 
- x/* advisories without a GHSA counterpart are also kept. 
- Packages without symbol evidence (pre-built SBOMs) keep today's module-granularity behavior

## Data Decisions
Where a GHSA covers the same module, the GHSA becomes the single, symbol-enriched record. 

- Each GHSA affected package matching a GO record's module is patched with the go-imports qualifier.
- Patching is decided per exact match on `affected[].package.name`
- The GO record is written only if packages remain that were NOT found in the GHSA.
- Ranges: the GHSA's win by default  
- Ranges exception: a GHSA range pinned to the pseudo-version of the fix commit can never match a real tagged release (v5.21.1 sorts above v0.0.0-… — a guaranteed FN). When the GO record carries that exact pseudo-version in its standard range plus exactly one custom range in `ecosystem_specific.custom_ranges` we patch in the custom_range
- Every GHSA amendment is recorded on the blob's modifications audit trail ({url, changes[]) changes can be symbol additions and range replacements.

## Schema: 
- DB model 6.1.9 adds GoImports to package qualifiers
- db search 1.1.7 adds go_imports
- db search vuln 1.0.6 adds modifications

### Known limitations / follow-ups (tracked separately)                                                                                                                                                                  
- Non-semver version strings (~20 govulndb custom_ranges strings plus 14 + 45 on the GHSA side


grype/db/v6/blobs.go
```diff
@@ -160,6 +174,21 @@ type PackageQualifiers struct {
 	// RootIO indicates that the vulnerability applies only to Root IO packages (packages with Root IO fixes).
 	// When true, standard packages will not match this vulnerability (NAK pattern).
 	RootIO *bool `json:"rootio,omitempty"`
+
+	// GoImports lists the packages and symbols within an affected Go module that contain the vulnerability
+	// (from govulndb's ecosystem_specific.imports). When set, packages carrying binary symbol evidence only
+	// match if at least one of the listed symbols is present in the binary.
+	GoImports []GoImport `json:"go_imports,omitempty"`
```
spiffcs, line 181: In the grype report what of the symbol data should we end up including? Just what the intersection was that caused the match or the full table for the package?

grype/db/v6/build/writer.go
```diff
@@ -28,6 +28,15 @@ type writer struct {
 	// which normalize casing so the invariant can't drift.
 	severityCache map[string]db.Severity
 
+	// goGHSAEntries holds back Go-ecosystem GHSA entries (keyed by lowercase
+	// GHSA ID, with goGHSAOrder preserving arrival order for deterministic
+	// writes) and govulndbEntries holds back GO-* entries, so the govulndb↔GHSA
+	// overlap can be reconciled at Close time regardless of provider processing
+	// order. See govulndb_merge.go.
```
spiffcs: I might take another stab at this write strategy tomorrow

grype/db/v6/build/govulndb_merge.go
```diff
@@ -0,0 +1,365 @@
...
+	existing := make(map[string]bool)
+	if blob.Qualifiers != nil {
+		for _, imp := range blob.Qualifiers.GoImports {
+			existing[imp.Path] = true
+		}
+	}
+	var added []db.GoImport
+	for _, imp := range imports {
+		if existing[imp.Path] {
+			continue
```
wagoodman, line 430: I doubt this is likely, but I'll mention it anyway: if there are two go records aliasing the same GHSA that list the same import path with **different** symbol sets then the second set is silently dropped instead of being combined.

grype/db/v6/build/writer.go
```diff
@@ -104,6 +114,18 @@ func (w *writer) Write(entries ...data.Entry) error {
...
+	}
+
+	return w.writeEntryToBatch(entry)
+}
+
+// writeEntryToBatch fills in missing severity and adds the entry to the write
+// batches. Held govulndb/GHSA entries also come back through here at Close
+// time, after reconciliation (see govulndb_merge.go), so severity fill always
+// runs with the complete NVD cache.
+func (w *writer) writeEntryToBatch(entry transformers.RelatedEntries) error {
```
wagoodman, line 124: why hold all of the go and ghsa records? We have a mechanism for ordering providers one before another (today we have NVD before others). I feel that we could write ghsa's as usual then patch overlapping ghsa records when we write the go vuln records --which doesn't requiring holding any/all records in memory and write them at close. 

I did not look at the *provider caches that we have and how they are affected. 

Was there specific frictions/reasons why to hold writes for all records?

grype/db/v6/build/transformers/osv/testdata/GO-2026-4610.json
```diff
@@ -0,0 +1 @@
...r","symbols":["defaultSystemPluginDirs"]}],"custom_ranges":[{"type":"ECOSYSTEM","events":[{"introduced":"19.03.0+incompatible"}]}]}},{"package":{"name":"github.com/docker/compose","ecosystem":"Go"},"ranges":[{"type":"SEMVER","events":[{"introduced":"0"}]}],"ecosystem_specific":{}},{"package":{"name":"github.com/docker/compose/v2","ecosystem":"Go"},"ranges":[{"type":"SEMVER","events":[{"introduced":"2.31.0"}]}],"ecosystem_specific":{}},{"package":{"name":"github.com/docker/compose/v5","ecosystem":"Go"},"ranges":[{"type":"SEMVER","events":[{"introduced":"0"},{"fixed":"5.1.0"}]}],"ecosystem_specific":{}}],"references":[{"type":"ADVISORY","url":"https://github.com/docker/cli/security/advisories/GHSA-p436-gjf2-799p"},{"type":"FIX","url":"https://github.com/docker/cli/commit/13759330b1f7e7cb0d67047ea42c5482548ba7fa"},{"type":"FIX","url":"https://github.com/docker/cli/pull/6713"},{"type":"FIX","url":"https://github.com/docker/compose/pull/12300"},{"type":"WEB","url":"https://docs.docker.com/desktop/release-notes"},{"type":"WEB","url":"https://www.zerodayinitiative.com/advisories/ZDI-CAN-28304"}],"database_specific":{"url":"https://pkg.go.dev/vuln/GO-2026-4610","review_status":"REVIEWED"}}
```
willmurphyscode: Is it worth indenting / formatting this stuff? it's impossible to read dense JSON in a diff.

spiffcs: Yea let me pull Alex's test changes and I can push a commit that formats these

grype/db/v6/build/transformers/osv/transform_go_vuln_db.go
```diff
@@ -115,25 +100,100 @@ func govulndbReferences(vuln unmarshal.OSVVulnerability) []db.Reference {
...
 	}
 	var aphs []db.AffectedPackageHandle
+	// replacements carries the pseudo-version reconciliation context per package, keyed by the
+	// blob pointer since sorting below reorders the handles
+	replacements := make(map[*db.PackageBlob]transformers.GoVulnDBAffectedPackage)
 	for _, affected := range vuln.Affected {
-		if !govulndbEmits(affected.Package.Name) {
-			continue
+		var qualifiers *db.PackageQualifiers
+		if imports := govulndbImports(affected, vuln.ID); len(imports) > 0 {
```
willmurphyscode, line 113: Is it worth also capturing an architecture qualifier, e.g. https://vuln.go.dev/ID/GO-2022-0187.json is only affected on `amd64`. That's probably out of scope here but i wanted to check what others think. My vote would be to add it as a fast follow but still wanted to mention it.

grype/db/v6/build/transformers/osv/testdata/GO-2026-4610.json
spiffcs: These are now updated so it's easy for a reviewer or future author to see the JSON diff

grype/db/v6/build/writer.go
spiffcs: We talked about this on a sync, but to record the decision here:
- We're not holding all the ghsa records, just the ones from the Go ecosystem
- Write once was deemed a better path than write then patch because of the additional complexity of adding new patch code. 

If we run into issues with write once we can come back on a follow up to add the PATCH methods to the db builder/reconciler

grype/db/v6/build/transformers/osv/transform_go_vuln_db.go
spiffcs: I'll write up an issue and do this as a follow up to this PR

cmd/grype/cli/commands/root.go
```diff
@@ -295,6 +297,35 @@ func warnWhenDistroHintNeeded(pkgs []pkg.Package, context *pkg.Context) {
...
+
+	if withoutSymbols == 0 {
+		return ""
+	}
+
+	return fmt.Sprintf("%d Go binary package(s) have no function symbols; matching falls back to module "+
+		"granularity and may report false positives. Regenerate the SBOM with symbol capture enabled "+
+		"(SYFT_GOLANG_CAPTURE_SYMBOLS=all syft ...) for more precise Go results.", withoutSymbols)
+}
+
```
wagoodman: This isn't right; the lack of symbols is not proof of `syft | grype` usage. If we're going to populate a warning like this it should be in the package decode path (deeper in execution, at a lower level of abstraction), not in the `cmd` package.

I'm able to trigger this message without piping syft output to grype.

grype/pkg/golang_metadata.go
```diff
@@ -9,6 +9,11 @@ type GolangBinMetadata struct {
 	H1Digest          string        `json:"h1Digest,omitempty" cyclonedx:"h1Digest"`
 	MainModule        string        `json:"mainModule,omitempty" cyclonedx:"mainModule"`
 	GoCryptoSettings  []string      `json:"goCryptoSettings,omitempty" cyclonedx:"goCryptoSettings"`
+
+	// Symbols are the function symbols compiled into the binary, grouped by the import path of the package
+	// that owns them; each value is the list of symbol names local to that package (import path prefix
+	// stripped). Populated only when syft cataloged the binary with symbol capture enabled.
+	Symbols map[string][]string `json:"symbols,omitempty"`
```
wagoodman: this is a blocker: I'm seeing symbols in the JSON output. We don't want to include this information in the raw artifact. We MIGHT distill it down to if there are symbols present or not, but all of that low level information is not useful and will clog up the output significantly.

grype/pkg/qualifier/gosymbols/qualifier.go
```diff
@@ -0,0 +1,161 @@
...
+func presentSymbols(p pkg.Package) (map[string]map[string]struct{}, bool) {
+	m, ok := p.Metadata.(pkg.GolangBinMetadata)
+	if !ok || len(m.Symbols) == 0 {
+		return nil, false
+	}
+	present := make(map[string]map[string]struct{}, len(m.Symbols))
+	for importPath, locals := range m.Symbols {
+		set := make(map[string]struct{}, len(locals))
+		for _, local := range locals {
+			set[normalizeSymbol(local)] = struct{}{}
```
wagoodman: doing normalization in a nested for loop here could be expensive (I did not pprof this, but worth trying) and probably is not necessary. We could do this in the grype.pkg.Provider path once for every symbol and not within the qualifier search.

spiffcs: I think this change does a bit better and is less expensive than what we had. It builds the index per package when we call: `golangMetadataFromPkg`

Maybe we can sync if there was a different direction you had in mind.

(https://github.com/anchore/grype/pull/3509/commits/cfd7b163d169cad21ad7e230485be8052fde33df)

grype/pkg/golang_metadata.go
spiffcs: This removes Symbols from the json output. We can talk tomorrow about if there is any distilling we want to do onto the package, but I think that would just confuse the user given we already have `matchedSymbols` for match details:
```
  {
    "type": "exact-direct-match",
    "matcher": "go-module-matcher",
    "searchedBy": { "language": "go", "package": { "name": "github.com/tidwall/gjson", "version": "v1.9.2" } },
    "found": {
      "vulnerabilityID": "GHSA-ppj4-34rq-v8j9",
      "versionConstraint": "<1.9.3 (go)",
      "matchedSymbols": "github.com/tidwall/gjson.Get, github.com/tidwall/gjson.Result.Get, github.com/tidwall/gjson.parseObject, github.com/tidwall/gjson.queryMatches"
    }
  }
```
[d5aa272](https://github.com/anchore/grype/pull/3509/commits/d5aa272743645cf5a3164b7b605635543e3fa5ca)

cmd/grype/cli/commands/root.go
spiffcs: [b934c85](https://github.com/anchore/grype/pull/3509/commits/b934c85e6a772d4a1ce91569112b77fbe44b654a)

We can sync in the AM, but I think this is correct now where we do a single warning.

spiffcs:
I noticed on this change that there was a gap in the schema updates. I just opened https://github.com/anchore/grype/pull/3574 which probably should be merged first. Then I can rebase these changes on top of those to get the correct order of schema changes accounted.

wagoodman (approved):

Issue #3510, opened 2026-06-17
False positive: rangeless `govulndb` (GO-*) Go matches reported as unfixed when an aliased GHSA in the same DB carries the fix


philroche:
### What happened

For some Go advisories the recently added `govulndb` provider stores an affected-package record
with **no version range and no fix**. A rangeless record matches *every* version of the module, so
grype reports the module as vulnerable no matter how new it is, even when the **same vulnerability**
is also in the DB via its GHSA (github provider) with a correct fix version that would exclude the
installed version.

The two records are the same vulnerability, linked by alias, and both live in the **same database**.
This is visible directly with `grype db search` (public Anchore DB, no auth):

```sh
grype db search --pkg github.com/aws/aws-sdk-go --vuln GHSA-7f33-f4f5-xwgw
```
```
VULNERABILITY        PACKAGE                    ECOSYSTEM  NAMESPACE           VERSION CONSTRAINT
GHSA-7f33-f4f5-xwgw  github.com/aws/aws-sdk-go  go-module  github:language:go  <1.34.0
```
```sh
grype db search --pkg github.com/aws/aws-sdk-go --vuln GO-2022-0635
```
```
VULNERABILITY  PACKAGE                    ECOSYSTEM  NAMESPACE
GO-2022-0635   github.com/aws/aws-sdk-go  go-module  govulndb:language:go
```

The `github` record has `VERSION CONSTRAINT <1.34.0` (JSON: `fix.version: 1.34.0`, `state: fixed`);
the `govulndb` record has an empty constraint (its blob is just `{"cves":[...]}`, no `ranges`).

[GO-2022-0635](https://pkg.go.dev/vuln/GO-2022-0635) == [CVE-2020-8912](https://nvd.nist.gov/vuln/detail/CVE-2020-8912) == [GHSA-7f33-f4f5-xwgw](https://github.com/advisories/GHSA-7f33-f4f5-xwgw),
patched in `github.com/aws/aws-sdk-go` `v1.34.0` (the GHSA lists "Patched versions: 1.34.0"). grype
already records the alias relationship (`relatedVulnerabilities` lists the CVE and GHSA) and the
GHSA's fix range is in the DB, yet grype reports the rangeless GO match as unfixed against `v1.55.8`,
roughly 40 minor releases past the fix.

### How to reproduce it

Public Anchore DB, single up-to-date Go module, no auth, no custom data:

```sh
grype pkg:golang/github.com/aws/aws-sdk-go@v1.55.8
```
[...]

PR #3514, opened 2026-06-17
fix(go-vulndb): merge in custom ranges when applicable

willmurphyscode:
Go OSV data has the standard OSV ranges, but also has custom ranges. Packages that do non-standard go versioning, for example by pushing a tag that starts with v2.x.y but not adding a /v2 in their module path, get relegated to the custom ranges field. In other words, there is one affected range for "strictly go versioning compliant versions" and another set of ranges for "version control tags that are not perfectly compliant".

Previously, Grype just ignore this second bucket of version ranges, resulting in a meaningful number of false positives. Fix this by, if custom ranges are present, stripping out unbounded (e.g. > 0) ranges from the regular ranges and then unioning the two range sets.

https://pkg.go.dev/vuln/GO-2025-4153 is a good example of a record that is a false positive today that this branch fixes.

See https://github.com/anchore/grype/issues/3510


grype/db/v6/build/transformers/osv/transform_go_vuln_db.go
```diff
@@ -122,6 +153,68 @@ func govulndbAffectedPackages(vuln unmarshal.OSVVulnerability) []db.AffectedPack
...
+		return nil
+	}
+	raw, ok := affected.EcosystemSpecific["custom_ranges"]
+	if !ok {
+		return nil
+	}
+
+	// custom_ranges arrives as []any of map[string]any; round-trip through JSON
+	// to reuse the osvmodel.Range shape and the shared range-normalization path.
+	encoded, err := json.Marshal(raw)
```
kzantow: I think this is fine given where this is being used; it would be nice to not have to re-encode stuff.

It looks like mapstructure (which is already in the codebase) might work without re-serializing:
```
config := &mapstructure.DecoderConfig{
		Metadata: nil,
		Result:   &customRanges,
		TagName:  "json", // Instructs mapstructure to look at `json:"..."` tags
	}
```

grype/db/v6/build/transformers/osv/transform_go_vuln_db.go
```diff
@@ -110,6 +111,36 @@ func govulndbAffectedPackages(vuln unmarshal.OSVVulnerability) []db.AffectedPack
...
+
+		// ecosystem_specific.custom_ranges is go.dev's escape hatch for the
+		// "+incompatible" case: when a module ships vN tags for N>1 but never
+		// moved its path to /vN, go.dev cannot map the source advisory's
+		// versions onto canonical Go module semver. Its presence is the signal
+		// that the standard `ranges` are an unreliable best-effort artifact, so
+		// custom_ranges (which mirrors the advisory's own affected windows, per
+		// the record's details text) is authoritative.
+		//
+		// We do not blindly replace the standard ranges, though: the two often
```
kzantow: I _really, really_ would love succinct comments (this goes for all places in this PR). These multi-paragraph things are a pretty big time waster for humans when most of the time a 1 or 2 sentence thing could suffice to understand why something is in the code. It was pretty rare to have multi-paragraph comments before Ai, from my experience; even for Claude when there's more text than is necessary, that's bad.

willmurphyscode: Yep will address.

grype/db/v6/build/transformers/osv/transform_go_vuln_db.go
```diff
@@ -122,6 +115,124 @@ func govulndbAffectedPackages(vuln unmarshal.OSVVulnerability) []db.AffectedPack
...
+		return nil
+	}
+	var customRanges []osvmodel.Range
+	if err := json.Unmarshal(encoded, &customRanges); err != nil {
+		return nil
+	}
+
+	var ranges []db.Range
+	for _, r := range customRanges {
+		bounded, openEnded := govulndbRanges(r, "go")
```
kzantow: It looks like other places call `govulndbRangeType(r.Type)`, should this do the same thing?

grype/db/v6/build/transformers/osv/transform_go_vuln_db.go
```diff
@@ -106,9 +95,13 @@ func govulndbAffectedPackages(vuln unmarshal.OSVVulnerability) []db.AffectedPack
 	}
 	var aphs []db.AffectedPackageHandle
 	for _, affected := range vuln.Affected {
-		var ranges []db.Range
-		for _, r := range affected.Ranges {
-			ranges = append(ranges, getGrypeRangesFromRange(r, govulndbRangeType(r.Type))...)
+		ranges := govulndbAffectedRanges(affected)
+		if len(ranges) == 0 {
```
kzantow: I understand why using custom ranges we want to omit the unbounded ranges, but it looks like we might have `introduced: 0` and no end version, because something isn't fixed. I think those will also be omitted with the current implementation. Generally, if there are records in the grype DB without ranges at all, we consider those "vulnerable for all versions", it looks like we will be dropping those now, is that right? I'm not really sure which is better. Definitely agree that we should not use them when there are custom ranges.

kzantow (approved):
LGTM

PR #3519, opened 2026-06-22
feat: add vulnerable version ranges to CycloneDX output

### LINKED ISSUE #3512: Include vulnerable ranges in CycloneDX output format

Include [vulnerable ranges](https://cyclonedx.org/docs/1.7/json/#vulnerabilities_items_affects_items_versions_items_range) in CycloneDX ouptut format per each vulnerability.

**Why is this needed**:

This helps customers to identify vulnerabilities with available fix, differentiate vulnerabilities with and without fixes, and automate remediation steps.

I specifically would like to notify my system administrators about available security updates. Currently it isn't possible to determine from Grype's CycloneDX output if a vulnerability has a patch available.

**Additional context**:

The information about available fixes is already present in the json format, so it shouldn't be too hard to add. The version ranges must follow the [Package URL Version Range syntax (vers)](https://www.packageurl.org/docs/vers/how-to-parse).

somaz94:
This populates the `affects[].versions[]` entries in Grype's CycloneDX output, addressing #3512.

Today the CycloneDX `affects` array only carries a `ref`, so consumers cannot tell from the
CycloneDX document whether a vulnerability has a fix available. This change adds:

- the matched package version, recorded with status `affected`
- for vulnerabilities with known fixed-in versions, each fix as a `vers`-syntax range
  (`vers:<scheme>/>=<version>`) with status `unaffected`, so consumers can spot available
  fixes and automate remediation

The `vers` scheme is derived from the package PURL type and falls back to `generic`. When
neither a version nor a fix is known, the `versions` array is omitted (no change for those).

One design point worth your read: I model the fix boundary as `>=<fixedVersion>` / `unaffected`
rather than a `<<fixedVersion>` / `affected` range, because Grype reliably knows the fix version
and the matched version but not the introduced version. Happy to switch to the affected-range
form if you prefer.

Validation (run locally):
- `go test ./grype/presenter/cyclonedx/...` passes; added table-driven tests covering the
  affected / single fix / multiple fixes / not-fixed / missing-PURL / empty cases
- regenerated the CycloneDX golden files with `-update` and reviewed the diff
- `go build ./...`, `gofmt`, `go vet`, and `golangci-lint run` on the package are clean (no new findings)

closes #3512


somaz94:
Friendly ping — this has been green (14/14) for about five weeks with no review yet.

It adds vulnerable version ranges to the CycloneDX output, so downstream consumers can tell which versions a finding actually applies to. Happy to rebase, or to split the schema change from the formatting change if that would make it easier to review.

spiffcs (approved):

spiffcs:
Apologies @somaz94. We're a small team so it's hard to get to all the PR across all our tools. Thank you for the contribution here! You should see it in the next release.

PR #3527, opened 2026-06-23
fix(govulndb): only emit records for stdlib

willmurphyscode:
For third party go packages, go vuln db has some very strange version ranges, and this has caused a number of false positive reports. Additionally, for most of them, Grype currently reports both a GHSA and a GO record, resulting in user confusion.

However, there is one key class of record available in the go vuln db that is not in GHSA: records affecting the go stdlib itself (that is, the core parts of the langage that are statically linked into every go binary).

Therefore, for now, emit only records relating to the go stdlib. Future enhancements can pull in additional data related to third party go packages in go vuln db.


grype/matcher/golang/testdata/govulndb-go/govulndb/results/go-2022-0617.json
```diff
@@ -1,62 +0,0 @@
-{
```
kzantow, line 1: Do these records need to get removed? Should they be omitted by the transformer? Having a lot of churn in these records seems like it would be problematic eventually

kzantow (approved):

wagoodman (approved):

PR #3531, opened 2026-06-25
fix: use full namespaced name for db search PURLs

### LINKED ISSUE #3508: grype db search does not properly parse golang PURL
The following:
``` sh
[0000] DEBUG config:
  log:
      quiet: false
      level: trace
      file: ""
  dev:
      profile: ""
  db:
      ca-cert: ""
      auto-update: true
      validate-age: true
      require-update-check: true
      update-available-timeout: 30s
      update-download-timeout: 5m0s
  exp: {}
  dev:
      db:
          debug: false
No results found
```

``` sh
syft version
Application:   syft
Version:       1.43.0
GitDescription: v1.43.0
Platform:      linux/amd64
GoVersion:     go1.26.2
Compiler:      gc
SchemaVersion: 16.1.3
```
K Thx Bye!

arpitjain099:
Fixes #3508.

`grype db search --pkg` with a golang PURL doesn't match anything. The package specifier was built straight from the raw PURL fields using `purl.Name`, and for ecosystems that carry part of the name in the PURL namespace that drops the rest of it. So `pkg:golang/github.com/gin-gonic/gin` gets searched as just `gin`, while the DB keys the record under the full module path.

Per @kzantow's review, this now handles PURLs the same way scanning does rather than reimplementing name reconstruction inside the command:

- `pkg.Provide` decodes the PURL string into a package, so it carries the information grype expects from a PURL.
- `name.PackageNames` then reports the DB search names for that package, and each name becomes a package specifier plus a CPE specifier.

Golang module paths, npm scopes, and Maven group:artifact coordinates resolve the way they do during a scan, and ecosystem-specific normalization comes along with it (Python picks up PEP 503, for instance). The ecosystem comes from the decoded syft package type, falling back to the PURL type if the decoded package doesn't carry one. The per-ecosystem helper from the first version of this PR is gone.

### Testing

Table cases in `TestDBSearchPackagesPostLoad` cover the golang module path, the npm scope, Maven group:artifact, and pypi PEP 503 normalization. The plain-name and invalid-PURL cases were already covered.

```
$ go test ./cmd/grype/cli/options/... ./grype/db/v6/name/... ./grype/pkg/
ok      github.com/anchore/grype/cmd/grype/cli/options   1.078s
ok      github.com/anchore/grype/grype/db/v6/name        1.042s
ok      github.com/anchore/grype/grype/pkg               5.548s
```

`main` is merged in as of `159bf2d`, which touched `grype/pkg` and the name resolver, so those two packages are included above.


cmd/grype/cli/options/database_search_packages.go
```diff
@@ -82,3 +83,32 @@ func (o *DBSearchPackages) PostLoad() error {
...
+//     "github.com/gin-gonic/gin".
+//   - npm scoped packages parse to Namespace="@scope" and Name="name", and are
+//     stored as "@scope/name".
+//   - Maven packages parse to Namespace="groupId" and Name="artifactId", and are
+//     stored as "groupId:artifactId".
+//
+// Without this, a search for a namespaced PURL only used purl.Name and silently
+// failed to match. This mirrors the same reconstruction the openvex build
+// transformer performs (grype/db/v6/build/transformers/openvex).
+func packageNameFromPURL(purl *packageurl.PackageURL) string {
```
kzantow: Hi @arpitjain099 -- we talked about this on the [livestream](https://www.youtube.com/watch?v=_Y3XOPZNTDQ) and the consensus is we should handle PURLs the same as when scanning here. There are a number of other nuances that occur when you _run grype against a PURL_ compared to when you specify a PURL here. The simple solution would be to [use the SBOM decoder to get a package](https://github.com/anchore/grype/blob/main/grype/pkg/provider.go#L22), which will have all the information from the PURL as we expect it in a package. After that, we probably want to use the [PackageNames function](https://github.com/anchore/grype/blob/main/grype/db/v6/name/resolver.go#L46) to get all the search names for a given package. This will result in correctly searching for NPM, Java, Golang, and other packages and avoid these functions drifting

kzantow (dismissed):
Per [the comment](https://github.com/anchore/grype/pull/3531/changes#r3520207506) we should convert PURLs to packages and use the PackageNames function to search the same way matching will

arpitjain099:
Thanks @kzantow, that makes a lot of sense. I reworked it to go through the same path as scanning: decode the PURL into a package via `pkg.Provide`, then feed each package to `name.PackageNames` to get the search names, and build the specifiers from those. So Golang module paths, npm scopes, and Maven group:artifact now resolve the way they do during a scan, and Python even picks up PEP 503 normalization for free. The custom per-ecosystem helper is gone. Updated the tests to cover those cases (build/vet/tests pass locally).

cmd/grype/cli/options/database_search_packages.go
arpitjain099: Thanks for the pointer to the SBOM decoder and `PackageNames`, that's what this does now: `pkg.Provide` decodes the PURL into a package, then `name.PackageNames` supplies the search names. The per-ecosystem helper this thread was on is gone, so there's nothing left here to drift.

One thing I should own: the PR description still described that old helper until just now, so the body and the diff disagreed for a couple of weeks. I've rewritten the description to match what the code actually does. Apologies if you looked in the meantime and it read as though nothing had changed.

I've also merged current main in, since it touched `grype/pkg` and the name resolver, and CI is green. Ready for another look whenever you have time.

noqt (changes requested):
Reviewing from NOQT. The package-name half is on the right path: exact head `eda84395a44bc5cc52a7d95e0d270ca883871199` passes its focused option test, vet, and diff hygiene, and it merges cleanly onto current main `f6f2ce0a2293e65c97bf5468926cb0c9ddad822f`, where the focused test and vet also pass.

There is one lookup regression to fix before merge. `appendPURLSpecs` sends every `name.PackageNames` result into both `PkgSpecs` and `CPESpecs`. For `pkg:golang/github.com/gin-gonic/gin@v1.9.0`, current behavior searches CPE product `gin` with target software `golang`; this head instead searches product `github.com/gin-gonic/gin` with target `go-module`. `PackageNames` is right for the DB package lookup, but it is not a CPE product mapping.

A reviewer-only regression that keeps the full package search name while preserving the existing CPE fallback fails with:

```text
expected CPE product: gin
actual CPE product:   github.com/gin-gonic/gin
```

Please keep the provider/`PackageNames` path for `PkgSpecs`, build the CPE spec once from the parsed PURL (`purl.Name`, `purl.Type`), and add that regression. The rest of the focused current-main qualification is green.

arpitjain099:
@noqt thanks, that is a real regression and you described it exactly. Fixed in e6587ec.

`appendPURLSpecs` was appending to `CPESpecs` inside the `name.PackageNames` loop, so the search names leaked into the CPE product. The CPE spec is now built once, outside that loop, from the parsed PURL:

```go
o.CPESpecs = append(o.CPESpecs, &v6.PackageSpecifier{CPE: &cpe.Attributes{Part: "a", Product: purl.Name, TargetSW: purl.Type}})
```

`PackageNames` still drives `PkgSpecs`, which was the point of the change. `appendPURLSpecs` now takes the already-parsed `packageurl.PackageURL` rather than re-deriving the type, and the doc comment says why the two sides differ.

For `pkg:golang/github.com/gin-gonic/gin@v1.9.0` the result is now the package name `github.com/gin-gonic/gin` in `go-module` and the CPE product `gin` with target software `golang`, which is what main does today. That case is the regression test.

The npm, maven and pypi expectations moved for the same reason, and one of them was informative: `pkg:pypi/Django_extensions` parses to `django-extensions` because purl normalisation already applies PEP 503, so that CPE product is unchanged.

`go test ./cmd/grype/cli/options/...` and `go vet` on the same path are both green.

PR #3533, opened 2026-06-25
feat(RHEL): respect multiple RHSAs based on stream in version

willmurphyscode:
Previously, when multiple RHSAs fixed the same RPM in the same major version of RHEL, Grype was only able to match based on the one with the highest epoch-version-release (EVR). Change grype to now choose the right RHSA if multiple are available based on the elN_M suffix that the fixed versions carry.

### DIFF
--- a/grype/db/v6/build/transformers/os/transform.go
+++ b/grype/db/v6/build/transformers/os/transform.go
@@ -134,6 +134,13 @@ func getPackages(vuln unmarshal.OSVulnerability) ([]db.AffectedPackageHandle, []
 		}
 		aph.BlobValue.Ranges = ranges
 		afs = append(afs, aph)
+
+		// Emit an "= <evr>" unaffected handle for every distinct fix version the advisories
+		// called out (UnaffectedVersions). When several same-base streams collapse to one
+		// affected record ("< highest fix"), a host running exactly its own stream's fix build
+		// otherwise reports as vulnerable; these exact-match unaffected records let the matcher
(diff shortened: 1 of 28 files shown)


grype/db/v6/build/transformers/os/transform.go
```diff
@@ -143,6 +150,59 @@ func getPackages(vuln unmarshal.OSVulnerability) ([]db.AffectedPackageHandle, []
...
 }
 
+// exactUnaffectedHandles turns each FixedIn.AdditionalAdvisoryFixes entry into an unaffected
+// package handle constrained to that exact version ("= <evr>"), carrying the advisory that
+// shipped that build as a fix reference. These ride alongside the affected handle (which still
+// carries the coarse "< highest fix" constraint) so that a host at exactly a called-out fix
+// build matches an unaffected record and the matcher can drop the otherwise-false-positive
+// disclosure - and report which advisory's fix the host already carries. The rpm-modularity
+// qualifier mirrors the affected handle so modular hosts match the right stream.
+func exactUnaffectedHandles(group groupIndex, fixedIns []unmarshal.OSFixedIn, vuln unmarshal.OSVulnerability, qualifiers *db.PackageQualifiers) []db.UnaffectedPackageHandle {
```
willmurphyscode, line 160: Can we use package qualifiers to keep the matcher from having to be too smart here? Like a "RHEL Y Stream reachability" qualifier or something?

kzantow:
I'm pretty sure this was implemented by https://github.com/anchore/grype/pull/3542 apologies if this was meant to remain open 🙏

PR #3542, opened 2026-07-01
feat(rhel): duplicate RHSAs to all applicable RHEL minor versions

willmurphyscode:
Historically, affected package handles for RHEL have been emitted at major versions only, modeling RHEL 9, for example, as a single namespace within which to search for vulnerabilities. However, this representation cannot correctly match on situations where multiple RHSAs fix the same RPM on the same (CVE, RHEL major, RPM name) tuple. In order to create this capability without necessitating a client side matching behavior change, implement the following workaround: When multiple RHSAs are released for different RHEL minor versions but the same major, compute which minor versions should have which RHSAs, and emit copies of the affected package handle for each RHEL minor version they should match on.

So in the example of CVE-2023-38545 for curl, an affected package handle for RHSA-2023:5763's version is emitted for rhel 9.0, 9.1, and 9.2 namespaces, and for RHSA-2023:6745's version is emitted for 9.3 and all subsequent 9.x minor versions, effectively gaining the ability to improve matching on older Grype versions at the expense of increased database size.

In the rare case where there are multiple applicable RHSAs _and_ one of them cannot be assigned to a set of RHEL minor versions, guess the set of RHEL minor versions based on EVR ordering with the other RHSAs.


grype/db/internal/provider/unmarshal/os_vulnerability.go
```diff
@@ -28,6 +28,17 @@ type OSFixedIn struct {
...
+}
+
+// OSAdvisory is a per-stream fix candidate emitted by vunnel for multi-RHSA packages.
+// The target minor and support channels are parsed from the CSAF product_id. Minor is
+// populated for .elN_M / modular builds; GA .elN builds carry Minor == nil.
+type OSAdvisory struct {
+	Advisory string   `json:"Advisory"`
+	Version  string   `json:"Version"`
+	Minor    *int     `json:"Minor"`
+	Channels []string `json:"Channels"`
```
wagoodman, line 47: channels is never read anywhere here, should we just remove it for now?

wagoodman: Same with Advisory

grype/db/v6/build/transformers/os/transform.go
```diff
@@ -143,6 +143,280 @@ func getPackages(vuln unmarshal.OSVulnerability) ([]db.AffectedPackageHandle, []
...
+//
+// Erring low is safe: a host on a minor beyond the span finds no exact row and falls back
+// to the major-only row (today's coarse, roll-forward-to-highest behavior). Erring high
+// only wastes rows (extra minors carry the highest fix, same as the major fallback).
+//
+// PROTOTYPE: this should be derived from the published-minor / EOL feed rather than
+// hardcoded. See [transformer-per-minor-must-be-cumulative] in the design notes.
+var rhelMinorSpan = map[string]int{
+	"5": 11, "6": 11, "7": 11, "8": 11, "9": 11, "10": 4,
+}
```
wagoodman: this is listed as "prototype" and is probably best graduated to a go generate approach that we run periodically (or have a better idea here?)

grype/db/v6/build/transformers/os/transform.go
```diff
@@ -143,6 +143,280 @@ func getPackages(vuln unmarshal.OSVulnerability) ([]db.AffectedPackageHandle, []
...
+		Version: db.Version{
+			Type:       versionFormat,
+			Constraint: deriveConstraintFromFix(versionutil.CleanConstraint(v), vulnID),
+		},
+		Fix: &db.Fix{
+			Version: versionutil.CleanFixedInVersion(v),
+			State:   db.FixedStatus,
+		},
+	}
+}
```
wagoodman, line 354: should we put on a FixDetail to capture the ref? (might be wrong / overkill)

grype/matcher/rpm/testdata/rhel-multi-rhsa/rhel/results/rhel@9/cve-9999-0001.json
wagoodman, line 1: this case is not based on a real record, can we pick from an actual artifact? of not that's ok, lets mention that as a code comment in the test then

grype/matcher/rpm/testdata/rhel-multi-rhsa/rhel/results/rhel@9/cve-9999-0002.json
wagoodman, line 1: this case is not based on a real record, can we pick from an actual artifact? of not that's ok, lets mention that as a code comment in the test then

grype/matcher/rpm/testdata/rhel-multi-rhsa/rhel/results/rhel@8/cve-9999-0003.json
wagoodman, line 1: this case is not based on a real record, can we pick from an actual artifact? of not that's ok, lets mention that as a code comment in the test then

grype/db/internal/provider/unmarshal/os_vulnerability.go
willmurphyscode: This is a bug - those are meant to be forwarded into matching. will send rev.

willmurphyscode: On both counts - the advisory we actually compared with should wire all the way into `-o json` output and I believe the eus matcher needs to know about channels.

grype/db/v6/build/transformers/os/transform.go
willmurphyscode: I think we should do something like that, yes.

grype/matcher/rpm/testdata/rhel-multi-rhsa/rhel/results/rhel@9/cve-9999-0001.json
willmurphyscode: I've tried to go back and find real data wherever possible.

wagoodman (approved):

PR #3562, opened 2026-07-10
feat: add minimum severity report filtering

### LINKED ISSUE #3529: Ability to make grype report smaller by filtering out low and medium

Grype report is too big to be uploaded as an artifact in gitlab when run on an sbom prepared by syft. I'm not interested in low and medium ones. At the moment, the only way is to use only-fixed or only-notfixed flag which is not useful in this case. 

**Why is this needed**:

A feature to limit the size of grype report (sarif or json etc) by filtering for severity. 

**Additional context**:

krishgera:
## Summary

Adds a minimum severity filter for Grype report output.

Users can configure the threshold through:

- CLI: `--min-severity <level>`
- YAML: `min-severity`
- Environment: `GRYPE_MIN_SEVERITY`

Supported values are `negligible`, `low`, `medium`, `high`, and `critical`.

## Behavior

- The threshold is inclusive.
- Findings with Unknown severity are excluded when filtering is enabled.
- Both regular and ignored matches are filtered.
- Filtering is applied once before the report is passed to the output presenters, keeping table, JSON, SARIF, template, and CycloneDX output consistent.
- Scanning, ignore/VEX processing, and `--fail-on` behavior remain unchanged.

## Testing

Added coverage for:

- all severity thresholds;
- case-insensitive configuration;
- invalid and Unknown values;
- regular and ignored matches;
- empty filtered reports and metadata preservation;
- YAML, environment, and CLI precedence;
- `--only-fixed` and `--only-notfixed`;
- independent `--fail-on` behavior;
- table, JSON, SARIF, template, and CycloneDX output.

Validation performed:

- `make format`
- `make lint`
- focused unit tests
- compiled-binary CLI tests
- `git diff --check`

Closes #3529

### DIFF
--- a/cmd/grype/cli/commands/report_filter.go
+++ b/cmd/grype/cli/commands/report_filter.go
@@ -1,0 +1,28 @@
+package commands
+
+import (
+	"slices"
+
+	"github.com/anchore/grype/grype/presenter/models"
+	"github.com/anchore/grype/grype/vulnerability"
+)
+
+func filterDocumentByMinSeverity(doc *models.Document, minSeverity *vulnerability.Severity) {
+	if minSeverity == nil {
+		return
+	}
+
+	doc.Matches = slices.DeleteFunc(doc.Matches, func(m models.Match) bool {
+		return vulnerability.ParseSeverity(m.Vulnerability.Severity) < *minSeverity
+	})
+	if doc.Matches == nil {
+		doc.Matches = make([]models.Match, 0)
+	}
+
+	doc.IgnoredMatches = slices.DeleteFunc(doc.IgnoredMatches, func(m models.IgnoredMatch) bool {
+		return vulnerability.ParseSeverity(m.Vulnerability.Severity) < *minSeverity
+	})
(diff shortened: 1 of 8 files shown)


noqt (approved):
Reviewing from NOQT. This is a coherent report-filtering boundary: the threshold is inclusive, Unknown is excluded only when filtering is configured, active and ignored records are filtered once before every presenter, and `--fail-on` remains independent at the matcher layer.

I traced the CLI/YAML/environment wiring, validation, filter placement, and merged-tree API calls. The tests cover every threshold, metadata-preserving empty reports, config precedence, fix-state composition, fail-on independence, and table/JSON/SARIF/template/CycloneDX output. Exact head `d3bb81b5f62d17f3f315714d97fe24f93444300a` merges cleanly with current main `f6f2ce0a2293e65c97bf5468926cb0c9ddad822f`, and the exact-range diff passes `git diff --check`.

Only DCO is available as a hosted check, and Go isn't available in this environment, so I'm not claiming an independent test pass. I found no code blocker; please run the hosted suite before merging.

PR #3568, opened 2026-07-13
Add match details to CycloneDX vulnerabilities

### LINKED ISSUE #3513: Include match details in the CycloneDX output format

Include matchDetails object in the CycloneDX output format as well.

<details><summary>matchDetails object example</summary>

```json
      "matchDetails": [
        {
          "type": "exact-indirect-match",
          "matcher": "rpm-matcher",
          "searchedBy": {
            "distro": {
              "type": "fedora",
              "version": "23"
            },
            "package": {
              "name": "krb5",
              "version": "1.14.1-6.fc23"
            },
            "namespace": "fedora:distro:fedora:23"
          },
          "found": {
            "vulnerabilityID": "CVE-2016-3120",
            "versionConstraint": "< 0:1.14.3-8.fc23||< 0:1.14.3-4.fc23 (rpm)"
          },
          "fix": {
            "suggestedVersion": "0:1.14.3-4.fc23"
          }
        }
      ]
```

</details>

**Why is this needed**:

Match details include important information about the match, including type of the match. Marking package as vulnerable based on a CVE issued to an upstream package may result in false positives and less confidence. Including information about direct/indirect match would allow consumers to understand how the match was made and avoid false positives.

**Additional context**:

The match details object is already part of the json format so I recon it's part of a public stable API. As such I don't see a reason to not include it in the CycloneDX format as well. CycloneDX allows setting custom properties for each vulnerability finding (see [spec](https://cyclonedx.org/docs/1.7/json/#vulnerabilities_items_properties)). Because only string values are allowed, I propose stringifying the matchDetails object. Syft already makes use of proterties to store `syft:package:type`.

Although I have only use case for the match type, I don't see a reason to not include the whole object. 

Previously discussed on Discourse: [Additional properties in CycloneDX-json Grype output](https://anchorecommunity.discourse.group/t/additional-properties-in-cyclonedx-json-grype-output)

msnandhis:
## Summary

Include Grype match details in CycloneDX vulnerability entries as a custom property.

Each vulnerability with match details now has a `grype:match:details` property whose value is the complete `matchDetails` array encoded as JSON. Vulnerabilities without match details do not receive the property.

## Motivation

CycloneDX output currently omits the evidence that explains how Grype matched a vulnerability. Consumers need that information to distinguish direct and indirect matches and make better-informed triage decisions.

## Approach

- Serialize the existing `models.MatchDetails` collection without changing its shape.
- Use one `grype:match:details` property so the complete ordered collection survives both CycloneDX JSON and XML output.
- Propagate serialization errors instead of silently dropping match evidence.
- Update CycloneDX presenter snapshots.

The issue confirms that CycloneDX properties are the intended extension point, but it does not prescribe the exact property name or whether the data should be flattened. This draft uses a single JSON-valued property to preserve nested match evidence losslessly. Maintainer feedback on that representation is welcome before this is marked ready.

## Testing

- `go test ./grype/presenter/cyclonedx -count=1`
- `make static-analysis`
- `git diff --check`

Tests cover direct and indirect match details, omission for empty details, JSON and XML round trips, and serialization errors.

Closes #3513

## AI assistance

AI-assisted tools were used during implementation and test development. I reviewed the resulting diff, verified the behavior locally, and take responsibility for the contribution.

### DIFF
--- a/grype/presenter/cyclonedx/testdata/snapshot/TestCycloneDxPresenterDir.golden
+++ b/grype/presenter/cyclonedx/testdata/snapshot/TestCycloneDxPresenterDir.golden
@@ -118,6 +118,12 @@
         {
           "ref": "a246fd2054833c93"
         }
+      ],
+      "properties": [
+        {
+          "name": "grype:match:details",
+          "value": "[{\"type\":\"exact-direct-match\",\"matcher\":\"dpkg-matcher\",\"searchedBy\":{\"distro\":{\"type\":\"ubuntu\",\"version\":\"20.04\"}},\"found\":{\"constraint\":\"\\u003e= 20\"},\"fix\":{\"suggestedVersion\":\"1.2.1\"}}]"
+        }
       ]
     },
     {
@@ -158,6 +164,12 @@
       "affects": [
         {
           "ref": "pkg:deb/package-2@2.2.2?package-id=74378afe15713625"
+        }
+      ],
+      "properties": [
+        {
+          "name": "grype:match:details",
+          "value": "[{\"type\":\"exact-indirect-match\",\"matcher\":\"dpkg-matcher\",\"searchedBy\":{\"cpe\":\"somecpe\"},\"found\":{\"constraint\":\"somecpe\"}}]"
         }
       ]
     }
--- a/grype/presenter/cyclonedx/testdata/snapshot/TestCycloneDxPresenterImage.golden
+++ b/grype/presenter/cyclonedx/testdata/snapshot/TestCycloneDxPresenterImage.golden
@@ -119,6 +119,12 @@
         {
           "ref": "a246fd2054833c93"
         }
+      ],
+      "properties": [
+        {
+          "name": "grype:match:details",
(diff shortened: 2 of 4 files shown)


noqt (changes requested):
I reviewed this for NOQT against current `main` (`f6f2ce0a`). The CycloneDX property approach matches the direction in #3513, and the JSON/XML round-trip coverage is the right contract.

There is one blocking integration conflict in `grype/presenter/cyclonedx/vulnerability.go`. The clean resolution is to keep this PR's `encoding/json` import, `generateCDXProperties`, and `Properties: properties`, while retaining current main's `vulnerability` import and `Range: generateAffectedVersions(m)`. Git then applies the two test and two snapshot changes without another source conflict; `git diff --cached --check` is clean.

Please rebase with that combined resolution and refresh the two CycloneDX golden snapshots. I also attempted the four new match-details tests locally, but the cold Windows Go build did not reach the test binary before I stopped it, so I'm not treating that as a pass or a code failure. The focused command to rerun is:

```console
go test -p 1 -vet=off ./grype/presenter/cyclonedx -run 'TestNewVulnerability_(IncludesMatchDetailsProperty|OmitsMatchDetailsPropertyWhenEmpty|MatchDetailsPropertySerializesInCycloneDXFormats|ReturnsErrorForUnencodableMatchDetails)$' -count=1
```

Once the rebase and focused test are green, this looks ready for maintainer review.

PR #3572, opened 2026-07-14
fix: rhel version streams

kzantow:
This PR adjusts the redhat stream fixes to be applied only to the specific minor version. For example:
A fix for CVE-2023-4813 issued in RHSA-2023:5453 was published in a 9_2.7 release, but this was after later builds were likely available on the main branch. The development branches and main EL9 fix landed in `0:2.34-100.el9` noted in RHBA-2024:2413 whereas the 9.2 stream fix landed in: `0:2.34-60.el9_2.7`. A user who is on 9.0 as I understand it will be able to get a mainline fix but not the 9_2 stream fix, so including the 9_2 fixed version range needs to only apply to RHEL 9.2, previously the fix was showing up in all prior versions. With this change, it looks like: 
```
CVE-2023-4813  glibc         rpm        redhat@9        < 0:2.34-100.el9      
CVE-2023-4813  glibc         rpm        redhat@9.0      < 0:2.34-100.el9   
CVE-2023-4813  glibc         rpm        redhat@9.1      < 0:2.34-100.el9   
CVE-2023-4813  glibc         rpm        redhat@9.10     < 0:2.34-100.el9      
CVE-2023-4813  glibc         rpm        redhat@9.2      < 0:2.34-60.el9_2.7   
CVE-2023-4813  glibc         rpm        redhat@9.3      < 0:2.34-100.el9      
CVE-2023-4813  glibc         rpm        redhat@9.4      < 0:2.34-100.el9      
CVE-2023-4813  glibc         rpm        redhat@9.5      < 0:2.34-100.el9      
CVE-2023-4813  glibc         rpm        redhat@9.6      < 0:2.34-100.el9      
CVE-2023-4813  glibc         rpm        redhat@9.7      < 0:2.34-100.el9      
CVE-2023-4813  glibc         rpm        redhat@9.8      < 0:2.34-100.el9      
CVE-2023-4813  glibc         rpm        redhat@9.9      < 0:2.34-100.el9
```

PR #3585, opened 2026-07-20
Add fix date to rhel minor records created from rhsa

wagoodman:
This is a fix related to the recent changes with #3542 and #3572, which expand major-only RHEL records to also emit per-minor version records. In the process of emitting these new records, the fix available information that carries fix date information was not carried to the new records. This PR addresses this by preserving the fix available object from the major record to all created records.


kzantow (approved):
🙏

PR #3587, opened 2026-07-20
fix: prevent false positives for import-scoped Go advisories without symbol data

### LINKED ISSUE #3573: False Positive: GO-2026-5932
**Vulnerability ID**:

GO-2026-5932

**Package URL or steps to reproduce**:

Minimal one-line reproduction via Package URL:

```console
$ grype pkg:golang/golang.org/x/crypto@v0.54.0
NAME                 INSTALLED  TYPE       VULNERABILITY  SEVERITY  EPSS  RISK
golang.org/x/crypto  v0.54.0    go-module  GO-2026-5932   Unknown   N/A   N/A
```

The finding is reported for the whole `golang.org/x/crypto` module, but
GO-2026-5932 only applies to the `golang.org/x/crypto/openpgp` sub-package.

Full repository with a buildable/scannable image (binary imports **only**
`golang.org/x/crypto/ssh`, never `openpgp`):
https://github.com/krzysztof-oneleet/grype-openpgp-false-positive

Image-level reproduction (analog of `grype <image>`):

```console
$ docker build -t grype-openpgp-fp:latest .
$ grype grype-openpgp-fp:latest
NAME                 INSTALLED  TYPE       VULNERABILITY  SEVERITY  EPSS  RISK
golang.org/x/crypto  v0.54.0    go-module  GO-2026-5932   Unknown   N/A   N/A
```

The binary provably does not link `openpgp`:

```console
$ go tool nm app | grep -i openpgp     # (no output — not linked)
$ go version -m app | grep x/crypto
        dep     golang.org/x/crypto     v0.54.0 h1:YLIA59K4fiNzHzjnZt2tUJQjQtUWfWbeHBqKtk3eScw=
```

No special grype configuration was used (default config, default DB).

This is a **false positive** caused by module-level matching where the advisory
is scoped to a single sub-package.

- GO-2026-5932 applies **only** to `golang.org/x/crypto/openpgp` (and its
  `packet`, `armor`, `clearsign`, `errors`, `elgamal`, `s2k` sub-packages). It
  does **not** apply to other sub-packages such as `golang.org/x/crypto/ssh`.
  Advisory: https://pkg.go.dev/vuln/GO-2026-5932
- The advisory has **no fixed version** — the remediation is "don't use the
  `openpgp` package," which is unmaintained/unsafe by design. As a result, every
  consumer of *any* part of `golang.org/x/crypto` is flagged permanently with no
  upgrade path, even when `openpgp` is never imported or linked.
- Grype reads only the module path + version from the Go binary's build info
  (equivalent to `go version -m`), which does not record which sub-packages the
  linker actually included, so it cannot distinguish `ssh`-only usage from
  `openpgp` usage.
- `govulncheck` (call-graph aware) correctly reports **0** affecting
  vulnerabilities for the same code:

  ```console
  $ govulncheck ./...
  === Symbol Results ===

  No vulnerabilities found.

  Your code is affected by 0 vulnerabilities.
  This scan also found 1 vulnerability in packages you import and 2
  vulnerabilities in modules you require, but your code doesn't appear to call
  these vulnerabilities.
  ```

- Same behavior was reported upstream for Trivy:
  https://github.com/aquasecurity/trivy/discussions/10944

Note: GO-2026-5932 has no CVE/GHSA alias — it exists only in the Go
vulnerability database — yet it is present in the grype DB (see version below)
and matches at the module level.

**Environment**:
- Output of `grype version`:
  ```
  Application:         grype
  Version:             0.115.0
  BuildDate:           2026-06-26T10:16:27Z
  GitCommit:           Homebrew
  GoVersion:           go1.26.4
  Compiler:            gc
  Syft Version:        v1.46.0
  Supported DB Schema: 6
  ```
  Vulnerability DB: schema v6.1.7, built 2026-07-14T06:57:40Z
  (vulnerability-db_v6.1.7_2026-07-14T00:35:10Z).
- OS (host running grype): macOS 26.5.2 (build 25F84), arm64.
  Note: the finding is OS-independent (a Go module match); it also reproduces
  inside the `gcr.io/distroless/static-debian12:nonroot` image built above.

texasich:
When a Go vulnerability advisory carries ecosystem_specific.imports scoping
the affected sub-packages (e.g., GO-2026-5932 only affects
golang.org/x/crypto/openpgp, not the entire module) but the scanned
binary has no captured symbol evidence, the gosymbols qualifier
previously fell back to module-granularity matching. This caused
false positives: every binary importing the module was flagged even
when it provably does not link the vulnerable sub-package.

The fix changes evaluate() to report scoped=true when imports are
present even without symbol evidence. Satisfied() then correctly
returns false instead of falling back to module-level matching.

Module-level advisories (no imports) continue to match without
symbol evidence — this change only affects advisories that
declare sub-package scoping.

Closes #3573

### DIFF
--- a/grype/pkg/qualifier/gosymbols/qualifier.go
+++ b/grype/pkg/qualifier/gosymbols/qualifier.go
@@ -26,14 +26,16 @@ func New(imports []Import) qualifier.Qualifier {
 	return &gosymbolsQualifier{imports: imports}
 }
 
-// Satisfied reports whether the package plausibly uses any of the vulnerable symbols. The check
-// only applies when the scanned package carries binary symbol evidence (go binaries cataloged with
-// symbol capture enabled); packages without symbol evidence always satisfy the qualifier so that
-// module-granularity matching behavior is preserved.
+// Satisfied reports whether the package plausibly uses any of the vulnerable symbols. When the
+// advisory carries import-scoped qualifiers (ecosystem_specific.imports), the package must carry
+// binary symbol evidence and use at least one of the listed imports or symbols. When symbol
+// evidence is absent and the advisory is import-scoped, the qualifier is not satisfied — this
+// avoids false positives for advisories that only affect specific sub-packages. Module-level
+// advisories (no imports) continue to match without symbol evidence.
 func (q *gosymbolsQualifier) Satisfied(p pkg.Package) (bool, error) {
 	matched, scoped := q.evaluate(p)
 	if !scoped {
-		// no symbol scoping (advisory lists no imports, or the package carries no symbol evidence):
+		// advisory has no imports — pure module-level advisory,
 		// preserve module-granularity matching
 		return true, nil
 	}
@@ -62,10 +64,11 @@ func MatchedSymbols(quals []qualifier.Qualifier, p pkg.Package) []string {
 
 // evaluate is the single source of truth for both the filter (Satisfied) and the report
 // (MatchedSymbols): it returns the sorted, de-duplicated set of fully-qualified vulnerable symbols
-// the package is found to use, plus whether symbol scoping applied at all. scoped is false when the
-// advisory references no imports or the package carries no symbol evidence — both fall back to
-// module-granularity matching. When scoped is true, a non-empty result means at least one vulnerable
-// symbol (or a whole-package import path) is present; empty means the package uses none of them.
+// the package is found to use, plus whether the advisory carries import scoping. scoped is true
+// when the advisory references imports (even if the package carries no symbol evidence — in that
+// case matched will be empty, causing Satisfied() to correctly filter out the vulnerability rather
+// than falling back to module-granularity matching). scoped is false only when the advisory has
+// no imports at all (pure module-level advisory).
 func (q *gosymbolsQualifier) evaluate(p pkg.Package) (matched []string, scoped bool) {
 	if len(q.imports) == 0 {
 		return nil, false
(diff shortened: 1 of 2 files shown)


wagoodman:
I'm going to close this PR as ultimately it's trading FPs for FNs, and hinges on the "absence of evidence can be interpreted as evidence of absence" fallacy. 

Take CVE-2024-45337 for example... the real x/crypto/ssh auth bypass (an SSH server using `ServerConfig.PublicKeyCallback` can be tricked into authorizing the wrong key). High severity, absolutely something you would want flagged by grype.

Assume we scan the same genuinely-vulnerable SSH server binary two ways:

1. `grype ./sshd` in which case grype forces symbol capture and will see the offending `ServerConfig.PublicKeyCallback` linked in, thus it will raise up that we're vulnerable (correct).

2. `syft ./sshd -o json | grype` and since syft's symbol capture is off by default, the `x/crypto` package in that SBOM carries no symbol info at all. This PR interprets no symbols in a way where the package qualifier returns "not satisfied"  thus the GO-2024-3321 advisory is suppressed (grype reports nothing! a new FN). Ideally we should interpret no symbols as "can't prove the ssh code is unused" and fall back to a plain module+version match (still flag GO-2024-3321/CVE-2024-45337) 

So we don't want to treat "no symbols present" as "the vulnerable symbol isn't used." But no symbols almost always just means the capture didn't run (default SBOMs, source scans, stdlib-only capture, other SBOM tools, etc). We need to fail open in these cases (assume vulnerable, not safe when there are unknowns).

#3592 is trying to fix the same symptom a different way: sometimes we're raising up FPs because they are found two different ways... through the GHSA and the GO vuln records, especially when the aliasing is indirect (say through the CVE id only, not referencing each other).

PR #3592, opened 2026-07-22
Scope Go GHSA twins by shared CVE

wagoodman:
Fixes a class of Go false positives in the govulndb/GHSA merge, plus a related version-parsing false negative that turned up alongside it.

The main one: a binary that links only `golang.org/x/crypto/sha3` (nothing ssh-related reachable) still drew 14 phantom `x/crypto` ssh CVEs. The build-time merge scopes GHSA records with the Go symbol info from their aliased govulndb twin, so an import-scoped advisory doesn't match with the plain module+version.

The bug: it only found the twin when the govulndb record aliased the GHSA id directly, and recent `x/crypto/ssh` records (e.g. [`GO-2026-5013`](https://pkg.go.dev/vuln/GO-2026-5013)) alias only the CVE... so their GHSA twins were never scoped and leaked through at module level, reproducing exactly the false positives the gosymbols feature exists to remove.

- the merge now resolves the twin by the whole alias group (shared CVE **and** GHSA id), not GHSA id alone, so symbol scope reaches every twin
- the covered govulndb record is deduped as before, leaving the scoped GHSA as the only match
- self-limiting by design: a GHSA that only shares a CVE but names a different module gets no symbols and doesn't cover, so nothing over-merges
- version ranges are left alone; only the symbol qualifier moves across

With the fixes from this PR, the `sha3-only` x/crypto binary drops from 14 findings to 0, and no import-scoped advisory loses its module-level fallback when symbols are absent (so no new false negatives)

The second, smaller fix: grype's Go version comparator couldn't parse a prerelease tag glued straight onto the version with no `-` separator (e.g. `9.6.0b1`), so any advisory range carrying such a bound went inert and matched nothing.

- normalize the glued tag (`9.6.0b1` converts to `9.6.0-b1`) during parsing; it then sorts correctly before its release
- only versions that currently fail to parse are touched... real Go tags are canonical semver, so nothing else moves
- fixes a real miss where [`github.com/redis/go-redis/v9` @ `9.6.0`](https://vuln.go.dev/ID/GO-2025-3540.json) (affected, fixed in `9.6.3`) matched nothing at all

_(Shout out to @texasich and @krzysztof-oneleet to bringing these cases up in #3573 and #3587 ! 🙌 )_

### DIFF
--- a/grype/db/v6/build/govulndb_merge.go
+++ b/grype/db/v6/build/govulndb_merge.go
@@ -52,6 +52,10 @@ type goVulnDBMerger struct {
 	goGHSAEntries   map[string]*transformers.RelatedEntries
 	goGHSAOrder     []string
 	govulndbEntries []*transformers.RelatedEntries
+
+	// cveToGHSAKeys indexes held GHSA keys by the CVE ids they alias, so a GO
+	// record can reach its GHSA twin by shared CVE. Built in reconcile.
+	cveToGHSAKeys map[string][]string
 }
 
 func newGoVulnDBMerger() *goVulnDBMerger {
@@ -105,6 +109,8 @@ func hasGoModulePackages(entry transformers.RelatedEntries) bool {
 // merger's held state is reset before returning. Called once from Close; the
 // writer persists the returned entries through its own batching path.
 func (m *goVulnDBMerger) reconcile() []transformers.RelatedEntries {
+	m.cveToGHSAKeys = m.buildCVEIndex()
+
 	var surviving []*transformers.RelatedEntries
 	for _, entry := range m.govulndbEntries {
 		if m.handleEntry(entry) {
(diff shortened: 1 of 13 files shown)


grype/db/v6/build/govulndb_merge.go
```diff
@@ -52,6 +52,10 @@ type goVulnDBMerger struct {
 	goGHSAEntries   map[string]*transformers.RelatedEntries
 	goGHSAOrder     []string
 	govulndbEntries []*transformers.RelatedEntries
+
+	// cveToGHSAKeys indexes held GHSA keys by the CVE ids they alias, so a GO
+	// record can reach its GHSA twin by shared CVE. Built in reconcile.
+	cveToGHSAKeys map[string][]string
```
kzantow: nit: should this be any alias and not just CVE? this same question would end up trickling down to a number of places where we are looking specifically for cve- instead of including all aliases

kzantow (approved):
LGTM -- one question about making this more generic for all aliases

grype/db/v6/build/govulndb_merge.go
wagoodman: you're right, this can be much broader with no real risk

PR #3596, opened 2026-07-23
fix(db): surface the update error when no DB exists to fall back on

### LINKED ISSUE #3399: First scan after install fails due to error initializing database
Hello,

I just installed version 0.111.1 to scan a docker image that I loaded into my local daemon by using `docker save` on a remote host and `docker load` to load the tar archive on my localhost. I am on MacOS 26.2 on an m4 MBP (ARM). I could not scan the image because the db failed to complete.

```bash
¯\_(ツ)_/¯❯ grype my_image:latest                                                                       
 ✔ Loaded image                                                                                   my_image:latest
 ✔ Parsed image                           sha256:c936f5454a179d042ea93f72432273035afe163e9c14f9cf917c39d15501e30d
 ✔ Cataloged contents                            35f7f0bf6282f97bab81cf8efaad1183d326c670666a32e39e4fbcd993e38852
   ├── ✔ Packages                        [403 packages]
   ├── ✔ Executables                     [714 executables]
   ├── ✔ File metadata                   [5,270 locations]
   └── ✔ File digests                    [5,270 files]  [0070]  WARN error updating db error=unable to update vulner
[0070] ERROR failed to load vulnerability db: database does not exist
```

However, I ran `grype db update` and it succeeded and took about as much time as I would have expected to download the 111MB of vulnerability data.

```bash
¯\_(ツ)_/¯❯ grype db update                                                                            
```

now when I rerun `grype my_image:latest`, it completes as expected. So I guess there is a bug that affects database updates during scans, possibly when initializing the database for the first time? Or perhaps this is expected behavior and the README.md needs updating?

RajeshRajendiran:
## Summary
- On a first-run scan (no DB downloaded yet), when the update check fails and `RequireUpdateCheck` is false (the default for the scan command), the code only logs the update failure as a `WARN` and falls through to the DB status check.
- Since there's no existing DB to fall back on, `Status()` resolves to the generic `ErrDBDoesNotExist`, and the real, actionable update error is discarded — only visible in a `WARN` log line, not in the final error returned/printed to the user.
- This makes a transient update failure on a fresh install look like an opaque "database does not exist" problem, which is confusing given that a subsequent `grype db update` typically succeeds right after (see #3399).
- Fix: when `Status()` fails specifically because the DB doesn't exist, and there was a prior (warned-but-swallowed) update error, join the two so the underlying cause is part of the returned error instead of being lost.

## Test plan
- [x] Added `TestStatusError` covering: no update error, update error + missing DB (error is surfaced), update error + unrelated status error (update error is not spuriously attached)
- [x] `go test ./grype/...` passes
- [x] `go vet` / `gofmt` clean on changed files

Fixes #3399

### DIFF
--- a/grype/load_vulnerability_db.go
+++ b/grype/load_vulnerability_db.go
@@ -1,6 +1,7 @@
 package grype
 
 import (
+	"errors"
 	"fmt"
 
 	v6 "github.com/anchore/grype/grype/db/v6"
@@ -20,6 +21,7 @@ func LoadVulnerabilityDB(distCfg v6dist.Config, installCfg v6inst.Config, update
 		return nil, nil, fmt.Errorf("unable to create curator: %w", err)
 	}
 
+	var updateErr error
 	if update {
 		updated, err := c.Update()
 		if err != nil {
@@ -27,6 +29,7 @@ func LoadVulnerabilityDB(distCfg v6dist.Config, installCfg v6inst.Config, update
 				return nil, nil, fmt.Errorf("unable to update db: %w", err)
 			}
 			log.WithFields("error", err).Warn("error updating db")
+			updateErr = err
 		}
 		if !updated {
 			log.Debug("no db update found")
@@ -37,7 +40,7 @@ func LoadVulnerabilityDB(distCfg v6dist.Config, installCfg v6inst.Config, update
 
 	s := c.Status()
 	if s.Error != nil {
-		return nil, nil, s.Error
+		return nil, nil, statusError(s.Error, updateErr)
 	}
 
 	rdr, err := c.Reader()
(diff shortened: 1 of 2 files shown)


noqt (approved):
I reviewed this for NOQT at `eda8462ac1a5d44dec8c6827a539e2acbbeb112c`.

This fixes the confusing first-run path from #3399 without changing the fallback policy: when the update fails and no database exists, the returned error now preserves both the missing-database sentinel and the underlying download/update failure. That gives people the actionable cause instead of only “database does not exist.”

The focused regression and `./grype` package passed at the exact head. I also merged it cleanly onto current `main` (`f6f2ce0`) and passed the focused test, `./grype`, the CLI commands package, and vet there. Approved.

Tiny non-blocking cleanup: the test table’s `wantNil` field is unused. Hosted checks currently show DCO only, so the normal integration CI still matters.

PR #3601, opened 2026-07-25
Fix Rust false positives for local Cargo packages

3nesdeniz:
## Summary

Prevent Rust ecosystem matching from treating workspace and path packages in `Cargo.lock` as same-named crates.io packages.

## Motivation

In anchore/syft#3598, an in-repository crate named `telemetry` at version `0.1.0` is reported as the unrelated crates.io package and matched to `GHSA-hpcx-3pw8-g3j2`.

Cargo omits the `source` field for workspace and path packages. Grype already receives that typed `RustCargoLockEntry` metadata from Syft, so this change preserves the source as Rust-specific package metadata and uses it to distinguish a known-local package from a package whose provenance is simply unavailable.

Syft PR anchore/syft#5105 corrects the emitted Cargo PURL for the same packages. This Grype change is still necessary: an A/B test with the current vulnerability database confirmed that unmodified Grype continues to match the local package by Rust ecosystem, even when consuming a patched Syft document with no PURL.

## Changes

- map the Cargo.lock source into `pkg.RustMetadata` during package synthesis;
- expose the matching evidence through the JSON presenter's `metadataType` and `metadata` fields;
- keep the local-package decision inside the Rust matcher instead of the ecosystem-agnostic package API;
- skip Rust ecosystem/name matching only when typed Cargo.lock metadata explicitly identifies a local package;
- preserve legacy matching when Cargo.lock metadata is unavailable;
- preserve matching for registry, git, and Rust binary-audit packages; and
- preserve the existing opt-in CPE matching path for users with `match.rust.using-cpes: true`.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Testing

- `make unit` — passed, 74.3% coverage
- `make lint`
- `make static-analysis`
- `go test -race ./grype/matcher/... ./grype/pkg`
- `go test ./grype/pkg ./grype/matcher/rust ./grype/internal/packagemetadata ./grype/presenter/models`
- `go test -race ./grype/pkg ./grype/matcher/rust ./grype/internal/packagemetadata ./grype/presenter/models`
- `go generate ./grype/internal` followed by a clean worktree check
- exact CLI A/B with the current Grype DB:
  - unmodified Grype reports `GHSA-hpcx-3pw8-g3j2` for local `telemetry 0.1.0`;
  - this branch reports no finding for that package;
- regression coverage for crates.io, git, local Cargo.lock, metadata-free, binary-audit, opt-in CPE, and JSON serialization cases
- `git diff --check`

## Format boundary

The decision relies on typed `RustCargoLockEntry` source metadata. Native Syft JSON and CycloneDX retain enough provenance for this behavior. Generic SPDX conversion currently loses that source distinction, so an SPDX document may retain the legacy name/version match; addressing provenance loss across formats is outside this focused matcher fix.

## Checklist

Related to anchore/syft#3598.


grype/pkg/package.go
```diff
@@ -62,6 +62,11 @@ type Package struct {
 
 	// Related packages may be used for scanning
 	RelatedPackages map[artifact.RelationshipType][]*Package
+
+	// RustCargoLockSource is present only when the package originated from a
+	// Cargo.lock entry. An empty value distinguishes workspace and path packages
+	// from packages whose source metadata is unavailable.
+	RustCargoLockSource *string `json:"-"`
```
wagoodman: is there a reason why we have `json:"-"` here? if it affects matching then this evidence it useful generally in the json output.

wagoodman: Also this should be in the rust package metadata struct since its ecosystem-specific

grype/pkg/package.go
```diff
@@ -106,6 +114,12 @@ func New(p syftPkg.Package, enhancers ...Enhancer) Package {
 	return out
 }
 
+// IsRustCargoLockLocalPackage reports whether a package came from a workspace
+// or path entry in Cargo.lock. Cargo omits the source field for those entries.
+func (p Package) IsRustCargoLockLocalPackage() bool {
```
wagoodman: since this is an ecosystem specific question it should not be a helper on package (which is ecosystem agnostic)

grype/matcher/internal/common.go
```diff
@@ -19,16 +19,26 @@ func MatchPackageByEcosystemAndCPEs(store vulnerability.Provider, p pkg.Package,
...
-		cpeMatches, cpeIgnores, cpeErr := MatchPackageByCPEs(store, p, matcher)
-		switch {
-		case errors.Is(cpeErr, ErrEmptyCPEMatch):
-			log.Debugf("attempted CPE search on %s, which has no CPEs. Consider re-running with --add-cpes-if-none", p.Name)
-		case cpeErr != nil:
-			log.Debugf("could not match by package CPE (package=%+v): %v", p, cpeErr)
-		}
-		matches = append(matches, cpeMatches...)
-		ignored = append(ignored, cpeIgnores...)
+	cpeMatches, cpeIgnores, _ := MatchPackageByCPEsIfEnabled(store, p, matcher, includeCPEs)
```
spiffcs: I don't think we should be touching the `grype/matcher/internal/common.go` for this change. There is also no test change to go along with this diff.

From my initial read this just a cosmetic refactor. 
Are we actually fixing a bug in the original flow? 

If there is a subtle bug with the original can we illustrate the fix with a test?

grype/pkg/package.go
spiffcs: We should not be adding this field onto the main `Package` struct. This struct is kept pure of ecosystem implementation/proof details. For the correct inclusion see the `Metadata` field on package. It looks like we don't have a Rust specific metadata type over in grype now that we decorate the matches with so this PR would have to add that with the new field `RustCargoLockSource`. 

We also want to make sure this is included in the json so the consumer has context for the fields that influenced our matching decision.

spiffcs (commented):
Thanks @3nesdeniz for the PR here. I've left a few comments that should be addressed before we can get this merged. Thank you for the contribution.

3nesdeniz:
Thanks for the review — I moved the Cargo.lock source evidence into Rust-specific metadata and JSON, kept the decision inside the Rust matcher, restored the shared matcher unchanged, and added focused regression coverage. Unit, lint, static-analysis, and race checks pass locally.

3nesdeniz:
Hi @wagoodman @spiffcs — following up on the re-review when you have a chance. The Cargo.lock source evidence now lives in Rust-specific metadata/JSON as suggested, with the shared matcher restored unchanged. Happy to iterate further if anything else stands out.

3nesdeniz:
Rebased on latest `main`. To recap how the review feedback is addressed: the source evidence now lives on a Rust-specific `RustMetadata` type (`json:"rustCargoLockSource"`, so it surfaces in JSON output) rather than the ecosystem-agnostic `Package` struct; `grype/matcher/internal/common.go` is no longer touched; and there's matcher + presenter/model test coverage for the local-vs-registry Cargo distinction. `go test ./grype/pkg/... ./grype/matcher/rust/... ./grype/internal/packagemetadata/...` is green. Happy to adjust further.

noqt (approved):
Reviewing from NOQT. This closes the false-positive path at the right boundary: typed Cargo.lock metadata identifies workspace/path packages, and only those packages skip the crates.io ecosystem/name match. Registry, git, metadata-free, binary-audit, and opt-in CPE paths remain covered, while the source evidence is retained in JSON for consumers.

I checked exact head `3085d4c3420ec5a30920d94905a79b8bbbc5840a` with Go 1.26.3:

- the Rust matcher, metadata registry, presenter model, and changed `TestNew` package path pass;
- focused vet and diff hygiene pass;
- the broader `grype/pkg` run reaches only four unchanged Docker fixture failures because no Docker daemon is available.

The branch merges cleanly onto current main `f6f2ce0a2293e65c97bf5468926cb0c9ddad822f`; the same focused tests and vet pass there. Syft's identity half is already merged in anchore/syft#5105. I found no blocker.

PR #3584, opened 2026-07-18
Fix filtered vulnerability progress counts

Eljees:
## Summary

Fix vulnerability progress summaries after matcher-provided ignore filters remove matches.

## Motivation

Matcher-level ignore filters, including explicit distro NAKs, run after the progress monitor has already counted raw matches by severity and fix state. Grype updated the final match total after filtering but left the other counters unchanged, which could produce output such as `0 vulnerability matches`, non-zero severity counts, and a negative `not-fixed` count.

## Changes

- Reconcile total, fixed, and severity counters from the deduplicated, post-filter match set.
- Preserve the existing semantics for user-provided ignores and hard-coded dropped matches.
- Extend the APK NAK ownership regression test to cover the reported match, fixed, and critical counters.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Validation

- `go test ./grype ./cmd/grype/cli/ui`
- `.tool/golangci-lint run --tests=false` (0 issues in an LF checkout)

## Checklist

- [x] I have added unit tests that cover changed behavior
- [x] I have tested the affected matcher and UI packages
- [x] I have followed the existing code style

Fixes #3523

PR #3602, opened 2026-07-27
fix: exclude the distro Go toolchain stdlib module from OS packages

Eljees:
## Summary

- exclude an embedded Go module when a comprehensive distro package owns it
- add a regression case for a distro package and Go module whose versions differ

## Why

The overlap relationship already establishes that the distro package owns the embedded Go module. The previous version-prefix check ran before the comprehensive-distro ownership rule, so a legitimate owned module was retained when its module version differed from the distro package version.

This can leave duplicate package representations available for vulnerability matching even though the OS package is the authoritative representation.

## Validation

- `git diff --check`
- `go test ./grype/pkg` was attempted with `golang:1.26.3`; dependency downloads failed with upstream TLS handshake timeouts before the test package could run

### DIFF
--- a/grype/pkg/package.go
+++ b/grype/pkg/package.go
@@ -2,9 +2,11 @@ package pkg
 
 import (
 	"fmt"
+	"path"
 	"regexp"
 	"slices"
 	"sort"
+	"strconv"
 	"strings"
 
 	"github.com/anchore/grype/grype/distro"
@@ -256,9 +258,14 @@ func excludePackage(p *Package, parent *Package) bool {
 
 	comprehensiveDistroOwner := distroFeedIsComprehensive(parent.Distro) && isOSPackage(parent)
 
-	// Go module versions describe the embedded modules rather than the owning
-	// distro package, so they cannot be compared to determine ownership.
-	if comprehensiveDistroOwner && p.Type == syftPkg.GoModulePkg {
+	// Go module versions describe the embedded module rather than the owning
+	// distro package, so the version-similarity check below cannot recognize an
+	// owned duplicate. Only treat the module as a duplicate when it appears to
+	// be the distro package's own main module (e.g. the containerd package
+	// owning github.com/containerd/containerd); dependency modules embedded in
+	// the same binaries must stay matchable, since a distro feed does not track
+	// vulnerabilities for a third-party package's vendored dependencies.
+	if comprehensiveDistroOwner && p.Type == syftPkg.GoModulePkg && goModuleIsOwnersMainModule(p.Name, parent.Name) {
 		return true
 	}
 
@@ -320,6 +327,25 @@ var comprehensiveDistros = []distro.Type{
 	distro.Mariner,
 	distro.RedHat,
 	distro.Ubuntu,
+}
+
+// goModuleIsOwnersMainModule indicates whether a Go module found within an OS
+// package's files appears to be that package's own main module rather than an
+// embedded dependency. The comparison is based on the module path's base name
+// (accounting for /vN major-version suffixes), e.g. the "containerd" package
+// matches "github.com/containerd/containerd". The Go standard library
+// ("stdlib") is only considered owned by a distro's Go toolchain package.
+func goModuleIsOwnersMainModule(modulePath, packageName string) bool {
+	if modulePath == "stdlib" {
+		return strings.EqualFold(packageName, "go") || strings.EqualFold(packageName, "golang")
+	}
+	base := path.Base(modulePath)
+	if len(base) > 1 && base[0] == 'v' {
+		if _, err := strconv.Atoi(base[1:]); err == nil {
+			base = path.Base(path.Dir(modulePath))
+		}
+	}
+	return strings.EqualFold(base, packageName)
 }
 
 func isOSPackage(p *Package) bool {
--- a/grype/pkg/package.go
+++ b/grype/pkg/package.go
@@ -254,6 +254,14 @@ func excludePackage(p *Package, parent *Package) bool {
 	// python      3.9.2      binary
 	// python3.9   3.9.2-1    deb
 
+	comprehensiveDistroOwner := distroFeedIsComprehensive(parent.Distro) && isOSPackage(parent)
+
+	// Go module versions describe the embedded modules rather than the owning
+	// distro package, so they cannot be compared to determine ownership.
+	if comprehensiveDistroOwner && p.Type == syftPkg.GoModulePkg {
+		return true
+	}
+
 	// If the version is not approximately the same, keep both
 	if !strings.HasPrefix(parent.Version, p.Version) && !strings.HasPrefix(p.Version, parent.Version) {
 		return false
@@ -263,7 +271,7 @@ func excludePackage(p *Package, parent *Package) bool {
 	// for distros that have a comprehensive feed. That is, distros that list
 	// vulnerabilities that aren't fixed. Otherwise, the child package might
 	// be needed for matching.
-	if distroFeedIsComprehensive(parent.Distro) && isOSPackage(parent) && !isOSPackage(p) {
+	if comprehensiveDistroOwner && !isOSPackage(p) {
 		return true
 	}
 
--- a/grype/pkg/package_test.go
+++ b/grype/pkg/package_test.go
@@ -1523,11 +1523,18 @@ func Test_RemovePackagesByOverlap(t *testing.T) {
 			expectedPackages: []string{"rpm:python3-rpm@4.14.3-26.el8"},
 		},
 		{
-			name: "distro package owns a Go module with a different version",
+			name: "distro package owns its own Go main module with a different version",
 			sbom: withLinuxRelease(catalogWithOverlaps(
 				[]string{"deb:containerd@1.7.24~ds1-4ubuntu1", "go-module:github.com/containerd/containerd@1.7.22"},
 				[]string{"deb:containerd@1.7.24~ds1-4ubuntu1 -> go-module:github.com/containerd/containerd@1.7.22"}), "ubuntu"),
 			expectedPackages: []string{"deb:containerd@1.7.24~ds1-4ubuntu1"},
+		},
+		{
+			name: "distro package keeps embedded dependency Go modules",
+			sbom: withLinuxRelease(catalogWithOverlaps(
+				[]string{"deb:gitlab-ce@15.6.1-ce.0", "go-module:golang.org/x/crypto@v0.0.0-20220525230936-793ad666bf5e"},
+				[]string{"deb:gitlab-ce@15.6.1-ce.0 -> go-module:golang.org/x/crypto@v0.0.0-20220525230936-793ad666bf5e"}), "ubuntu"),
+			expectedPackages: []string{"deb:gitlab-ce@15.6.1-ce.0", "go-module:golang.org/x/crypto@v0.0.0-20220525230936-793ad666bf5e"},
 		},
 		{
 			name: "amzn linux doesn't remove packages in this way",
--- a/grype/pkg/package_test.go
+++ b/grype/pkg/package_test.go
@@ -1523,6 +1523,13 @@ func Test_RemovePackagesByOverlap(t *testing.T) {
 			expectedPackages: []string{"rpm:python3-rpm@4.14.3-26.el8"},
 		},
 		{
+			name: "distro package owns a Go module with a different version",
+			sbom: withLinuxRelease(catalogWithOverlaps(
+				[]string{"deb:containerd@1.7.24~ds1-4ubuntu1", "go-module:github.com/containerd/containerd@1.7.22"},
+				[]string{"deb:containerd@1.7.24~ds1-4ubuntu1 -> go-module:github.com/containerd/containerd@1.7.22"}), "ubuntu"),
+			expectedPackages: []string{"deb:containerd@1.7.24~ds1-4ubuntu1"},
+		},
+		{
 			name: "amzn linux doesn't remove packages in this way",
 			sbom: withLinuxRelease(catalogWithOverlaps(
 				[]string{"rpm:python3-rpm@4.14.3-26.el8", "python:rpm@4.14.3"},


spiffcs (commented):
@Eljees thanks for the PR. Can you provide an example of an incorrect grype vulnerability report that existed before this PR and a new report that highlights the fix?

```
The previous version-prefix check ran before the comprehensive-distro ownership rule, 
so a legitimate owned module was retained when 
--> its module version differed from the distro package version.

This can leave duplicate package representations available for vulnerability matching 
even though the OS package is the authoritative representation.
```

Let me know if I'm reading the above spec/description correctly.

If we find there a module version that is different from the distro package version then that suggests there were two modules discovered. We would have a package for both the distro owned package (distro package version) AND the other package that had a version mis match. 

If you have an example of an SBOM or vulnerability report that this fixes I'd like to look at those cases to check if we SHOULD be eliminating a package here, or if it's correct to have both.

Eljees:
Thanks @spiffcs — you were right to push on this. I compared before/after on the pinned quality SBOMs with the same imported DB (candidate build vs v0.116.0), and the evidence does not support the change in its current form.

Concrete counter-example from the quality corpus (`docker.io/gitlab/gitlab-ce@sha256:04d4219d…`, Ubuntu-based): the omnibus `gitlab-ce 15.6.1-ce.0` deb owns 1,836 embedded Go modules via ownership-by-file-overlap (gitaly, gitlab-shell, …). With this PR all of their go-module matches are dropped — 877 on this image alone, including CVE-2023-48795 (Terrapin) against `golang.org/x/crypto` — and 869 of them have no corresponding OS-package match, because Ubuntu's feed doesn't track vulnerabilities in a third-party deb's vendored Go code. Corpus-wide the change removes 1,271 go-module matches and adds none; 1,195 have no distro-level replacement. So these are real false negatives, not duplicate cleanup: `distroFeedIsComprehensive` tells us about the distro type, not whether the owning package is actually covered by the distro feed.

I'll narrow this to the case the issue is actually about: only exclude a go-module child when it is the parent package's own main module (e.g. the containerd rpm ↔ `github.com/containerd/containerd`), keeping dependency modules matchable. Converting to draft while I rework it.

Eljees:
Reworked in 056da3df as described above. The exclusion now only applies when the embedded Go module appears to be the owning package's *own main module* — the module path basename matches the package name (containerd rpm ↔ `github.com/containerd/containerd`), plus `stdlib` when it is owned by the distro's Go toolchain package. Dependency modules embedded in third-party packages stay matchable.

Validation against the pinned quality SBOMs with a shared DB (same binary pair, before = this branch's base):

- `gitlab-ce` (Ubuntu omnibus deb owning 1,836 embedded Go modules): **zero match changes** — the previous revision dropped 877 matches there, 869 of them with no distro-level replacement (incl. CVE-2023-48795 on `golang.org/x/crypto`).
- `azurelinux3` test image: only the `stdlib go1.22.5` ↔ `golang 1.22.5-1.azl3` rpm duplicates are removed (49 matches; 11 have alias-equivalent CVE matches on the rpm, the rest are absent from the Azure Linux feed, which is authoritative for a comprehensive distro).
- Unit tests updated: the containerd case still excludes; a new case asserts a `gitlab-ce`-style dependency module is kept.

Marking ready for review again — happy to adjust scope further if you'd prefer the stdlib case split out.

Eljees:
@spiffcs you're reading it right, and the answer is that the original patch was wrong — it produces false negatives, not duplicate cleanup.

I built this branch and its base with the same imported DB and ran both against the pinned quality-gate SBOM corpus. On the `gitlab/gitlab-ce` image the omnibus `gitlab-ce 15.6.1-ce.0` deb owns 1,836 embedded Go modules by file overlap (gitaly, gitlab-shell, ...). The first revision dropped 877 matches on that image alone — 869 of them with no corresponding OS-package match, including CVE-2023-48795 (Terrapin) on `golang.org/x/crypto`. Corpus-wide it removed 1,271 go-module matches and added none, 1,195 of which had no distro-level replacement.

So the assumption the patch leaned on doesn't hold: `distroFeedIsComprehensive` tells us about the distro type, not about whether the owning package's vendored code is covered by that feed. Ubuntu doesn't track vulnerabilities in a third-party deb's vendored Go code, so dropping those children loses real matches.

056da3df narrows it to the case the issue is actually about: an embedded Go module is only excluded when it looks like the owning package's own main module — module path basename matches the package name, e.g. the containerd rpm and `github.com/containerd/containerd` — plus `stdlib` when it's owned by the distro's Go toolchain package. Dependency modules embedded in a third-party package's binaries stay matchable.

Re-run on the same binary pair:

- `gitlab-ce`: zero match changes.
- `azurelinux3`: only the `stdlib go1.22.5` / `golang 1.22.5-1.azl3` rpm duplicates are removed (49 matches; 11 have alias-equivalent CVE matches on the rpm, the rest are absent from the Azure Linux feed).

Unit tests updated — the containerd case still excludes, and a new case asserts a gitlab-ce-style dependency module is kept. Happy to split the stdlib half out if you'd rather scope it tighter.

Eljees:
Ping on this one and #3584, #3606 - the oldest has been open since 18 July. They are independent; happy to rebase any of them.

Eljees:
@spiffcs — I said on #3584 that I'd stop pinging, so this isn't one. It's an offer to close.

Where this stands: you asked for a concrete before/after, and the honest answer was that my original patch was wrong. Measured against the pinned quality-gate SBOM corpus with a shared imported DB, it removed 1,271 go-module matches corpus-wide and added none — 1,195 with no distro-level replacement, including CVE-2023-48795 on `golang.org/x/crypto`. `distroFeedIsComprehensive` tells us about the distro type, not whether the owning package's vendored code is covered by that feed, which is where the assumption broke.

056da3d narrows it to the case the issue is actually about: exclude an embedded Go module only when it looks like the owning package's own main module, plus stdlib owned by the distro's Go toolchain package. On the same corpus that's zero change on gitlab-ce and only the azurelinux3 stdlib duplicates.

If that scope still isn't what you want, say so and I'll close it — no explanation needed, and I won't re-open the topic. If it is what you want, it's green and mergeable and I'll rebase whenever suits you.

noqt (changes requested):
Reviewing from NOQT at exact head `056da3df17b22cd5fd23b7b09d8a6acbc76d348d`.

The rework correctly preserves embedded dependency modules, but the basename heuristic still recreates the same coverage error for third-party OS packages whose own main module matches the package name. `comprehensiveDistroOwner` knows only the image distro and package type; it has no repository or source provenance for the deb/rpm.

I added a reviewer-only regression with an Ubuntu image, parent `deb:vault@1.16.2-1`, and child `go-module:github.com/hashicorp/vault@v1.16.2`. Current main keeps both packages. This head drops the Go module because `vault == path.Base(module)`, even though an Ubuntu feed cannot be assumed to cover an arbitrary third-party `vault` deb. That can hide the upstream Go findings the child exists to match.

Please retain the main-module child when versions differ unless there is evidence that the owning package is actually covered by the selected distro feed, and add this third-party-package boundary as a regression. If the toolchain/stdlib case has a stronger provenance signal, that can be split or narrowed independently.

The PR's focused overlap suite, affected-package vet, diff hygiene, and clean merge onto current main `f6f2ce0a2293e65c97bf5468926cb0c9ddad822f` all pass. The reviewer regression passes on current main and fails on this head; that false-negative boundary is the blocker.

Eljees:
@noqt I reproduced your scenario as a test case on this exact head, and you are right about the mechanics: with `deb:vault@1.16.2-1` owning `go-module:github.com/hashicorp/vault@v1.16.2` on an Ubuntu image, the module is removed.

Before concluding that is a defect introduced here, I ran the same scenario with a `binary` child instead of a Go module, on this branch's base (no changes of mine involved):

```
deb:vault@1.16.2-1  ->  binary:vault@1.16.2      (ubuntu)
actual: []string{"deb:vault@1.16.2-1"}
```

The binary child is absorbed too, by the pre-existing overlap rule (version-prefix similarity plus the comprehensive-distro check). The boundary you describe — an OS package the distro feed does not actually cover absorbing its child — is a property of the existing mechanism, because the SBOM carries no repository provenance for a deb/rpm; `comprehensiveDistroOwner` can only see the image's distro and the package type, exactly as you note. This change deliberately mirrors that same trade-off for the one Go-module shape where the version check cannot fire (module versions carry `v` prefixes and lack distro revisions), rather than inventing a new, stricter rule for Go modules than the one binaries already live under.

On the specific suggestion to retain the main-module child when versions differ: that gate would also retain `deb:containerd@1.7.24~ds1-4ubuntu1` vs `go-module:github.com/containerd/containerd@1.7.22`, since distro revisions and upstream module tags essentially never match textually — which reintroduces the duplicate this PR (and the linked issue) is about. A version gate that is loose enough to catch the containerd case is loose enough to drop your vault case as well; the two are not separable on version evidence alone.

Where the line could actually be drawn correctly is provenance: if syft ever records the package's origin repository, gating the exclusion (both the existing binary rule and this one) on "the owning package really comes from the distro" would fix your scenario and keep the duplicate suppression. That is out of scope for grype's data today.

If the maintainers prefer to be conservative here, the change can be narrowed to the `stdlib`-owned-by-toolchain case only, which is provenance-safe and still removes the bulk of the observed duplicates (the azurelinux3 stdlib/golang pairs). @spiffcs — happy to go either way; the corpus numbers for both variants are in the earlier comments.

noqt:
You're right about the current baseline. I reproduced the binary case on clean main `f6f2ce0`: `deb:vault@1.16.2-1` already absorbs `binary:vault@1.16.2` through the version-prefix rule.

That changes my framing, not the merge recommendation. The provenance gap is pre-existing; this PR expands that accepted false-negative trade-off to a new package class whose differing versions previously kept it matchable. Existing binary behavior isn't evidence that the expansion is safe.

Your conservative option is the right repair from NOQT's perspective: narrow this PR to the `stdlib`-owned-by-`go`/`golang` toolchain case and leave the main-module basename heuristic out until package-origin provenance exists. I withdraw my suggestion that a version gate can separate containerd from the third-party case—you've shown that it can't.

So the changes-requested verdict remains, but with the repair path narrowed to the stdlib slice you proposed. Thanks for testing the counterexample instead of hand-waving it away.

Eljees:
Agreed on the binary case being pre-existing behaviour. Since we now agree the stdlib-owned-by-toolchain slice is the provenance-safe subset, I can push a revision narrowed to just that case whenever @spiffcs wants it — the corpus impact of that slice is the azurelinux3 stdlib/golang pairs measured earlier, and it sidesteps the third-party-package boundary entirely. Holding until the maintainers weigh in, since the main-module-basename heuristic is the part still genuinely in dispute and that call is theirs to make.

Eljees:
@noqt @spiffcs pushed the narrowing in `ac3c8cb`, as a follow-up commit rather than a force push so the earlier rounds stay readable.

I said on 1 September that I would hold this until a maintainer weighed in, because the main-module heuristic was the part genuinely in dispute and that call is yours. Three weeks on, holding a changes-requested PR indefinitely does not serve anyone, and the narrowing is the conservative direction either way: it removes behaviour rather than adding it, and the removed half is recorded in this thread if you want it back.

What is left is the `stdlib`-owned-by-toolchain slice only. `isDistroToolchainStdlib` returns true for exactly one shape - the module path is `stdlib` and the owning package is named `go` or `golang`. Everything the basename heuristic used to catch now falls through to the existing version-similarity rule and is kept, which is what `main` does today.

The boundary is pinned by four cases in `Test_RemovePackagesByOverlap`. Run against three different `grype/pkg/package.go` with one unchanged test file, so the difference is the rule and nothing else:

**1. `upstream/main` at `41a43d8a`**

```
--- FAIL: Test_RemovePackagesByOverlap/distro_Go_toolchain_owns_the_standard_library_it_ships
    expected: []string{"deb:golang@1.22.5-1"}
    actual  : []string{"deb:golang@1.22.5-1", "go-module:stdlib@go1.22.5"}
```

One failure. `main` keeps the stdlib module next to the toolchain package that ships it, because `"1.22.5-1"` and `"go1.22.5"` are not prefixes of each other. The other three cases already pass on `main` - including the vault one, which is the point below.

**2. this branch at `056da3df`, the basename heuristic**

```
--- FAIL: Test_RemovePackagesByOverlap/distro_package_keeps_its_own_Go_main_module
    expected: []string{"go-module:github.com/hashicorp/vault@v1.16.2", "deb:vault@1.16.2-1"}
    actual  : []string{"deb:vault@1.16.2-1"}
```

One failure, and it is your scenario reproduced as a test rather than as prose: `deb:vault@1.16.2-1` owning `go-module:github.com/hashicorp/vault@v1.16.2` on ubuntu. `main` keeps both, this head drops the module. That is the false negative, in the repository's own harness.

**3. `ac3c8cb`, the narrowed rule**

```
ok  	github.com/anchore/grype/grype/pkg	0.054s
ok  	github.com/anchore/grype/grype/pkg	0.277s
```

All four cases pass, and so does the rest of `grype/pkg` (the second line is the package with `TestSyftLocationExcludes` left out - it shells out to `docker build`, which this sandbox has no daemon for).

The four cases, so the intent is legible without reading the diff:

| owner | module | narrowed rule | on `main` |
|---|---|---|---|
| `deb:golang@1.22.5-1` | `stdlib@go1.22.5` | dropped | kept |
| `deb:vault@1.16.2-1` | `stdlib@go1.22.5` | kept | kept |
| `deb:vault@1.16.2-1` | `github.com/hashicorp/vault@v1.16.2` | kept | kept |
| `deb:gitlab-ce@15.6.1-ce.0` | `golang.org/x/crypto@v0.0.0-2022...` | kept | kept |

So the only behaviour this PR now changes is the first row.

The corpus measurement from earlier still describes that row: on the pinned quality SBOMs the azurelinux3 `stdlib go1.22.5` / `golang 1.22.5-1.azl3` pairs are the 49 matches that go, and `gitlab-ce` sees zero match changes.

The title is updated to match the narrowed scope. If the maintainers would rather have the main-module half too once package-origin provenance exists, that is a separate change and a separate discussion - it is not smuggled into this one.

PR #3631, opened 2026-08-05
fix: make --by-cve merge pick a stable record

### LINKED ISSUE #3630: `--by-cve`: which advisory record survives the merge varies between runs
When more than one advisory record exists for the same CVE and package, `--by-cve` collapses them into a single match, but which record's metadata survives changes between identical runs. The finding set is stable; the content attached to each finding is not.

### Reproduction

Built from `main` at `970b4d10`. Database v6.1.9 (built 2026-08-05), verified identical across every run by comparing `descriptor.db.status` in the output.

`repro.cdx.json` — one component reachable through both a purl and a CPE:

```json
{
  "bomFormat": "CycloneDX",
  "specVersion": "1.5",
  "version": 1,
  "metadata": { "component": { "type": "application", "name": "repro", "version": "1.0.0" } },
  "components": [
    {
      "type": "library",
      "name": "stdlib",
      "version": "go1.21.0",
      "purl": "pkg:golang/stdlib@1.21.0",
      "cpe": "cpe:2.3:a:golang:go:1.21.0:*:*:*:*:*:*:*"
    }
  ]
}
```

```
export GRYPE_MATCH_GOLANG_ALWAYS_USE_CPE_FOR_STDLIB=true
for i in $(seq 1 20); do
  grype sbom:repro.cdx.json --by-cve -o json -q > run$i.json
done
```

`match.golang.always-use-cpe-for-stdlib` is only needed to give this particular component two sources; it defaults to `false` on main and defaulted to `true` in v0.112.0. Any package matched through both an ecosystem advisory and a CPE reaches the same code path.

**Observed**, comparing the 20 runs on `vulnerability.{namespace,dataSource,description,fix,urls,severity,relatedVulnerabilities}` keyed by (id, package, version):

- 20 distinct outputs from 20 runs; **0 of 190 run pairs are identical**
- 65 of 83 findings vary in `namespace`, `dataSource`, `fix` and `urls`; 24 also vary in `description`
- between 21 and 43 findings differ per run pair

`CVE-2023-45288` / `stdlib` `go1.21.0` alternates between:

| | run A | run B |
|---|---|---|
| `dataSource` | `https://go.dev/issue/65051` | `https://nvd.nist.gov/vuln/detail/CVE-2023-45288` |
| `namespace` | `govulndb:language:go` | `nvd:cpe` |
| `fix.available[].kind` | `advisory` (dated 2024-04-03) | `first-observed` (dated 2025-09-04) |

**Control:** the same command without `--by-cve`, 20 runs, is byte-identical in content across all 190 pairs. It is also stable when only one source matches the package, so the two competing records are what triggers it.

**Expected:** identical input and identical database produce identical output.

### Cause

Without `--by-cve` these are two separate findings — `GO-2024-2687` in `govulndb:language:go` found by `exact-direct-match`, and `CVE-2023-45288` in `nvd:cpe` found by `cpe-match` — and both report the same fix versions, `1.21.9, 1.22.2`.

1. `Matches.Enumerate` ranges over the `byFingerprint` map, so it yields matches in Go's randomized map order (`grype/match/matches.go:170-179`).

2. `findDBMatches` feeds that enumeration straight into the normalization loop, so which match becomes the merge receiver is decided by map order (`grype/vulnerability_matcher.go:154-158`).

3. `normalizeByCVE` rewrites `Vulnerability.ID` and `Vulnerability.Namespace` to the aliased CVE record's, and nothing else (`grype/vulnerability_matcher.go:362-364`). `Reference.Internal`, `Vulnerability.Metadata` and `Vulnerability.Fix` still describe the record the match was found in.

4. `Fingerprint` is `{vulnerabilityID, vulnerabilityNamespace, packageID}` plus the joined `Fix.Versions` (`grype/match/match.go:29-38`). Because both records report the same fix versions, the rewritten record now collides with the NVD record on the *full* fingerprint, so `addOrMerge` takes case A and calls `existingMatch.Merge(newMatch)` (`grype/match/matches.go:99-107`).

5. `Match.Merge` unions `RelatedVulnerabilities`, `Details` and `CPEs`, and leaves every other field of the receiver's `Vulnerability` alone (`grype/match/match.go:40-88`). So the record enumerated first keeps its `Metadata` and `Fix`; the other contributes only references and details.

The JSON presenter emits `m.Vulnerability.Metadata` directly when it is populated (`grype/presenter/models/match.go:51`), and `namespace`/`dataSource`/`description`/`urls`/`severity` come from that blob (`grype/presenter/models/vulnerability_metadata.go:62-73`), which is why the emitted `namespace` can read `govulndb:language:go` even though `Vulnerability.Namespace` was rewritten to `nvd:cpe`.

`Sorted()` sorts at the end, which stabilizes the *order* of the output but not the *content* of each merged match — which is why the finding set looks stable.

This is the same class of problem as #1046, which fixed related vulnerabilities and CPEs being lost during the merge and added the `// for stable output` sorts now in `Match.Merge`; the remaining fields were not covered.

### Impact

Anyone diffing grype output across runs — drift detection, cached comparisons, or a report expected to be reproducible from the same inputs — sees changes that did not come from a data or version change. The losing record is not preserved anywhere a consumer could recover it from, so the choice cannot be normalized downstream. When the NVD record wins, the curated fix metadata is what is lost: in the control run above, all 140 `fix.available` entries on the ecosystem records carry `kind: advisory` with real advisory dates, while all 171 on the NVD records carry `kind: first-observed`.

### Question

Two directions seem possible, and I do not want to assume which fits better:

- have the normalization loop (or `Enumerate` itself) iterate a sorted fingerprint list, which is narrow, though the survivor then becomes whichever fingerprint sorts first — for CVE-vs-GHSA/GO that is consistently the NVD record, so it would be stable but would settle on the record with the less specific fix data;
- pick the survivor in the case-A merge by an explicit rule, preferring the record matched directly against an ecosystem advisory over one reached only by CPE, which is stable and keeps the curated fix data, but changes merge semantics slightly outside `--by-cve`.

Would either be acceptable, or is there a preferred direction? Happy to open a PR.

Dashtid:
Fixes #3630.

`--by-cve` collapses several advisory records for one CVE and package into a single match, but which record's metadata survived changed between identical runs.

`normalizeByCVE` rewrites an ecosystem record's `ID` and `Namespace` to the CVE it aliases and nothing else — `Reference.Internal`, `Vulnerability.Metadata` and `Vulnerability.Fix` still describe the record the match was found in. When that rewritten record reports the same fix versions as the NVD record for the same CVE and package, the two collide on the full `Fingerprint` (which includes the joined `Fix.Versions`), so `addOrMerge` takes case A. `Match.Merge` unions `RelatedVulnerabilities`, `Details` and `CPEs` and leaves the receiver's remaining `Vulnerability` fields alone, and the receiver is whichever match was added first — which, for the normalization loop at `grype/vulnerability_matcher.go:156`, is Go's randomized map order, because `Matches.Enumerate` ranges over `byFingerprint`.

### Change

Pick the survivor explicitly in case A rather than letting it fall out of arrival order, preferring the record matched directly against an ecosystem advisory over one reached only by CPE.

Records that rank equally fall back to a stable comparison of the record they came from. A record whose metadata could not be resolved loses to one that can be identified, since `getCachedMetadata` failing leaves `Vulnerability.Metadata` nil (`grype/db/v6/vulnerability_provider.go:501-506`) and such a record has no description or data source to contribute.

Preferring the direct match is not only for stability. In the reproduction below, every `fix.available` entry on the ecosystem records carries `kind: advisory` with real advisory dates (140 entries), while every entry on the NVD records carries `kind: first-observed` (171 entries) — so when the NVD record won the coin flip, the curated fix metadata was what got dropped.

This is deliberately the second of the two directions raised in #3630. The narrower alternative — iterating a sorted fingerprint list — is also stable, but the survivor then becomes whichever fingerprint sorts first, which for CVE-vs-GHSA/GO is consistently the NVD record. Happy to switch to that if it is preferred; it is a small change from here.

### Verification

Built from this branch and from `main` at `970b4d10`, against the same database (v6.1.9, built 2026-08-05, confirmed identical across runs via `descriptor.db.status`), scanning a one-component SBOM matched through both a purl and a CPE. Comparing runs on `vulnerability.{namespace,dataSource,description,fix,urls,severity,relatedVulnerabilities}`:

| | before | after |
|---|---|---|
| distinct outputs from 20 runs | 20 | **1** |
| run pairs identical in content | 0 / 190 | **190 / 190** |
| findings whose content varies | 65 / 83 | **0 / 83** |

The finding set itself was already stable before and is unchanged after; the surviving record is now consistently the ecosystem one (`https://go.dev/issue/65051`, `kind: advisory`), and both match details are still unioned onto it.

### Tests

`TestMatches_MergeSurvivorIsIndependentOfAddOrder` in `grype/match/matches_test.go` builds the colliding pair directly — same fingerprint, different metadata and fix provenance — and adds them in both orders, asserting the same record wins each time. It is not a scan-twice-and-compare test, so it cannot pass by luck: on `main` the `nvd record added first` case fails deterministically.

`grype/match`, `grype`, `grype/presenter/models` and `cmd/grype/cli/...` pass. I could not get a clean signal from the full suite locally on Windows — a number of matcher and db packages fail there on `t.TempDir()` cleanup ("The process cannot access the file because it is being used by another process") and on `internal/dbtest` fixture tooling, identically with and without this change, and `make static-analysis` fails at `go mod tidy -diff` on a clean checkout too. Those all look environmental rather than related to this change, but CI is the better judge.

### DIFF
--- a/grype/match/matches.go
+++ b/grype/match/matches.go
@@ -98,13 +98,21 @@ func (r *Matches) addOrMerge(newMatch Match, newFp Fingerprint) {
 
 	if existingMatch, exists := r.byFingerprint[newFp]; exists {
 		// case A
-		if err := existingMatch.Merge(newMatch); err != nil {
-			log.WithFields("original", existingMatch.String(), "new", newMatch.String(), "error", err).Warn("unable to merge matches")
-			// at least capture the additional details
-			existingMatch.Details = append(existingMatch.Details, newMatch.Details...)
+		// Merge keeps the receiver's vulnerability metadata, and matches are not always added in a
+		// deterministic order (Matches.Enumerate ranges over a map), so pick the survivor explicitly
+		// rather than letting it fall out of arrival order.
+		keep, absorb := existingMatch, newMatch
+		if prefersVulnerabilityOf(newMatch, existingMatch) {
+			keep, absorb = newMatch, existingMatch
 		}
 
-		r.byFingerprint[newFp] = existingMatch
+		if err := keep.Merge(absorb); err != nil {
+			log.WithFields("original", keep.String(), "new", absorb.String(), "error", err).Warn("unable to merge matches")
+			// at least capture the additional details
+			keep.Details = append(keep.Details, absorb.Details...)
+		}
+
+		r.byFingerprint[newFp] = keep
 	} else if existingFingerprints, exists := r.byCoreFingerprint[newFp.coreFingerprint]; exists {
 		// case B
 		if !r.mergeCoreMatches(newMatch, newFp, existingFingerprints) {
@@ -192,6 +200,54 @@ func (r *Matches) Sorted() []Match {
 // Count returns the total number of matches in a result
 func (r *Matches) Count() int {
 	return len(r.byFingerprint)
+}
+
+// prefersVulnerabilityOf reports whether candidate's vulnerability metadata should survive a merge
+// with incumbent, given that the two share a fingerprint.
+//
+// Sharing a fingerprint does not mean the two matches describe the same upstream record. Normalizing
+// by CVE rewrites an ecosystem record's ID and namespace to the CVE it aliases, which makes it collide
+// with the NVD record for that CVE and package. Only the identity is rewritten: the description, data
+// source, severity and fix data still describe the record the match was originally found in, so which
+// of the two survives is visible in the output.
+//
+// Prefer the record found by matching the package directly against an ecosystem advisory. Those records
+// carry curated fix data, while a record reached only by CPE describes the same CVE more loosely.
+func prefersVulnerabilityOf(candidate, incumbent Match) bool {
+	candidateRank, incumbentRank := vulnerabilityRank(candidate), vulnerabilityRank(incumbent)
+	if candidateRank != incumbentRank {
+		return candidateRank > incumbentRank
+	}
+
+	// equally ranked records still need a stable winner, otherwise the result depends on which of the
+	// two happened to be added first. A record whose metadata could not be resolved has no
+	// description or data source to contribute, so it loses to one that can be identified.
+	candidateSource, incumbentSource := vulnerabilitySource(candidate), vulnerabilitySource(incumbent)
+	if candidateSource == "" || incumbentSource == "" {
+		return incumbentSource == "" && candidateSource != ""
+	}
+
+	return candidateSource < incumbentSource
+}
+
+// vulnerabilityRank orders matches by how directly the vulnerability record was tied to the package.
+func vulnerabilityRank(m Match) int {
+	switch {
+	case hasMatchType(m.Details, ExactDirectMatch):
+		return 2
+	case hasMatchType(m.Details, ExactIndirectMatch):
+		return 1
+	}
+	return 0
+}
+
+// vulnerabilitySource identifies the record a match's vulnerability metadata came from. Two matches
+// that share a fingerprint can still originate from different providers.
+func vulnerabilitySource(m Match) string {
+	if m.Vulnerability.Metadata == nil {
+		return ""
+	}
+	return m.Vulnerability.Metadata.Namespace + "|" + m.Vulnerability.Metadata.DataSource
 }
 
 func hasMatchType(details Details, ty Type) bool {
--- a/grype/match/matches_test.go
+++ b/grype/match/matches_test.go
@@ -702,3 +702,143 @@ func TestMatches_Add_Merge(t *testing.T) {
 		})
 	}
 }
+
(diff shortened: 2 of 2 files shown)


Dashtid:
Superseded by #3634, which landed the same preference inside `Match.Merge` rather than in `addOrMerge`. Closing.

The upstream version is the better placement. This PR chose the survivor at the one call site where the collision was observed; `compareRecordAuthority` decides it inside `Merge`, so it holds for every path that merges and for more than two records -- which `addOrMerge` alone does not, since a third record would be compared against a union that was never an input.

Verified on `f6f2ce0a` before closing: the reproduction from #3630 now gives 45 of 45 identical run pairs, against 0 of 190 when it was filed. Detail in the issue.

PR #3632, opened 2026-08-05
fix: don't reorder a match's fix versions while sorting

Dashtid:
`ByElements.Less` sorts both matches' `Fix.Versions` in place before joining them for comparison:

```go
fixVersions1 := m[i].Vulnerability.Fix.Versions
fixVersions2 := m[j].Vulnerability.Fix.Versions
sort.Strings(fixVersions1)
sort.Strings(fixVersions2)
```

The local variables copy only the slice header, so `sort.Strings` reorders the slice each `Match` actually holds.

### What this causes

`Less` only reaches that comparison when two matches tie on vulnerability ID, package name, version and type, so the effect is confined to findings that collide there. For those, `Sorted()` hands back matches whose `Fix.Versions` are in sorted rather than stored order, and the presenters emit them as they find them (`grype/presenter/cyclonedx/vulnerability.go:206` iterates the slice directly). A comparison function rewriting its operands is the underlying problem; the reordered output is how it shows.

`Sorted()` is called per matcher in `searchDBForMatches`, so this is on the normal scan path rather than test-only.

### Change

Compare a sorted copy instead of sorting in place. The ordering `Less` produces is unchanged — it still compares the sorted, comma-joined versions — only the mutation goes away.

The test pins that both operands keep their original version order through a `Less` call and through `Sorted()`. Both subtests fail on unpatched `main` and pass with the change.

### History

As first opened, this PR also described three consequences through `Fingerprint`, which at the time joined `Fix.Versions`: a `Less` call changing a match's fingerprint, `byFingerprint` keys drifting from their values after `Sorted()`, and deduplication depending on a prior sort. #3634 removed the fix versions from `Fingerprint`, so none of those is reachable any more and the tests that pinned them passed on unpatched `main`. Rebased and cut back to the mutation that still discriminates.

Found while looking at #3630, but independent of it — different file, different mechanism, and it stands on its own.

### Verification

`go test ./grype/match/...` and `go vet` pass on the rebased head. Reverting `sort.go` to `main` makes both subtests fail. `golangci-lint` panics on package load with the toolchain here (Go 1.26.5), so CI is the better judge of lint.

### DIFF
--- a/grype/match/sort.go
+++ b/grype/match/sort.go
@@ -1,6 +1,7 @@
 package match
 
 import (
+	"slices"
 	"sort"
 	"strings"
 )
@@ -22,12 +23,8 @@ func (m ByElements) Less(i, j int) bool {
 				if m[i].Package.Type == m[j].Package.Type {
 					// this is an approximate ordering, but is not accurate in terms of semver and other version formats
 					// but stability is what is important here, not the accuracy of the sort.
-					fixVersions1 := m[i].Vulnerability.Fix.Versions
-					fixVersions2 := m[j].Vulnerability.Fix.Versions
-					sort.Strings(fixVersions1)
-					sort.Strings(fixVersions2)
-					fixStr1 := strings.Join(fixVersions1, ",")
-					fixStr2 := strings.Join(fixVersions2, ",")
+					fixStr1 := sortedFixVersions(m[i].Vulnerability.Fix.Versions)
+					fixStr2 := sortedFixVersions(m[j].Vulnerability.Fix.Versions)
 
 					if fixStr1 == fixStr2 {
 						loc1 := m[i].Package.Locations.ToSlice()
@@ -57,4 +54,20 @@ func (m ByElements) Less(i, j int) bool {
 // Swap swaps the elements with indexes i and j.
 func (m ByElements) Swap(i, j int) {
 	m[i], m[j] = m[j], m[i]
+}
+
+// sortedFixVersions renders fix versions in a stable order for comparison.
+//
+// This deliberately works on a copy. The slice belongs to the Match being compared, and the
+// presenters emit Fix.Versions in the order the match holds them, so sorting it in place here would
+// let a comparison rewrite what a caller later reads back from its own match.
+func sortedFixVersions(versions []string) string {
+	if len(versions) < 2 {
+		return strings.Join(versions, ",")
+	}
+
+	sorted := slices.Clone(versions)
+	slices.Sort(sorted)
+
+	return strings.Join(sorted, ",")
 }
--- a/grype/match/sort_test.go
+++ b/grype/match/sort_test.go
@@ -1,0 +1,55 @@
+package match
+
+import (
+	"testing"
+
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/anchore/grype/grype/pkg"
+	"github.com/anchore/grype/grype/vulnerability"
+	syftPkg "github.com/anchore/syft/syft/pkg"
+)
+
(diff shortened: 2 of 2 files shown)


Dashtid:
#3634 (merged 2026-08-25) dropped the joined fix versions from `Fingerprint`, which invalidates most of the impact described in this PR's body. Re-checked against `f6f2ce0a`.

The mutation is still there: `ByElements.Less` calls `sort.Strings` on `m[i].Vulnerability.Fix.Versions` directly, so comparing two matches still reorders slices the caller owns, and `Sorted()` still leaves `fix.versions` in sorted rather than database order in the emitted document.

What no longer holds is everything downstream of it. `Fingerprint` is now `(vulnerabilityID, vulnerabilityNamespace, packageID)` with `vulnerabilityFixes` removed, so a `Less` call cannot change a fingerprint, the `byFingerprint` index cannot drift from its keys, and deduplication no longer depends on the order fix versions are listed in.

Running this PR's test against **unpatched** `f6f2ce0a`, only one of the three subtests still discriminates:

- `Less leaves fix versions and fingerprints alone` -- fails, on the two `Fix.Versions` assertions
- `Sorted keeps every match consistent with the key it is stored under` -- passes unpatched
- `deduplication does not depend on a prior Sorted call` -- passes unpatched

So as written this would land two tests that cannot fail, which I would rather not leave in the tree.

I can rebase and cut it back to the mutation alone -- a comparison function writing to its operands, with the visible effect now limited to fix-version ordering in the output -- or close it if that on its own is not worth the change. Which would you prefer?

noqt (changes requested):
The comparator fix is still useful, but this needs a rebase and a narrower regression before merge.

I replayed exact head `357b142ac90ada3619db041854aeaf1312a7c951` and current main `f6f2ce0a2293e65c97bf5468926cb0c9ddad822f` independently. On current main, only the caller-owned slice mutation remains: `Less` turns `2.0.0, 1.0.0` into `1.0.0, 2.0.0`. The fingerprint-invariant and deduplication cases already pass because #3634 removed fix versions from `Fingerprint`.

Please rebase, keep the copy-before-sort change, and cut the test to the mutation that still discriminates. The helper comment should also stop saying `Fingerprint` joins `Fix.Versions`, because that is no longer true on main. A focused check that both operands retain their original version order after `Less` (and, if useful, after `Sorted`) would pin the remaining behavior cleanly.

Exact-head `go test ./grype/match` and diff hygiene pass. Reviewing from NOQT.

Dashtid:
Rebased onto `f6f2ce0a` and cut back to the mutation, at `0c06b6a9`.

The test now checks one thing: both operands keep their original `Fix.Versions` order, once through a direct `Less` call and once through `Sorted()`. The fixtures tie on ID, package name, version and type so the comparison is guaranteed to reach the fix-versions branch, and differ in namespace so they stay two distinct matches. Both subtests fail with `sort.go` reverted to `main` and pass with the change.

The helper comment and the commit message no longer mention `Fingerprint`; the PR body now records what was true when it was opened and what #3634 changed, rather than restating the old claims.

`grype/match` tests and `go vet` pass on the rebased head. `golangci-lint` panics on package load with the toolchain here, so I have left lint to CI. Is there a reason the `Sorted()` subtest should key on something other than namespace?

Dashtid:
Status, since nothing has moved here since 2026-08-31: head `0c06b6a9` is the mutation-only version. The three workflows for that head (`Validations`, `Validate GitHub Actions`, `CodeQL`) have sat at `action_required` since 2026-08-31 with no jobs executed, so DCO is the only check that has run on this PR since the rebase. The branch is 23 commits behind `main` with no conflicts; happy to rebase if a fresh head is preferred. Is anything outstanding on the change itself, or is this waiting on a workflow approval?

PR #3633, opened 2026-08-06
fix: mark NVD records as disputed status based on cveTags

### LINKED ISSUE #3248: Set disputed status for grype db vulnerability handles where the entry is flagged as disputed in nvd
**What happened**:

Currently, there are no vulnerability handle records from the `nvd` provider that are marked with a status of `disputed` even though there are records tagged as disputed from the NVD api.  This is because currently the grype vulnerability handle status is only calculated from the [vulnStatus](https://github.com/anchore/grype/blob/9d54499415ad6f34426ce82082fa205a2e6bc45a/grype/db/internal/provider/unmarshal/nvd/cve.go#L60) property of the NVD record and for that particular API the status isn't used to mark as disputed, but rather a `"disputed"` label is added into `cveTags` from https://csrc.nist.gov/schema/nvd/api/2.0/cve_api_json_2.0.schema.

So, we should update the NVD unmarshaller at https://github.com/anchore/grype/blob/9d54499415ad6f34426ce82082fa205a2e6bc45a/grype/db/internal/provider/unmarshal/nvd/cve.go#L43 to capture the `cveTags` and then consider them when determining status in https://github.com/anchore/grype/blob/9d54499415ad6f34426ce82082fa205a2e6bc45a/grype/db/v6/build/transformers/nvd/transform.go#L96

NVD records flagged as disputed should have the vulnerability handle status set to `disputed` in the grype database

https://nvd.nist.gov/vuln/detail/CVE-2001-1517 as an example of an  NVD CVE entry that is marked as `disputed` and which we get the following vunnel result for:

```json
{
  "schema": "https://raw.githubusercontent.com/anchore/vunnel/main/schema/vulnerability/nvd/schema-1.0.1.json",
  "identifier": "2001/cve-2001-1517",
  "item": {
    "cve": {
      "id": "CVE-2001-1517",
      "sourceIdentifier": "cve@mitre.org",
      "published": "2001-12-31T05:00:00.000",
      "lastModified": "2025-04-03T01:03:51.193",
      "vulnStatus": "Deferred",
      "cveTags": [
        {
          "sourceIdentifier": "cve@mitre.org",
          "tags": [
            "disputed"
          ]
        }
      ],
      "descriptions": [
        {
          "lang": "en",
          "value": "RunAs (runas.exe) in Windows 2000 stores cleartext authentication information in memory, which could allow attackers to obtain usernames and passwords by executing a process that is allocated the same memory page after termination of a RunAs command.  NOTE: the vendor disputes this issue, saying that administrative privileges are already required to exploit it, and the original researcher did not respond to requests for additional information"
        }
      ],
      "metrics": {
        "cvssMetricV2": [
          {
            "source": "nvd@nist.gov",
            "type": "Primary",
            "cvssData": {
              "version": "2.0",
              "vectorString": "AV:L/AC:L/Au:N/C:P/I:N/A:N",
              "baseScore": 2.1,
              "accessVector": "LOCAL",
              "accessComplexity": "LOW",
              "authentication": "NONE",
              "confidentialityImpact": "PARTIAL",
              "integrityImpact": "NONE",
              "availabilityImpact": "NONE"
            },
            "baseSeverity": "LOW",
            "exploitabilityScore": 3.9,
            "impactScore": 2.9,
            "acInsufInfo": false,
            "obtainAllPrivilege": false,
            "obtainUserPrivilege": false,
            "obtainOtherPrivilege": false,
            "userInteractionRequired": false
          }
        ]
      },
      "weaknesses": [
        {
          "source": "nvd@nist.gov",
          "type": "Primary",
          "description": [
            {
              "lang": "en",
              "value": "NVD-CWE-Other"
            }
          ]
        }
      ],
      "configurations": [
        {
          "nodes": [
            {
              "operator": "OR",
              "negate": false,
              "cpeMatch": [
                {
                  "vulnerable": true,
                  "criteria": "cpe:2.3:o:microsoft:windows_2000:*:*:*:*:*:*:*:*",
                  "matchCriteriaId": "4E545C63-FE9C-4CA1-AF0F-D999D84D2AFD"
                },
                {
                  "vulnerable": true,
                  "criteria": "cpe:2.3:o:microsoft:windows_2000:*:sp1:*:*:*:*:*:*",
                  "matchCriteriaId": "294EBA01-147B-4DA0-937E-ACBB655EDE53"
                },
                {
                  "vulnerable": true,
                  "criteria": "cpe:2.3:o:microsoft:windows_2000:*:sp2:*:*:*:*:*:*",
                  "matchCriteriaId": "4E8B7346-F2AA-434C-A048-7463EC1BB117"
                }
              ]
            }
          ]
        }
      ],
      "references": [
        {
          "url": "http://archives.neohapsis.com/archives/vulnwatch/2001-q4/0041.html",
          "source": "cve@mitre.org",
          "tags": [
            "Patch",
            "Vendor Advisory"
          ]
        },
        {
          "url": "http://cert.uni-stuttgart.de/archive/bugtraq/2001/11/msg00100.html",
          "source": "cve@mitre.org"
        },
        {
          "url": "http://www.iss.net/security_center/static/7531.php",
          "source": "cve@mitre.org",
          "tags": [
            "Patch"
          ]
        },
        {
          "url": "http://www.securityfocus.com/bid/3184",
          "source": "cve@mitre.org"
        },
        {
          "url": "http://archives.neohapsis.com/archives/vulnwatch/2001-q4/0041.html",
          "source": "af854a3a-2127-422b-91ae-364da2661108",
          "tags": [
            "Patch",
            "Vendor Advisory"
          ]
        },
        {
          "url": "http://cert.uni-stuttgart.de/archive/bugtraq/2001/11/msg00100.html",
          "source": "af854a3a-2127-422b-91ae-364da2661108"
        },
        {
          "url": "http://www.iss.net/security_center/static/7531.php",
          "source": "af854a3a-2127-422b-91ae-364da2661108",
          "tags": [
            "Patch"
          ]
        },
        {
          "url": "http://www.securityfocus.com/bid/3184",
          "source": "af854a3a-2127-422b-91ae-364da2661108"
        }
      ]
    }
  }
}
```

pujitha24:
Motivation:
NVD API records can be labeled "disputed" via the cveTags field
(https://csrc.nist.gov/schema/nvd/api/2.0/cve_api_json_2.0.schema)
independently of the vulnStatus field. Grype's NVD unmarshaller never
captured cveTags, and the v6 transformer only derived vulnerability
handle status from vulnStatus, so these records ended up with an
incorrect status (e.g. unknown) instead of disputed. For example,
CVE-2001-1517 has vulnStatus "Deferred" (not a status grype maps to
anything) but cveTags contains a "disputed" tag, so its grype db
status should be disputed.

Approach:
- Add a CveTags field (and CveTag struct) to CveItem in
  grype/db/internal/provider/unmarshal/nvd/cve.go, matching the NVD
  API schema shape, plus an IsDisputed() helper that does a
  case-insensitive scan for a "disputed" tag.
- In grype/db/v6/build/transformers/nvd/transform.go, getVulnStatus
  now checks IsDisputed() before falling back to the vulnStatus-based
  switch, since a record can carry a disputed cveTag while vulnStatus
  is any other value (e.g. "Analyzed" or an unrecognized one like
  "Deferred"). This means a disputed tag takes precedence over any
  other vulnStatus mapping, including "rejected" - this is an
  intentional choice since disputed and rejected are independent,
  non-mutually-exclusive signals in the NVD data and disputed is the
  more actionable/specific one to surface.

Validation:
Ran `go build ./...` (passes) and `go test ./grype/db/...` (all
packages pass), including the new grype/db/v6/build/transformers/nvd
TestGetVulnStatus table test, which reproduces the exact scenario
from the issue (vulnStatus "Deferred" + a disputed cveTag ->
db.VulnerabilityDisputed), matching CVE-2001-1517's real NVD data.
Also fed the exact JSON payload from the issue through the updated
CveItem JSON decoding (in a local scratch test, not included in this
commit) and confirmed CveTags populates correctly and IsDisputed()
returns true.

This change affects the shared unmarshal/transform packages used to
build the grype v6 vulnerability database; it does not alter grype's
own scan-time behavior in this repo, only the derived status of
disputed NVD records once a database is rebuilt with this logic.

Report: https://github.com/anchore/grype/issues/3248
Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

Fixes #3248

### DIFF
--- a/grype/db/internal/provider/unmarshal/nvd/cve.go
+++ b/grype/db/internal/provider/unmarshal/nvd/cve.go
@@ -2,6 +2,7 @@ package nvd
 
 import (
 	"sort"
+	"strings"
 
 	"github.com/Masterminds/semver/v3"
 	"github.com/jinzhu/copier"
@@ -47,6 +48,7 @@ type CveItem struct {
 	// CisaRequiredAction    *string         `json:"cisaRequiredAction,omitempty"`
 	// CisaVulnerabilityName *string         `json:"cisaVulnerabilityName,omitempty"`
 	Configurations []Configuration `json:"configurations,omitempty"`
+	CveTags        []CveTag        `json:"cveTags,omitempty"`
 	Descriptions   []LangString    `json:"descriptions"`
 	// EvaluatorComment      *string         `json:"evaluatorComment,omitempty"`
 	// EvaluatorImpact       *string         `json:"evaluatorImpact,omitempty"`
@@ -169,6 +171,15 @@ type Weakness struct {
 	Type        string       `json:"type"`
 }
 
+// CveTag captures source-provided labels for a CVE record, such as "disputed" or "unsupported-when-assigned".
+// see https://csrc.nist.gov/schema/nvd/api/2.0/cve_api_json_2.0.schema
+type CveTag struct {
+	SourceIdentifier string   `json:"sourceIdentifier,omitempty"`
+	Tags             []string `json:"tags,omitempty"`
+}
+
+const disputedTag = "disputed"
+
 func (o CveItem) Description() string {
 	for _, d := range o.Descriptions {
 		if d.Lang == englishLanguage {
@@ -176,6 +187,19 @@ func (o CveItem) Description() string {
 		}
 	}
 	return ""
+}
+
+// IsDisputed returns true if any source has tagged this CVE as "disputed" via cveTags. Unlike vulnStatus, this
+// tag is not mutually exclusive with other statuses (e.g. a CVE can be "Analyzed" and disputed at the same time).
+func (o CveItem) IsDisputed() bool {
+	for _, t := range o.CveTags {
+		for _, tag := range t.Tags {
+			if strings.EqualFold(tag, disputedTag) {
+				return true
+			}
+		}
+	}
+	return false
 }
 
 type CvssSummary struct {
--- a/grype/db/internal/provider/unmarshal/nvd/cve_test.go
+++ b/grype/db/internal/provider/unmarshal/nvd/cve_test.go
@@ -1,6 +1,7 @@
 package nvd
 
 import (
+	"encoding/json"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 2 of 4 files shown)


pujitha24:
Just checking in on this — it'\''s rebased and green, happy to make any changes that would help review go smoother.

noqt (changes requested):
Reviewing from NOQT at exact head `45a1a0eab06b6f1fa16d971d9545948e940fcd90`.

The `cveTags` ingestion and the Deferred/Analyzed handling are useful. I found one correctness blocker in the precedence, though: `getVulnStatus` returns `Disputed` before examining `vulnStatus`, so a record with `vulnStatus: "Rejected"` plus a disputed tag becomes `Disputed`.

That reverses Grype's own safety meaning: [`VulnerabilityRejected`](https://github.com/anchore/grype/blob/f6f2ce0edb7d2fc9de89e022dd9e95df63f57e34/grype/db/v6/enumerations.go#L18-L22) means the record must not be acted upon, while `VulnerabilityDisputed` may still be actionable. A reviewer-only table case expecting `Rejected` fails with `Disputed` at this head.

Please give `Rejected` precedence, then apply the disputed tag to non-rejected statuses, and add the `Rejected + disputed` regression. The focused unmarshal/transformer tests and vet pass otherwise, and this head merges cleanly onto current `main`.

Non-blocking: promoting the scratch JSON-decode check into a permanent test would protect the new ingestion boundary.

noqt (approved):
Reviewing the repaired head `930aa007268246b780355c844452116461bca74e` from NOQT.

This closes my precedence blocker cleanly. `Rejected` is now handled before the disputed-tag override, and the added `Rejected + disputed` regression passes while Deferred/Analyzed disputed records retain the intended behavior.

I also independently decoded the documented NVD `cveTags` JSON shape and confirmed `IsDisputed()` sees it. The focused unmarshal/transformer tests and vet pass on the exact head and after a clean merge onto current `main`; diff hygiene is clean.

The wider `grype/db/...` run reached only unrelated Windows-host limitations (locked SQLite/temp files, unavailable `zstd`/`make`, symlink privilege, and path separators), not failures in the changed packages.

I found no remaining blocker. Thanks for the quick, focused repair.

noqt (approved):
Re-reviewing the newest head `577dd31f20ccff84fde6f67afc82025e1ceeb4df` from NOQT.

The committed JSON regression is the right follow-through on the earlier suggestion. It exercises the documented NVD `cveTags` shape end to end, verifies the decoded source/tag values through `IsDisputed()`, and covers the no-tag case. The rejected-status precedence repair remains intact.

The focused unmarshal and transformer packages pass with vet on the exact head and after a clean merge onto current `main`; diff hygiene is clean. I found no new blocker and am renewing the approval at this head.

PR #3634, opened 2026-08-06
feat(apk/matcher): apk matcher does alias aware aggregation

### LINKED ISSUE #3630: `--by-cve`: which advisory record survives the merge varies between runs

crosleyzack:
### What

Adds alias aware deduplication in apk matcher

### Why

Currently, apk matcher will only identify vulnerabilities as duplicates if they share the exact same vuln id (CVE-2029-1234). This however doesn't work with the Chainguard OSV feed, which uses CGA type vuln ids, resulting in two errors cases:
1. Double reporting of vulns
2. CGA fix events not suppressing upstream vulns (CVE / GHSA / GO)

This PR copies the methodology used by rpm matcher and the existing result set object to perform the deduplication, resulting in correct output.

### Example

Take this below example using grype built with this branch and a build of grype-db using OSV feed:

```
✦3 ❯ CG_GRYPE_GRYPE_BIN="/home/zackary-crosley/anchore/grype/bin/grype" cg-grype --osv-db ghcr.io/chainguard-images/scanner-test:unfixed-vulnerabilities-many-chainguard -q

NAME           INSTALLED  FIXED IN   TYPE  VULNERABILITY       SEVERITY  EPSS          RISK
git            2.39.0-r0             apk   CVE-2025-48384      High      2.8% (85th)   81.4    KEV
git            2.39.0-r0  2.39.1-r0  apk   CGA-vg9m-2fc2-xv7w  Critical  56.3% (98th)  53.0
openssl        3.0.7-r0   3.3.2-r0   apk   CGA-92jc-655c-2rx8  High      66.6% (99th)  49.9
openssl        3.0.7-r0   3.1.0-r0   apk   CGA-p5cv-98wp-p27p  High      59.5% (99th)  44.3
openssl        3.0.7-r0   3.1.1-r0   apk   CGA-3pjw-698m-wq8g  Medium    75.1% (99th)  43.2
git            2.39.0-r0  2.39.1-r0  apk   CGA-wh8v-26rh-7874  Critical  44.3% (98th)  41.6
git            2.39.0-r0  2.40.1-r0  apk   CGA-5h2q-xgq9-whpr  High      51.9% (98th)  38.9
openssl        3.0.7-r0   3.6.1-r0   apk   CGA-c6p4-q6jv-qjrr  High      47.6% (98th)  38.8
openssl        3.0.7-r0   3.3.0-r5   apk   CGA-g7wx-8452-qfp7  Medium    54.0% (98th)  29.4
git            2.39.0-r0  2.45.1-r0  apk   CGA-9pfc-qx4h-fmhh  Critical  29.2% (97th)  26.3
openssl        3.0.7-r0   3.1.0-r0   apk   CGA-3pmc-pfcf-4f9g  High      20.4% (97th)  15.3
openssl        3.0.7-r0   3.1.0-r0   apk   CGA-c376-f4w2-qjmv  Medium    16.2% (96th)  8.8
openssl        3.0.7-r0   3.3.1-r3   apk   CGA-r27g-x88q-7j3g  Critical  5.6% (92nd)   5.1
git            2.39.0-r0  2.40.1-r0  apk   CGA-w9c5-9xmc-q4r8  High      6.1% (92nd)   4.4
openssl        3.0.7-r0   3.6.3-r0   apk   CGA-4wjr-v347-69w7  High      5.2% (91st)   4.3
openssl        3.0.7-r0   3.1.0-r0   apk   CGA-ww6f-6vqq-vwp5  High      4.5% (90th)   3.4
openssl        3.0.7-r0   3.1.1-r3   apk   CGA-vxcm-3662-jgv4  Medium    6.5% (93rd)   3.4
openssl        3.0.7-r0   3.1.0-r1   apk   CGA-x7mr-4j9g-2xqg  High      3.7% (88th)   2.7
openssl        3.0.7-r0   3.3.3-r0   apk   CGA-5hfj-hjrq-fjc2  Medium    5.8% (92nd)   2.7
openssl        3.0.7-r0   3.6.1-r0   apk   CGA-826m-52gp-hm9m  Medium    4.5% (90th)   2.5
openssl        3.0.7-r0   3.1.4-r0   apk   CGA-p3hq-hch9-gc3f  High      3.3% (87th)   2.5
openssl        3.0.7-r0   3.1.4-r1   apk   CGA-xx8w-f5jh-v433  Medium    4.5% (90th)   2.3
openssl        3.0.7-r0   3.3.1-r0   apk   CGA-47j4-6g2h-8vfm  High      2.9% (85th)   2.2
openssl        3.0.7-r0   3.2.1-r0   apk   CGA-h3v7-jg5r-3grr  Medium    3.2% (86th)   1.7
openssl        3.0.7-r0   3.1.1-r4   apk   CGA-jwrp-9qw3-4739  Medium    3.0% (86th)   1.6
openssl        3.0.7-r0   3.1.0-r0   apk   CGA-9f26-9xq6-w8hx  High      1.8% (76th)   1.4
openssl        3.0.7-r0   3.1.0-r0   apk   CGA-qpqj-hwhh-xg9x  High      1.8% (76th)   1.4
openssl        3.0.7-r0   3.1.0-r0   apk   CGA-hh5v-rr6v-w7p9  High      1.8% (76th)   1.4
openssl        3.0.7-r0   3.4.1-r0   apk   CGA-95p7-74f6-gq84  Medium    2.4% (82nd)   1.4
openssl        3.0.7-r0   3.2.1-r0   apk   CGA-22p8-hhxj-9pcc  Medium    2.3% (81st)   1.3
openssl        3.0.7-r0   3.5.4-r0   apk   CGA-c4v2-6rpm-vq95  High      1.7% (75th)   1.3
openssl        3.0.7-r0   3.5.4-r0   apk   CGA-cpj2-wf29-8hr5  Medium    2.2% (81st)   1.3
openssl        3.0.7-r0   3.3.0-r5   apk   CGA-fcf2-xh9w-8793  Medium    2.3% (81st)   1.3
openssl        3.0.7-r0   3.5.4-r0   apk   CGA-vgxf-c52j-3xr7  Medium    2.3% (81st)   1.2
git            2.39.0-r0  2.45.1-r0  apk   CGA-5rfv-8hjv-cw3q  High      1.3% (67th)   1.0
openssl        3.0.7-r0   3.0.7-r1   apk   CGA-w8m6-wvqh-fpxh  High      1.2% (65th)   0.9
openssl        3.0.7-r0   3.6.2-r0   apk   CGA-85p3-36wg-c582  High      1.2% (65th)   0.9
openssl        3.0.7-r0   3.6.3-r0   apk   CGA-wpj5-f6vw-h7r5  High      1.2% (64th)   0.9
openssl        3.0.7-r0   3.1.0-r2   apk   CGA-8954-r89r-f89v  Medium    1.6% (73rd)   0.8
openssl        3.0.7-r0   3.6.2-r0   apk   CGA-2c5c-rx22-cxxf  High      1.1% (61st)   0.8
openssl        3.0.7-r0   3.6.3-r0   apk   CGA-xm6c-wcqr-7hgw  High      1.0% (60th)   0.8
openssl        3.0.7-r0   3.6.2-r0   apk   CGA-r9mm-rw3c-wqh3  High      1.0% (60th)   0.8
openssl        3.0.7-r0   3.6.2-r0   apk   CGA-rff8-35jv-r39r  High      1.0% (60th)   0.8
openssl        3.0.7-r0   3.6.3-r0   apk   CGA-r88x-fhmc-m5p6  High      1.0% (60th)   0.8
openssl        3.0.7-r0   3.1.0-r0   apk   CGA-mj8v-4crj-mm6x  Medium    1.5% (71st)   0.7
git            2.39.0-r0             apk   CVE-2025-48385      High      0.9% (55th)   0.7
git            2.39.0-r0  2.45.1-r0  apk   CGA-q59g-6cmh-h273  High      0.9% (56th)   0.7
git            2.39.0-r0  2.39.2-r1  apk   CGA-973m-xjh9-w8h6  Medium    1.1% (63rd)   0.6
openssl        3.0.7-r0   3.6.1-r0   apk   CGA-ggqp-v682-3mmx  High      0.8% (54th)   0.6
openssl        3.0.7-r0   3.6.2-r0   apk   CGA-2m6v-xf35-w7r6  High      0.8% (53rd)   0.6
openssl        3.0.7-r0   3.3.0-r8   apk   CGA-97xx-3w7g-6gm5  Medium    1.1% (63rd)   0.6
openssl        3.0.7-r0   3.6.1-r0   apk   CGA-8h52-q2fx-gx4p  High      0.8% (52nd)   0.6
openssl        3.0.7-r0   3.6.3-r0   apk   CGA-gf26-66cr-92gv  Medium    1.0% (59th)   0.5
openssl        3.0.7-r0   3.1.0-r5   apk   CGA-qp33-5hvr-984x  Medium    1.0% (57th)   0.5
binutils       2.39-r4    2.40-r3    apk   CGA-qwq7-9xhr-q7g4  Medium    0.9% (56th)   0.5
openssl        3.0.7-r0   3.6.3-r0   apk   CGA-f678-5v67-vhcj  High      0.6% (46th)   0.5
openssl        3.0.7-r0   3.6.3-r0   apk   CGA-4p89-mwph-mpph  High      0.6% (45th)   0.5
openssl        3.0.7-r0   3.6.3-r0   apk   CGA-6mcp-mm5g-jxrj  High      0.6% (44th)   0.4
openssl        3.0.7-r0   3.6.1-r0   apk   CGA-vmpm-7qr6-cwvf  Medium    0.7% (51st)   0.4
git            2.39.0-r0             apk   CVE-2024-52005      High      0.5% (40th)   0.4
openssl        3.0.7-r0   3.6.3-r0   apk   CGA-rgfj-c398-6gv5  High      0.5% (40th)   0.4
binutils       2.39-r4    2.46-r0    apk   CGA-v247-xrvq-gxvg  High      0.5% (40th)   0.4
git            2.39.0-r0  2.39.2-r1  apk   CGA-7rch-f29w-7fpg  Medium    0.7% (49th)   0.4
binutils       2.39-r4    2.40-r0    apk   CGA-v38c-29jm-xcv3  High      0.5% (39th)   0.4
openssl        3.0.7-r0   3.6.1-r0   apk   CGA-pww6-v8cf-x32w  High      0.4% (36th)   0.3
git            2.39.0-r0  2.45.1-r0  apk   CGA-wrrq-54cr-4pc5  Low       1.0% (57th)   0.3
openssl        3.0.7-r0   3.6.1-r3   apk   CGA-g6w5-5h7c-xmc7  High      0.4% (35th)   0.3
openssl        3.0.7-r0   3.1.1-r2   apk   CGA-99cx-6vv8-jc44  Medium    0.6% (46th)   0.3
openssl        3.0.7-r0   3.6.3-r0   apk   CGA-v77x-mff6-j494  Critical  0.4% (27th)   0.3
openssl        3.0.7-r0   3.6.3-r0   apk   CGA-72qw-77pm-grmj  Medium    0.6% (43rd)   0.3
openssl        3.0.7-r0   3.6.2-r0   apk   CGA-gxj3-3w3x-vx9h  Critical  0.3% (23rd)   0.3
openssl        3.0.7-r0   3.4.0-r6   apk   CGA-qxwv-h54f-gwqm  Medium    0.6% (45th)   0.3
git            2.39.0-r0  2.48.1-r0  apk   CGA-jr7c-fw8r-7h4f  Low       1.0% (60th)   0.3
openssl        3.0.7-r0   3.6.1-r0   apk   CGA-rpw7-8rcj-25xj  Medium    0.5% (40th)   0.3
sysstat        12.6.2-r0  12.7.4-r0  apk   CGA-642j-mjm6-jqp7  High      0.3% (25th)   0.3
git            2.39.0-r0  2.54.0-r0  apk   CGA-8pxj-3rxx-c2qc  High      0.3% (23rd)   0.2
openssl        3.0.7-r0   3.6.1-r0   apk   CGA-8724-74qp-7vp4  Medium    0.4% (33rd)   0.2
openssl        3.0.7-r0   3.6.3-r0   apk   CGA-hq5g-cpx3-w3c3  Medium    0.4% (32nd)   0.2
openssl        3.0.7-r0   3.6.3-r0   apk   CGA-3vhj-mvw4-x4mj  Low       0.6% (44th)   0.2
openssl        3.0.7-r0   3.5.1-r0   apk   CGA-m454-q539-fg8c  Medium    0.3% (24th)   0.2
git            2.39.0-r0             apk   CVE-2025-48386      Medium    0.3% (24th)   0.2
git            2.39.0-r0  2.45.1-r0  apk   CGA-fqh4-7gg7-fm33  Low       0.5% (41st)   0.2
openssl        3.0.7-r0   3.6.3-r0   apk   CGA-r8v3-h6jr-g5gg  Medium    0.4% (28th)   0.2
openssl        3.0.7-r0   3.6.3-r0   apk   CGA-6jgq-q8x4-jwm9  High      0.2% (14th)   0.2
git            2.39.0-r0  2.48.1-r0  apk   CGA-v67j-6c3f-3xqw  Low       0.7% (47th)   0.2
openssl        3.0.7-r0   3.6.3-r0   apk   CGA-m98w-7xrq-cqw4  Low       0.5% (37th)   0.2
openssl        3.0.7-r0   3.6.3-r0   apk   CGA-mm8v-2mxp-4988  Medium    0.3% (22nd)   0.2
binutils       2.39-r4    2.46.1-r2  apk   CGA-h43x-cqw6-4cjp  Medium    0.3% (16th)   0.1
binutils       2.39-r4    2.46-r0    apk   CGA-v7f8-23wj-rx8v  Medium    0.3% (17th)   0.1
openssl        3.0.7-r0   3.6.2-r0   apk   CGA-mfrg-992g-qvmr  Medium    0.2% (15th)   0.1
binutils       2.39-r4    2.46-r2    apk   CGA-m933-q4g4-vp8v  High      0.2% (6th)    0.1
binutils       2.39-r4    2.46.1-r0  apk   CGA-wgch-chvr-vmf4  Medium    0.2% (15th)   0.1
libnghttp2-14  1.69.0-r0  1.70.0-r0  apk   CGA-43r8-f2xj-r388  Medium    0.2% (10th)   0.1
libnghttp2-14  1.69.0-r0  1.70.0-r0  apk   CGA-8pjw-3mvg-2x6m  Medium    0.2% (10th)   0.1
binutils       2.39-r4    2.46.1-r2  apk   CGA-4pwv-cm48-hfjq  Medium    0.2% (8th)    0.1
binutils       2.39-r4    2.46.1-r0  apk   CGA-8586-9447-2gjh  Medium    0.2% (6th)    < 0.1
binutils       2.39-r4    2.46.1-r2  apk   CGA-xm8j-4g99-w9p5  Medium    0.2% (6th)    < 0.1
openssl        3.0.7-r0   3.6.1-r0   apk   CGA-h6xg-wcqx-j6v6  Medium    0.2% (7th)    < 0.1
openssl        3.0.7-r0   3.6.1-r0   apk   CGA-j87x-qhgx-7pf8  Medium    0.1% (4th)    < 0.1
openssl        3.0.7-r0   3.6.1-r0   apk   CGA-fgvw-px6h-97cp  Medium    0.2% (4th)    < 0.1
binutils       2.39-r4               apk   CGA-g89r-r3vh-pg54  Medium    0.1% (4th)    < 0.1
binutils       2.39-r4    2.45.1-r0  apk   CGA-r5fp-h96x-qx75  Low       0.3% (17th)   < 0.1
binutils       2.39-r4    2.45.1-r0  apk   CGA-7hrx-fpr7-mvvm  Low       0.3% (17th)   < 0.1
binutils       2.39-r4    2.45.1-r0  apk   CGA-9jhx-3jc3-mv9r  Low       0.3% (17th)   < 0.1
binutils       2.39-r4    2.45.1-r0  apk   CGA-hrr6-r28w-hw3c  Low       0.2% (14th)   < 0.1
binutils       2.39-r4    2.45.1-r0  apk   CGA-vwwc-43p7-phcf  Low       0.2% (14th)   < 0.1
binutils       2.39-r4               apk   CGA-9gj8-gjxv-82qf  Medium    0.1% (1st)    < 0.1
binutils       2.39-r4    2.45.1-r0  apk   CGA-938x-j2mq-32r7  Low       0.2% (12th)   < 0.1
openssl        3.0.7-r0   3.6.1-r0   apk   CGA-wm9p-95p7-7xrw  Medium    0.1% (1st)    < 0.1
binutils       2.39-r4    2.45.1-r0  apk   CGA-xch7-3q9x-33j4  Low       0.2% (10th)   < 0.1
binutils       2.39-r4    2.45.1-r0  apk   CGA-fx9h-pvr4-6vc5  Low       0.2% (9th)    < 0.1
binutils       2.39-r4    2.45.1-r0  apk   CGA-537h-qc6v-h92v  Low       0.2% (8th)    < 0.1
binutils       2.39-r4    2.45.1-r0  apk   CGA-g6h8-w68g-p295  Low       0.2% (8th)    < 0.1
binutils       2.39-r4    2.45.1-r0  apk   CGA-hpm3-r37p-4m5w  Low       0.2% (8th)    < 0.1
binutils       2.39-r4    2.45.1-r0  apk   CGA-fgp2-3735-hw88  Low       0.2% (7th)    < 0.1
```

Almost all of these are CGAs, which we would expect as they supercede the upstream. The few exceptions (CVE-2025-48384, CVE-2025-48385, CVE-2025-48386) do not appear in our advisories, so these referring to upstream are expected. The CGA vulnerabilities have been evaluated and make sense.

I ran testing against all our `ghcr.io/chainguard-images/scanner-test` images and found the expected changes from secdb - arch filtering and origin grouping


philroche (changes requested):
Adding request changes to avoid accidental merge while we continue to validate internally @ Chainguard and discuss the approach with Anchore

grype/matcher/internal/cpe.go
```diff
@@ -37,12 +37,26 @@ func alpineCPEComparableVersion(version string) string {
...
+	return results.ToMatches(), ignore, nil
+}
+
+// FindResultsByCPEs retrieves all vulnerabilities that match any of the provided package's CPEs as a
+// result.Set, so callers can reconcile them against other sources by vulnerability identity. Records
+// the CPE search finds but whose version is not affected are returned as "CPE not vulnerable" ignores.
 //
 //nolint:funlen
-func MatchPackageByCPEs(vulnProvider vulnerability.Provider, p pkg.Package, upstreamMatcher match.MatcherType) ([]match.Match, []match.IgnoreFilter, error) {
+func FindResultsByCPEs(vulnProvider vulnerability.Provider, p pkg.Package, upstreamMatcher match.MatcherType) (result.Set, []match.IgnoreFilter, error) {
```
crosleyzack, line 39: MatchPackagesByCPEs created a result.Set but then converted it into an array. This would then require me to convert it back to a result.Set. Separating out this function which MatchPackageByCPE uses allows the code to be simplified

grype/matcher/internal/result/results.go
```diff
@@ -268,6 +280,22 @@ func (s Set) Update(incoming Set, shouldUpdate func(existing Result, incoming Re
...
+	for id, results := range s {
+		updated := make([]Result, len(results))
+		for i, r := range results {
+			fn(&r)
+			updated[i] = r
+		}
+		out[id] = updated
+	}
+	return out
+}
```
crosleyzack, line 272: there is no existing method to do this, but it is useful for operations like clearing out fixed versions for upstream matches

grype/matcher/internal/cpe.go
```diff
@@ -110,18 +122,46 @@ func MatchPackageByCPEs(vulnProvider vulnerability.Provider, p pkg.Package, upst
...
+				byFingerprint[fingerprint] = len(out)
+				out = append(out, result.Result{
+					ID:              r.ID,
+					Package:         r.Package,
+					Vulnerabilities: []vulnerability.Vulnerability{v},
+					Details:         append([]match.Detail(nil), r.Details...),
+				})
+			}
 		}
+		return out
```
crosleyzack: This is a very ugly way to replicate the `matchesByFingerprint` when using result sets. Not sure if there is a better way to do this.

grype/matcher/internal/result/provider.go
```diff
@@ -205,3 +238,41 @@ func buildMatchDetails(matcher match.MatcherType, distroMatchType match.Type, co
...
+
+	pkgVersion := version.New(catalogedPkg.Version, pkg.VersionFormat(catalogedPkg))
+	matchedCPEs := make([]cpe.CPE, 0, len(vuln.CPEs))
+	for _, c := range vuln.CPEs {
+		if c.Attributes.Version == wfn.Any || c.Attributes.Version == wfn.NA {
+			matchedCPEs = append(matchedCPEs, c)
+			continue
+		}
+
+		constraint, err := version.GetConstraint(c.Attributes.Version, pkgVersion.Format)
```
philroche: The old filterCPEsByVersion applied transformJvmVersion (using the CPE update field) before building this constraint for JVM-format packages; matchedCPEsForSearch does not. This only affects which of the vulnerability's CPEs are reported in the match detail for Java packages, not the match decision. Was dropping the transform intentional, or lost in the move?

grype/matcher/internal/cpe_test.go
```diff
@@ -1019,70 +1019,6 @@ func TestFindMatchesByPackageCPE(t *testing.T) {
 	}
 }
 
-func TestFilterCPEsByVersion(t *testing.T) {
```
philroche, line 1022: TestFilterCPEsByVersion is deleted here, but its logic lives on as matchedCPEsForSearch in result/provider.go with no equivalent test. Could you port this test to cover matchedCPEsForSearch? The cases carry over almost unchanged.

grype/matcher/apk/matcher_test.go
```diff
@@ -470,6 +470,80 @@ func TestMatcherApk_NvdCanceledByUpstreamSecdbNak(t *testing.T) {
...
+			WithRelationshipType(artifact.OwnershipByFileOverlapRelationship)
+	})
+}
+
+// TestMatcherApk_UpstreamCveDedupedByVulnerableAlias langfuse-3-worker is below the CGA fix
+// (3.152.0-r0 < 3.153.0-r0), so both the distro CGA record and the NVD CVE-2026-24398 CPE
+// record consider it vulnerable. The two share identity via the CGA's alias set, so the
+// upstream NVD finding is deduped into the authoritative distro match rather than surfacing
+// separately - only the CGA distro match is returned.
+func TestMatcherApk_UpstreamCveDedupedByVulnerableAlias(t *testing.T) {
```
philroche, line 526: Suggestion: both fixtures use an advisory with a fixed version. A third variant using an advisory with no fixed event (introduced-only, like an open detection record) would cover the most common record shape in the Chainguard OSV feed, and exercises the keep-then-deduplicate path where the feed still considers the package vulnerable at every version.

philroche (changes requested):
The rewrite addresses everything from my earlier review: the no-fixed-version assumption is gone (those records are now handled as ordinary still-vulnerable records), the end-to-end alias tests cover both the drop and the deduplicate outcomes, and the restructured flow is much easier to follow.

Three small items before I approve, left as inline comments:

1. A question on the Java-specific handling that did not survive the move of the CPE detail filtering into the result provider.
2. The test for that moved logic was deleted without an equivalent; porting it looks cheap.
3. A suggested third fixture variant for an advisory with no fixed version, which for the Chainguard OSV feed is the most common record shape.

We are also re-running our side-by-side validation (secdb-built versus OSV-built database scans on the same images) against this branch as it stands now, since our earlier numbers were taken before the rewrite. I will follow up with the results and approve once these are settled.

philroche:
Following up with our validation results, run against this branch as it stands (4273a3c).

We rebuilt grype from the branch and re-ran our side-by-side comparison on two images (the public ghcr.io/chainguard-images/scanner-test:unfixed-vulnerabilities-many-chainguard image and a production Python image), scanning each with a database built from the legacy secdb feed and one built from the Chainguard OSV feed.

1. No change on existing feeds: with the secdb-built database, this branch produces row-for-row identical results to released grype on both images.
2. The fix works: with the OSV-built database, the 13 NVD false positives on glibc that released grype resurfaces drop to zero, and each finding is reported once instead of under two or three identifiers (204 to 120 matches on scanner-test, 34 to 10 on the Python image, converging with the secdb baselines).
3. Nothing lost: an identity-aware coverage check found every finding from the secdb baseline still reported, except two binutils CVEs where the secdb feed itself wrongly attributes subpackage fixes to the origin package - cases where this branch's output is the correct one.

Validation successful from our side. The only pending items are the three inline review comments above; I will approve once those are settled.

grype/matcher/internal/result/provider.go
crosleyzack: Interested in @kzantow's thoughts on this, as I believe this matcher is only for APKs and thus we shouldn't need special handling for JVM versioning here

grype/db/v6/vulnerability_provider.go
```diff
@@ -716,8 +716,24 @@ func filterUnaffectedCPEVersions(constraintMatcher search.VersionConstraintMatch
...
+			// preserve the prior "implicitly vulnerable to all versions" behavior on the
+			// unexpected case that the empty constraint cannot be constructed.
+			return false, nil
+		}
+		matches, err := matcher.MatchesConstraint(constraint)
+		if err != nil {
+			log.WithFields("error", err).Debug("match constraint error on ranges-less package")
+		}
+		// filter if this doesn't match
+		return !matches, nil
```
crosleyzack, line 736: found this edge case during testing. We should not always say that we are vulnerable, we should still ask the matcher. For instance, the `nakConstraint` is defined as:
```
nakConstraint = search.ByConstraintFunc(func(c version.Constraint) (bool, error) {
	return c.String() == nakVersionString, nil   // nakVersionString == "< 0 (apk)"
})
```

since this does not literally have the nakVersionString as its value, we should not match on this. Anything that is doing a version comparison, however, we should.

grype/matcher/apk/matcher_test.go
```diff
@@ -544,6 +544,40 @@ func TestMatcherApk_UpstreamCveDedupedByVulnerableAlias(t *testing.T) {
...
+			AsDistroSearch()
+
+		// ...the aliased upstream NVD CVE is deduped away.
+		findings.DoesNotHaveAnyVulnerabilities("CVE-2026-30000")
+
+		// the ranges-less advisory must NOT be treated as an apk NAK: it is affected at every
+		// version, not "not affected". The "< 0" NAK filter no longer matches a record that
+		// carries no "< 0" constraint, so no Explicit APK NAK ignores are emitted.
+		findings.Ignores().IsEmpty()
+	})
```
crosleyzack, line 578: test validates that for an `"introduced": "0"` vulnerability, we expect the CGA but not the CVE, as the CGA should suppress the upstream CVE

grype/matcher/apk/matcher_test.go
crosleyzack: Added in 1f29ae4f4e9ca49c3cedbe0ad49f05ae4eec2044. Found an edge case that required fixing as a side effect of this as well!

grype/matcher/internal/cpe_test.go
crosleyzack: Test added in 1f29ae4f4e9ca49c3cedbe0ad49f05ae4eec2044

philroche (approved):
All three items from my review are addressed: the ported test for the moved CPE filtering, the introduced-only fixture (which uncovered and fixed a real issue - ranges-less records satisfying arbitrary constraint searches, so an unfixed advisory could be misread as a not-affected marker), and the Java handling question credibly answered pending kzantow's confirmation (only the apk, rpm and dpkg matchers use this CPE path, and none handle JVM-format packages).

We also re-ran our full validation against this exact head (1f29ae4): with a secdb-built database the results remain row-for-row identical to released grype on both test images, and with the Chainguard OSV-built database the results are unchanged from our previous validation - false positives suppressed, one row per finding, nothing lost.

Approving. Thank you @crosleyzack for the quick turnaround and @kzantow for the review support.

grype/matcher/internal/result/provider.go
kzantow: I don't think this was being used for the JVM versioning, but there was a separate issue I pushed a fix for

grype/matcher/internal/result/results.go
```diff
@@ -32,46 +32,51 @@ type Result struct {
...
-	merged.Details = NewMatchDetailsSet(merged.Details...).ToSlice()
-	return merged
-}
-
 func (s Set) ToMatches() []match.Match {
 	var out []match.Match
 	for _, results := range s {
-		merged := unionIntoResult(results)
+		out = append(out, resultsToMatches(results)...)
+	}
```
crosleyzack: would it be possible to have two different `results` in `s` have a match with the same fingerprint? This would not catch that case

crosleyzack: Also, this would run for every matcher now, but originally only existed for apks. Do we want fingerprint deduplication for every matcher?

grype/matcher/internal/result/results_test.go
```diff
@@ -604,6 +604,116 @@ func TestSet_ToMatches(t *testing.T) {
...
+				{
+					Vulnerability: vulnerability.Vulnerability{
+						Reference: vulnerability.Reference{ID: "CVE-2021-1"},
+						Fix:       vulnerability.Fix{Versions: []string{"3.0.0"}},
+					},
+					Package: testPkg,
+					Details: match.Details{{Type: match.CPEMatch, Found: match.CPEResult{VersionConstraint: "< 3.0.0"}}},
+				},
+			},
+		},
```
crosleyzack: would these be two CPEs that differ in some way other than version? I would assume two identical CPEs except for version should never occur, and then I would expect a warning or something

kzantow: Right, this would be a weird data case, not common but is possible: the same namespace tells us 2 conflicting things, maybe due to searching more than once with slightly different parameters.

grype/matcher/internal/result/results.go
kzantow: I did a lot of thinking about this and from what I can tell, we do want fingerprint deduplication at the match level, but today that fingerprint included the _fix version_, which I could not find any support for including, so I've removed that and now the results deduplicate per specific vulnerability (id, namespace) and package. I think this should end up changing further to not include namespace, but I didn't want to add this larger behavior change to this PR, and I believe the change here is a step in the right direction.

grype/matcher/internal/cpe.go
kzantow: Agree -- this was part of what motivated consolidating the match merge behavior in the merge package

grype/matcher/apk/matcher.go
```diff
@@ -33,229 +44,260 @@ func (m *Matcher) Type() match.MatcherType {
...
+		vulnerable = vulnerable.Map(func(r *result.Result) {
+			details := make([]match.Detail, len(r.Details))
+			for i, d := range r.Details {
+				if d.Type == match.ExactDirectMatch {
+					d.Type = match.ExactIndirectMatch
 				}
-				finalCpeMatches = append(finalCpeMatches, nvdOnlyMatch)
-			}
-			continue
-		}
```
crosleyzack, line 122: Code that I had based removing the fixed version off of

wagoodman (approved):

Issue #2863, opened 2025-08-11
Return vulnerabilities for packages with unknown versions


kzantow:
**What would you like to be added**:
Add a configuration option to treat unknown versions like `0` for matching purposes.

**Why is this needed**:
For packages without version information, we are not surfacing vulnerabilities so users are not made aware of potentially critical vulnerabilities that affect their software.

**Additional context**:
We [added a binary classifier for Vault](https://github.com/anchore/syft/pull/4121) because there was no flag to surface a vulnerability we believe is critical for users to know about. Other critical vulnerabilities should be surfaced to users who want to know these potential vulnerabilities and manually determine if they are affected.

PR #3641, opened 2026-08-07
fix(match): skip packages with an empty version like the "unknown" sentinel

sueun-dev:
`isUnknownVersion` in the shared matcher helpers only recognizes the literal `"unknown"` string, so a package that reaches the distro or language matcher with an empty version slips past the skip guard. `OnlyVulnerableVersions` then sees a blank version and returns a match-everything predicate ("if no version is provided, match everything"), so the package gets every disclosure recorded for its name reported as a match.

That path is reachable from an imported SBOM whose component omits a version (e.g. `grype sbom:./bom.json`), or any provider that leaves the version blank. The effect is a flood of false positives: an `activerecord` gem with no version comes back matched against every activerecord CVE in the feed, regardless of the affected ranges.

The rpm matcher already handles this — its own `isUnknownVersion` is `v == "" || strings.ToLower(v) == "unknown"`. And per #2863 the intended default for versionless packages is to *not* surface vulnerabilities (surfacing them is a requested opt-in, not the current behavior). This change brings the distro and language matchers in line: an empty version is treated the same as `unknown`, so the package is skipped.

The CPE matcher is left as-is on purpose. It supports versionless matching by design — `MatchPackageByCPEs` has explicit empty-version test cases and `filterCPEsByVersion` has a "do not filter on empty version" case — so it keeps calling `isUnknownVersion` directly.

## Test

`TestMatchPackageByDistroSkipsMissingVersion` and `TestMatchPackageByLanguageSkipsMissingVersion` cover both the empty and `unknown` forms. Both fail before this change (the empty-version package produces matches) and pass after.

### DIFF
--- a/grype/matcher/internal/distro.go
+++ b/grype/matcher/internal/distro.go
@@ -27,7 +27,7 @@ func MatchPackageByDistro(provider vulnerability.Provider, searchPkg pkg.Package
 		return nil, nil, nil
 	}
 
-	if isUnknownVersion(searchPkg.Version) {
+	if isMissingVersion(searchPkg.Version) {
 		log.WithFields("package", searchPkg.Name).Trace("skipping package with unknown version")
 		return nil, nil, nil
 	}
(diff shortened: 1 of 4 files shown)


noqt (approved):
Independent review from NOQT at `9dd90cc38fbc7e23690f7e7b9e2b91100af7994f`.

This closes a real false-positive path at the right boundary. Distro and language range matchers now treat an empty version like the existing `unknown` sentinel, so an imported versionless component can’t match every historical disclosure for its name. The CPE path remains unchanged, preserving its deliberate versionless behavior. That also matches the current product policy in #2863: broad matching for unknown versions is a future opt-in, not today’s default.

Validation performed:

- the new distro and language regressions pass on the exact head
- the full affected matcher package passes and vets cleanly
- Git produces a clean merge with current main
- the full affected package and vet pass again on that merged tree
- exact-range diff hygiene passes

The change is focused, preserves the intentional CPE exception, and I found no blocking issue.

PR #3643, opened 2026-08-08
fix: treat Amazon Linux's ALAS feed as comprehensive when deduplicati…

### LINKED ISSUE #3612: amazon linux should be treated as having a comprehensive feed in distrofeediscomprehensive()
**What happened**:

on amazon linux 2023, syft catalogs rpm-pkg python libraries twice, once from the rpm db & once from t installed egg-info or dist-info metadata & correctly links them with an ownership-by-file-overlap relationship, grype already suppresses the duplicate language package before matching via removepackagesbyoverlap() added in #1387 for #1373, but distrofeediscomprehensive() in grype/pkg/package.go currently hardcodes amazon linux to false

```go
if dst.Type == distro.AmazonLinux {
    // AmazonLinux shows "like rhel" but is not an rhel clone
    // and does not have an exhaustive vulnerability feed.
    return false
}
```

as a result, file level python duplicates are matched against ghsa & pypi using t upstream ver string, which by design cannot reflect amazon linux backporting model, every backported rpm that repackages a python libs creates false positives that t rpm level, alas aware, matching correctly avoids..

docker file

```
FROM public.ecr.aws/amazonlinux/amazonlinux:2023

RUN dnf install -y \
        python3 \
        python3-urllib3 \
        python3-setuptools \
        python3-cryptography \
        python3-lxml \
        shadow-utils \
    && dnf clean all
```

- Aalas explorer, https://explore.alas.aws.amazon.com/, now tracks both fixed & unfixed cves on dedicated per-cve pages, pending-fix cases have their own pages, such as https://explore.alas.aws.amazon.com/CVE-2026-56852.html, and not-affected statuses are published too, which matches distrofeediscomprehensive()'s stated criterion
- aws support team formally confirmed t false positive mechanism & the feed coverage across many cases

> third party vulnerability scanners often report false positives on amazon linux because they compare upstream package versions with cve databases without recognizing amazon linux backporting, they read the upstream version from the package metadata instead of the rpm package version that includes the backported security patches

- based on my understanding, grype already ingests alas through t amazon provider in grype db, the data is already in the database, only this gate prevents it from being treated as authoritative for overlap cases

**Environment**:
- Output of `grype version`: latest
- OS (e.g: `cat /etc/os-release` or similar): linux

hamodywe:
…ng packages

distroFeedIsComprehensive() hardcoded Amazon Linux to false, which meant excludePackage() never suppressed the language-level duplicate that syft also reports for an RPM-repackaged Python library on Amazon Linux (once from the RPM db, once from egg-info/dist-info metadata). The suppression that every other comprehensive distro gets skipped Amazon Linux, so its language-level duplicate is matched against upstream PyPI/GHSA advisories using the plain upstream version string, which cannot reflect Amazon's RPM backporting -- producing a false-positive CVE for every backported package repackaged this way.

grype-db already ingests the ALAS feed, which -- like the distros in comprehensiveDistros -- tracks both fixed and unfixed/pending-fix CVEs per package (see https://explore.alas.aws.amazon.com), the same criterion the rest of the list was derived from. The AmazonLinux override predates that feed's current coverage.

Removed the override and added distro.AmazonLinux to comprehensiveDistros. Amazon Linux's ID_LIKE varies by release (AL2023 reports "fedora", AL2 reports "centos rhel fedora") and never matches any Type constant in the list via the existing IDLike fallback, so an explicit entry is required rather than relying on that fallback.

Updated the existing "amzn linux doesn't remove packages in this way" test case in Test_RemovePackagesByOverlap, which encoded the old (incorrect) behavior as expected, to assert the corrected behavior instead -- it now behaves the same as the RHEL case directly above it in the table. Confirmed the updated assertion fails against the unmodified code (reverting only package.go, keeping the test change) with a clear mismatch on the expected package list, and passes with the fix applied.

fixes #3612

### DIFF
--- a/grype/pkg/package.go
+++ b/grype/pkg/package.go
@@ -284,11 +284,6 @@ func distroFeedIsComprehensive(dst *distro.Distro) bool {
 	if dst == nil {
 		return false
 	}
-	if dst.Type == distro.AmazonLinux {
-		// AmazonLinux shows "like rhel" but is not an rhel clone
-		// and does not have an exhaustive vulnerability feed.
-		return false
-	}
 	for _, d := range comprehensiveDistros {
 		if strings.EqualFold(string(d), dst.Name()) {
 			return true
@@ -305,7 +300,14 @@ func distroFeedIsComprehensive(dst *distro.Distro) bool {
 // computed by:
 // sqlite3 vulnerability.db 'select distinct namespace from vulnerability where fix_state in ("wont-fix", "not-fixed") order by namespace;' | cut -d ':' -f 1 | sort | uniq
 // then removing 'github'
+//
+// AmazonLinux was added separately (see https://github.com/anchore/grype/issues/3612):
+// the ALAS feed ingested by grype-db tracks both fixed and unfixed/pending-fix CVEs per
+// package (see https://explore.alas.aws.amazon.com), the same criterion used to derive
+// the rest of this list, but it was excluded by an explicit override that predates that
+// feed's current coverage.
 var comprehensiveDistros = []distro.Type{
+	distro.AmazonLinux,
 	distro.ArchLinux,
 	distro.Azure,
 	distro.Debian,
--- a/grype/pkg/package_test.go
+++ b/grype/pkg/package_test.go
@@ -1523,11 +1523,16 @@ func Test_RemovePackagesByOverlap(t *testing.T) {
 			expectedPackages: []string{"rpm:python3-rpm@4.14.3-26.el8"},
 		},
 		{
-			name: "amzn linux doesn't remove packages in this way",
+			// AmazonLinux's ALAS feed is comprehensive (tracks both fixed and unfixed CVEs per
+			// package, see https://github.com/anchore/grype/issues/3612), so this behaves the
+			// same as the "python bindings for system RPM install" case above: the language-level
+			// duplicate is removed rather than matched separately against the upstream PyPI
+			// version, which would misreport CVEs already backported by Amazon's RPM.
+			name: "amzn linux removes packages in this way, same as other comprehensive distros",
 			sbom: withLinuxRelease(catalogWithOverlaps(
 				[]string{"rpm:python3-rpm@4.14.3-26.el8", "python:rpm@4.14.3"},
 				[]string{"rpm:python3-rpm@4.14.3-26.el8 -> python:rpm@4.14.3"}), "amzn"),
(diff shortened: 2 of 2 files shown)


noqt (changes requested):
Reviewing from NOQT at exact head 7a6ac418a831ae0272871298d82afd4908ebc5b8.

This changes a security boundary on a premise the current ingestion path does not satisfy. Anchore's Vunnel Amazon provider reads the ALAS RSS feeds, downloads each advisory page, extracts concrete `new_packages` fixes, and emits `FixedIn` rows for those versions. It does not ingest the CVE Explorer's pending/unfixed or not-affected pages:

- https://github.com/anchore/vunnel/blob/e5681bf2d47c3beacfcc0cd3bf677d9f08053f56/src/vunnel/providers/amazon/__init__.py
- https://github.com/anchore/vunnel/blob/e5681bf2d47c3beacfcc0cd3bf677d9f08053f56/src/vunnel/providers/amazon/parser.py

That means Grype's Amazon namespace is not comprehensive in the sense this function requires. Marking it comprehensive drops the overlapping language package even when an upstream vulnerability has no ALAS fixed advisory, turning today's duplicate/false-positive risk into a false-negative risk.

Please keep the Amazon override until the data pipeline ingests and proves unfixed coverage, or narrow this suppression with evidence that the owning RPM/package is covered by ALAS. The changed unit test pins the new behavior; it cannot establish feed completeness.

The focused overlap test, affected-package vet, current-main merge, and diff hygiene pass. This data-coverage gap is the blocker.

willmurphyscode:
@noqt is right - the current vunnel ALAS provider is not comprehensive, so this change is incorrect at this time.

There's a [related vunnel issue](https://github.com/anchore/vunnel/issues/776) and if someone with more familiarity with Amazon Linux wanted to _make_ the vunnel provider include not-fixed disclosures, we'd be very glad. @noqt even if you could give us the URLs of better data sources, that would be a big help; last time I looked into it things felt kind of undocumented and web-scrapy and I went on to other projects without finding a decisively better plan.

@hamodywe thanks for the PR but it is incorrect at this time. Please feel free to open an issue with specific false positives the team can look into. For example, if there's a fix on an RPM that owns a PyPI package, Grype should suppress the PyPI GHSA match (because the distro is claiming to have backported a fix onto a vulnerable upstream), and I'd be particularly interested in gaps where that isn't happening.

kzantow:
I think at this point Amazon Linux feed is _not comprehensive_ and changing it as such would result in many false _negatives_. If this is an incorrect assessment, we should revisit this, or if the provider is updated in such a way that we consider it comprehensive, we will revisit at that time.

PR #3654, opened 2026-08-15
fix: keep the PURL namespace when searching the db by package URL

### LINKED ISSUE #3508: grype db search does not properly parse golang PURL

aron-intframe:
Closes #3508.

`DBSearchPackages.PostLoad` used `purl.Name` on its own and dropped `purl.Namespace`. A Go module arrives split across the two, so the search in the issue:

```
grype db search --pkg "pkg:golang/github.com/gin-gonic/gin@v1.9.0"
DEBUG searching for affected packages os=any pkg=package(name=gin, ecosystem=golang)
No results found
```

looked for `gin`, not `github.com/gin-gonic/gin`.

The reason I didn't just concatenate is that the join is ecosystem-specific and getting it wrong makes things worse elsewhere:

- maven joins on a colon, so `pkg:maven/org.apache.commons/commons-lang3` is `org.apache.commons:commons-lang3`. Took that from `grype/db/v6/name/java.go` rather than inventing it.
- golang, npm, composer, github, bitbucket and swift join on a slash. This also fixes scoped npm, where `pkg:npm/%40types/node` was searching for `node`.
- on an rpm or deb PURL the namespace is the distro, not part of the package name, so those keep using the name alone.

Anything else falls through to the old behaviour, since guessing at a convention I can't point to seemed worse than leaving it.

The CPE specifier still uses the bare `purl.Name` as the product. CPE products aren't namespaced, and `cpe:2.3:a:*:github.com/gin-gonic/gin:*` wouldn't match anything.

Added four cases to `TestDBSearchPackagesPostLoad`. Against the unpatched file the golang, npm and maven ones fail and the rpm one passes, which is the shape I wanted — the rpm case is there to catch a fix that over-corrects.

```
go test ./cmd/grype/cli/options/ -run TestDBSearchPackagesPostLoad
ok  github.com/anchore/grype/cmd/grype/cli/options  0.026s
```

### DIFF
--- a/cmd/grype/cli/options/database_search_packages.go
+++ b/cmd/grype/cli/options/database_search_packages.go
@@ -62,7 +62,9 @@ func (o *DBSearchPackages) PostLoad() error {
 				log.Warnf("ignoring version and qualifiers for package URL %q", purl)
 			}
 
-			o.PkgSpecs = append(o.PkgSpecs, &v6.PackageSpecifier{Name: purl.Name, Ecosystem: purl.Type})
+			name := packageNameFromPURL(purl)
+
+			o.PkgSpecs = append(o.PkgSpecs, &v6.PackageSpecifier{Name: name, Ecosystem: purl.Type})
 			o.CPESpecs = append(o.CPESpecs, &v6.PackageSpecifier{CPE: &cpe.Attributes{Part: "a", Product: purl.Name, TargetSW: purl.Type}})
 
 		default:
@@ -81,4 +83,31 @@ func (o *DBSearchPackages) PostLoad() error {
 	}
 
 	return nil
+}
+
+// packageNameFromPURL builds the package name the DB is keyed by from a parsed
+// package URL. For most ecosystems the PURL name is only the last segment, and
+// dropping the namespace searches for the wrong thing entirely: the Go module
+// "github.com/gin-gonic/gin" arrives as namespace "github.com/gin-gonic" plus
+// name "gin", and searching for "gin" alone finds nothing.
+//
+// How the two are joined is ecosystem-specific, so this only rejoins the types
+// whose convention is known. A namespace that is not part of the package
+// identity, such as the distro on an rpm or deb PURL, is left out.
+func packageNameFromPURL(purl packageurl.PackageURL) string {
+	if purl.Namespace == "" {
+		return purl.Name
+	}
+
+	switch purl.Type {
+	case packageurl.TypeMaven:
+		// Maven coordinates are "<group>:<artifact>", matching how the Java
+		// name resolver builds them (grype/db/v6/name/java.go).
+		return fmt.Sprintf("%s:%s", purl.Namespace, purl.Name)
+	case packageurl.TypeGolang, packageurl.TypeNPM, packageurl.TypeComposer,
+		packageurl.TypeGithub, packageurl.TypeBitbucket, packageurl.TypeSwift:
+		return fmt.Sprintf("%s/%s", purl.Namespace, purl.Name)
+	}
+
+	return purl.Name
 }
--- a/cmd/grype/cli/options/database_search_packages_test.go
+++ b/cmd/grype/cli/options/database_search_packages_test.go
@@ -43,6 +43,58 @@ func TestDBSearchPackagesPostLoad(t *testing.T) {
 			},
 		},
 		{
+			// the module path is split across namespace and name, and searching
+			// for "gin" alone finds nothing
+			name: "golang PURL keeps the full module path",
+			input: DBSearchPackages{
+				Packages: []string{"pkg:golang/github.com/gin-gonic/gin@v1.9.0"},
+			},
+			expectedPkg: v6.PackageSpecifiers{
+				{Name: "github.com/gin-gonic/gin", Ecosystem: "golang"},
+			},
+			expectedCPE: v6.PackageSpecifiers{
+				{CPE: &cpe.Attributes{Part: "a", Product: "gin", TargetSW: "golang"}},
+			},
+		},
+		{
+			name: "scoped npm PURL keeps the scope",
+			input: DBSearchPackages{
+				Packages: []string{"pkg:npm/%40types/node@20.0.0"},
+			},
+			expectedPkg: v6.PackageSpecifiers{
+				{Name: "@types/node", Ecosystem: "npm"},
+			},
+			expectedCPE: v6.PackageSpecifiers{
+				{CPE: &cpe.Attributes{Part: "a", Product: "node", TargetSW: "npm"}},
+			},
+		},
+		{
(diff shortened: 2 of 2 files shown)


noqt (approved):
This fixes the namespaced-PURL search at the right boundary. Go modules, scoped npm packages, and Maven coordinates now retain the identity Grype's database is keyed by, while RPM keeps its distro namespace out of the package name.

That ecosystem-aware split matters: joining every non-Maven namespace would turn an RPM PURL such as `pkg:rpm/redhat/openssl` into the wrong package name. The explicit type list and RPM negative regression make this variant the safer fix.

I checked exact head `ad804d02bb5910d36723a429ea6e754dee3da3bd` with Go 1.26.3:

- the focused PURL cases pass;
- the full CLI options package passes;
- `go vet` and exact-range diff hygiene pass.

The CPE lookup deliberately remains on the bare product name. I found no blocker.

Reviewing from NOQT.

aron-intframe:
I re-ran this rather than taking the approval as read, and it holds at `ad804d02bb5910d36723a429ea6e754dee3da3bd`: `go test ./cmd/grype/cli/options/` and `go vet ./cmd/grype/cli/options/` both pass (Go 1.24.6 host, toolchain auto-upgraded to the `go 1.26.3` pinned in go.mod), with all 11 `TestDBSearchPackagesPostLoad` subtests green.

I also went past the unit tests, since they only assert the string the specifier is built from and not that the string finds anything. I pulled the published v6 DB (`grype db update`, 2.0G cache) and diffed a build of this branch against a build with `cmd/grype/cli/options/database_search_packages.go` reverted to `HEAD~1`:

```
### BEFORE pkg:golang/github.com/gin-gonic/gin
No results found
```

Same DB, same queries, on this branch:

```
GHSA-2c4m-59x9-fr2g  github.com/gin-gonic/gin  go-module  github:language:go  >=1.3.1-0.20190301021747-ccb9e902956d,<1.9.1
GHSA-67hx-6x53-jw92  @babel/traverse  npm        github:language:javascript  <7.23.2
GHSA-j288-q9x7-2f5v  org.apache.commons:commons-lang3                                java-archive  github:language:java  >=3.0,<3.18.0
GHSA-5c58-w9xc-qcj9  symfony/http-kernel  php-composer  github:language:php  >=2.0.0,<2.3.27
GHSA-7fj7-39wj-c64f  github.com/apple/swift-nio  swift      github:language:swift  <2.29.1
```

So #3508 goes from nothing to five GHSAs; `pkg:npm/%40babel/traverse`, `pkg:composer/symfony/http-kernel` and `pkg:swift/github.com/apple/swift-nio@2.0.0` each go from `No results found` to hits; and the maven PURL gains the `github:language:java` row on top of the `nvd:cpe` row it already returned. `pkg:rpm/redhat/openssl` prints the same `No results found` on both builds, so the RPM negative case reflects real behaviour against real data, not just a table assertion.

One thing I turned up while checking, which is outside this PR but shouldn't sit unsaid. The build side already has a helper of the same name and shape, `packageNameFromPURL` in `grype/db/v6/build/transformers/openvex/transform.go`, except its switch covers only maven and npm, so golang falls through:

```
openvex build-side name for pkg:golang/github.com/gin-gonic/gin@v1.9.0     = "gin"
```

That is the exact mismatch its own doc comment warns about: the GHSA transformer stores `github.com/gin-gonic/gin`, so an OpenVEX statement for the same Go module lands on a different `packages` row. Pre-existing, and unaffected either way by this PR. If you would rather have one shared helper than two same-named ones with diverging type lists, I'll do that, here or as a separate PR, whichever you prefer.

What I did not verify: I ran only the options package, not the full suite or `test/cli`. `pkg:github/actions/checkout` returns `No results found` on both builds, so the `github` and `bitbucket` branches remain convention-only, unconfirmed against real data. Also worth knowing, the packageurl library rejects a swift PURL with no version (`version is required`), so the swift path is only reachable with a version that `PostLoad` then warns about and drops; unchanged by this PR.

No new commits pushed, the branch is still at `ad804d0`.

noqt:
That real-DB replay closes the remaining confidence gap for me—thanks for taking it past spec construction. It strengthens the case to land this head unchanged.

I reproduced the OpenVEX mismatch from NOQT with a reviewer-only Go regression:

```text
```

I'd keep that follow-up separate from #3654. This PR is narrowly correct and ready; making helper consolidation a condition would delay the user-facing lookup fix. A separate issue or PR should add the Go OpenVEX regression first, then either extend the build-side switch or share the mapping if that can be done without an awkward dependency boundary.

My recommendation remains: merge #3654 as-is, and track the OpenVEX mapping independently.

aron-intframe:
Agreed on both counts, and I didn't take the regression on faith: on this head (`ad804d0`) I wrote the same reviewer-only test against `packageNameFromPURL` in `grype/db/v6/build/transformers/openvex/transform.go`, and it fails with your exact message:

```text
--- FAIL: TestPackageNameFromPURLGolangNamespace (0.00s)
    reviewer_regression_test.go:19: packageNameFromPURL() = "gin", want "github.com/gin-gonic/gin"
FAIL	github.com/anchore/grype/grype/db/v6/build/transformers/openvex	0.014s
```

The switch there covers only `TypeMaven` and `TypeNPM`, so `pkg:golang/github.com/gin-gonic/gin@v1.9.0` falls through to the bare name. I also checked that the follow-up is as small as it looks: with that regression in place, extending the npm arm to `case packageurl.TypeNPM, packageurl.TypeGolang:` turns the package green:

```text
ok  	github.com/anchore/grype/grype/db/v6/build/transformers/openvex	0.014s
```

Both experiments were local-only and reverted, nothing pushed; the branch is still at `ad804d0`, where the options tests still pass:

```text
ok  	github.com/anchore/grype/cmd/grype/cli/options	0.016s
```

So: merge as-is, track the OpenVEX mapping separately, in your order (regression commit first, then the fix). One thing worth deciding in that follow-up rather than assuming: a shared helper would put `cmd/grype/cli/options` and `grype/db/v6/build` on opposite sides of a dependency boundary, and the CLI-side mapping also covers composer/github/bitbucket/swift, which the build side may not want to claim support for. So the small switch extension may age better than consolidation. I'll take the follow-up to a separate issue/PR once this lands, unless a maintainer gets to it first.

PR #3655, opened 2026-08-16
fix(version): prevent panic on portage versions without digits

ashvinctrl:
Portage version comparison panics on any version string with no digits in it.

`versionRegexp` needs a numeric component, so `FindStringSubmatch("none")` returns nil and `comparePortageVersions` indexes `match1[1]` without checking. On main this is `index out of range [1] with length 0` at portage_version.go:51. It happens on both sides of the comparison, the package version and the version being compared against.

What makes it worth fixing rather than leaving as a bad-input crash: `callMatcherSafely` turns a matcher panic into a fatal error, and `searchDBForMatches` returns an empty `match.Matches` when it sees one, so everything already collected is dropped. Running the default matchers over two portage packages against a mock provider, openssl with a real advisory plus one sibling package versioned `none`:

```
only openssl 1.1.1                matches=1  err=<nil>
openssl 1.1.1 + sibling "none"    matches=0  err=unable to find matches against vulnerability
                                  database: portage-matcher encountered a fatal error:
                                  runtime error: index out of range [1] with length 0
```

The openssl finding disappears and the scan fails. With this change both cases report the one match.

Reachability is narrow. Syft's portage cataloger cannot produce this, since `cpvRe` in parse_portage_contents.go requires `(\d+)` in the version group, so image and directory scans are fine. The path is grype ingesting package data it did not produce, so `grype sbom:...` with a portage package whose version has no digits.

## Approach

`newPortageVersion` now returns `(portageVersion, error)` the way apk, deb, rpm, pacman and the others already do, and the `PortageFormat` case in `getComparator` picks the error up. It was one of only two cases there ignoring one.

The check reuses the same unanchored `versionRegexp`, so nothing that compared successfully before is turned away. Only values that used to panic now return an error, which gets logged and the package skipped, matching how other malformed versions are already handled.

## Testing

Two tests added. The first covers `none`, `latest`, `rolling`, `stable`, `git`, `N/A` and `abc` on the package side, the compared-against side, and through a constraint check. It panics on main and passes with the change. The second pins versions that have to keep working (`1`, `1.2.3`, `1.2.3-r1`, `1.2.3_alpha1`, `1.2.3b`, `12.2`, `a1`) so the new rejection cannot be tightened into a regression.

`go test ./grype/version/...` passes. A full `go test ./...` before and after gives an identical set of 27 failing packages, all pre-existing failures on my Windows checkout (db, network, CLI and integration suites).

No linked issue, found while property-testing the version comparators for antisymmetry.

### DIFF
--- a/grype/version/portage_version.go
+++ b/grype/version/portage_version.go
@@ -1,6 +1,7 @@
 package version
 
 import (
+	"errors"
 	"math/big"
 	"regexp"
 	"strings"
@@ -20,10 +21,19 @@ type portageVersion struct {
 	version string
 }
 
-func newPortageVersion(raw string) portageVersion {
+// errNoNumericComponent is returned for portage versions that versionRegexp cannot match at all.
+// Every portage version starts with a numeric component, so a value without one (e.g. "none")
+// has nothing for comparePortageVersions to work with.
+var errNoNumericComponent = errors.New("no numeric version component")
+
+func newPortageVersion(raw string) (portageVersion, error) {
+	if versionRegexp.FindStringSubmatch(raw) == nil {
+		return portageVersion{}, invalidFormatError(PortageFormat, raw, errNoNumericComponent)
+	}
+
 	return portageVersion{
 		version: raw,
-	}
+	}, nil
 }
 
 func (v portageVersion) Compare(other *Version) (int, error) {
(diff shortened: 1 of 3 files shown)


ashvinctrl:
Only DCO has run on this PR so far. Everything else is still waiting on a maintainer to approve workflow runs for a first-time contributor, so CI has not exercised the change. Could someone start that?

For what it is worth, the panic reproduces on current main. Dropping this probe into `grype/version` on an unpatched checkout:

```go
func TestPortageDigitFreePanicProbe(t *testing.T) {
	a := NewVersion("none", PortageFormat)
	b := NewVersion("1.2.3", PortageFormat)
	a.Compare(b)
}
```

gives:

```
panic: runtime error: index out of range [1] with length 0
	grype/version/portage_version.go:51
	grype/version/portage_version.go:42
	grype/version/portage_version.go:34
```

On this branch `go build ./...` is clean and `go test ./grype/version/...` passes.

willmurphyscode (approved):

PR #3658, opened 2026-08-18
docs: point template authors at presenter models

### LINKED ISSUE #333: Docs for using templates should point to view models, not JSON output
Our current [documentation for using templates](https://github.com/anchore/grype#using-templates) points users to the JSON output to learn about what data is available to the template author. This is insufficient, because although it currently describes the correct shape of data, it leaves users guessing at the correct casing for field names.

A better source of truth than the JSON output is the presenter models themselves, starting with the [top-level data model](https://github.com/anchore/grype/blob/main/grype/presenter/models/document.go#L14-L19).

Example of user pain caused by this in our community Slack: https://anchorecommunity.slack.com/archives/C019BUXV7R6/p1623137850005500

VedantMadane:
## Summary
Point template authors at the presenter models instead of JSON output for field names.

- Document the presenter `Document` model in `templates/README.md`
- Clarify godoc on presenter model types so templates use Go field names
- Fix the broken `#using-templates` link in CLI option help

Fixes #333

### DIFF
--- a/cmd/grype/cli/options/grype.go
+++ b/cmd/grype/cli/options/grype.go
@@ -186,7 +186,9 @@ valid values are: registry, docker, podman`)
   - './out/**/*.json'
 same as --exclude`)
 	descriptions.Add(&o.File, `if using template output, you must provide a path to a Go template file
-see https://github.com/anchore/grype#using-templates for more information on template output
+see https://github.com/anchore/grype/blob/main/templates/README.md for template output guidance
+and https://github.com/anchore/grype/blob/main/grype/presenter/models for the presenter models
+(Go field names used by templates; not the -o json key casing)
 the default path to the template file is the current working directory
 output-template-file: .grype/html.tmpl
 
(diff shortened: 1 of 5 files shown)


noqt (changes requested):
Reviewing from NOQT at exact head bd25511e70338bc0241a8163f90ed29565c198e3.

The model links and Go-field-name guidance are useful, but the README examples are written at the wrong template level. A custom template receives models.Document at its root. That type has Matches, but it has no root Vulnerability or Artifact fields.

A reviewer-only execution of the two documented expressions fails with:

- cannot evaluate field Vulnerability in type models.Document
- cannot evaluate field Artifact in type models.Document

Please make the example structurally valid, for example by ranging over .Matches before using .Vulnerability.Severity and .Artifact.Name. A small presenter-template test that executes the documented shape would keep this from drifting.

The affected options, presenter-model, and template packages otherwise pass and vet cleanly on the exact head and on a clean current-main merge. Both replacement links resolve, and diff hygiene is clean. This is the only blocker I found.

PR #3661, opened 2026-08-20
chore: update the stdin subprocess tests

kzantow:
This PR updates some tests that are timing out in CI a lot lately


kzantow:
There was at least one other change to fix this already

willmurphyscode (approved):

PR #3662, opened 2026-08-21
fix(db search): use the full namespaced package name from a PURL

### LINKED ISSUE #3508: grype db search does not properly parse golang PURL

theadsingh:
`grype db search --pkg` drops the namespace from any PURL that has one, so the lookup runs against the bare name.

`DBSearchPackages.PostLoad` builds the package specifier from `purl.Name` alone:

```go
o.PkgSpecs = append(o.PkgSpecs, &v6.PackageSpecifier{Name: purl.Name, Ecosystem: purl.Type})
```

For `pkg:golang/github.com/gin-gonic/gin@v1.9.0` the packageurl library splits that into `Namespace="github.com/gin-gonic"` and `Name="gin"`, so the search asks for a Go module literally called `gin`. The DB stores Go modules under the full module path (the entries in `grype/db/v6/cpe_lookups.go` are all of the form `github.com/anchore/syft`), so nothing matches. That is the zero-result trace in the issue:

```
DEBUG searching for affected packages os=any pkg=package(name=gin, ecosystem=golang) vuln=any
TRACE fetched package record ... pkg=package(name=gin, ecosystem=go-module) records=0
```

Nothing downstream puts the namespace back. `searchQueryBuilder.normalizePackageName` runs `name.Normalize`, but `name.FromType` only returns a resolver for Python and Java, so for Go modules it is a passthrough.

The other namespaced ecosystems go the same way: `pkg:maven/org.apache.commons/commons-lang3` searched for `commons-lang3` instead of `org.apache.commons:commons-lang3`, and `pkg:npm/%40babel/core` searched for `core` instead of `@babel/core`.

This joins the namespace back on, colon-separated for Maven and slash-separated otherwise. The openvex transformer already does the same join for Maven and npm in its own `packageNameFromPURL`, though it leaves Go modules on the bare name; I have not touched that one, since the build pipeline is a separate path from the CLI search and I could not test a change there against real data.

The CPE specifier deliberately keeps the bare name. CPE product attributes are not namespaced, so `cpe:2.3:a:*:gin:*:*:*:*:*:golang:*:*` is the correct CPE search and only the name-based lookup was wrong.

The three added test cases fail on main with `gin`, `commons-lang3` and `core` respectively, and pass with the change. `go build ./...`, `go vet ./cmd/...` and `go test ./cmd/...` are clean.

Fixes #3508


noqt (changes requested):
This needs one ecosystem boundary before it is safe to merge. `packageNameFromPURL` slash-joins every non-Maven namespace, but an RPM namespace identifies the distro; it is not part of the database package name. Grype's own `db search` example explicitly treats `pkg:rpm/redhat/openssl` as equivalent to `--ecosystem rpm --pkg openssl`.

On exact head `504be32220019fed6acd2b642204aef5b6244530`, I added that documented case to `TestDBSearchPackagesPostLoad` and got:

```text
want Name: "openssl"
got  Name: "redhat/openssl"
```

The untouched PR's options-package suite passes, and vet plus exact-range diff hygiene are clean, so the problem is isolated to the unconditional fallback. Please join namespaces only for ecosystems where the namespace is part of package identity and add the RPM case as a negative regression.

Reviewing from NOQT.

theadsingh:
You're right, and it's grype's own documented behaviour I broke: `db_search.go` shows
`--pkg 'pkg:rpm/redhat/openssl'` as equivalent to `--ecosystem rpm --pkg openssl`. I
reproduced your result before changing anything, and deb goes the same way:

```
rpm  want "openssl"  got "redhat/openssl"
deb  want "curl"     got "debian/curl"
```

Swapped the unconditional fallback for an allowlist, so the join only happens where the
namespace is part of the package's identity in the DB:

- maven, joined with `:` (groupId:artifactId)
- golang and npm, joined with `/` (full module path, npm scope)
- everything else keeps the bare name

I kept it to those three rather than enumerating more, because those are the ones I could
evidence in-tree: the golang entries in `cpe_lookups.go` are full module paths, and the
openvex transformer's own `packageNameFromPURL` already does exactly this allowlist for
maven and npm. Anything I could not evidence now falls into the bare-name default, which is
the safer direction for a distro namespace.

Added the rpm case you asked for plus a deb one as negative regressions. Both fail on the
previous commit with the values above and pass now. `go build ./...`, `go vet ./cmd/...`
and `go test ./cmd/...` are clean.

The `CLI tests (Linux)` failure on the previous run was `TestSubprocessStdin` timing out at
84s, which does not touch this code path; the re-push should re-run it.

PR #3671, opened 2026-08-24
Document all Grype ignore-rule fields

### LINKED ISSUE #3663: Docs missing for include-aliases
The `include-aliases` option (under an `ignore` rule) is not mentioned on this page at all:
https://oss.anchore.com/docs/reference/grype/configuration/

However it is present in the [code](https://github.com/anchore/grype/blob/ffbca561d576c584b621f5421616337cad013d90/grype/match/ignore.go#L35)

Interestingly https://oss.anchore.com/docs/reference/grype/configuration/ says this is the full set of *supported* rule fields:
```
# This is the full set of supported rule fields:
#   - vulnerability: CVE-2008-4318
#     fix-state: unknown
#     package:
#       name: libcurl
#       version: 1.5.1
#       type: npm
#       location: "/usr/local/lib/node_modules/**"
```
so is `include-aliases` (and also perhaps `reason`, `namespace` & `match-type`) not actually supported and thus deliberately undocumented?

dev-rehaann:
## Summary

Document every supported field in the generated ignore-rule configuration reference.

## Motivation

The reference calls its example the full set of supported fields, but omits include-aliases and several other valid selectors.

## Changes

- document alias, reason, namespace, and match-type fields
- document package language and upstream-name fields

## Type of change

- [x] Documentation

## Checklist

## Testing

- go test ./cmd/grype/cli/options

Closes #3663


noqt (approved):
Reviewing from NOQT at exact head 8e8e92dac93e337d787a977bedfe50a873b81104.

This closes the documentation gap from #3663 cleanly. The generated configuration reference now shows every field accepted by the ignore-rule model, including aliases, reason, namespace, match type, package language and upstream name, while keeping the VEX-only fields separate.

I checked the tags in match.IgnoreRule and its package selector against the rendered grype config output. The focused options tests and vet pass with Go 1.26.3 on the exact head. The branch merges cleanly onto current main f6f2ce0a2293e65c97bf5468926cb0c9ddad822f; the same tests, vet, rendered-config check, and diff hygiene pass there.

DCO is green; no broader hosted CI has run. I found no blocker.

kzantow (approved):

PR #3673, opened 2026-08-26
feat: expose distro-fixed dropped matches via ignoredMatches, add --suppressed-sources filter

### LINKED ISSUE #3450: Expose distro-fixed dropped matches via ignoredMatches so `--show-suppressed` can surface them
## What would you like to be added

Expose the matches that grype produces and then drops via the distro-fixed ownership rule ([`OwnershipIgnores`](https://github.com/anchore/grype/blob/v0.111.0/grype/matcher/internal/ignores.go#L12) invoked with reason `"DistroPackageFixed"` at [`grype/matcher/internal/distro.go#L105`](https://github.com/anchore/grype/blob/v0.111.0/grype/matcher/internal/distro.go#L105)) through the `ignoredMatches` return of [`VulnerabilityMatcher.FindMatchesContext`](https://github.com/anchore/grype/blob/v0.111.0/grype/vulnerability_matcher.go#L92), the same way user-provided ignore rules already are.

Today the drop happens here in [`grype/vulnerability_matcher.go#L255-L256`](https://github.com/anchore/grype/blob/v0.111.0/grype/vulnerability_matcher.go#L255-L256):

```go
filtered, dropped := match.ApplyIgnoreFilters(allMatches, ignoredMatchFilter(allIgnorers))
logIgnoredMatches(dropped)
```

The `dropped` slice is logged and then discarded. The request is to flow it back to the caller instead.

Two reasonable shapes:

1. **Always include dropped matches in `ignoredMatches`**, with an `IgnoreRule` carrying the distro package, secfix version, and a reason such as `DistroPackageFixed` so consumers can render "suppressed because `<distro-package>` is FIXED for `<vuln>` at `<version>`". Simpler change, lights up `--show-suppressed` automatically.
2. **Add an opt-in flag** on `VulnerabilityMatcher` (for example `IncludeDistroSuppressed bool`) and a matching CLI flag (for example `--include-distro-suppressed`) that flips the dropped set into `ignoredMatches`. Slightly more conservative, same downstream effect.

Either shape unblocks the use case below; the first is simpler and is consistent with how user-rule-ignored matches behave today.

## Why is this needed

We rely on grype to validate that remediations on our APK packages stay remediated. Since `v0.110.0` ( #3282 ), a FIXED entry in our distro security feed for an APK suppresses the same CVE on every component bundled inside that APK on every subsequent rebuild. This is the right default for reducing noise, but the suppression is silent: the dropped matches do not appear in `ignoredMatches`, so they are invisible to `--show-suppressed` and to SDK consumers.

The downstream effect is that once a FIXED entry lands in `security.json` for an APK at version `X-rN`, every subsequent rebuild of that APK is implicitly assumed to preserve the fix. If a rebuild downgrades a bundled stdlib or vendored module to a version that re-introduces the CVE, grype reports no findings and there is no signal that a regression occurred. The blindspot is permanent for any APK that has ever had a FIXED secfix entry, across every component bundled inside it (Go modules, stdlib, Python packages, Java jars, Node modules).

Exposing the dropped set via `ignoredMatches` lets downstream consumers cross-check internal advisory state against grype's own assertion and reopen an advisory when the bundled-component state diverges, without having to re-implement the matching logic on the consumer side.

### Ignored vs dropped

The two kinds of filtered matches are intentionally distinguished in [`grype/vulnerability_matcher.go#L246-L248`](https://github.com/anchore/grype/blob/v0.111.0/grype/vulnerability_matcher.go#L246-L248):

```go
// note: there is a difference between "ignore" and "dropped" matches.
// ignored: matches that are filtered out due to user-provided ignore rules
// dropped: matches that are filtered out due to hard-coded rules
```

`ignored` flows back to callers via the second return of `FindMatches` / `FindMatchesContext`. `dropped` is captured by [`filtered, dropped := match.ApplyIgnoreFilters(...)`](https://github.com/anchore/grype/blob/v0.111.0/grype/vulnerability_matcher.go#L255) and then discarded after [`logIgnoredMatches(dropped)`](https://github.com/anchore/grype/blob/v0.111.0/grype/vulnerability_matcher.go#L256).

### Worked example

CVE-2026-25679 affects the Go stdlib and is fixed in `1.25.8` / `1.26.1` (see [GO-2026-4601](https://pkg.go.dev/vuln/GO-2026-4601)).

- `fuse-overlayfs-snapshotter-2.1.7-r3` was published with go 1.25.8+ and a FIXED secfix entry was published. The Chainguard security feed at <https://packages.cgr.dev/chainguard/security.json> records:

  ```json
  "2.1.7-r3": [
    "CVE-2026-25679",
    ...
  ]
  ```

- `fuse-overlayfs-snapshotter-2.1.7-r7` was rebuilt with go 1.24.13. The stdlib in r7 is again vulnerable to CVE-2026-25679.

Running this on r7:

```shell
grype --show-suppressed fuse-overlayfs-snapshotter-2.1.7-r7.apk
```

returns no findings and no ignored matches. The CPE match for `cpe:2.3:a:golang:go:1.24.13` against the DB entry `cpe:2.3:a:golang:go:*` with constraint `< 1.25.8 || >= 1.26.0-0, < 1.26.1` is produced by the matcher and then dropped, because the containing APK has a FIXED secfix record at r3 and r7 satisfies `>= 2.1.7-r3`. Today there is no way for a caller to see that match.

This same shape applies to every bundled language package inside any APK that has ever had a FIXED secfix entry. 

### References

- Original RFC for explicit reporting of negative matches: #1426
- Change that introduced the suppression: #3282
- Suppression code path (pinned to `v0.111.0`):
  - [`grype/matcher/internal/distro.go#L105`](https://github.com/anchore/grype/blob/v0.111.0/grype/matcher/internal/distro.go#L105) — `OwnershipIgnores(... "DistroPackageFixed" ...)` call site
  - [`grype/matcher/internal/ignores.go#L12`](https://github.com/anchore/grype/blob/v0.111.0/grype/matcher/internal/ignores.go#L12) — `OwnershipIgnores` function definition
  - [`grype/vulnerability_matcher.go#L255-L256`](https://github.com/anchore/grype/blob/v0.111.0/grype/vulnerability_matcher.go#L255-L256) — `filtered, dropped := match.ApplyIgnoreFilters(...)` immediately followed by `logIgnoredMatches(dropped)`, with `dropped` not returned
  - [`grype/vulnerability_matcher.go#L92`](https://github.com/anchore/grype/blob/v0.111.0/grype/vulnerability_matcher.go#L92) — `FindMatchesContext` signature, the public API affected
- Go advisory for the worked example: [GO-2026-4601](https://pkg.go.dev/vuln/GO-2026-4601)

philroche:
# PR: Expose distro-fixed dropped matches via `ignoredMatches`

Closes anchore/grype#3450.

Three visible additions:

1. `ignoredMatches` now includes matches that grype today silently drops via hard-coded distro-fixed rules from #3282's `OwnershipIgnores("DistroPackageFixed", ...)`. `--show-suppressed` surfaces them alongside user-rule ignores.
2. Each `IgnoredMatch` carries a new field `Sources []IgnoreSource` (JSON `sources`) identifying which class of rule fired: `user-rule`, `distro-fixed`, `explicit-exclusion`, `vex`, or `hardcoded-correction`.
3. New CLI flag `--suppressed-sources=<comma-list>` filters `--show-suppressed` output by category. Setting it implicitly enables `--show-suppressed`. Table format only; JSON and SARIF emit every entry regardless.

## Changes

- `grype/match/ignore.go`
  - New `IgnoreSource` enum, five values: `user-rule`, `distro-fixed`, `explicit-exclusion`, `vex`, `hardcoded-correction`.
  - New field `Sources []IgnoreSource` on `IgnoredMatch` (`json:"sources,omitempty"`).
  - Helpers `WithSource`, `TagIgnoredMatches`, `DedupeIgnoredMatches`. `DedupeIgnoredMatches` keys on `Match.Fingerprint`. On merge, `Sources` acts as a set (no duplicates, first-occurrence order) and `AppliedIgnoreRules` concatenates (no dedupe, order preserved). The retained `Match` is the first-occurrence instance; matcher-provided per-instance details on losing matches are not carried over. Deliberate tradeoff — happy to widen if reviewers prefer.
- `grype/vulnerability_matcher.go`
  - `searchDBForMatches` gains a third return `[]match.IgnoredMatch` holding every hard-coded drop. Two drop points feed it:
    - Per-package `ApplyExplicitIgnoreRules` at the current line 237, tagged `explicit-exclusion`.
    - Global `ApplyIgnoreFilters(allMatches, ignoredMatchFilter(allIgnorers))` at the current line 255, tagged `distro-fixed` when any applied rule's `Reason == "DistroPackageFixed"` (the reason string `OwnershipIgnores` uses for #3282), otherwise `hardcoded-correction`.
  - `findDBMatches` propagates and tags user-provided ignores as `user-rule`.
  - `findVEXMatches` tags newly-added ignored entries as `vex`.
  - `mergeIgnoredMatches` now calls `match.DedupeIgnoredMatches`, which also fixes existing double-counting under `NormalizeByCVE`.
- `grype/presenter/models/{ignore,document}.go`
  - New `Sources []string` field on the presenter's `IgnoredMatch` (`json:"sources,omitempty"`). `NewDocument` maps it through from `match.IgnoredMatch.Sources`.
- `cmd/grype/cli/options/grype.go`
  - New `--suppressed-sources=<comma-list>` flag. Comma-separated list of `IgnoreSource` categories that filter the `--show-suppressed` output. Setting it implicitly enables `--show-suppressed`. Unknown values return a helpful error listing every valid value at CLI parse time (`parseSuppressedSources` in `cmd/grype/cli/commands/root.go`).
- `internal/format/presenter.go` and `grype/presenter/table/presenter.go`
  - `PresentationConfig` gains `SuppressedSources []string`; the table presenter accepts it and filters `doc.IgnoredMatches` by `Sources` before rendering.
- Table-only. JSON and SARIF continue to emit every entry regardless of the flag, matching the existing `--show-suppressed` convention that flags shape human-readable output rather than raw data.

The table presenter otherwise renders every ignored entry that reaches it, so the newly-surfaced distro-fixed drops appear under `--show-suppressed` without any row-format change.

## Why this shape

In prior discussion with @kzantow (Anchore) about this class of suppression, he proposed exactly this direction:

> We could add an option to include distro suppressed results into the `ignoredMatches` section, which today only includes ignored results from user-configured rules. It's my opinion that we should probably do this by default to show that there was a match according to one data set but it was suppressed by another. If we did this, then they would be shown in the table view using `--show-suppressed` as Will suggested earlier.

Per that suggestion: `searchDBForMatches` always returns the drops. No opt-in flag. Consumers that never inspected the second return keep working. Consumers that want to differentiate use `Sources` rather than pattern-matching on `AppliedIgnoreRules[i].Reason`.

## Worked example

`fuse-overlayfs-snapshotter-2.1.7-r7.apk` bundles Go stdlib `go1.24.13`. NVD entry for CVE-2026-25679 says `< 1.25.8 || >= 1.26.0-0, < 1.26.1`. Chainguard secfix at `2.1.7-r3` marks the CVE FIXED, so `r7` (`>= r3`) is not directly vulnerable; the stdlib match against NVD is suppressed via `OwnershipIgnores(... "DistroPackageFixed" ...)` at `grype/matcher/internal/distro.go:105`.

Before this PR: `grype --show-suppressed fuse-overlayfs-snapshotter-2.1.7-r7.apk` returns no findings and no ignored matches. The stdlib CPE match is produced by the matcher and dropped without trace.

After this PR: the same command returns the stdlib CVE in `ignoredMatches`:

```json
{
  "sources": ["distro-fixed"],
  "appliedIgnoreRules": [
    {
      "vulnerability": "CVE-2026-25679",
      "reason": "DistroPackageFixed"
    }
  ],
  "vulnerability": { "id": "CVE-2026-25679", "namespace": "nvd:cpe" },
  "artifact": { "name": "stdlib", "version": "go1.24.13", "type": "go-module" }
}
```

## Backwards compatibility

- No public API removed or renamed.
- The new `Sources` field on both `match.IgnoredMatch` and `presenter/models.IgnoredMatch` is `omitempty` in JSON. Consumers reading existing keys see no change.
- Table presenter output is unchanged when `--show-suppressed` is not set.
- Table presenter output when `--show-suppressed` IS set now includes suppressions that were previously invisible. That is the intended behaviour.

## Testing

- New unit test `TestVulnerabilityMatcher_DistroFixedSuppressionSurfacesIgnoredMatch` in `grype/vulnerability_matcher_test.go` reproduces the `fuse-overlayfs-snapshotter` / CVE-2026-25679 shape end to end. It asserts on `Sources`, `AppliedIgnoreRules`, and the empty main-matches slice.
- Four subtests in the existing `TestVulnerabilityMatcher_FindMatches` were updated to include `Sources` where relevant: `pass on severity threshold with VEX` picks up `vex`; `normalize by cve -- ignore GHSA` and `ignore CVE (not normalized by CVE)` pick up `user-rule`; `normalize by cve -- ignore CVE` collapses two same-fingerprint entries into one under the new dedup, with an `AppliedIgnoreRules` union of length 2.
- Presenter snapshot fixtures required no regeneration; `Sources` is `omitempty` and no existing fixture exercises a distro-fixed drop path.
- New table-presenter tests cover the `--suppressed-sources` filter: `TestSuppressedSourceAllows` exercises the pure filter helper across seven cases (empty request, single-match, non-match, multi-source match, etc.) and `TestDisplaysIgnoredMatches_FilteredBySuppressedSource` runs three end-to-end presenter cases against the shared analysis fixture.
- `go build ./...` and `go test ./grype/... ./cmd/... ./internal/format/...` clean.

## Open questions for reviewers

- Naming: `IgnoreSource` vs `IgnoreOrigin` vs `IgnoreCategory`. Rename welcome.
- Should `hardcoded-correction` split further (e.g. `protobuf-correction`)? Kept as a single catch-all here to keep the enum surface small.
- Should the `explicit-exclusion` bucket carry the DB row's Reason? Not touched here; can add if useful.
- `--suppressed-sources` uses a comma-separated single-string form (matching grype's own `--ignore-states`). A repeatable `StringSliceVarP`-style flag is not offered because clio's FlagSet exposes only `StringArrayVarP`; happy to switch if clio grows the wrapper or you want to break the idiom.

## Downstream consumer

Chainguard's `cg scan apk --show-suppressed` (chainguard-dev/mono, feature branch on our fork) reads matches with `distro-fixed` in `Sources` to surface regression signals that PR #3282 hides today. Happy to point at the branch if reviewers want the concrete consumer view.

### DIFF
--- a/cmd/grype/cli/commands/root.go
+++ b/cmd/grype/cli/commands/root.go
@@ -120,10 +120,21 @@ var ignoreLinuxKernelHeaders = []match.IgnoreRule{
 
 //nolint:funlen
 func runGrype(ctx context.Context, app clio.Application, opts *options.Grype, userInput string) (errs error) {
+	suppressedSources, err := parseSuppressedSources(opts.SuppressedSources)
+	if err != nil {
+		return err
+	}
+
+	// --suppressed-sources implies --show-suppressed. Without this the filter
+	// would silently drop every row because the underlying rendering loop
+	// short-circuits when ShowSuppressed is false.
+	showSuppressed := opts.ShowSuppressed || len(suppressedSources) > 0
+
 	writer, err := format.MakeScanResultWriter(opts.Outputs, opts.File, format.PresentationConfig{
-		TemplateFilePath: opts.OutputTemplateFile,
-		ShowSuppressed:   opts.ShowSuppressed,
-		Pretty:           opts.Pretty,
+		TemplateFilePath:  opts.OutputTemplateFile,
+		ShowSuppressed:    showSuppressed,
+		SuppressedSources: suppressedSources,
+		Pretty:            opts.Pretty,
 	})
 	if err != nil {
 		return err
@@ -560,4 +571,23 @@ func applyVexRules(opts *options.Grype) error {
 	}
 
 	return nil
+}
+
+// parseSuppressedSources splits raw at commas, validates each entry against
+// the defined IgnoreSource values, and returns the resulting slice. An empty
+// or whitespace-only input yields a nil slice and no error, which the caller
+// treats as "no filter".
+func parseSuppressedSources(raw string) ([]string, error) {
+	entries := stringutil.SplitCommaSeparatedString(raw)
+	if len(entries) == 0 {
+		return nil, nil
+	}
+	out := make([]string, 0, len(entries))
+	for _, s := range entries {
+		if !match.IsValidIgnoreSource(s) {
+			return nil, fmt.Errorf("unknown --suppressed-sources %q: valid values are %v", s, match.ValidIgnoreSources())
+		}
+		out = append(out, s)
+	}
+	return out, nil
 }
--- a/cmd/grype/cli/commands/root.go
+++ b/cmd/grype/cli/commands/root.go
@@ -573,21 +573,27 @@ func applyVexRules(opts *options.Grype) error {
 	return nil
 }
 
-// parseSuppressedSources splits raw at commas, validates each entry against
-// the defined IgnoreSource values, and returns the resulting slice. An empty
-// or whitespace-only input yields a nil slice and no error, which the caller
-// treats as "no filter".
+// parseSuppressedSources splits raw at commas, trims each entry, drops
+// entries that are empty after trimming, and validates the remaining entries
+// against the defined IgnoreSource values. Empty or whitespace-only input
+// yields a nil slice and no error, which the caller treats as "no filter".
+// Trimming is what makes list forms like "user-rule, vex" work; without it
+// the leading space on the second entry would fail validation.
 func parseSuppressedSources(raw string) ([]string, error) {
 	entries := stringutil.SplitCommaSeparatedString(raw)
-	if len(entries) == 0 {
-		return nil, nil
-	}
 	out := make([]string, 0, len(entries))
 	for _, s := range entries {
+		s = strings.TrimSpace(s)
+		if s == "" {
+			continue
+		}
 		if !match.IsValidIgnoreSource(s) {
 			return nil, fmt.Errorf("unknown --suppressed-sources %q: valid values are %v", s, match.ValidIgnoreSources())
 		}
 		out = append(out, s)
+	}
+	if len(out) == 0 {
+		return nil, nil
 	}
 	return out, nil
 }
--- a/cmd/grype/cli/commands/root_test.go
+++ b/cmd/grype/cli/commands/root_test.go
@@ -367,3 +367,61 @@ func Test_applyVexRules(t *testing.T) {
 		})
 	}
 }
+
+func TestParseSuppressedSources(t *testing.T) {
+	tests := []struct {
+		name       string
+		raw        string
+		wantResult []string
+		wantErrSub string
+	}{
+		{
+			name:       "whitespace-only input yields no filter",
+			raw:        "   ",
+			wantResult: nil,
+		},
+		{
+			name:       "empty input yields no filter",
+			raw:        "",
+			wantResult: nil,
+		},
+		{
+			name:       "spaced comma list is trimmed before validation",
+			raw:        "user-rule, vex",
+			wantResult: []string{"user-rule", "vex"},
+		},
+		{
+			name:       "surrounding whitespace on each entry is trimmed",
+			raw:        "  distro-fixed  ,  vex  ",
+			wantResult: []string{"distro-fixed", "vex"},
+		},
+		{
+			name:       "empty entries in the list are dropped",
+			raw:        "user-rule,,vex,",
+			wantResult: []string{"user-rule", "vex"},
+		},
+		{
+			name:       "valid single value",
+			raw:        "distro-fixed",
+			wantResult: []string{"distro-fixed"},
+		},
+		{
+			name:       "invalid value reports the offending token and lists valid values",
+			raw:        "user-config",
+			wantErrSub: `unknown --suppressed-sources "user-config"`,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := parseSuppressedSources(tt.raw)
+			if tt.wantErrSub != "" {
+				require.Error(t, err)
+				assert.Contains(t, err.Error(), tt.wantErrSub)
+				return
+			}
+			require.NoError(t, err)
+			assert.Equal(t, tt.wantResult, got)
+		})
+	}
+}
--- a/cmd/grype/cli/options/grype.go
+++ b/cmd/grype/cli/options/grype.go
@@ -31,6 +31,7 @@ type Grype struct {
 	FailOn                     string             `yaml:"fail-on-severity" json:"fail-on-severity" mapstructure:"fail-on-severity"`
 	Registry                   registry           `yaml:"registry" json:"registry" mapstructure:"registry"`
 	ShowSuppressed             bool               `yaml:"show-suppressed" json:"show-suppressed" mapstructure:"show-suppressed"`
+	SuppressedSources          string             `yaml:"suppressed-sources" json:"suppressed-sources" mapstructure:"suppressed-sources"` // --suppressed-sources, comma-separated list of ignore-source categories to filter --show-suppressed output
 	ByCVE                      bool               `yaml:"by-cve" json:"by-cve" mapstructure:"by-cve"` // --by-cve, indicates if the original match vulnerability IDs should be preserved or the CVE should be used instead
 	SortBy                     SortBy             `yaml:",inline" json:",inline" mapstructure:",squash"`
 	Name                       string             `yaml:"name" json:"name" mapstructure:"name"`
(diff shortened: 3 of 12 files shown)


noqt (changes requested):
Thanks for adding the suppression-source path. I found two blocking issues on exact head `679d06d`:

1. The only failing hosted check is still actionable here. Staticcheck reports SA1019 on `newIgnoredMatchModels(... metadataProvider vulnerability.MetadataProvider)`. The suppression added inside the helper applies to the `newMatch(...)` call, not the helper declaration where the diagnostic is emitted. Please put the project-consistent directive immediately above `newIgnoredMatchModels`, as `NewDocument` already does:

```go
//nolint:staticcheck // MetadataProvider is deprecated but still used internally
func newIgnoredMatchModels(...)
```

2. `parseSuppressedSources` documents whitespace-only input as no filter, but `SplitCommaSeparatedString` neither trims nor drops whitespace-only entries. Consequently `--suppressed-sources " "` errors, and a natural list such as `user-config, vex` rejects the second value. Please trim entries before validation (and drop entries that become empty), with focused tests for whitespace-only, spaced comma lists, valid values, and an invalid value.

The affected match, presenter-model, table, and CLI-command packages otherwise pass locally, and the remaining hosted checks are green. Reviewing from NOQT.

philroche:
Closing in favour of #3705, which keeps only the behaviour change: dropped matches flow into the existing `ignoredMatches` return, gated behind an opt-in config boolean `include-matcher-suppressions`, no struct changes. The typed `Sources` field and `--suppressed-sources` flag are withdrawn; structured categorisation fits better in the negative-matches design (#1426). @noqt thank you for the review; both items you raised were in code the successor no longer carries (the presenter helper and the flag parser), and the plumbing tests move to #3705.

PR #3681, opened 2026-08-30
docs: document full set of ignore rule fields in --ignore help

### LINKED ISSUE #3663: Docs missing for include-aliases

rylena:
Closes #3663

The `--ignore` option help comment claims to enumerate the full set of
supported rule fields, but several fields that are part of the
`match.IgnoreRule` and `match.IgnoreRulePackage` structs are missing
from the example:

- Top-level: `include-aliases`, `reason`, `namespace`, `match-type`
- Under `package:`: `language`, `upstream-name`

This made it look like those fields are unsupported or undocumented,
even though Grype honors them when present in user config.

This change extends the example to list every supported field, with
indentation matching the YAML structure so the nesting under
`package:` is obvious.

### Verification

```
go test ./cmd/grype/cli/options/ -count=1 -short
go test ./grype/match/ -count=1 -short
go build ./cmd/grype/cli/options/
```

All pass.

Signed-off-by: Rylen Anil <rylen.anil@gmail.com>

### DIFF
--- a/cmd/grype/cli/options/grype.go
+++ b/cmd/grype/cli/options/grype.go
@@ -203,8 +203,14 @@ This is the full set of supported rule fields:
     package:
       name: libcurl
       version: 1.5.1
+      language: javascript
       type: npm
       location: "/usr/local/lib/node_modules/**"
+      upstream-name: libcurl
+    namespace: nvd:cpe
+    match-type: exact-direct-match
+    include-aliases: true
+    reason: "not applicable in this environment"
 
 VEX fields apply when Grype reads vex data:
   - vex-status: not_affected


noqt (approved):
This head is complete and consistent with the code it documents. I compared the help block against the YAML tags on `match.IgnoreRule` and `match.IgnoreRulePackage`; all 15 supported fields are represented, with the package-only fields still nested correctly.

On exact head `d5a9abdb2db9051512957c8cbadc8f76487f08bc` with Go 1.26.3, the options and match package tests pass, the affected package builds, and the exact diff is clean.

One integration note: #3671 covers the same six omissions and predates this PR. I wouldn’t merge both. Its field ordering follows the struct more closely, while this version is also valid, complete, and has no technical blocker. Maintainers can choose either variant.

Reviewing from NOQT.

rylena:
Closing this PR in favor of #3671, which was opened earlier and addresses the same set of missing fields in the `--ignore` help comment. Thanks to the reviewer for flagging the overlap. Happy to coordinate on a single fix if any of the open questions in #3671 need another contributor's input.

Issue #3682, opened 2026-08-30
OpenVEX transformer drops Go module namespaces from package handles


noqt:
**What happened**:

The v6 OpenVEX transformer drops the namespace from Go module PURLs. On current `main` (`f6f2ce0a2293e65c97bf5468926cb0c9ddad822f`), `packageNameFromPURL` only reconstructs namespaces for Maven and npm, so:

```text
pkg:golang/github.com/gin-gonic/gin@v1.9.0 -> gin
```

The GHSA transformer and Go matchers use `github.com/gin-gonic/gin`. That leaves the OpenVEX package handle on a different `packages` row, so a statement for the module can be missed.

**What you expected to happen**:

Go PURLs should retain the full module path: `github.com/gin-gonic/gin`. The mapping should stay ecosystem-aware; joining every namespace would incorrectly turn `pkg:rpm/redhat/openssl` into `redhat/openssl`.

**How to reproduce it (as minimally and precisely as possible)**:

Add this focused case beside the OpenVEX transformer tests:

```go
p, err := packageurl.FromString("pkg:golang/github.com/gin-gonic/gin@v1.9.0")
if err != nil {
    t.Fatal(err)
}
if got := packageNameFromPURL(&p); got != "github.com/gin-gonic/gin" {
    t.Fatalf("packageNameFromPURL() = %q, want %q", got, "github.com/gin-gonic/gin")
}
```

Then run:

```console
go test ./grype/db/v6/build/transformers/openvex
```

The assertion reports `got "gin"`.

**Anything else we need to know?**:

This surfaced while reviewing #3654. I’m reporting it separately so that narrow DB-search fix can land unchanged. A focused Go regression should come first; after that, either extend the build-side switch or share the mapping if that does not create an awkward dependency boundary.

Reported from NOQT.

**Environment**:
- Output of `grype version`: source at `main` commit `f6f2ce0a2293e65c97bf5468926cb0c9ddad822f`
- OS: Windows reviewer host; Go 1.26.3

spiffcs:
#3683 addressed this issue. I just did the test case on main and it looks like grype passes this now. If there is more to this issue that I'm missing that is not covered by the initial report tag me and I can reopen 😄

PR #3683, opened 2026-08-30
Keep the namespace on Go module PURLs in the OpenVEX transformer

### LINKED ISSUE #3680: OpenVEX transformer drops the namespace from Go module PURLs
**What happened**:

On current main `f6f2ce0a2293e65c97bf5468926cb0c9ddad822f`, `packageNameFromPURL` in `grype/db/v6/build/transformers/openvex/transform.go` reconstructs Maven and npm names but not namespaced Go modules. For `pkg:golang/github.com/gin-gonic/gin@v1.9.0`, it returns `gin`, so `getPackageHandle` emits a `go-module` package under the short name rather than `github.com/gin-gonic/gin`.

That diverges from the package identity used by the advisory and matcher paths and can put an OpenVEX statement on a different package row.

A namespaced Go PURL should retain the full module path, while ecosystems whose namespace is metadata rather than package identity should remain unchanged.

Add this focused regression in the `openvex` package:

```go
func TestGolangPURLKeepsNamespace(t *testing.T) {
    purl, err := packageurl.FromString("pkg:golang/github.com/gin-gonic/gin@v1.9.0")
    if err != nil {
        t.Fatal(err)
    }

    if got, want := packageNameFromPURL(&purl), "github.com/gin-gonic/gin"; got != want {
        t.Fatalf("packageNameFromPURL() = %q, want %q", got, want)
    }
}
```

Then run:

```text
go test ./grype/db/v6/build/transformers/openvex -run '^TestGolangPURLKeepsNamespace$' -count=1
```

Current result:

```text
```

I reproduced this from NOQT while reviewing #3654. It is pre-existing and outside that PR, so it should not block the ready DB-search fix. A focused regression here should come first; the implementation can then extend the build-side mapping or share the ecosystem mapping if that does not create an awkward dependency boundary.

Context: https://github.com/anchore/grype/pull/3654#issuecomment-5466249912

**Environment**:
- Grype source: current main at `f6f2ce0a2293e65c97bf5468926cb0c9ddad822f`
- Go: `go1.26.3 windows/amd64`
- OS: Windows amd64

arpitjain099:
Fixes #3680.

`packageNameFromPURL` rebuilds Maven and npm names from the purl namespace and falls through to `purl.Name` otherwise, so a Go module lost its path. The reproduction in the issue holds on `main`:

```
expected: "github.com/gin-gonic/gin"
actual  : "gin"
```

Go belongs with npm rather than in the fallthrough because its namespace is not metadata, it is the leading part of the module path. Checking how the purl library actually splits these before assuming:

```
pkg:golang/github.com/gin-gonic/gin@v1.9.0   ns="github.com/gin-gonic"  name="gin"
pkg:golang/k8s.io/client-go@v0.29.0          ns="k8s.io"                name="client-go"
pkg:golang/gopkg.in/yaml.v3@v3.0.1           ns="gopkg.in"              name="yaml.v3"
pkg:golang/stdlib@1.21.0                     ns=""                      name="stdlib"
```

so `Namespace + "/" + Name` reconstructs the module path, and `stdlib` returns early on the existing empty-namespace check and is untouched.

The comment above that switch already anticipates this ("Callers can extend this switch as more VEX providers come online for additional ecosystems"), so I extended the existing `TypeNPM` case rather than adding a separate branch that would do the same join.

## Tests

`Test_packageNameFromPURL` is a table covering the four Go shapes above plus four guards that the change does not widen: Maven still uses `:`, npm scoped names still rejoin, `pkg:deb/debian/curl` still drops a namespace that genuinely is metadata, and a flat ecosystem is unchanged. The Go cases fail on `main` with the actual/expected pair above.

`go build ./...`, `go vet`, `gofmt` and `go test ./grype/db/v6/build/...` are clean.

I have taken only #3680 here. #3682 covers the handle-level consequence in `getPackageHandle`, and I did not want to fold two issues into one PR; happy to follow up on it once you are happy with this, or to close it as covered if you consider the name fix sufficient.

### DIFF
--- a/grype/db/v6/build/transformers/openvex/transform.go
+++ b/grype/db/v6/build/transformers/openvex/transform.go
@@ -158,7 +158,13 @@ func packageNameFromPURL(purl *packageurl.PackageURL) string {
 	switch purl.Type {
 	case packageurl.TypeMaven:
 		return purl.Namespace + ":" + purl.Name
-	case packageurl.TypeNPM:
+	case packageurl.TypeNPM, packageurl.TypeGolang:
+		// For Go the namespace is not metadata, it is the leading part of the
+		// module path: pkg:golang/github.com/gin-gonic/gin splits into
+		// Namespace="github.com/gin-gonic" and Name="gin", and the module is
+		// only identified by the two joined. Dropping it would file the
+		// statement under "gin". A module with no namespace, such as
+		// pkg:golang/stdlib, is returned unchanged by the check above.
 		return purl.Namespace + "/" + purl.Name
 	}
 	return purl.Name
--- a/grype/db/v6/build/transformers/openvex/transform_test.go
+++ b/grype/db/v6/build/transformers/openvex/transform_test.go
@@ -8,6 +8,8 @@ import (
 	govex "github.com/openvex/go-vex/pkg/vex"
 	"github.com/stretchr/testify/require"
 
+	"github.com/anchore/packageurl-go"
+
 	"github.com/anchore/grype/grype/db/internal/provider/unmarshal"
 	"github.com/anchore/grype/grype/db/provider"
 	db "github.com/anchore/grype/grype/db/v6"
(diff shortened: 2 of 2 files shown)


noqt (approved):
Reviewed for NOQT against current `main` (`f6f2ce0a`). This is the right scoped fix for #3680: Go's PURL namespace is part of the module path, while the existing early return and ecosystem switch keep flat and metadata namespaces unchanged.

I replayed the parser-and-join contract with the three namespaced Go shapes, `stdlib`, scoped npm, and namespaced Debian. All six produced the expected package names. The exact head also merges cleanly with current main and `git diff --check` is clean. The new table covers the important identity cases and guards the Maven/npm/other-ecosystem behavior.

I couldn't complete the full OpenVEX package test locally because the Windows Go build did not reach the test binary within the bounded run, so I'm treating that as a local verification limit rather than claiming a package-test pass.

This same name feeds `getPackageHandle`, so #3682 appears to be the duplicate handle-level consequence and can be closed in favor of #3680 once this lands.

kzantow (approved):
Thanks for the contribution @arpitjain099

PR #3696, opened 2026-09-09
fix(openvex): reconstruct full Go module path from namespace+name

### LINKED ISSUE #3680: OpenVEX transformer drops the namespace from Go module PURLs

kokhlo:
Closes #3680.

On current main, `packageNameFromPURL` returns the short name for Go packages (e.g. `"gin"` for `pkg:golang/github.com/gin-gonic/gin`) because only Maven and npm had explicit switch cases. That diverges from the go advisory transformer, which stores the full module path, and can put the OpenVEX handle on a different `packages` row.

Adds `TypeGolang` to the switch, mirroring the npm join. The focused regression now passes:

```text
$ go test ./grype/db/v6/build/transformers/openvex/ -run TestGolangPURLKeepsNamespace -count=1
ok  	github.com/anchore/grype/grype/db/v6/build/transformers/openvex	0.493s
```

The entire suite is green (0.486s locally).

### DIFF
--- a/grype/db/v6/build/transformers/openvex/transform.go
+++ b/grype/db/v6/build/transformers/openvex/transform.go
@@ -151,6 +151,13 @@ func getPackageHandle(product *govex.Product, vuln *unmarshal.OpenVEXVulnerabili
 // FixedIn name verbatim ("@scope/name"). Joining Namespace+"/"+Name
 // reproduces that string. Callers can extend this switch as more VEX
 // providers come online for additional ecosystems.
+//
+// Go modules are stored under their full module path by the go advisory
+// path ("github.com/gin-gonic/gin"), which lives split across the purl
+// namespace and name; joining them keeps the OpenVEX handle on the same
+// packages row the go matcher searches. Ecosystems whose namespace is
+// metadata rather than package identity (e.g. github archives) keep the
+// short name and are not listed here.
 func packageNameFromPURL(purl *packageurl.PackageURL) string {
 	if purl.Namespace == "" {
 		return purl.Name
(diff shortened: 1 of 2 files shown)


kzantow:
Thanks for the contribution; the first of multiple PRs was merged

PR #3705, opened 2026-09-11
feat: optionally include internally dropped matches in ignoredMatches

### LINKED ISSUE #3450: Expose distro-fixed dropped matches via ignoredMatches so `--show-suppressed` can surface them

philroche:
Closes #3450.

## What is changing

Grype internally suppresses matches through three kinds of hard-coded filters:

- distro fixed: the distro reports the vulnerability fixed in the package that owns the matched files (`reason: "distro-fixed"`)
- APK NAK: the distro reports the package was never vulnerable (`reason: "distro-nak"`)
- the built-in false-positive list in `grype/match/explicit_ignores.go` (log4j lookalike packages, protobuf, ruby gem name collisions, zlib)

Today `searchDBForMatches` applies these filters, logs the suppressed matches at debug level, and discards them. This change adds an opt-in configuration boolean, `include-matcher-suppressions` (default false, no CLI flag yet). When enabled, the suppressed matches flow into the existing `ignoredMatches` return of `FindMatchesContext`, alongside the user-rule and VEX entries that already appear there, and from there into the `--show-suppressed` table and the JSON `ignoredMatches` array. The existing `reason` text on each applied rule says why each entry was hidden.

Since these reason values become consumer-visible for the first time in this PR, the second commit renames them from internal prose (`"DistroPackageFixed"`, `"Explicit APK NAK"`) to identifier-style tokens and exports them as constants, `match.IgnoreReasonDistroFixed` and `match.IgnoreReasonDistroNAK`, so Go consumers can import them rather than hard-code strings.

Default behaviour is unchanged. No changes to `IgnoreRule` or `IgnoredMatch` shapes, no CLI flag.

## Example usage

Enable via configuration:

```yaml
# .grype.yaml
include-matcher-suppressions: true
```

or the derived environment variable:

```shell
GRYPE_INCLUDE_MATCHER_SUPPRESSIONS=true grype --show-suppressed <target>
```

JSON consumers can isolate the distro-driven suppressions:

```shell
GRYPE_INCLUDE_MATCHER_SUPPRESSIONS=true grype -o json <target> \
  | jq '.ignoredMatches[] | select(any(.appliedIgnoreRules[]?; .reason == "distro-fixed" or .reason == "distro-nak"))'
```

Verified against the worked example below: scanning the SBOM of `fuse-overlayfs-snapshotter-2.1.7-r7` with the option enabled surfaces the suppressed Go stdlib entries (GO-2026-4601 and related) with `reason: "distro-fixed"`; with the option unset the output is unchanged.

## Why

A consumer of `FindMatchesContext` cannot distinguish "grype found no match" from "grype found a match and suppressed it". That distinction matters to anyone auditing why a vulnerability is absent from a report.

Our use case: detecting CVE regressions in Chainguard packages. Once a fixed or not-vulnerable entry is published in our secdb for a package, grype suppresses that CVE on every later build of the package, so a rebuild that reintroduces the CVE (for example a Go toolchain downgrade) produces a clean scan and no signal. With the suppressed matches surfaced, we filter `ignoredMatches` to the entries whose applied rules carry `reason: "distro-fixed"` or `reason: "distro-nak"`, which are exactly the suppressions caused by our own distro data, and cross-check those against what the package actually contains. User-rule and VEX suppressions stay out of that filter because they are deliberate choices rather than regression candidates.

Worked example: `fuse-overlayfs-snapshotter-2.1.7-r7.apk` bundles Go stdlib `go1.24.13`, vulnerable to CVE-2026-25679 per NVD. A Chainguard secdb fix entry at `2.1.7-r3` suppresses the stdlib match on every later build. Today `grype --show-suppressed` prints nothing for it. With this change enabled the entry appears in `ignoredMatches` with `reason: "distro-fixed"`, and our tooling flags that the fixed claim no longer holds.

## Relationship to other work

Supersedes #3673, which layered a typed source field and a filter flag on the same plumbing. Structured categorisation and per-record provenance belong in the negative-matches work (#1426), so this PR carries only the behaviour change behind an opt-in boolean.

### DIFF
--- a/cmd/grype/cli/commands/root.go
+++ b/cmd/grype/cli/commands/root.go
@@ -234,6 +234,7 @@ func runGrype(ctx context.Context, app clio.Application, opts *options.Grype, us
 		Alerts: grype.AlertsConfig{
 			EnableEOLDistroWarnings: opts.Alerts.EnableEOLDistroWarnings,
 		},
+		IncludeMatcherSuppressions: opts.IncludeMatcherSuppressions,
 	}
 
 	remainingMatches, ignoredMatches, err := vulnMatcher.FindMatchesContext(ctx, packages, pkgContext)
--- a/cmd/grype/cli/options/grype.go
+++ b/cmd/grype/cli/options/grype.go
@@ -31,7 +31,8 @@ type Grype struct {
 	FailOn                     string             `yaml:"fail-on-severity" json:"fail-on-severity" mapstructure:"fail-on-severity"`
 	Registry                   registry           `yaml:"registry" json:"registry" mapstructure:"registry"`
 	ShowSuppressed             bool               `yaml:"show-suppressed" json:"show-suppressed" mapstructure:"show-suppressed"`
-	ByCVE                      bool               `yaml:"by-cve" json:"by-cve" mapstructure:"by-cve"` // --by-cve, indicates if the original match vulnerability IDs should be preserved or the CVE should be used instead
+	IncludeMatcherSuppressions bool               `yaml:"include-matcher-suppressions" json:"include-matcher-suppressions" mapstructure:"include-matcher-suppressions"` // include matches suppressed internally by matchers (distro fixed/NAK records, built-in false-positive list) in the ignored matches output, default=false
+	ByCVE                      bool               `yaml:"by-cve" json:"by-cve" mapstructure:"by-cve"`                                                                   // --by-cve, indicates if the original match vulnerability IDs should be preserved or the CVE should be used instead
 	SortBy                     SortBy             `yaml:",inline" json:",inline" mapstructure:",squash"`
 	Name                       string             `yaml:"name" json:"name" mapstructure:"name"`
 	DefaultImagePullSource     string             `yaml:"default-image-pull-source" json:"default-image-pull-source" mapstructure:"default-image-pull-source"`
@@ -71,6 +72,7 @@ func DefaultGrype(id clio.Identification) *Grype {
 		CheckForAppUpdate:          true,
 		VexAdd:                     []string{},
 		MatchUpstreamKernelHeaders: false,
+		IncludeMatcherSuppressions: false,
 		SortBy:                     defaultSortBy(),
 		Timestamp:                  true,
 		Alerts:                     defaultAlerts(),
@@ -218,6 +220,7 @@ VEX fields apply when Grype reads VEX data:
 `)
 	descriptions.Add(&o.VexAdd, `VEX statuses to consider as ignored rules`)
 	descriptions.Add(&o.MatchUpstreamKernelHeaders, `match kernel-header packages with upstream kernel as kernel vulnerabilities`)
+	descriptions.Add(&o.IncludeMatcherSuppressions, `include matches suppressed internally by matchers (e.g. distro fixed/NAK records, the built-in false-positive list) in the ignored matches output`)
 }
 
 func (o Grype) FailOnSeverity() *vulnerability.Severity {
--- a/grype/vulnerability_matcher.go
+++ b/grype/vulnerability_matcher.go
@@ -46,6 +46,13 @@ type VulnerabilityMatcher struct {
 	VexProcessor          *vex.Processor
 	Alerts                AlertsConfig
 
+	// IncludeMatcherSuppressions includes matches that were dropped by
+	// hard-coded rules or matcher-provided ignore filters (e.g. distro
+	// fixed/NAK records suppressing matches on owned packages) in the
+	// ignoredMatches results. When false (the default) these are discarded
+	// after logging, preserving the previous behavior.
+	IncludeMatcherSuppressions bool
+
 	// tracked packages with distro issues (populated during FindMatches)
 	eolDistroPackages     []pkg.Package
 	distroDetectionFailed bool
@@ -138,7 +145,7 @@ func (m *VulnerabilityMatcher) findDBMatches(ctx context.Context, pkgs []pkg.Pac
 	var ignoredMatches []match.IgnoredMatch
 
 	log.Trace("finding matches against DB")
-	matches, err := m.searchDBForMatches(ctx, pkgs, progressMonitor)
+	matches, droppedMatches, err := m.searchDBForMatches(ctx, pkgs, progressMonitor)
 	if err != nil {
 		if match.IsFatalError(err) {
 			return nil, nil, err
@@ -166,6 +173,14 @@ func (m *VulnerabilityMatcher) findDBMatches(ctx context.Context, pkgs []pkg.Pac
 		ignoredMatches = m.mergeIgnoredMatches(originalIgnoredMatches, ignoredMatches)
 	}
 
+	if m.IncludeMatcherSuppressions {
+		// include matches dropped by hard-coded rules and matcher-provided filters
+		// (e.g. distro fixed/NAK records suppressing matches on owned packages) so
+		// callers can see that a match was found and then suppressed; user-rule
+		// ignores keep their existing positions at the front of the slice
+		ignoredMatches = m.mergeIgnoredMatches(ignoredMatches, droppedMatches)
+	}
+
 	return &matches, ignoredMatches, nil
 }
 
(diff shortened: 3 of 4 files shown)


philroche:
Second commit renames the two internal suppression reason values to identifier-style tokens and exports them as constants, `match.IgnoreReasonDistroFixed` (`"distro-fixed"`) and `match.IgnoreReasonDistroNAK` (`"distro-nak"`). These values reach output for the first time in this PR, so this is the one moment they can be renamed without breaking any consumer, and exporting them as constants lets Go consumers import them instead of hard-coding strings.

grype/match/ignore_test.go
```diff
@@ -1181,7 +1181,7 @@ func TestIgnoreRelatedPackage(t *testing.T) {
 	}
 
 	filter := IgnoreRelatedPackage{
-		Reason:           "Explicit APK NAK by Ownership",
+		Reason:           IgnoreReasonDistroNAK + " by ownership",
```
kzantow: Should this be a specific reason like the others: `distro-nak-by-ownership` ?

grype/match/ignore.go
```diff
@@ -18,6 +18,19 @@ type IgnoreFilter interface {
...
+// to consumers via VulnerabilityMatcher.IncludeMatcherSuppressions. Exported
+// so consumers can compare against them without repeating the literals.
+const (
+	// IgnoreReasonDistroFixed is the reason recorded when a distro fix record
+	// for an owning package suppresses a match on an owned package.
+	IgnoreReasonDistroFixed = "distro-fixed"
+
+	// IgnoreReasonDistroNAK is the reason recorded when a distro record
+	// declaring a package not vulnerable suppresses a match.
+	IgnoreReasonDistroNAK = "distro-nak"
```
kzantow: We should not add these to the public API; can we move these to `grype/matcher/internal`? (or maybe `grype/internal/ignore_reasons.go` since they are referenced in the `grype` package for tests) All of the individual matcher constants can just be removed and reference these constants

philroche: Moved in 4b8906b: the constants now live in the internal package `grype/internal/ignorereasons` (`ignorereasons.DistroFixed`, `ignorereasons.DistroNAK`), out of the public `match` package, and the matcher-local alias constants are removed in favour of referencing the shared ones. External consumers rely on the string values in output rather than importable symbols, which keeps the API surface clear of the internal suppression mechanism ahead of its rework.

grype/match/ignore_test.go
philroche: Done in 4b8906b: the fixture reason is now the distinct token `distro-nak-by-ownership` in both places the concatenated form appeared (`ignore_test.go` and `vulnerability_matcher_test.go`).

philroche: Follow-up in 5b24af0: the `distro-nak-by-ownership` fixture value existed only as test input, nothing builds it, so both fixtures now use `ignorereasons.DistroNAK` and the constants are the only reason values in the tree. Also added the missing `IncludeAliases` assertion so the whole emitted rule is verified.

kzantow (approved):
Good addition! Thanks, @philroche  👍

